PR #202, opened 2020-10-05
Select: initial setup for discussion

benoitgrelard:
Hey all 👋 

**Important**: This PR contains a version of `Select` which is grossly incomplete and inaccessible (only mouse action). However, I wanted to put it up early so we can discuss the specific approach I have taken here around tackling one of the many challenges that `Select` brings us.

Also, we're not focusing on naming here, but rather on what's below 👇 

---

With this out of the way, let me explain here what I'd like us to focus on for this discussion:

One of the many challenges that we face with `Select` is generally obtaining information about the options which are useful for all sorts of accessibility features (typeahead, etc). Our `collection` package handles most of this, collection data about each option and providing us with an `index`.

However, one such need is that we need to be able to render the currently selected option's label inside the select's button. To be able to do so, we need to have collected the information about each option. To be able to do this, the options must be rendered.

That said, if we follow the same approach as our other similar components like `Popover`, `Tooltip`, or even `Dialog`, we typically don't render anything unless it's "open". This is good because when closed then nothing is happening, and code complexity is kept low.

However, this is not enough here…

---

This is in essence, what I tried to tackle here. I have gone around this by rendering alternative version of components down the tree when "closed". This lets us collect the necessary information even if the select is "closed".

There are advantages and caveats, just as everything else.
I will leave some comments inline around the implementation and we can then discuss the solution.

Thanks! 🙏

### DIFF
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -174,7 +174,7 @@ const SelectPosition = forwardRef<typeof POSITION_DEFAULT_TAG, SelectPositionPro
     return context.open ? (
       <SelectPositionImpl ref={forwardedRef} {...props} />
     ) : (
-      <>{props.children}</>
+      <div hidden>{props.children}</div>
     );
   }
 );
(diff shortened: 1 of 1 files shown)


vitalikda:
@benoitgrelard great work here! Could you clarify, is there any ETA for Select?

benoitgrelard:
> @benoitgrelard great work here! Could you clarify, is there any ETA for Select?

With the holidays coming, probably sometime in Jan.

PR #278, opened 2020-12-03
Slider: Remove `number` type from value props

jjenzz:


benoitgrelard:
> @jjenzz I know this is old but why is value an array at all?

Because you can have multiple thumbs.

PR #347, opened 2020-12-17
[react-*]: Remove extended internal data attributes

jjenzz:
What do we think of this approach? If we're happy, I'll apply it across all of the others as part of this PR.


packages/react/polymorphic/src/forwardRefWithAs.ts
```diff
@@ -49,6 +49,8 @@ interface ForwardRefExoticComponentWithAs<
   <As extends React.ElementType<any>>(
     props: MergeWithDOMProps<As, OwnProps & { as: As }>
   ): JSX.Element;
+
+  displayName: string;
```
jjenzz, line 53: I've made this required on this polymorphic thing... I'm not sure this is a good idea tbh. It might be better to just do `CollapsibleContent.displayName!`. Thoughts?

chaance: Wouldn't you get an error when you initially define the component because it's missing that property? This has never worked for me in the past, but maybe it's a tsconfig thing? 🧐

jjenzz: > Wouldn't you get an error when you initially define the component

Yes, because I've made it required so I would expect that you would get an error telling you to add it. That's why I'm saying I don't think this is really a good idea here because some people might want to use this polymorphic stuff but don't care about `displayName`s.

I'm thinking the `CollapsibleContent.displayName!` approach would be better where I call `removePartDataAttrObj` but that also feels weird assuming we've correctly added the display name haha. 

Open to other ideas here of course!

chaance: I think I'm with you. We shouldn't need this frequently so the occasional `!` is totes fine.

chaance: The only thing I can think of is that some apps strip out `displayName` in their production builds to save a few bytes. No idea if those sort of tools should have an impact on `node_modules` or if it's something to consider, but it would break if that were in fact the case.

jjenzz: LOL, good point! doh. Hmmm. I'm not sure either but I'll import the consts instead to be sure. Good spot :+1:

jjenzz:
@chaance We've come full circle here and added a prop like you originally suggested. However, it's not just a prop to solve our internal issues now as I had an idea to make it useful for consumers: https://github.com/radix-ui/primitives/pull/397#issuecomment-760242925

This feels much better, gives us the prop you wanted and makes me happy because we're not just exposing something to solve an internal problem. Consumers can customise selectors now 🎉

benoitgrelard:
♥️ this new approach.
I think if we go for that we should rename that `namespaced` util, feels a bit general in global context like that. But otherwise it's awesome.

jjenzz:
@benoitgrelard sure, what naming do you think works better for that? Maybe just `getSelector`?

benoitgrelard:
Yeah probably something like that, although it is probably used for other stuff too. So either we find a more applicable generic name, or I don't mind if we just do `const getSelector = val => namespaced(val);` either.

jjenzz:
Applying this to all of our components really wore me down on Friday haha. Thoughts on me creating a `Primitive` component that all of our primitives re-compose? 

I could do this then:


```jsx
type PrimitiveOwnProps = {
	/**
	 * A string to use as the component selector for CSS purposes. It will be added as
	 * a data attribute. Pass `null` to remove selector.
	 *
	 * @defaultValue radix-{component-display-name} - For example, `AccordionItem` 
	 * selector will be `radix-accordion-item`.
	 */	
	selector?: string | null;
}
```

jjenzz:
This PR is insanely big now, sorry! Hopefully easier to review each commit but if not I can try split it into individual PRs.

I've not bothered with allowing customisation of the CSS properties in this PR. I felt we can probably tackle that separately to avoid bloating this further.

benoitgrelard:
is it actually ready to review now?

jjenzz:
@benoitgrelard yep, think so 🙈  haha

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -84,32 +87,26 @@ type ContextMenuContentOwnProps = {
...
 const ContextMenuContent = forwardRefWithAs<typeof MenuPrimitive.Root, ContextMenuContentOwnProps>(
   (props, forwardedRef) => {
-    const {
-      anchorRef,
-      disableOutsidePointerEvents = true,
-      side = 'bottom',
-      align = 'start',
-      ...contentProps
-    } = props;
+    const { anchorRef, ...contentProps } = props;
```
jjenzz, line 90: Changes in this file made me realise this `anchorRef` prop isn't being used. Is this a bug?

packages/react/popover/src/Popover.tsx
```diff
@@ -343,24 +349,25 @@ PopoverContent.displayName = CONTENT_NAME;
...
+        onClick={composeEventHandlers(props.onClick, () => context.setOpen(false))}
+      />
+    );
+  }
+);
 
 PopoverClose.displayName = CLOSE_NAME;
 
-/* -----------------------------------------------------------------------------------------------*/
+/* ---------------------------------------------------------------------------------------------- */
```
benoitgrelard, line 370: nit: unintended change?

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -259,15 +261,14 @@ type ScrollAreaInternalProps = {
 };
 
 type ScrollAreaNativeProps = ScrollAreaInternalProps &
-  Omit<ScrollAreaOwnProps, 'unstable_forceNative'> & {
+  Omit<ScrollAreaOwnProps, 'unstable_forceNative' | 'selector'> & {
```
benoitgrelard, line 264: technically should `as` be omitted here too then?

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -301,7 +303,7 @@ const ScrollAreaNative = forwardRefWithAs<typeof ROOT_DEFAULT_TAG, ScrollAreaNat
 );
 
 type ScrollAreaImplProps = ScrollAreaInternalProps &
-  Omit<ScrollAreaOwnProps, 'unstable_forceNative'>;
+  Omit<ScrollAreaOwnProps, 'unstable_forceNative' | 'selector'>;
```
benoitgrelard, line 306: Why are we omitting `selector` here? Isn't this the non-native scrollarea?

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -490,191 +492,191 @@ const ScrollAreaImpl = forwardRefWithAs<typeof ROOT_DEFAULT_TAG, ScrollAreaImplP
...
+      stateContext.domSizes.position.width,
+      stateContext.domSizes.viewport.height,
+      stateContext.domSizes.viewport.width,
+    ]
+  );
 
-    return (
+  return (
+    <div
+      {...(selector ? getSelectorObj(selector + '-position') : undefined)}
```
benoitgrelard, line 626: 👍 Seeing this now, can we also do the same for the popper wrapper then? Then my PR isn't needed either.

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -282,7 +283,8 @@ const ScrollAreaNative = forwardRefWithAs<typeof ROOT_DEFAULT_TAG, ScrollAreaNat
     const ref = useComposedRefs(scrollAreaRef, forwardedRef);
 
     return (
-      <Comp
+      <Primitive
+        selector={getSelector(ROOT_NAME)}
         ref={ref}
```
benoitgrelard, line 288: nit: ref after spread

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -466,7 +467,8 @@ const ScrollAreaImpl = forwardRefWithAs<typeof ROOT_DEFAULT_TAG, ScrollAreaImplP
         <ScrollAreaRefsContext.Provider value={refs}>
           <ScrollAreaContext.Provider value={context}>
             <ScrollAreaStateContext.Provider value={reducerState}>
-              <Comp
+              <Primitive
+                selector={getSelector(ROOT_NAME)}
                 {...domProps}
                 ref={ref}
```
benoitgrelard, line 473: nit: ref after spread

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -490,191 +492,191 @@ const ScrollAreaImpl = forwardRefWithAs<typeof ROOT_DEFAULT_TAG, ScrollAreaImplP
...
-            // https://blog.alexandergottlieb.com/overflow-scroll-and-the-right-padding-problem-a-css-only-solution-6d442915b3f4
-            display: 'table',
-            width: '100%',
-            paddingBottom: `var(${SCROLL_AREA_CSS_PROPS.scrollbarXOffset})`,
-            paddingRight: `var(${SCROLL_AREA_CSS_PROPS.scrollbarYOffset})`,
-          }}
-        >
-          <Comp {...getPartDataAttrObj(VIEWPORT_NAME)} ref={ref} {...domProps} />
-        </div>
+        <Primitive selector={selector} ref={ref} {...domProps} />
```
benoitgrelard, line 654: nit: ref after spread

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -490,191 +492,191 @@ const ScrollAreaImpl = forwardRefWithAs<typeof ROOT_DEFAULT_TAG, ScrollAreaImplP
...
+  );
+});
 
-const ScrollAreaViewport = forwardRefWithAs<typeof VIEWPORT_DEFAULT_TAG>(
+const ScrollAreaViewport = forwardRefWithAs<typeof ScrollAreaViewportImpl>(
   function ScrollAreaViewport(props, forwardedRef) {
-    const { as: Comp = VIEWPORT_DEFAULT_TAG, ...domProps } = props;
     return useNativeScrollArea() ? (
-      <Comp {...getPartDataAttrObj(VIEWPORT_NAME)} ref={forwardedRef} {...domProps} />
+      <Primitive selector={getSelector(VIEWPORT_NAME)} ref={forwardedRef} {...props} />
```
benoitgrelard, line 663: nit: ref after spread

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -490,191 +492,191 @@ const ScrollAreaImpl = forwardRefWithAs<typeof ROOT_DEFAULT_TAG, ScrollAreaImplP
...
     ) : (
-      <ScrollAreaViewportImpl ref={forwardedRef} as={Comp} {...domProps} />
+      <ScrollAreaViewportImpl ref={forwardedRef} {...props} />
```
benoitgrelard, line 665: nit: ref after spread

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -1460,34 +1445,45 @@ const ScrollAreaButton = forwardRefWithAs<typeof BUTTON_DEFAULT_TAG, ScrollAreaB
...
+  typeof ScrollAreaButton,
+  ScrollAreaButtonStartOwnProps
+>(function ScrollAreaButtonStart(props, forwardedRef) {
+  return (
+    <ScrollAreaButton
+      selector={getSelector(BUTTON_START_NAME)}
+      {...props}
+      name={BUTTON_START_NAME}
+      direction="start"
+      ref={forwardedRef}
```
benoitgrelard, line 1467: nit: ref after spread

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -1460,34 +1445,45 @@ const ScrollAreaButton = forwardRefWithAs<typeof BUTTON_DEFAULT_TAG, ScrollAreaB
...
+const ScrollAreaButtonEnd = forwardRefWithAs<typeof ScrollAreaButton, ScrollAreaButtonEndOwnProps>(
   function ScrollAreaButtonEnd(props, forwardedRef) {
     return (
       <ScrollAreaButton
+        selector={getSelector(BUTTON_END_NAME)}
         {...props}
         name={BUTTON_END_NAME}
         direction="end"
-        {...getPartDataAttrObj(BUTTON_END_NAME)}
         ref={forwardedRef}
```
benoitgrelard, line 1487: nit: ref after spread

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -1500,80 +1496,78 @@ ScrollAreaButtonEnd.displayName = BUTTON_END_NAME;
...
-const ScrollAreaCorner = forwardRefWithAs<typeof CORNER_DEFAULT_TAG, ScrollAreaCornerProps>(
-  function ScrollAreaCorner(props, forwardedRef) {
-    return useNativeScrollArea() ? null : <ScrollAreaCornerImpl ref={forwardedRef} {...props} />;
-  }
-);
+const ScrollAreaCorner = forwardRefWithAs<typeof ScrollAreaCornerImpl>(function ScrollAreaCorner(
+  props,
+  forwardedRef
+) {
+  return useNativeScrollArea() ? null : <ScrollAreaCornerImpl ref={forwardedRef} {...props} />;
```
benoitgrelard, line 1569: nit: ref after spread

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -256,7 +251,7 @@ const TooltipContentImpl = forwardRefWithAs<typeof PopperPrimitive.Root, Tooltip
 
 TooltipContent.displayName = CONTENT_NAME;
 
-/* ------------------------------------------------------------------------------------------------*/
+/* ---------------------------------------------------------------------------------------------- */
```
benoitgrelard, line 254: unintended change?

benoitgrelard (commented):
This is AWESOME! ♥️ 

Don't be scared by the number of comments, it's mostly a bunch of really tiny questionable nits 😀

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -130,28 +134,24 @@ type CheckboxIndicatorOwnProps = {
...
-);
+const CheckboxIndicator = forwardRefWithAs<
+  typeof INDICATOR_DEFAULT_TAG,
+  MergeOwnProps<typeof Primitive, CheckboxIndicatorOwnProps>
+>((props, forwardedRef) => {
+  const { forceMount, ...indicatorProps } = props;
+  const checked = useCheckboxContext(INDICATOR_NAME);
+  return (
+    <Presence present={forceMount || checked === 'indeterminate' || checked === true}>
+      <Primitive
```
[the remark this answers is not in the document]
jjenzz: Yeah, that was a neat idea but tbh, the abstraction here actually confused me when I was doing this because it made me think something more complex was going on. I spent some time trying to figure out what that was.

If there is no need for that additional mental computation I think it's helpful to just avoid it by keeping it inline till we know the abstraction is needed.

We just need to make sure we keep on top of that in PR reviews I guess 😬

packages/react/dialog/src/Dialog.tsx
```diff
@@ -125,14 +128,11 @@ const DialogOverlay = forwardRefWithAs<typeof DialogOverlayImpl, DialogOverlayOw
...
 
-const DialogOverlayImpl = forwardRefWithAs<typeof OVERLAY_DEFAULT_TAG>((props, forwardedRef) => {
-  const { as: Comp = OVERLAY_DEFAULT_TAG, ...overlayProps } = props;
-  return (
-    <Portal>
-      <Comp {...getPartDataAttrObj(OVERLAY_NAME)} ref={forwardedRef} {...overlayProps} />
-    </Portal>
-  );
-});
+const DialogOverlayImpl = forwardRefWithAs<typeof Primitive>((props, forwardedRef) => (
```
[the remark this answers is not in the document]
jjenzz: I didn't do it here because this uses `Portal` which may have its own hooks and things internally that would need to "reset" on mount/unmount (e.g. the `forceUpdate()`). 

I only removed the abstraction when it was a component that just rendered `Primitive` because those don't do anything.

packages/react/menu/src/Menu.tsx
```diff
@@ -351,111 +348,115 @@ MenuLabel.displayName = LABEL_NAME;
...
 type MenuItemOwnProps = {
   disabled?: boolean;
   textValue?: string;
+  /**
+   * A string to use as the component selector for CSS purposes. It will be added as
+   * a data attribute. Pass `null` to remove selector.
+   *
+   * @defaultValue radix-menu-item
+   */
+  selector?: string | null;
```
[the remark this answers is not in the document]
jjenzz: Oops, I meant to remove this! Good spot!

packages/react/scroll-area/src/ScrollArea.tsx
jjenzz: Oooh good catch. I shouldn't have removed `selector` here. They should be able to pass `selector` and `as`.

packages/react/scroll-area/src/ScrollArea.tsx
jjenzz: We shouldn't be lol... will add it back in.

packages/react/scroll-area/src/ScrollArea.tsx
jjenzz: this one is already but will do the others :+1:

packages/react/checkbox/src/Checkbox.tsx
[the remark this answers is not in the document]
benoitgrelard: Yeah I was thinking the same. I feel like it's just something that is pretty easy to catch or eventually realise that we're doing too much work in a component that's always rendered, and in this case remediate. Rather than premeditate the changes.

packages/react/scroll-area/src/ScrollArea.tsx
benoitgrelard: haha 🤦‍♂️ got too trigger happy there 😀

benoitgrelard (approved):
All looks great to me! Thanks for the changes! ♥️

packages/react/checkbox/src/Checkbox.tsx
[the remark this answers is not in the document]
jjenzz: 💯 :+1:

jjenzz:
Whoooop 🎉

PR #349, opened 2020-12-17
Generate GitHub Release [step 1]

chaance:
This PR will introduce the first step in keeping changelogs via GitHub releases. it uses [this action](https://github.com/marketplace/actions/generate-changelog-action) as suggested by @jjenzz.

With this addition, here's what needs to happen:
- In every PR, any change affecting a package will need to be documented in `RELEASE.md`. We will ask contributors to add this for now before merging a PR.
- When releasing a new version, we'll need to add the correct version to the title in `RELEASE.md`, then create a version tag (e.g., `v0.2.1`) and push that tag to GH. That tag will trigger the action, which creates the GH release with the content from `RELEASE.md`.
- After the release is done, we can run `yarn create-release-notes` to reset the release notes to blank for the next release.

This is intended as an incremental step. As a next step I'd like to use tags and PR titles to populate `RELEASE.md` after merging so we can remove that step for contributors.

I tested this briefly with a dummy commit just to make sure the action works as expected:

image

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,6 +5,7 @@
 - [ ] Add or edit tests to reflect the change (run `yarn test`).
 - [ ] Add or edit Storybook examples to reflect the change (run `yarn start`).
 - [ ] Add documentation to support any new features.
+- [ ] Add a brief overview of the change in `RELEASE.md`.
 
 This pull request:
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,8 +2,11 @@ name: Release
 
 on:
   push:
-    tags:
-      - 'v[0-9]*'
+    branches:
+      - changelogs
+      - 'releases/*'
+    # tags:
+    #   - 'v*'
 
 jobs:
   build:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,25 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v[0-9]*'
+
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          body_path: ./current-release.md
+          draft: false
+          prerelease: false
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,11 +2,8 @@ name: Release
 
 on:
   push:
-    branches:
-      - changelogs
-      - 'releases/*'
-    # tags:
-    #   - 'v*'
+    tags:
+      - 'v*'
 
 jobs:
   build:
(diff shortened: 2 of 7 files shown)


chaance:
Note: This approach probably needs to be tweaked if we're planning on versioning packages independently.

jjenzz:
> Note: This approach probably needs to be tweaked if we're planning on versioning packages independently.

Ah, we are versioning packages independently right? 😬

jjenzz:
> This is intended as an incremental step. As a next step I'd like to use tags and PR titles to populate `RELEASE.md` after merging so we can remove that step for contributors.

Love this approach btw :+1:

chaance:
> Ah, we are versioning packages independently right? 😬

I should clarify: are we planning on releasing independently?

Here's where it gets complicated: What if `Collapsible` gets a breaking API change that doesn't impact the API for `Accordion`, but `Accordion` still needs to update internally to use the new `Collapsible` API? Does `Collapsible` get a major bump but `Accordion` sticks with the same major? 

If so, I don't think we can use GitHub releases because it sticks a release tag on the repo as a whole. I think that's fine if some packages just don't get a new version when a release goes out, but if we strictly follow semver on a package-by-package basis then we'll probably need a different strategy.

benoitgrelard:
> > Ah, we are versioning packages independently right? 😬
> 
> I should clarify: are we planning on releasing independently?
> 
> Here's where it gets complicated: What if `Collapsible` gets a breaking API change that doesn't impact the API for `Accordion`, but `Accordion` still needs to update internally to use the new `Collapsible` API? Does `Collapsible` get a major bump but `Accordion` sticks with the same major?
>

This is where I love the yarn thing because we have full control over that granular decision when we do the version check. It's entirely at our discretion. For example in the past, I added a new export in utils, and used it in a package. yarn will flag all the components using utils as a dep, but I, as a human can safely decide to exclude all the packages that I know don't use this new export.

> If so, I don't think we can use GitHub releases because it sticks a release tag on the repo as a whole. I think that's fine if some packages just don't get a new version when a release goes out, but if we strictly follow semver on a package-by-package basis then we'll probably need a different strategy.

Yeah not sure github tags make much sense for this to be honest as they're attached to the repo as a whole. I remember what lerna used to do is they'd create tags with the name and version of each components otherwise…

jjenzz:
Hmm yeah this is an interesting one. So as I understand it, we want to:

- automate releases
- have a changelog that's updated through some sort of PR flow 
- version packages independently 

Thinking about this more I'm not sure if we need Github releases.

Microsoft have a monorepo manager called [Rush](https://rushjs.io/). It looks pretty much the same as what we're doing with Yarn 2, except it allows you to include a changelog description in the versioning file you commit. It looks like [this](https://github.com/microsoft/rushstack/pull/2399/files#diff-a604c12535928fdf5b1f6a46edd726f46a9546e69d46d61f3fc7e4031fa554d3) and is automatically generated when you answer a bunch of questions prompted by `rush change`. 

The `rush publish` command curates these to form a CHANGELOG.md **in each package** and also appears to manage versions of dependent packages (would need to test):

> It does quite a few things internally to make it happen: gather all change files to figure out what kind of version increase is needed, what packages need to have version increase, **increase the versions of dependencies**, clean up change files, and so on

This all sounds promising but do we need one overarching `CHANGELOG.md` that lists the changes to all packages also? This part confuses me because our monorepo isn't a package, it isn't versioned and released... the individual packages are.

I was thinking individual `CHANGELOG`s are fine with a Github Action that triggers a release from a PR. I dunno though?

benoitgrelard:
Closing this for now as we haven't needed it yet.

PR #353, opened 2020-12-21
Speed up pre-commit hooks with lint-staged

chaance:

This pull request:

- [x] Other


benoitgrelard:
@chaance So the main speed-up is from the fact that we're only going to re-lint / typecheck the files that are being staged rather than the whole codebase?

benoitgrelard (commented):
Looks good!

One small nit: Could we move this into the `"lint-staged"` key in the `package.json` to minimise the proliferation of config files at the root?

lint-staged.config.js
```diff
@@ -0,0 +1,7 @@
+const lint = 'eslint --quiet --fix --cache';
+
+// https://github.com/okonet/lint-staged/issues/825#issuecomment-620018284
+module.exports = {
+  '*.{js}': [lint],
+  '*.{ts,tsx}': [lint, () => 'tsc --skipLibCheck --noEmit'],
+};
```
jjenzz, line 7: I believe we're able to put this on `package.JSON` right? Just thinking if configs are small perhaps they're fine there? We have others there at least.

jjenzz (changes requested):
Love this. Been meaning to chat about this. Thoughts on moving it to push rather than commit too?

chaance:
@benoitgrelard See the comment I left in the config file; in order for `tsc` to work properly with `lint-staged` (and for our scripts to run cross platform) one of those keys we have to use the function syntax in a js config file. I kind of hate it too but I couldn't get it to work in package.json without running bash in the command, which wouldn't work on Windows and would exclude contributors. 😕

chaance:
> Love this. Been meaning to chat about this. Thoughts on moving it to push rather than commit too? 

Even better IMO!

lint-staged.config.js
benoitgrelard: haha I left exactly the same comment, but it seems there's an issue with that, hence why the comment above (I missed it too)

benoitgrelard:
> > Love this. Been meaning to chat about this. Thoughts on moving it to push rather than commit too?
> 
> Even better IMO!

We can't really move this to pre-push hook right? otherwise how would the `--fix` work?

chaance:
@benoitgrelard I'm thinking we could keep `prettier` and `eslint` in pre-commit. Those actually run pretty fast anyway when only executed on staged changes. `tsc` is significantly slower and I think we can move that to pre-push.

chaance:
@benoitgrelard @jjenzz Updated a bit. Per my comment above I kept `prretty-quick` and `eslint` in the pre-commit hook, runs only on staged files and is way faster now. `tsc` goes in pre-push because it was much slower, and that let me delete the config file and keep everything in `package.json`.

package.json
```diff
@@ -9,6 +9,7 @@
   ],
   "scripts": {
     "lint": "eslint packages --ext .js,.ts,.tsx",
+    "types:check": "tsc --skipLibCheck --noEmit",
```
chaance, line 12: Putting the script here instead of running `tsc` directly from husky is what let me delete the config file with function syntax.

benoitgrelard (approved):
Even better!

jjenzz:
🎉 :+1:

PR #360, opened 2020-12-22
Build with Preconstruct

chaance:
This PR explores using [Preconstruct](https://preconstruct.tools/) in place of Parcel for building packages. I'm leaving this is draft mode for now because I'm not even 100% sure if it's a net win for us, but I wanted to open it for discussion. Here's my initial thoughts:

- Preconstruct is specifically designed for monorepos, whereas Parcel is not. This is evident to me in improved DX on the command line + optimized build speed. This build runs super fast compared to running Parcel independently on every package. This was also dead simple to set up and use. The few error messages I've seen have been far more readable and actionable so far.
- Preconstruct uses Rollup under the hood, which is simpler and less magical than Parcel. It uses Babel as well but doesn't pre-configure anything there, so that's now exposed to us directly.
- Hopefully we never see a Yarn dependency graph issue again. 🤞

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-  "presets": ["@babel/preset-typescript"]
-}
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-typescript"]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ node_modules
 dist
 test-output.*
 .cache
-.parcel-cache
 storybook-static
 .env
 .eslintcache
--- a/.yarn/versions/dcda9e1d.yml
+++ b/.yarn/versions/dcda9e1d.yml
@@ -1,0 +1,41 @@
+releases:
+  "@radix-ui/popper": patch
+  "@radix-ui/react-accessible-icon": patch
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-announce": patch
+  "@radix-ui/react-arrow": patch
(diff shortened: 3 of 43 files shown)


jjenzz:
> whereas Parcel is not

It does AFAIK:

https://twitter.com/devongovett/status/1181306747283103745?s=21

I think our pain with it might be because we are using `yarn workspaces foreach` to run `parcel build` for each package instead of passing a wildcard to a parcel build at root (my bad).

I was also interested in seeing the comparison with Preconstruct though so thanks for doing this 👍

It would be good to compare this with building the monorepo using Parcel the way they intend it (which might fix our dependency graph issue). 

I've also been getting frustrated with Parcel lately so I'm by no means attached to it but just want to make sure it's not because we're using it wrong haha.

chaance:
> It does AFAIK:
> 
> https://twitter.com/devongovett/status/1181306747283103745?s=21

Oh this is nice! Good find. Definitely tough to find these things when you're running pre-release software with imcomplete documentation 😅 

> It would be good to compare this with building the monorepo using Parcel the way they intend it (which might fix our dependency graph issue).
> 
> I've also been getting frustrated with Parcel lately so I'm by no means attached to it but just want to make sure it's not because we're using it wrong haha.

Totally agreed, I can try to make some tweaks to our existing build to see if we gain anything here or if it's just a horizontal move.

chaance:
I think we should go ahead and consider merging https://github.com/radix-ui/primitives/pull/361 and consider this PR separately after the fact. There may be no need to switch from Parcel after all, but I would like to circle back to this one for discussion when @jjenzz is back.

chaance:
Closing this for now, seems like we're all pretty happy to stick with Parcel for the time being 👍

PR #361, opened 2020-12-22
Build with Parcel monorepo script

chaance:
This PR is a potential alternative to https://github.com/radix-ui/primitives/pull/360, or simply an incremental change we could merge while considering the merits of Preconstruct vs. Parcel.

At any rate, this uses Parcel's built-in API for building monorepo packages.

Some notes on my experience here:
- Parcel's error handling leaves a lot to be desired, which is probably to be expected since we're still using a beta release here. But it took much longer to work through the errors to get this working than it did with Preconstruct, which has some pretty smart tools for dealing with issues built in (as well as human-readable messages all along the way).
- ~One of Parcel's transformers had a type error that was breaking the build at first. I had to fix it using `yarn patch`. Gonna try to replicate that and submit a PR upstream later.~ Looks like this was eventually already fixed, so I pinned to the nightly version we were already using accordion to our lockfile. Build is passing now without my patch.
- Ultimately this speeds up our build quite a bit, which is super nice. Couldn't tell a huge difference with speed of Parcel vs. Preconstruct when using this approach.


Makefile
```diff
@@ -0,0 +1,2 @@
+build:
+	parcel build packages/*/*/
```
chaance, line 2: This is a bit of a hack, but if you put this in the package script, Yarn borks because it can't deal with the glob pattern. Unsure why.

chaance: That said, I think make is kind of nice for composing scripts in a more readable format than all of the package.json noise. I wouldn't complain if we decided to use it more 🙂

chaance:
@benoitgrelard Think you could take a peek here when you have a sec? This isn't much of an overhaul to our current build, but it's more inline with how Parcel is intended to be used with monorepos and it should simplify things a bit (hopefully squashing those annoying dependency graph errors as a bonus).

Makefile
jjenzz: I've just had a look at this and it's zsh that throws on zero-match glob patterns. The trick is to quote the glob so that it's actually **parcel** that expands the glob instead of your shell.

```jsx
"build": "parcel build 'packages/*/*/'",
```

package.json
```diff
@@ -15,10 +14,7 @@
     "storybook": "start-storybook -p 9009 --ci",
     "dev": "yarn storybook",
     "build-storybook": "build-storybook",
-    "build": "yarn build:config && yarn build:packages && yarn build:cleanup",
-    "build:config": "mv tsconfig.json tsconfig.tmp.json && mv tsconfig.production.json tsconfig.json",
```
jjenzz, line 19: Unfortunately I think we still need this `tsconfig.json` dance because of https://github.com/parcel-bundler/parcel/issues/1251#issuecomment-701546994

I've just tested this PR locally and the issue is happening again:

image

A bit frustrating that we have to add this workaround for now 🙁

jjenzz (changes requested):
Couple of comments but this is SO much quicker. LOVE it 🎉

package.json
chaance: Ah bummer, good catch though. I'll revert this and add a comment somewhere to note for future reference.

.yarn/versions/970578a9.yml
```diff
@@ -0,0 +1,41 @@
...
+  "@radix-ui/react-scroll-area": patch
+  "@radix-ui/react-separator": patch
+  "@radix-ui/react-slider": patch
+  "@radix-ui/react-switch": patch
+  "@radix-ui/react-tabs": patch
+  "@radix-ui/react-toggle-button": patch
+  "@radix-ui/react-tooltip": patch
+  "@radix-ui/react-utils": patch
+  "@radix-ui/react-visually-hidden": patch
+  "@radix-ui/utils": patch
```
benoitgrelard, line 38: I think we can decline all those no? There aren't any change in the packages to be published.

benoitgrelard (commented):
🤯🤯🤯 This is out of this world! From 3min to 14seconds on my MacBook Pro 16".
Bonus point for the fact that the fans don't kick off at all now 😀

Just one small comment on the version check.

.yarn/versions/970578a9.yml
chaance: @benoitgrelard I went with patch to be safe, as it's not entirely safe to assume that the output code won't be different from before since we're pinning to a different version of Parcel. Unsure if it ultimately matter in this case as I imagine we'll be rolling out a new version of all primitives for the next release (I'd like to add `babel-plugin-dev-expressions` in a follow-up PR to strip some stuff that will almost certainly touch every package anyway).

benoitgrelard: ha I see, let's roll with it then!

benoitgrelard (approved):

devongovett:
Hey, nice to see you're using Parcel here! 😍 

> Parcel's error handling leaves a lot to be desired

Do you have some details about what issues you ran into? Would love to fix them! 😄

benoitgrelard:
Hey @devongovett, nice seeing you here!
I'm not too sure which errors @chaance encountered in the process, but at the very least we are much happier since we've switched to the monorepo build, it's miles faster! (see here: https://github.com/radix-ui/primitives/pull/361#pullrequestreview-565745208)

chaance:
@devongovett 👋 I'd say overall it wasn't that there were any problems to fix, just user errors on my part that could have likely been anticipated, caught, and handled in some elegant way. It wasn't to suggest that anything is broken on your part, it was more because I had just gone through the same exercise with Preconstruct which has some really nice DX touches when you are *doing it wrong*.

When using tools like Parcel that are supposed to "just work", it can be frustrating to get stack trace errors pointing back to the source. Happy to follow up again with something more specific when it happens again 🙂.

benoitgrelard:
Hey @devongovett, we've recently hit an issue (#406) where our published ts types were seemingly wrong.
Turns out there's already an open bug for it: https://github.com/parcel-bundler/parcel/issues/5244

Do you know of plans to fix this?

We used to use the cache, then had issues, then thought we'd turn it back one with newer releases of parcel, but it turns out there are still issues.

All this makes it a bit hard to trust so for now we've turned it off again.

PR #364, opened 2020-12-23
Test: Avatar

chaance:
- [x] Add or edit tests to reflect the change (run `yarn test`).

This pull request:

- [x] Other

### DIFF
--- a/packages/react/avatar/src/Avatar.test.tsx
+++ b/packages/react/avatar/src/Avatar.test.tsx
@@ -7,6 +7,7 @@ import * as Avatar from './Avatar';
 
 const ROOT_TEST_ID = 'avatar-root';
 const FALLBACK_TEXT = 'AB';
+const IMAGE_ALT_TEXT = 'Fake Avatar';
 const DELAY = 300;
 
 describe('given an Avatar with fallback and no image', () => {
@@ -41,8 +42,9 @@ describe('given an Avatar with fallback and no image', () => {
 
 describe('given an Avatar with fallback and a working image', () => {
   let rendered: RenderResult;
+  let image: HTMLElement | null = null;
+  const orignalGlobalImage = window.Image;
 
-  const orignalImage = window.Image;
   beforeAll(() => {
     jest.useFakeTimers();
     (window.Image as any) = class MockImage {
@@ -59,14 +61,14 @@ describe('given an Avatar with fallback and a working image', () => {
 
   afterAll(() => {
     jest.useRealTimers();
-    window.Image = orignalImage;
+    window.Image = orignalGlobalImage;
   });
 
   beforeEach(() => {
     rendered = render(
       <Avatar.Root data-testid={ROOT_TEST_ID}>
         <Avatar.Fallback>{FALLBACK_TEXT}</Avatar.Fallback>
-        <Avatar.Image data-testid="tst-123">{FALLBACK_TEXT}</Avatar.Image>
+        <Avatar.Image src="/test.jpg" alt={IMAGE_ALT_TEXT} />
       </Avatar.Root>
     );
   });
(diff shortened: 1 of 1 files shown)


packages/react/avatar/src/Avatar.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  // TODO: MockImage isn't working as expected. Come back and figure out why.
+  // it('should render the image after it has loaded', async () => {
+  //   let image: HTMLElement | null = null;
+  //   await waitFor(() => {
+  //     jest.advanceTimersByTime(DELAY);
+  //   });
+  //   image = rendered.queryByTestId('tst-123');
+  //   expect(image).toBeInTheDocument();
+  // });
```
anuraghazra, line 92: I was debugging this for a bit and found out that the onload of the `useImageLoadingStatus` function isn't even getting called because in the test we are not providing "src" prop to Avatar.Image, and looking at the logic of `useImageLoadingStatus` it will just return from the useEffect if src isn't passed. 

https://github.com/radix-ui/primitives/blob/8663e7602f60a925fce34bfed3ff032c08ac2e06/packages/react/avatar/src/Avatar.tsx#L108-L111

packages/react/avatar/src/Avatar.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  beforeEach(() => {
+    rendered = render(
+      <Avatar.Root data-testid={ROOT_TEST_ID}>
+        <Avatar.Fallback>{FALLBACK_TEXT}</Avatar.Fallback>
+        <Avatar.Image data-testid="tst-123">{FALLBACK_TEXT}</Avatar.Image>
+      </Avatar.Root>
+    );
+  });
```
anuraghazra, line 72: 🤔 img tags can't have children.

I think this should be :-

```jsx
  beforeEach(() => {
    rendered = render(
      <Avatar.Root data-testid={ROOT_TEST_ID}>
        <Avatar.Fallback>{FALLBACK_TEXT}</Avatar.Fallback>
        <Avatar.Image src="https://img.png" data-testid="tst-123" />
      </Avatar.Root>
    );
  });
```

anuraghazra:
Doing the above changes i was able to get the test passing. 

```jsx
  it('should render the image after it has loaded', async () => {
    await waitFor(() => {
      jest.advanceTimersByTime(DELAY);
    });

    expect(rendered.getByTestId('tst-123')).toBeInTheDocument();
  });
```

packages/react/avatar/src/Avatar.test.tsx
chaance: Great find, thank you for catching that!

packages/react/avatar/src/Avatar.test.tsx
chaance: Oversight on my end 😅 thanks for catching!

anuraghazra (approved):

packages/react/avatar/src/Avatar.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+  it('should render a fallback after the delay', async () => {
+    await waitFor(() => {
+      jest.advanceTimersByTime(DELAY);
+      fallback = rendered.queryByText(FALLBACK_TEXT);
+    });
+    expect(fallback).toBeInTheDocument();
+  });
+});
+
+export default EventTarget;
```
benoitgrelard, line 146: Is this needed? I can't see what it's here for.

benoitgrelard (commented):
All LGTM, just one small comment.

packages/react/avatar/src/Avatar.test.tsx
chaance: Nope, I honestly don't know why this is here but I'll pull it out real quick 😅

benoitgrelard (approved):

PR #369, opened 2020-12-23
FocusScope blur fix

### LINKED ISSUE #365: [focus-scope] Last element inside a Popover is not triggering blur event
### Current Behavior

If I have a Popover open with a couple input elements, the last input element never triggers the `blur` event - seems to be prevented by the [focus-scope](https://github.com/radix-ui/primitives/blob/main/packages/react/focus-scope/src/createFocusScope.tsx#L44-L48).

### Expected behavior

I expect the last element to fire the `blur` event.

### Reproducible example

[CodeSandbox example](https://codesandbox.io/s/hardcore-surf-9oser?file=/src/App.tsx)

https://user-images.githubusercontent.com/76673/103000150-3ae46180-4509-11eb-8d73-df7f9e61d9c8.mp4

### Your environment

| Radix Package(s) | @radix-ui/react-popover         | 0.0.1        |
| React            | n/a     | 17         |
| Browser          |         | Any browser |

chaance:
Fixes #365. I tested across all our stories that use `FocusScope` under the hood and, as far as I can tell, everything still works as expected. @benoitgrelard will definitely want to have a look at this as he may be aware of some edge cases I'm not as it relates to the changes made in the `focusout` handler.

- [x] Add or edit tests to reflect the change (run `yarn test`).
- [x] Add or edit Storybook examples to reflect the change (run `yarn dev`).

This pull request:

- [x] Fixes a bug in an existing package
- [ ] Other

### DIFF
--- a/packages/react/focus-scope/src/FocusScope.stories.tsx
+++ b/packages/react/focus-scope/src/FocusScope.stories.tsx
@@ -1,13 +1,35 @@
 import React from 'react';
 import { FocusScope } from './FocusScope';
+import { composeEventHandlers } from '@radix-ui/react-utils';
 
 export default { title: 'Components/FocusScope' };
 
 export const Basic = () => {
   const [trapped, setTrapped] = React.useState(false);
+  const [highlightAction, setHighlightAction] = React.useState<'null' | 'blur' | 'focus'>('null');
+  const InputWrapper =
+    highlightAction === 'blur'
+      ? BlurAnnouncement
+      : highlightAction === 'focus'
+      ? FocusAnnouncement
+      : React.Fragment;
 
   return (
     <>
+      <RadioGroup
+        name="highlight"
+        checked={highlightAction}
+        onChange={(newValue) => {
+          setHighlightAction(newValue as any);
+        }}
+      >
+        <Legend>Highlight focus changes</Legend>
+        <Radio value="null">Never</Radio>
+        <Radio value="blur">On Blur</Radio>
+        <Radio value="focus">On Focus</Radio>
+      </RadioGroup>
+      <hr />
+
       <div>
         <button type="button" onClick={() => setTrapped(true)}>
           Trap
@@ -29,12 +51,20 @@ export const Basic = () => {
                 border: '2px solid',
               }}
             >
-              <input type="text" placeholder="First name" />
-              <input type="text" placeholder="Last name" />
-              <input type="number" placeholder="Age" />
-              <button type="button" onClick={() => setTrapped(false)}>
-                Close
-              </button>
+              <InputWrapper>
+                <input type="text" placeholder="First name" />
+              </InputWrapper>
+              <InputWrapper>
+                <input type="text" placeholder="Last name" />
+              </InputWrapper>
+              <InputWrapper>
+                <input type="number" placeholder="Age" />
+              </InputWrapper>
+              <InputWrapper>
+                <button type="button" onClick={() => setTrapped(false)}>
+                  Close
+                </button>
+              </InputWrapper>
             </form>
           )}
         </FocusScope>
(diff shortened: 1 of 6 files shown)


chaance:
FWIW I also spent about 2 hours trying to write tests for this thing and started to pull my hair out, because the fake DOM is my worst enemy. I'll continue those in a separate branch and sync up with someone after holibobz.

chaance:
Update: fixed the tests, just needed to mock rAF.

benoitgrelard:
Hmm, yeah I'm going to need to dig into this quite thoroughly tomorrow because I remember we were doing this specifically to prevent some other issue… 😅 

I'm with you on the tests, testing anything to do with focus is almost impossible as it's not real focus, etc.

chaance:
> Hmm, yeah I'm going to need to dig into this quite thoroughly tomorrow because I remember we were doing this specifically to prevent some other issue

If you can replicate, would be great to document the tradeoff so we can reference this somewhere. I know we fixed an issue by using `focusin` and `focusout` instead of `focus` and `blur`, but I haven't changed that here.

FWIW I tinkered with this in all of the packages that use FocusScope and didn't notice any of the usual edge cases, but feel free to give it a spin when you can!

packages/react/focus-scope/src/FocusScope.stories.tsx
```diff
@@ -264,3 +294,162 @@ export const WithOptions = () => {
...
+}
+
+function CheckedInput({ children, ...props }: React.ComponentProps<'input'>) {
+  return (
+    <label style={{ display: 'flex', marginTop: '0.25em' }}>
+      <input {...props} />
+      <span style={{ display: 'inline-block', marginLeft: '0.25em' }}>{children}</span>
+    </label>
+  );
+}
```
benoitgrelard, line 455: Not sure if you meant to leave this all here or if it was just for your own testing but I feel like we should remove this as it detracts massively from the story itself. If it's to check that other focus/blur events do happen properly I feel like an inline `onFocus`, `onBlur` handler on the things we want to check with a console.log is good enough and doesn't detract much from the story.

benoitgrelard (commented):
Finally got around to testing this, and I'm still unsure about it…

It makes sense that it solves the "new" issue because we've remove the `stopImmediatePropagation` (in the blur case)…

but…

We now have to use `requestAnimationFrame` to prevent the `useFocusOut` hook in `DismissableLayer` to trigger (because we're now not preventing blur).

So to me it all really feels like a replacing a hack with another hack to counter the issue that we've had from the start that these 2 things are related and have friction between each other.

I'd be keen to look a bit deeper at trying to figure out a proper fix because I can see that this will keep biting us…

What do you think?

packages/react/focus-scope/src/FocusScope.stories.tsx
chaance: Yeah we can totally remove it. Let's keep it until we settle on a proper fix for quick testing, but I'm glad to pull it out before we merge.

chaance:
> So to me it all really feels like a replacing a hack with another hack to counter the issue that we've had from the start that these 2 things are related and have friction between each other.
> 
> I'd be keen to look a bit deeper at trying to figure out a proper fix because I can see that this will keep biting us…
> 
> What do you think?

Yep, happy to put some more energy into it next week. I don't love timing hacks either, just couldn't land on anything better at the time. 

I'm going to pull these tests out and submit a separate PR, figure they're valuable either way!

jjenzz:
Do we know why the `stopImmediatePropagation` was there in the first place yet? What problem did it solve? I'm planning on taking a look at this at some point but wanna make sure I understand the whole story here.

benoitgrelard:
Yes, I can take you through it, it was something we did together.

benoitgrelard:
Closing this one as it was superseded by #395

PR #378, opened 2021-01-07
[react-utils]: Fix return value from useCallbackRef

chaance:

This pull request:

- [x] Fixes a bug in an existing package
- [ ] Other

### DIFF
--- a/packages/react/utils/src/useCallbackRef.tsx
+++ b/packages/react/utils/src/useCallbackRef.tsx
@@ -1,11 +1,10 @@
 import * as React from 'react';
 
 /**
- * A custom hook that converts a callback to a ref
- * to avoid triggering re-renders when passed as a prop
- * or avoid re-executing effects when passed as a dependency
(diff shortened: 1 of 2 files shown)


packages/react/utils/src/useCallbackRef.tsx
```diff
@@ -1,11 +1,10 @@
...
 
 /**
- * A custom hook that converts a callback to a ref
- * to avoid triggering re-renders when passed as a prop
- * or avoid re-executing effects when passed as a dependency
+ * A custom hook that converts a callback to a ref to avoid triggering re-renders when passed as a
+ * prop or avoid re-executing effects when passed as a dependency
  */
-export function useCallbackRef<T extends (...args: any[]) => any>(callback?: T): T {
+export function useCallbackRef<T extends (...args: any[]) => any>(callback: T | undefined): T {
```
chaance, line 7: I think this typing makes more sense, as we should always expect an argument here but it might be undefined.

benoitgrelard (approved):

Issue #351, opened 2020-12-18
[ScrollArea] Double-pass render breaks event handlers on nested children


chaance:
### Current Behavior

If you try to bind an event listener (or any other method, I guess) to a DOM ref for an element inside `ScrollArea`, it may not work as expected because the initially rendered DOM is discarded on the second pass if `ScrollArea` features are supported.

```tsx
function App() {
  const someRef = React.useRef(null);

  React.useEffect(() => {
    const someEl = someRef.current;
    const handleMouseEnter = () => console.log("mouse entered");
    const handleMouseDown =  () => console.log("mouse down");
    if (someEl) {
      someEl.addEventListener("mouseenter", handleMouseEnter);
      someEl.addEventListener("mousedown", handleMouseDown);
      return function() {
        someEl.removeEventListener("mouseenter", handleMouseEnter);
        someEl.removeEventListener("mousedown", handleMouseDown);
      }
    }
  }, []);

  return (
    <div className="App">
      <ScrollArea.Root>
        <ScrollArea.Viewport>
          <div ref={someRef}>
            Click me with the console open, you should see messages (but you won't)
          </div>
        </ScrollArea.Viewport>
      </ScrollArea.Root>
    </div>
  );
}
```

### Expected behavior

This should work as expected using the first code example above.

### Reproducible example

https://codesandbox.io/s/romantic-easley-h8cb9?file=/src/App.tsx

### Additional context

The current workaround is to use a callback ref and triggers a re-renders when the ref is placed, then re-call the effect.

```tsx
function App() {
  const someRef = React.useRef(null);
  const [element, setElement] = React.useState(null);

  React.useEffect(() => {
    const someEl = someRef.current;
    const handleMouseEnter = () => console.log("mouse entered");
    const handleMouseDown =  () => console.log("mouse down");
    if (someEl) {
      someEl.addEventListener("mouseenter", handleMouseEnter);
      someEl.addEventListener("mousedown", handleMouseDown);
      return function() {
        someEl.removeEventListener("mouseenter", handleMouseEnter);
        someEl.removeEventListener("mousedown", handleMouseDown);
      }
    }
  }, [element]);

  const someCallbackRef = React.useCallback((node) => {
    someRef.current = node;
    setElement(node)
  }, []);

  return (
    <div className="App">
      <ScrollArea.Root>
        <ScrollArea.Viewport>
          <div ref={someCallbackRef}>
            Click me with the console open, you should see messages (this should work, but it's super weird)
[...]

PR #379, opened 2021-01-07
ScrollArea: Temporary workaround for double render (#351)

chaance:
This PR exports a wrapper component that will bypass the double render caused by our `ScrollArea` implementation. This is intended as a stop-gap solution to an issue @StephenHaney is having with internal event listeners not running because we throw away the initially rendered DOM.

Long term I'm leaning towards removing the native fallback altogether and exposing a hook to detect compatibility instead, leaving it to the consumer to handle feature detection so they can wrap it however they'd like. That would let us get rid of the weird ref issues in the simplest way IMO. I'd like to wait for @jjenzz to get back to review and offer feedback as well.

- [x] Add or edit Storybook examples to reflect the change (run `yarn dev`).

This pull request:

- [x] Adds additional features/functionality to an existing package
- [ ] Other

### DIFF
--- a/packages/react/scroll-area/src/ScrollArea.tsx
+++ b/packages/react/scroll-area/src/ScrollArea.tsx
@@ -301,7 +301,7 @@ const ScrollAreaNative = forwardRefWithAs<typeof ROOT_DEFAULT_TAG, ScrollAreaNat
 );
 
 type ScrollAreaImplProps = ScrollAreaInternalProps &
-  Required<Omit<ScrollAreaOwnProps, 'unstable_forceNative'>>;
+  Omit<ScrollAreaOwnProps, 'unstable_forceNative'>;
 
 const initialSize = { width: 0, height: 0 };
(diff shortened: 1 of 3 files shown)


chaance:
cc @colmtuite

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -50,6 +50,14 @@ const SCROLL_AREA_CSS_PROPS_LIST = [
   'cornerHeight',
 ] as const;
 const SCROLL_AREA_CSS_PROPS = SCROLL_AREA_CSS_PROPS_LIST.reduce(reduceToCssProperties, {} as any);
+const AUTO = 'auto';
```
chaance, line 53: There are a lot of instances of these repeated strings, so I did this so that we might be able to save a few bytes here and there with minification.

benoitgrelard (approved):

PR #394, opened 2021-01-12
[Popover][DropdownMenu] Fix race condition when dismissing via click outside and `disableOutsidePointerEvents` is off. 

benoitgrelard:
I finally managed to fix this one! 🎉 

## Background

This was an issue that was reported by Lucas with regards to the `DropdownMenu`.
He was trying to turn off `disableOutsidePointerEvents` on `DropdownMenu.Content` so you can click through to things, and although it did work (for example button actions would be called), it wouldn't let him focus inside an input straight up.

I was sure I made this work with all the work I did on `DismissableLayer` and `FocusScope`, I even had some stories.

After reviewing this all with him, I realised that we had the same problem in `Popover` too.
But then I could see that my stories in `DismissableLayer.stories` which cover this specifically were working fine…

> **NOTE**: we have fake `Dialog` and fake `Popover` in this story file (basically crude versions of them to see how `DismissableLayer` and `FocusScope` interacted with each other).

So, considering the pieces were the same, there would have been some difference somewhere… 🤔

After a few days going through this, I realised the difference was `Presence` which is in our real components but not in these crude ones. and sure enough removing it fixed the issue.

So basically, because unmount has to be slightly delayed because of `Presence`, it gives us a race condition, which is basically that the focus trap is still active for a little bit longer than normal and so traps the focus when trying to focus the outside input.

## Solution

This solution I propose here is that I basically use the fact that we know when we want to `skipCloseAutoFocus`, and when that is the case, I disable the focus trap (effectively in proper time, rather than delayed unmount time). This works really well given that we have declarative access to whether the focus trap is active or not.

Let me know your thoughts on this, feels a little bit like a patch, but I think @jjenzz and I have been over this before and I don't think there would be any way for us to remove the delay from `Presence`.

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -136,7 +136,7 @@ const DropdownMenuContent = forwardRefWithAs<
       open={context.open}
       onOpenChange={context.setOpen}
       anchorRef={anchorRef || context.triggerRef}
-      trapFocus
+      trapFocus={skipCloseAutoFocus ? false : true}
       onCloseAutoFocus={(event) => {
         if (skipCloseAutoFocus) {
           event.preventDefault();
@@ -145,7 +145,7 @@ const DropdownMenuContent = forwardRefWithAs<
         }
       }}
       disableOutsidePointerEvents={disableOutsidePointerEvents}
-      onPointerDownOutside={(event) => {
+      onPointerDownOutside={composeEventHandlers(onPointerDownOutside, (event) => {
         const wasTrigger = context.triggerRef.current?.contains(event.target as HTMLElement);
 
         // skip autofocus on close if clicking outside is allowed and it happened
@@ -155,15 +155,13 @@ const DropdownMenuContent = forwardRefWithAs<
         // as it's already setup to close, otherwise it would close and immediately open.
         if (wasTrigger) {
           event.preventDefault();
-        } else {
-          onInteractOutside?.(event);
         }
 
         if (event.defaultPrevented) {
           // reset this because the event was prevented
           setSkipCloseAutoFocus(false);
         }
-      }}
+      })}
       onInteractOutside={onInteractOutside}
       disableOutsideScroll={disableOutsideScroll}
       portalled={portalled}
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -121,7 +121,6 @@ const DropdownMenuContent = forwardRefWithAs<
     ...contentProps
   } = props;
   const context = useDropdownMenuContext(CONTENT_NAME);
-  const [skipCloseAutoFocus, setSkipCloseAutoFocus] = React.useState(false);
   return (
     <MenuPrimitive.Root
       ref={forwardedRef}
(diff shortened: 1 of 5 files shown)


packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -145,7 +145,7 @@ const DropdownMenuContent = forwardRefWithAs<
         }
       }}
       disableOutsidePointerEvents={disableOutsidePointerEvents}
-      onPointerDownOutside={(event) => {
+      onPointerDownOutside={composeEventHandlers(onPointerDownOutside, (event) => {
```
benoitgrelard, line 148: Also realised we were not composing this one.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -155,15 +155,13 @@ const DropdownMenuContent = forwardRefWithAs<
         // as it's already setup to close, otherwise it would close and immediately open.
         if (wasTrigger) {
           event.preventDefault();
-        } else {
```
benoitgrelard, line 158: I also removed this as I don't think it makes sense, as that event is already called internally in all cases (focus outside, pointer down outside).

packages/react/popover/src/Popover.tsx
```diff
@@ -238,7 +238,7 @@ const PopoverContentImpl = forwardRefWithAs<typeof PopperPrimitive.Root, Popover
               <DismissableLayer
                 disableOutsidePointerEvents={disableOutsidePointerEvents}
                 onEscapeKeyDown={onEscapeKeyDown}
-                onPointerDownOutside={(event) => {
+                onPointerDownOutside={composeEventHandlers(onPointerDownOutside, (event) => {
```
benoitgrelard, line 241: Also realised we were not composing this one.

packages/react/popover/src/Popover.tsx
```diff
@@ -250,15 +250,13 @@ const PopoverContentImpl = forwardRefWithAs<typeof PopperPrimitive.Root, Popover
                   // as it's already setup to close, otherwise it would close and immediately open.
                   if (wasTrigger) {
                     event.preventDefault();
-                  } else {
```
benoitgrelard, line 253: I also removed this as I don't think it makes sense, as that event is already called internally in all cases (focus outside, pointer down outside).

packages/react/popover/src/Popover.tsx
```diff
@@ -224,7 +224,7 @@ const PopoverContentImpl = forwardRefWithAs<typeof PopperPrimitive.Root, Popover
       <PortalWrapper>
         <ScrollLockWrapper>
           <FocusScope
-            trapped={trapFocus}
+            trapped={skipCloseAutoFocus ? false : trapFocus}
```
jjenzz, line 227: can we rename `skipCloseAutoFocus` to something like `isInteractOutsideClose` or `isCloseOnInteractOutside` 😅 🙏  ... 

I'm trying to name it something that reads like:

```jsx
trapped={if closing via an outside interaction then don't trap, otherwise trap}
````

Seems to fit `onCloseAutoFocus` too:


```jsx
  if (isCloseOnInteractOutside) {
    event.preventDefault();
  } else {
    context.triggerRef.current?.focus();
  }
}}
```

I'm sure you can come up with something better but hopefully this explains what I'm trying to get at.

jjenzz (approved):
One small naming thing to help understand intent a bit better but otherwise :+1: love the way this uses the two component APIs to explicitly wire the expected behaviour.

benoitgrelard:
Heads up that I started working on the renaming stuff Jenna asked about and also re-jigging something a bit but I won't be able to get back to it until after 7pm GMT.

packages/react/menu/src/Menu.tsx
```diff
@@ -189,15 +194,39 @@ const MenuImpl = forwardRefWithAs<typeof PopperPrimitive.Root, MenuImplOwnProps>
...
           >
             {(focusScopeProps) => (
               <DismissableLayer
                 disableOutsidePointerEvents={disableOutsidePointerEvents}
                 onEscapeKeyDown={onEscapeKeyDown}
-                onPointerDownOutside={onPointerDownOutside}
+                onPointerDownOutside={composeEventHandlers(
+                  onPointerDownOutside,
+                  (event) => {
+                    const isNotRightClick =
```
jjenzz, line 217: `const isLeftClick` ? I always struggle with `isNot` because we have `!is` for that.

Also, if we type the event above this line will that prevent your wrapping here? 🙈 haha

```jsx
(event: MouseEvent) => {
	const isLeftClick = event.button === 0 && event.ctrlKey === false;
	...
})
```

jjenzz (approved):
a lil nit but again, all good from me :+1:

packages/react/menu/src/Menu.tsx
benoitgrelard: haha I know, same here!

Except in this case, it can be a `TouchEvent` and `event.button` doesn't exist on a `TouchEvent` (that's why I'm casting as `MouseEvent`).
So saying `isLeftClick` would not really be correct.
Whereas calling it `isNotRightClick` is actually correct in all cases.

Can you think of another way?

chaance (approved):

PR #396, opened 2021-01-13
[react-polymorphic] Replace `forwardRefWithAs` util with type

jjenzz:
A while ago, Benoit and I were looking into tree-shaking for our components. We realised that Parcel will automatically annotate `React.forwardRef` as `PURE` for us to tree-shake them however, we have replaced this with our own `forwardRefWithAs` utility that breaks tree-shaking.

We could create our own babel plugin to annotate these for us (which we looked into and worked well) however, on my quest to figure out how to apply props with generic types for `ToggleButtonGroup` it dawned on me that we could just cast our components to this type instead of using the util and we would get `PURE` annotation without any plugins.

I personally prefer this approach because the `forwardRefWithAs` utility implies functionality, but it is purely for types. This more explicitly communicates that.

**Note: I've only applied this to `Accordion` and `Collapsible` for now to discuss. If we're happy I'll apply across the codebase.**

- [X] Add or edit tests to reflect the change (run `yarn test`).

This pull request:

- [X] Fixes a bug in an existing package (tree-shaking)
- [ ] Other

### DIFF
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -7,11 +7,11 @@ import {
   useControlledState,
   useId,
 } from '@radix-ui/react-utils';
-import { forwardRefWithAs } from '@radix-ui/react-polymorphic';
 import { Primitive } from '@radix-ui/react-primitive';
 import { Collapsible, CollapsibleButton, CollapsibleContent } from '@radix-ui/react-collapsible';
 
-import type { OwnProps } from '@radix-ui/react-polymorphic';
+import type * as Polymorphic from '@radix-ui/react-polymorphic';
+import type { Merge } from '@radix-ui/utils';
 
 /* -------------------------------------------------------------------------------------------------
  * Root level context
@@ -35,21 +35,27 @@ const [AccordionContext, useAccordionContext] = createContext<AccordionContextVa
 
 const ITEM_NAME = 'AccordionItem';
 
-type AccordionItemOwnProps = {
-  /**
-   * Whether or not an accordion item is disabled from user interaction.
-   *
-   * @defaultValue false
-   */
-  disabled?: boolean;
-  /**
-   * A string value for the accordion item. All items within an accordion should use a unique value.
-   */
-  value: string;
-  open: never;
-  defaultOpen: never;
-  onOpenChange: never;
-};
+type AccordionItemOwnProps = Merge<
+  Omit<Polymorphic.OwnProps<typeof Collapsible>, 'open' | 'defaultOpen' | 'onOpenChange'>,
+  {
+    /**
+     * Whether or not an accordion item is disabled from user interaction.
+     *
+     * @defaultValue false
+     */
+    disabled?: boolean;
+    /**
+     * A string value for the accordion item. All items within an accordion should use a unique value.
+     */
+    value: string;
+  }
+>;
+
+type AccordionItemPrimitive = Polymorphic.ForwardRefComponent<
+  Polymorphic.IntrinsicElement<typeof Collapsible>,
+  AccordionItemOwnProps
+>;
+
 type AccordionItemContextValue = { open?: boolean; buttonId: string };
 
 const [AccordionItemContext, useAccordionItemContext] = createContext<AccordionItemContextValue>(
@@ -60,38 +66,34 @@ const [AccordionItemContext, useAccordionItemContext] = createContext<AccordionI
 /**
  * `AccordionItem` contains all of the parts of a collapsible section inside of an `Accordion`.
  */
-const AccordionItem = forwardRefWithAs<typeof Collapsible, AccordionItemOwnProps>(
-  (props, forwardedRef) => {
-    const { value, children, ...accordionItemProps } = props;
-    const accordionContext = useAccordionContext(ITEM_NAME);
-    const generatedButtonId = `accordion-button-${useId()}`;
-    const buttonId = props.id || generatedButtonId;
-    const open = (value && value === accordionContext.value) || false;
-    const disabled = accordionContext.disabled || props.disabled;
+const AccordionItem = React.forwardRef((props, forwardedRef) => {
+  const { value, children, ...accordionItemProps } = props;
+  const accordionContext = useAccordionContext(ITEM_NAME);
+  const generatedButtonId = `accordion-button-${useId()}`;
+  const buttonId = props.id || generatedButtonId;
+  const open = (value && value === accordionContext.value) || false;
+  const disabled = accordionContext.disabled || props.disabled;
 
-    const itemContext: AccordionItemContextValue = React.useMemo(() => ({ open, buttonId }), [
-      open,
-      buttonId,
-    ]);
+  const itemContext: AccordionItemContextValue = React.useMemo(() => ({ open, buttonId }), [
+    open,
+    buttonId,
+  ]);
 
-    return (
-      <Collapsible
-        selector={getSelector(ITEM_NAME)}
-        {...accordionItemProps}
-        ref={forwardedRef}
-        data-state={open ? 'open' : 'closed'}
-        data-disabled={disabled || undefined}
-        disabled={disabled}
-        open={open}
-        onOpenChange={() => accordionContext.setValue(value)}
-      >
-        <AccordionItemContext.Provider value={itemContext}>
-          {children}
-        </AccordionItemContext.Provider>
-      </Collapsible>
-    );
-  }
-);
+  return (
+    <Collapsible
+      selector={getSelector(ITEM_NAME)}
+      {...accordionItemProps}
+      ref={forwardedRef}
+      data-state={open ? 'open' : 'closed'}
+      data-disabled={disabled || undefined}
+      disabled={disabled}
+      open={open}
+      onOpenChange={() => accordionContext.setValue(value)}
+    >
+      <AccordionItemContext.Provider value={itemContext}>{children}</AccordionItemContext.Provider>
+    </Collapsible>
+  );
+}) as AccordionItemPrimitive;
 
 AccordionItem.displayName = ITEM_NAME;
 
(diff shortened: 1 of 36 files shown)


chaance:
This is great, and I agree probably a better way to go overall.

My main question is, will the inner function still be typed if we're casting at the end? I'd hate to not have any TS support internally. I didn't pull this branch down, but as written it appears as though both `props` and `forwardedRef` would be inferred as `any`, no?

packages/react/accordion/src/Accordion.tsx
```diff
@@ -104,10 +101,10 @@ const HEADER_DEFAULT_TAG = 'h3';
  * `AccordionHeader` contains the content for the parts of an `AccordionItem` that will be visible
  * whether or not its content is collapsed.
  */
-const AccordionHeader = forwardRefWithAs<typeof HEADER_DEFAULT_TAG>((props, forwardedRef) => {
+const AccordionHeader = React.forwardRef((props, forwardedRef) => {
```
chaance, line 104: Per my comment in the main thread, would we not need type annotations on the inner function params here internally?

benoitgrelard: I've just checked out the PR and as far as I can tell all the inner typing is working correctly. You get full types for props and forwardedRef.

packages/react/polymorphic/src/forwardRefWithAs.ts
```diff
@@ -35,7 +53,7 @@ interface ForwardRefExoticComponentWithAs<
    * events are typed for consumers.
    */
   <As extends keyof JSX.IntrinsicElements>(
-    props: MergeWithDOMProps<As, OwnProps & { as: As }>
+    props: MergeWithDOMProps<As, ExtendedProps<E, P> & { as: As }>
```
benoitgrelard, line 56: Why did we move to single letter generics? I quite liked how they were more descriptive before.

benoitgrelard (commented):
Absolutely ♥️ this! The intent is much better.
Just left a small comment regarding the changes in the types.

packages/react/polymorphic/src/forwardRefWithAs.ts
jjenzz: Ah this is a habit of mine because I find that single letter communicates more clearly that it is using a generic and not some predefined type somewhere. I always get more confused when generics are named strangely. Happy to change it back though if it's easier for you. Will do that 🙂

packages/react/accordion/src/Accordion.tsx
jjenzz: Yeah, this work because of this sneaky little bit at the bottom 😛 

https://github.com/radix-ui/primitives/pull/396/files#diff-fe43cc8774290b4daed95c5a7b644d0f1e4d25984d70e152499a592613884b7bR71

packages/react/polymorphic/src/forwardRefWithAs.ts
benoitgrelard: I can see arguments for both, I think in some of these it was quite useful though yeah.

benoitgrelard:
This is getting somewhere, @jjenzz and I figured out a new a way on a call which we think is even better.

packages/react/accordion/src/Accordion.tsx
```diff
@@ -35,21 +35,27 @@ const [AccordionContext, useAccordionContext] = createContext<AccordionContextVa
...
+    /**
+     * A string value for the accordion item. All items within an accordion should use a unique value.
+     */
+    value: string;
+  }
+>;
+
+type AccordionItemPrimitive = Polymorphic.ForwardRefComponent<
+  Polymorphic.IntrinsicElement<typeof Collapsible>,
+  Omit<AccordionItemOwnProps, 'open' | 'defaultOpen' | 'onOpenChange'>
```
benoitgrelard, line 56: @jjenzz I wonder if this should be up too,  as really again in my mind `AccordionItemOwnProps` contains all the props `AccordionItem` accepts apart from the DOM props.

packages/react/menu/src/Menu.tsx
```diff
@@ -66,256 +74,259 @@ const Menu = forwardRefWithAs<typeof MenuImpl, MenuOwnProps>((props, forwardedRe
       </MenuImpl>
     </Presence>
   );
-});
+}) as MenuPrimitive;
+
+type MenuImplOwnProps = Merge<
+  Polymorphic.OwnProps<typeof PopperPrimitive.Root>,
+  {
+    anchorRef: React.ComponentProps<typeof PopperPrimitive.Root>['anchorRef'];
```
benoitgrelard, line 82: It looks like this one is actually not needed right? Would you mind double-checking?

packages/react/menu/src/Menu.tsx
```diff
@@ -351,14 +374,21 @@ const ITEM_NAME = 'MenuItem';
 const ENABLED_ITEM_SELECTOR = `[data-${getSelector(ITEM_NAME)}]:not([data-disabled])`;
 const ITEM_SELECT = 'menu.itemSelect';
 
-type MenuItemOwnProps = {
-  disabled?: boolean;
-  textValue?: string;
-  onSelect?: (event: Event) => void;
-  onSelectCapture: never;
```
benoitgrelard, line 358: I think we missed omitting `onSelectCapture` here?

packages/react/polymorphic/package.json
```diff
@@ -18,6 +18,7 @@
     "prepublish": "yarn clean && yarn build"
   },
   "devDependencies": {
+    "@radix-ui/react-primitive": "workspace:*",
```
benoitgrelard, line 21: we don't normally put dev deps (of our packages) in here otherwise it screws up with yarn's versioning, ie, when changing primitives, it'll ask for release of polymorphic, which is wrong.

packages/react/polymorphic/src/forwardRefWithAs.ts
```diff
@@ -3,28 +3,34 @@ import * as React from 'react';
 /* -------------------------------------------------------------------------------------------------
  * Utility types
  * -----------------------------------------------------------------------------------------------*/
-type NeverKeys<T> = { [P in keyof T]: T[P] extends never ? P : never }[keyof T];
-type MergeProps<P1 = {}, P2 = {}> = Omit<P1, keyof P2> &
-  (NeverKeys<P2> extends never ? P2 : Omit<P2, NeverKeys<P2>>);
+type Merge<P1 = {}, P2 = {}> = Omit<P1, keyof P2> & P2;
```
benoitgrelard, line 6: Is this the same as the one in our types utils? Perhaps it was duplicated on purpose to avoid the dependency?

packages/react/polymorphic/src/forwardRefWithAs.ts
```diff
@@ -46,56 +52,14 @@ interface ForwardRefExoticComponentWithAs<
...
-  <As extends React.ElementType<any>>(
-    props: MergeWithDOMProps<As, OwnProps & { as: As }>
-  ): JSX.Element;
-}
-
-/* -------------------------------------------------------------------------------------------------
- * forwardRefWithAs
- * -----------------------------------------------------------------------------------------------*/
-
-/**
```
benoitgrelard, line 58: haha, nothing left in this file now 😀

I'm wondering though, do we still need the `never` thing? because how else would we remove native props? Let's catchup on that once I'm done with the review as I've left comments related to that elsewhere too.

packages/react/popover/src/Popover.tsx
```diff
@@ -190,155 +126,233 @@ const PopoverContent = forwardRefWithAs<typeof PopoverContentImpl>((props, forwa
...
+     * Whether the `Popover` should render in a `Portal`
+     * (default: `true`)
+     */
+    portalled?: boolean;
+
+    /**
+     * Used to force mounting when more control is needed. Useful when
+     * controlling animation with React animation libraries.
+     */
+    forceMount?: true;
```
benoitgrelard, line 201: I think `forceMount` here should be up in `PopoverContentOwnProps`.

packages/react/radio-group/src/Radio.tsx
```diff
@@ -21,21 +21,22 @@ const RADIO_NAME = 'Radio';
 const RADIO_DEFAULT_TAG = 'button';
 
 type InputDOMProps = React.ComponentProps<'input'>;
-type RadioOwnProps = {
-  checked?: boolean;
-  defaultChecked?: boolean;
-  required?: InputDOMProps['required'];
-  readOnly?: InputDOMProps['readOnly'];
-  onCheckedChange?: InputDOMProps['onChange'];
-  onChange: never;
```
benoitgrelard, line 30: Missed omitting `onChange`?

packages/react/slider/src/Slider.tsx
```diff
@@ -524,14 +545,13 @@ SliderTrack.displayName = TRACK_NAME;
 const RANGE_NAME = 'SliderRange';
 const RANGE_DEFAULT_TAG = 'span';
 
-type SliderRangeOwnProps = {
-  children: never;
-};
-
-const SliderRange = forwardRefWithAs<
+type SliderRangeOwnProps = Merge<Polymorphic.OwnProps<typeof Primitive>, { children?: never }>;
```
benoitgrelard, line 548: So yeah again related to my previous question, we do `never` here, but won't it get re-added anyway via the DOM props if we don't remove the never keys ourselves in our types internally?

packages/react/switch/src/Switch.tsx
```diff
@@ -20,24 +20,25 @@ const SWITCH_NAME = 'Switch';
 const SWITCH_DEFAULT_TAG = 'button';
 
 type InputDOMProps = React.ComponentProps<'input'>;
-type SwitchOwnProps = {
-  checked?: boolean;
-  defaultChecked?: boolean;
-  required?: InputDOMProps['required'];
-  readOnly?: InputDOMProps['readOnly'];
-  onCheckedChange?: InputDOMProps['onChange'];
-  onChange: never;
```
benoitgrelard, line 29: again, do we need to omit `onChange` somehow?

benoitgrelard (commented):
This is sooooo much better ♥️ 

I love it!

I have left a few comments inline, these are mainly all about the same question that I still have left in my head. Perhaps better to catchup on a call to clarify.

packages/react/polymorphic/src/forwardRefWithAs.ts
jjenzz: Yeah it's the same and that's exactly why I duplicated it. Utils package has a lot of stuff in it and seemed like an unnecessary overhead for this simple util.

packages/react/polymorphic/src/forwardRefWithAs.ts
jjenzz: Haha yeah, wanted to chat to you about this one actually.

benoitgrelard (approved):

PR #397, opened 2021-01-14
[Popper + dependents] Add wrapper data attribute

### LINKED ISSUE #368: Portalled tooltips appear behind modals/overlays that have z-index
## Bug report
Using a portalled tooltip in an overlay or modal that has a z-index means the tooltip gets positioned behind the modal.

### Current Behavior

Portalled tooltips are placed at end of `<body>` with `<radix-portal>` element and a `<div>` with positioning. These cannot be styled specifically for tooltip instances and therefore can't have z-index set on them. So they can't appear above any item that created them that has a z-index.

To attempt to fix I set the tooltip portalled to `false`, but this also illustrates the problem from this [issue](https://github.com/radix-ui/primitives/issues/367) where a not portalled tooltip inside a positioned element such as a modal will not be positioned correctly:

Screen Shot 2020-12-23 at 1 23 20 PM

### Expected behavior

Portalled tooltips appear above the elements that create them and/or can have a z-index set on them and are positioned correctly.

### Reproducible example

[CodeSandbox](https://codesandbox.io/s/radix-primitives-tailwind-forked-e095i?file=/src/App.tsx)

### Suggested solution

Give the positioning div a `[radix]` attribute to allow styling with a z-index.

### Your environment

| Radix Package(s) | Tooltip | 0.01 |
| React            | n/a     | 16.12  |
| Browser          |  Chrome, FF, Safari, Edge       | Latest |
| Assistive tech   | n/a        |         |
| Node             | n/a     |  10.17       |
| npm/yarn         | Yarn | 1.22.10 |
| Operating System | Mac + Windows | Latest |

benoitgrelard:
Closes #368 

This PR adds a `data-radix-*` attribute to the non-exposed part of `Popper` (the positioning wrapper).
This is in response to what we have discussed over on this issue: [Portalled tooltips appear behind modals/overlays that have z-index #368](https://github.com/radix-ui/primitives/issues/368#issuecomment-754195352)

I have passed the right name to each dependent component too so you end up with `data-radix-tooltip-content-wrapper` for example. This way users can precisely target only certain components.

A few questions here:

- is this enough? What if they only need to set some rules to only "some" tooltips, or "some" menus, etc
- we could provide a `wrapperClassname` escape hatch for that I suppose, or even do just that and not bother with the data attribute

Also, doing all this with the part name passed as a prop and the `{...getPartDataAttrObj(partName) }` internally made me think that perhaps this is a better solution to this issue: [[react-*]: Remove extended internal data attributes #347](https://github.com/radix-ui/primitives/pull/347)

Thoughts?

### DIFF
--- a/.yarn/versions/9673d3b6.yml
+++ b/.yarn/versions/9673d3b6.yml
@@ -1,0 +1,10 @@
+releases:
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-popper": patch
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
--- a/.yarn/versions/9673d3b6.yml
+++ b/.yarn/versions/9673d3b6.yml
@@ -5,6 +5,30 @@ releases:
   "@radix-ui/react-popover": patch
   "@radix-ui/react-popper": patch
   "@radix-ui/react-tooltip": patch
+  "@radix-ui/utils": patch
 
 declined:
   - primitives
+  - "@radix-ui/popper"
+  - "@radix-ui/react-accordion"
+  - "@radix-ui/react-alert-dialog"
+  - "@radix-ui/react-announce"
+  - "@radix-ui/react-arrow"
+  - "@radix-ui/react-aspect-ratio"
+  - "@radix-ui/react-avatar"
+  - "@radix-ui/react-checkbox"
+  - "@radix-ui/react-collapsible"
+  - "@radix-ui/react-dialog"
+  - "@radix-ui/react-focus-scope"
+  - "@radix-ui/react-label"
+  - "@radix-ui/react-progress"
+  - "@radix-ui/react-radio-group"
+  - "@radix-ui/react-roving-focus"
+  - "@radix-ui/react-scroll-area"
+  - "@radix-ui/react-separator"
+  - "@radix-ui/react-slider"
+  - "@radix-ui/react-switch"
+  - "@radix-ui/react-tabs"
+  - "@radix-ui/react-toggle-button"
+  - "@radix-ui/react-utils"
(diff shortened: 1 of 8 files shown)


jjenzz:
I believe @chaance proposed similar for #347 but in that case:
 
image

🙈  Happy to go with it if you're both considering it though. We can have a `selector` prop on all our components that defaults to `namespaced(partName)`.

That would allow consumers to rename it if they wanted to which is actually a nice feature and makes me feel more comfortable because we're not just exposing it to fix something internal then, we're giving consumers something useful.

```jsx
<Accordion selector="my-org-accordion" />
// Outputs
<div data-my-org-accordion />
```

I would vote for renaming `wrapperPartName` to `wrapperSelector` too.

benoitgrelard:
Oh that's an interesting idea!

benoitgrelard:
Updated as per @jjenzz's feedback and as per upcoming #347 changes.

benoitgrelard:
We should wait for #347 to be done, to update this one accordingly as the extra prop might not be warranted as @jjenzz puts it:

> jenna  23:43
I noticed we probably don’t need an extra wrapperSelector prop…
23:43
internally you can just do {...getSelectorObj(selector + '-wrapper')}  on your wrapper element
jenna  11:08
this should work no? {...(selector ? getSelectorObj(selector + '-wrapper') : undefined}

@jjenzz let me know if you'd rather do this straight in your PR otherwise.

chaance:
@benoitgrelard One of the benefits of the prop would be for consumers to override our internal selectors if they wanted. This could be very handy if they decide to wrap our components with a different set of naming conventions, and if they'd like to completely close off certain parts in their design systems. We couldn't do that without the `selector` prop, could we?

benoitgrelard:
@jjenzz's point here was that this one is an escape hatch for a part that they normally wouldn't need access to. So it would be fine to derive it from the main part. But yeah that does mean you can't rename it fully like the others.

What do you think Jenna?

benoitgrelard:
Closing this one now as #347 has addressed it already.

PR #400, opened 2021-01-14
[react-accordion] Allow disabled `AccordionItem` when `Accordion` is `disabled={false}`

jjenzz:
This was raised on Discord.

If the `Accordion.Root` had `disabled={false}`, the `Accordion.Item`s would all be enabled regardless of whether they explicitly had a `disabled={true}` prop themselves.

- [X] Add or edit tests to reflect the change (run `yarn test`).

This pull request:

- [X] Fixes a bug in an existing package
- [ ] Other

### DIFF
--- a/.yarn/versions/77e1fa7a.yml
+++ b/.yarn/versions/77e1fa7a.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-accordion": patch
+
+declined:
+  - primitives
--- a/packages/react/accordion/src/Accordion.test.tsx
+++ b/packages/react/accordion/src/Accordion.test.tsx
@@ -156,6 +156,64 @@ describe('given an Accordion with a change callback', () => {
   });
 });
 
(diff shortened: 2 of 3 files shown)


chaance (changes requested):
Looks good to me! Think we can go ahead and add a `CHANGELOG` here per the Slack conversation?

benoitgrelard (approved):

jjenzz:
@chaance There is no `CHANGELOG` file so I'm not really sure what the plan is there. 

We should probably create a separate CHANGELOG PR with everything merged since last release that demonstrates the format we can follow. I'm not sure if we want a `CHANGELOG` per package or just in the root for example so this is all stuff we can probably solve separately and then each PR would follow it once merged?

benoitgrelard:
Yeah I was going to say I think this is something we should start after we do our next release with all the changes that are currently queued up already.

chaance (approved):

PR #403, opened 2021-01-25
[Dialog][AlertDialog] Fix regression

benoitgrelard:
I've spotted a small regression with `Dialog` 0.0.2 (and therefore `AlertDialog)` (closing when tabbing even though focus is trapped).

It was introduced introduced after we made the changes in `FocusScope` to not cancel out `focusout` event.
We had made the necessary changes in `Popover` and `Menu`, etc (to `preventDefault` when `trapFocus` was true) but forgot to do it always in the case of `Dialog` because focus is always trapped.


jjenzz (approved):
:+1:

PR #404, opened 2021-01-25
[ContextMenu][DropdownMenu] Fix issue introduced by new `selector` prop

### LINKED ISSUE #402: [DropdownMenu] Opening focus and arrow keys does not work
### Current Behavior

**Version 0.0.2 only**. When opening a menu and pressing the down arrow, the focus is not moved to the first item.

### Expected behavior

Opening a menu dropdown and immediately pressing the down arrow should move focus to the first non-disabled item.

### Reproducible example

https://codesandbox.io/s/great-shannon-d6smv?file=/src/App.tsx

### Suggested solution

Arrow down should focus the first non-disabled item in the menu.

### Additional context

N/A

### Your environment

| Radix Package(s) |   `@radix-ui/react-dropdown-menu`      |  0.0.2       | 
| Browser          |   All      |         |
| Assistive tech   |   n/a      |         |
| npm/yarn         |   n/a      |         |
| Operating System |   n/a      |         |

benoitgrelard:
Closes #402 

After checking a reported issue #402, I realised we have introduced a bug in dependents of `Menu` (`ContextMenu`, `DropdownMenu`) when we added the `selector` prop.

This is because the expectation on what attributes are present changed, and we didn't account for it here.
Namely, before we were expecting the `data-radix-menu-item` to be present even in dependents of `Menu`.
This is no longer a safe assumption to make obviously, therefore we cannot rely on this selector.

Looking at other places we build selectors like this (roving focus behaviour, or typeahead behaviour) they both add their own data attribute to ensure they can safely select items. So it seems in this case we should be doing the same somehow. It feels a bit more annoying given that we know there is already a default selectors for items… but obviously now specific to each dependent of `Menu`).

I haven't made any change yet, wondering what your thoughts were.

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -370,6 +370,7 @@ MenuLabel.displayName = LABEL_NAME;
  * -----------------------------------------------------------------------------------------------*/
 
 const ITEM_NAME = 'MenuItem';
+// HERE: 👇
 const ENABLED_ITEM_SELECTOR = `[data-${getSelector(ITEM_NAME)}]:not([data-disabled])`;
 const ITEM_SELECT = 'menu.itemSelect';
 
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -304,10 +304,6 @@ const MenuImpl = React.forwardRef((props, forwardedRef) => {
                       }
                     }
                   })}
-                  // focus the menu if the mouse is moved over anything else than an item
-                  onMouseMove={composeEventHandlers(menuProps.onMouseMove, (event) => {
-                    if (!isItemOrInsideItem(event.target)) menuRef.current?.focus();
-                  })}
                 >
                   <RovingFocusGroup
                     reachable={itemsReachable}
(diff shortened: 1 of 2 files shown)


jjenzz:
Ah, oops. It would be good to have a call about this one because I've had a look at the code and wondering if we could get rid of `ENABLED_ITEM_SELECTOR` altogether. 

I commented out the following code and struggling to decipher the difference so would be good to run through it with you: 

```jsx
// focus the menu if the mouse is moved over anything else than an item
onMouseMove={composeEventHandlers(menuProps.onMouseMove, (event) => {
  if (!isItemOrInsideItem(event.target)) menuRef.current?.focus();
})}
```

Also, the `onKeyDown` handler could probably use a new `itemsRef.current` array that each `MenuItem` would append to in the context.

```jsx
React.useEffect(() => {
  context.itemsRef.current = [...context.itemsRef.current, ref.current];
}, [context.itemsRef, ref]);
```


If all else fails, adding a new stable selector for this makes sense to me 🙂

packages/react/menu/src/Menu.tsx
```diff
@@ -304,10 +304,6 @@ const MenuImpl = React.forwardRef((props, forwardedRef) => {
                       }
                     }
                   })}
-                  // focus the menu if the mouse is moved over anything else than an item
```
benoitgrelard, line 307: Always nice to be able to remove unused code, well spotted @jjenzz!

jjenzz (approved):

PR #408, opened 2021-01-26
Explicitly declare default values for `as` and `selector` props

jjenzz:
We noticed that Stitches passes an explicit `as={undefined}` prop to our components which wipes them out. We've decided that this shouldn't be possible and to explicitly handle this case by always falling back to a default.

We're going to make sure all our other defaults are handled correctly in a separate PR to avoid similar situations like this in future.

This pull request:

- [ ] Other

### DIFF
--- a/packages/react/scroll-area/src/ScrollArea.tsx
+++ b/packages/react/scroll-area/src/ScrollArea.tsx
@@ -287,6 +287,7 @@ type ScrollAreaNativePrimitive = Polymorphic.ForwardRefComponent<
 
 const ScrollAreaNative = React.forwardRef(function ScrollAreaNative(props, forwardedRef) {
   const {
+    selector = getSelector(ROOT_NAME),
     overflowX,
     overflowY,
     scrollbarVisibility,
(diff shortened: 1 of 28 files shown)


packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -693,7 +695,11 @@ type ScrollAreaViewportPrimitive = Polymorphic.ForwardRefComponent<
 
 const ScrollAreaViewport = React.forwardRef(function ScrollAreaViewport(props, forwardedRef) {
   return useNativeScrollArea() ? (
-    <Primitive selector={getSelector(VIEWPORT_NAME)} {...props} ref={forwardedRef} />
+    <Primitive
+      {...props}
+      selector={props.selector || getSelector(VIEWPORT_NAME)}
```
benoitgrelard, line 700: I'm sure there's a reason but can't this one also be a default arg like the others? Shouldn't it also get passed to `ScrollAreaViewportImpl`?

benoitgrelard (approved):

packages/react/scroll-area/src/ScrollArea.tsx
jjenzz: `ScrollAreaViewportImpl` receives all the `{...props}` and does the `selector = getSelector` thing internally itself.

PR #410, opened 2021-01-29
[Portal] Fix potential out of the box layering issues

benoitgrelard:
If the `Portal` is injected in `body` (default behaviour), we will now assume we want whatever is portalled to appear on top of everything.

Ideally this would be handled by making sure the app root creates a new stacking context, however this is quite hard to automate and we have had many users getting tripped up by this already (see #368, https://github.com/modulz/composer/pull/97 for example).

For this reason, we have opted for setting the max z-index on the portal itself in such cases.

Also, standardizes the portal element to follow the usual convention of other primitives, with regards to as, types and selector.

> 🔥 NOTE: This is a potential breaking change as the element rendered used to be `<radix-portal>` and will now be `<div data-radix-portal>`

### DIFF
--- a/.yarn/versions/8ca9debf.yml
+++ b/.yarn/versions/8ca9debf.yml
@@ -1,0 +1,12 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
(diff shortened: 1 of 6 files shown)


jjenzz (approved):

peduarte:
Good decision 💯

PR #409, opened 2021-01-28
Add `Slot` component

jjenzz:
Allows consumers to turn any of our components into a "slot" when wrapping to create more closed APIs. For example:

```jsx
function Tooltip({ label, children }) {
	return (
		<TooltipPrimitive.Root>
			<TooltipPrimitive.Trigger as={Slot}>
				{children}
			</TooltipPrimitive.Trigger>
			<TooltipPrimitive.Content>
				{label}
			</TooltipPrimitive.Content>
		</TooltipPrimitive.Root>
	);
}

// Usage
<Tooltip label="Notifications" sideOffset={5}>
	<button onClick={handleNotificationsClick}>
		<svg><path /></svg>
	</button>
</Tooltip>
```

- [X] Use a meaningful title for the pull request. Include the name of the package modified.
- [X] Test the change in your own code.
- [ ] Add or edit tests to reflect the change (run `yarn test`).
- [X] Add or edit Storybook examples to reflect the change (run `yarn dev`).
- [ ] Add documentation to support any new features.

This pull request:

- [ ] Fixes a bug in an existing package
- [ ] Adds additional features/functionality to an existing package
- [ ] Updates documentation or example code
- [X] Other

PR #414, opened 2021-02-01
[Menu] Fix missing children

benoitgrelard:
@jjenzz that one was my fault, I missed it in the review (https://github.com/radix-ui/primitives/pull/409/files#r567683307)

> Note: better to review with whitespace turned off


jjenzz (approved):
🙈  ooops, looks like I forgot to do a codebase search for `.Provider` instead of `{children}` which is why this one slipped through the net!

benoitgrelard:
Haha no worries, we just need to make sure we have some basic smoke tests at the very least (working on that).

PR #415, opened 2021-02-01
[Popper] Fix collapsing layout issues

benoitgrelard:
We noticed an issue with Popover flashing (sliding) when positioning.
This was due to the changes we made to `Portal` adding styles to position above things.
The main issue is that now we have 2 parents with `position: absolute` and so intrinsic-sized elements would collapse. 

> Update: we've now made a more surgical fix in popper itself.

### DIFF
--- a/.yarn/versions/c0ee0213.yml
+++ b/.yarn/versions/c0ee0213.yml
@@ -1,0 +1,11 @@
+releases:
+  "@radix-ui/popper": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-popover": patch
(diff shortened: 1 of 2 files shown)


packages/react/portal/src/Portal.tsx
```diff
@@ -58,6 +58,8 @@ const Portal = React.forwardRef((props, forwardedRef) => {
             ? {
                 position: 'absolute',
                 top: 0,
+                right: 0,
+                bottom: 0,
```
jjenzz, line 62: This looks like it will put a layer on top of everything... which would break `disableOutsidePointerEvents={false}` on things right? because this element will prevent click through?

benoitgrelard: Ooooh, I didn't think of this

jjenzz (approved):

PR #425, opened 2021-02-03
[Collapsible] Add chromatic story

benoitgrelard:


packages/react/collapsible/src/Collapsible.stories.tsx
```diff
@@ -79,11 +133,31 @@ const fadeOut = css.keyframes({
...
 });
+
+const styles = {
+  backgroundColor: 'rgba(0, 0, 255, 0.3)',
+  border: '2px solid blue',
+  padding: 10,
+
+  '&[data-state="closed"]': { borderColor: 'red' },
+  '&[data-state="open"]': { borderColor: 'green' },
+  '&:disabled': { opacity: 0.5 },
```
benoitgrelard, line 152: test for native disabled

packages/react/collapsible/src/Collapsible.stories.tsx
```diff
@@ -79,11 +133,31 @@ const fadeOut = css.keyframes({
...
+  '&[data-disabled]': { fontSize: '1.5em' },
```
benoitgrelard, line 153: test for disabled data attribute (at all levels)

packages/react/collapsible/src/Collapsible.test.tsx
```diff
@@ -2,7 +2,6 @@ import React from 'react';
 import { axe } from 'jest-axe';
```
benoitgrelard, line 2: Let me know if you think the tests that remain are ok to stay. I feel like these aren't covered so should stay.

jjenzz (approved):

PR #427, opened 2021-02-03
[AspectRatio] Add chromatic story

benoitgrelard:


packages/react/aspect-ratio/src/AspectRatio.tsx
```diff
@@ -28,6 +28,7 @@ const AspectRatio = React.forwardRef((props, forwardedRef) => {
         width: '100%',
         paddingBottom: `${100 / ratio}%`,
       }}
+      {...(selector ? getSelectorObj(selector + '-wrapper') : undefined)}
```
benoitgrelard, line 31: I've gone ahead and done the same thing as we've done for `Popper` to expose an escape hatch on the wrapper.

jjenzz (approved):

PR #428, opened 2021-02-03
Add version check to pre-push

jjenzz:
It has been bugging us that the version check always trips us up but we only find out it's an issue during CI. This will hopefully catch this issue sooner before we switch to different tasks!


benoitgrelard (approved):
♥️ ♥️ ♥️

PR #429, opened 2021-02-03
[Avatar] Add chromatic story

benoitgrelard:


packages/react/avatar/src/Avatar.test.tsx
```diff
@@ -91,15 +77,6 @@ describe('given an Avatar with fallback and a working image', () => {
...
-  it('should have a radix attribute on the image', async () => {
-    const partDataAttr = `data-${getSelector('AvatarImage')}`;
-    await waitFor(() => {
-      jest.advanceTimersByTime(DELAY);
-    });
-    image = rendered.queryByRole('img');
-    expect(image).toHaveAttribute(partDataAttr);
-  });
-
   it('should have alt text on the image', async () => {
```
jjenzz, line 80: Probably don't need this test either because it will be covered by the accessibility violations test I would have thought?

jjenzz (approved):
one small question but otherwise :+1:

packages/react/avatar/src/Avatar.test.tsx
benoitgrelard: I'll test if it's covered or not.

PR #430, opened 2021-02-03
[Checkbox] Add chromatic story

benoitgrelard:


packages/react/checkbox/src/Checkbox.test.tsx
```diff
@@ -47,52 +33,19 @@ describe('given a default Checkbox', () => {
...
-
-    it('should have a data-state attribute on the container set to `checked`', () => {
-      expect(checkbox).toHaveAttribute('data-state', 'checked');
-    });
-
-    it('should have a data-state attribute on the indicator set to `checked`', () => {
-      expect(indicator).toHaveAttribute('data-state', 'checked');
-    });
-
-    describe('and clicking the checkbox again', () => {
```
jjenzz, line 64: Do we want to keep this one somehow because it's testing interaction?

jjenzz (approved):
One question, otherwise :+1:

packages/react/checkbox/src/Checkbox.test.tsx
benoitgrelard: I'll check this now.

benoitgrelard: Yeah good shout, I had remove it cause it was checking data-state, but I've rewritten to check for the indicator

jjenzz (approved):

PR #432, opened 2021-02-04
[AlertDialog] Add chromatic story

benoitgrelard:


packages/react/alert-dialog/src/AlertDialog.stories.tsx
```diff
@@ -48,24 +48,121 @@ export const Controlled = () => {
...
+          <AlertDialogAction className={actionAttrClass}>Confirm</AlertDialogAction>
+          <AlertDialogCancel className={cancelAttrClass}>Cancel</AlertDialogCancel>
+        </AlertDialogContent>
+      </AlertDialog>
+    </div>
+  </div>
+);
+Chromatic.parameters = { chromatic: { disable: false } };
+
+const triggerClass = css({});
```
jjenzz, line 165: Doesn't look like we need this? But maybe you did it for consistency?

packages/react/alert-dialog/src/AlertDialog.stories.tsx
```diff
@@ -103,21 +200,45 @@ const StyledContent = styled('div', {
...
   border: 'none',
   backgroundColor: '$red',
   color: '$white',
 });
 
-const StyledTitle = styled('h2', {});
+const titleClass = css({});
 
-const StyledDescription = styled('p', {});
+const descriptionClass = css({});
```
jjenzz, line 220: Same question here, looks like you purposefully left them for some reason?

jjenzz (approved):
couple questions but otherwise :+1:

jjenzz:
One other thing, I was playing with the canvas view here: https://www.chromatic.com/component?appId=601857eb614bae0021c9b9dd&name=Components%2FAlertDialog&mode=interactive&buildNumber=50

I noticed that when I click cancel on the Uncontrolled dialog, it also closes the data attribute dialog and I can never re-open it after that. Do we have some re-used `id`s somewhere?

packages/react/alert-dialog/src/AlertDialog.stories.tsx
benoitgrelard: Yeah we do this everywhere for consistency.

benoitgrelard:
> One other thing, I was playing with the canvas view here: https://www.chromatic.com/component?appId=601857eb614bae0021c9b9dd&name=Components%2FAlertDialog&mode=interactive&buildNumber=50
> 
> I noticed that when I click cancel on the Uncontrolled dialog, it also closes the data attribute dialog and I can never re-open them after that. Do we have some re-used `id`s somewhere?

No it's because there are multiple open but not on top of each other and then the controlled one has no `onOpenChange` so can't be closed. These aren't meant to be interactive really.

jjenzz:
Cool, feel free to push the button 🎉

PR #440, opened 2021-02-04
[Slider] Add chromatic story

benoitgrelard:


jjenzz (approved):
Can we add an extra story here with the thumb at `0%` or `100%` please? Means we can see if the thumb offset logic has been broken because it makes sure the thumb edge aligns with the track edge.

All good otherwise :+1:

benoitgrelard:
> Can we add an extra story here with the thumb at `0%` or `100%` please? Means we can see if the thumb offset logic has been broken because it makes sure the thumb edge aligns with the track edge.
> 
> All good otherwise 👍

Good idea, would a single one with 2 thumbs at each extremity suffice?

jjenzz:
Yep, that'd be perfect 🙂

benoitgrelard:
That's done.

jjenzz:
Thanks :+1:

PR #451, opened 2021-02-10
New component: Toolbar (#412)

jjenzz:


jjenzz:
@Nazeh Just a heads up that we had tried to get this into our planned release today but while testing it again locally, Benoit noticed that the roving index prevents arrow keys from working inside the input. 

This is a bit of an accessibility concern which led us to test if our other components would work (e.g. `RadioGroup`) and we spotted some roving index issues. With the potential of these conflicts in mind, we decided to go with the following changes:

- `ToolbarItem` renamed to `ToolbarButton`. Any of our components that provide a `Trigger` can do `<ToolbarButton as={Component.Trigger}>` as you had already demonstrated in the stories.
- Added `ToolbarRadioGroup` and `ToolbarRadioItem` components. These allow `Toolbar` to control roving focus.
- Added a `ToolbarLink` part so that consumers get the space key functionality you added to the story out of the box.
- Removed the `input` example from the story. After researching, this feels like a bit of an anti-pattern but we're not entirely sure. For now, we're not going to promote the idea anywhere until we can explore it further. We may provide a `ToolbarInput` component later with more keyboard accessibility bindings.
- Removed `ToolbarGroup`. It didn't provide any accessibility/functional enhancements and seems to be covered by `ToolbarRadioGroup` and `ToggleButtonGroup` that we have in PR atm. 

If you have an questions/change requests/further ideas, please do let us know 🙂

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -73,24 +76,27 @@ const RadioGroup = React.forwardRef((props, forwardedRef) => {
...
-      <RovingFocusGroup loop>
-        <Primitive
-          {...groupProps}
-          selector={selector}
-          ref={forwardedRef}
-          role="radiogroup"
-          aria-labelledby={labelledBy}
-        />
-      </RovingFocusGroup>
+      {rovingFocus ? <RovingFocusGroup loop>{primitive}</RovingFocusGroup> : primitive}
```
jjenzz, line 99: I was torn between doing this or adding a `ConditionalWrap` component. Which do we prefer?

benoitgrelard: Historically we've stuck with this for now. I'm happy to keep this.

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -103,58 +109,89 @@ RadioGroup.displayName = RADIO_GROUP_NAME;
...
+  );
+}) as RadioGroupItemPrimitive;
+
+type RadioGroupRovingFocusItemOwnProps = Polymorphic.OwnProps<typeof RadioGroupItemImpl>;
+type RadioGroupRovingFocusItemPrimitive = Polymorphic.ForwardRefComponent<
+  Polymorphic.IntrinsicElement<typeof RadioGroupItemImpl>,
+  RadioGroupRovingFocusItemOwnProps
+>;
+
+const RadioGroupRovingFocusItem = React.forwardRef((props, forwardedRef) => {
```
benoitgrelard, line 133: I like this new separation now, and seeing the event handlers only for the roving focus one 👍

packages/react/toolbar/src/Toolbar.tsx
```diff
@@ -0,0 +1,286 @@
...
+      role="group"
+      data-orientation={orientation}
+      {...props}
+      selector={getSelector(GROUP_NAME)}
+      ref={forwardedRef}
+    />
+  );
+}) as ToolbarGroupPrimitive;
+
+ToolbarGroup.displayName = GROUP_NAME;
```
benoitgrelard, line 110: I thought we decided to remove this one? Is it a bad merge/rebase?

packages/react/toolbar/src/Toolbar.tsx
```diff
@@ -0,0 +1,286 @@
...
+const ToolbarRadioItem = React.forwardRef((props, forwardedRef) => (
+  <RadioGroupPrimitive.Item
+    {...props}
+    as={ToolbarButton}
+    selector={getSelector(RADIO_GROUP_ITEM_NAME)}
+    ref={forwardedRef}
+  />
+)) as ToolbarRadioPrimitive;
+
+ToolbarRadioItem.displayName = RADIO_GROUP_ITEM_NAME;
```
benoitgrelard, line 261: If we're not doing anything special, should we use the `extendComponent` util?

jjenzz: I am doing `as={ToolbarButton}` here and merging types which is why I needed to expand it like this

benoitgrelard: Oh I missed that. Does that mean that `as` is taken then and can't be used on `ToolbarRadioItem`?

jjenzz: Ah, good point, let me fix that ...

jjenzz: Fixed :+1: good spot 👀

colmtuite (approved):

benoitgrelard (approved):

PR #455, opened 2021-02-12
Add SSR testing playground setup

benoitgrelard:
This PR:

- update our yarn version to latest as it seem to be a blocker when installing deps in setting up a nextjs project
- adds a new workspace called `ssr-testing` with a nextjs project
- sets up nextjs so that it picks up our local packages sources instead of builds

> NOTE: If I'm completely honest, I don't fully understand how the `with-transpiled-modules` thing work, considering I'm not giving it any modules in the array, but it works… 🤷‍♂️


ssr-testing/next.config.js
```diff
@@ -0,0 +1,19 @@
...
+        : rule.use.loader === 'next-babel-loader'
+    );
+
+    if (babelRule) {
+      babelRule.include.push(path.resolve('../'));
+    }
+
+    return config;
+  },
+});
```
benoitgrelard, line 19: This took me ages to figure out, it works but I don't really get how…

ssr-testing/pages/radio-group.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { RadioGroup, RadioGroupItem, RadioGroupIndicator } from '@radix-ui/react-radio-group';
```
benoitgrelard, line 2: We're able to import the packages like this, and they import the source and everything rebuilds and hot reloads when making changes.

ssr-testing/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../tsconfig.json",
```
benoitgrelard, line 2: Extending our base config to get the path mapping

benoitgrelard:
awww man 😢  now getting this build error…

image

ssr-testing/next.config.js
jjenzz: well if it works, I guess that's all we need haha. It's just an internal thing anyway 🤷‍♀️

ssr-testing/pages/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+      <Head>
+        <title>Home</title>
+        <link rel="icon" href="/favicon.ico" />
+      </Head>
+
+      <main>
+        <h1>Components</h1>
+        <ul>
+          <li>
+            <Link href="./radio-group">RadioGroup</Link>
```
jjenzz, line 17: Getting this error 

image

Looks like we don't need the dot `.`?

jjenzz (approved):
Other than the build error and the following, looks good :+1:

Obviously there is the id inconsistency in `RadioGroup` but we'll fix that separately I presume!

package.json
```diff
@@ -31,7 +32,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
-    "@stitches/react": "canary",
+    "@stitches/react": "0.0.3-canary.4",
```
benoitgrelard, line 35: I had to fix this to latest alpha because they've been publishing new betas and as I've busted the yarn lock we were getting the new API. I don't want to do this work now which will muddy the waters of what this PR is about.

package.json
```diff
@@ -67,6 +68,9 @@
     "ts-jest": "^26.2.0",
     "typescript": "^3.9.7"
   },
+  "resolutions": {
+    "chokidar": "3.4.3"
+  },
```
benoitgrelard, line 73: So, even after all my efforts trying to install `next` manually, I couldn't get it to work (still got that hunk error) and any `yarn` from scratch would give that error anyway.

Adding this resolution fixes the issue as suggested here: https://github.com/vercel/next.js/pull/21343#issuecomment-777629123 (although I suspect that typed the version number wrong because they're referring to the new one rather than old one)

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -1852,7 +1855,7 @@ function scrollAreaStateReducer(
 // Alternatively we could reconsider exposing the Position component, or we could allow users to
 // pass their own positionRef with a prop.
 function useExtendedScrollAreaRef(
-  forwardedRef: React.ForwardedRef<any>,
+  forwardedRef: React.Ref<any>,
```
benoitgrelard, line 1858: No idea what happened with this file! But somehow this TS error got flagged because `React.ForwardedRef` is not a thing. I have no idea how this went through for months…

On top of that all the other changes are just prettier, again not sure how it went in not formatted…

For both reasons above I've declined a package update for this as it's just internal.

package.json
benoitgrelard: For posterity:

As the above linked issue suggests, yarn 2.4.0 fixed that issue but we have another issue which prevents us from updating. https://github.com/parcel-bundler/parcel/issues/4064#issuecomment-739900240

That seems to have appeared between 2.1.1 (which we are on right now) and 2.2.0.

jjenzz (approved):

PR #464, opened 2021-02-19
[SSR] Add new react-ssr package to house ssr specific stuff

benoitgrelard:
This PR adds a new `@radix-ui/react-ssr` package to house all SSR related stuff.
This is step 1 of making sure our components offer better SSR compliance.

> Note: this is mostly copied over from `react-aria`'s implementation of SSR.


packages/react/ssr/src/SSR.tsx
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return React.useMemo(
+    () => deterministicId || `radix-id-${context.prefix}-${++context.current}`,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [deterministicId]
+  );
+}
+
+function canUseDOM() {
```
benoitgrelard, line 49: We have the same function in our dom utils, but I think it makes more sense for it to be in this module. This is also in view of starting to break things up more and create less dependencies, so eventually I'll use this one instead of the other one and remove the old one.

packages/react/ssr/src/SSR.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+function useSSRSafeId(deterministicId?: string): string {
+  const context = React.useContext(SSRContext);
+  const isDOMEnv = canUseDOM();
+
+  if (!isDOMEnv && context === defaultSSRContext) {
+    console.warn(
+      'When server rendering, you must wrap your application in an <SSRProvider> to ensure consistent ids are generated between the client and server.'
+    );
```
jjenzz, line 39: Do we want to do this in production? Just wondering if a `process.env` check is worthwhile.

packages/react/ssr/src/SSR.tsx
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return React.useMemo(
+    () => deterministicId || `radix-id-${context.prefix}-${++context.current}`,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jjenzz, line 44: why aren't we passing `context` as a hook dependency?

packages/react/ssr/src/SSR.tsx
```diff
@@ -0,0 +1,53 @@
...
+      prefix: isRootSSRProvider ? 0 : ++currentContext.prefix,
+      current: 0,
+    }),
+    [isRootSSRProvider, currentContext]
+  );
+
+  return <SSRContext.Provider value={context} {...props} />;
+};
+
+function useSSRSafeId(deterministicId?: string): string {
```
jjenzz, line 32: Is the plan to use this inside our `useId` hook? As in, we wouldn't really interface with this directly?

packages/react/ssr/src/SSR.tsx
benoitgrelard: hmm not sure with this one. It's pretty critical to get that right, so perhaps still worth living it in prod no? I imagine a lot of folks might not test stuff properly in dev either so it might be handy to see this when popping open devtools on their prod site when one of their users reports something isn't working…

packages/react/ssr/src/SSR.tsx
benoitgrelard: Yeah exactly, there would be a friendlier api.

packages/react/ssr/src/SSR.tsx
benoitgrelard: That was inherited from `react-aria`. My guess is that otherwise everytime the prefix or context gets incremented it would have to re-run these so ids would keep increasing for no reason right? The intention seems to be that once generated it won't change (it would only change if they have passed their own id).

jjenzz (approved):

PR #465, opened 2021-02-20
[Presence] Improve `Presence` perf

jjenzz:
---------------

ℹ️&nbsp;&nbsp;Easier to review with hidden whitespace changes.

---------------

We had report of some perf issues with nested `Collapsible`s and I realised `Presence` was causing some problems there. 

I've prevented some `useEffect` logic from running unnecessarily on mount and removed `@xstate/fsm`. 

I had noticed that `@xstate/fsm` really eagerly re-renders components during my Carousel work so we can probably do without that and the additional bundle size it adds for such a simple machine here.

These changes managed to **double** the speed in my tests.

### DIFF
--- a/packages/react/presence/src/useStateMachine.tsx
+++ b/packages/react/presence/src/useStateMachine.tsx
@@ -1,27 +1,17 @@
 import * as React from 'react';
-import { createMachine, interpret } from '@xstate/fsm';
-import type { StateMachine, EventObject, Typestate } from '@xstate/fsm';
 
-export function useStateMachine<
-  C extends object,
-  E extends EventObject = EventObject,
-  S extends Typestate<C> = { value: any; context: C }
->(fsmConfig: StateMachine.Config<C, E, S>) {
-  const [machine] = React.useState(() => interpret(createMachine<C, E, S>(fsmConfig)));
-  const [state, setState] = React.useState<S['value']>(fsmConfig.initial);
-  const [context, setContext] = React.useState<C | undefined>(fsmConfig.context);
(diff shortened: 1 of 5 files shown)


packages/react/presence/src/useStateMachine.tsx
```diff
@@ -1,27 +1,20 @@
...
 
-  return React.useMemo(() => ({ state, context, send: machine.send }), [state, context, machine]);
+export function useStateMachine<M>(
+  initialState: MachineState<M>,
+  machine: M & Machine<MachineState<M>>
+) {
+  return React.useReducer((state: MachineState<M>, event: MachineEvent<M>): MachineState<M> => {
+    const nextState = (machine[state] as any)[event];
+    return nextState ?? state;
+  }, initialState);
```
benoitgrelard, line 19: ♥️ how minimal this is haha, so good 🙂

packages/react/presence/src/useStateMachine.tsx
```diff
@@ -1,27 +1,20 @@
...
-      setContext(state.context);
-    });
-    machine.start();
-    return () => {
-      subscription.unsubscribe();
-      machine.stop();
-    };
-  }, [machine]);
+// 🤯 https://fettblog.eu/typescript-union-to-intersection/
+type UnionToIntersection<T> = (T extends any ? (x: T) => any : never) extends (x: infer R) => any
```
benoitgrelard, line 8: I'm gonna still need some time to process this but I'll just do on my own time. I've played with it and it works very well! 👍

benoitgrelard (approved):

PR #469, opened 2021-02-23
[Collapsible] Use SSR-safe ids

benoitgrelard:
> Note: This PR builds upon #467 so will need to wait for it to be merged first.

I'll also do the version check once the other one is merged.

### DIFF
--- a/packages/react/collapsible/src/Collapsible.tsx
+++ b/packages/react/collapsible/src/Collapsible.tsx
@@ -1,13 +1,9 @@
 import * as React from 'react';
-import {
-  createContext,
-  useId,
-  composeEventHandlers,
-  useControlledState,
-} from '@radix-ui/react-utils';
+import { createContext, composeEventHandlers, useControlledState } from '@radix-ui/react-utils';
 import { getSelector } from '@radix-ui/utils';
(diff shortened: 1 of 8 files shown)


packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -152,31 +140,19 @@ type CollapsibleContentPrimitive = Polymorphic.ForwardRefComponent<
...
         <Primitive
-          data-state={getState(open)}
-          data-disabled={disabled ? '' : undefined}
+          data-state={getState(context.open)}
+          data-disabled={context.disabled ? '' : undefined}
           {...contentProps}
           selector={selector}
           ref={forwardedRef}
-          id={id}
+          id={id || context.contentId}
```
benoitgrelard, line 155: I wasn't sure what pattern to use here?
Would you prefer I do `id={context.contentId}` and move it up just before the `{...contentProps}` to show what can be overriden (like `data-state`, `data-disabled`, etc)?

jjenzz: yeah makes sense to move it up I suppose like our others :+1:

jjenzz (approved):

PR #471, opened 2021-02-24
[Dialog] Use SSR-safe ids

benoitgrelard:


packages/react/alert-dialog/src/AlertDialog.tsx
```diff
@@ -40,26 +40,26 @@ const [AlertDialogContext, useAlertDialogContext] = createContext<AlertDialogCon
...
 
 /* -------------------------------------------------------------------------------------------------
  * AlertDialog
  * -----------------------------------------------------------------------------------------------*/
 
 const AlertDialog: React.FC<React.ComponentProps<typeof Dialog>> = (props) => {
-  const { id: idProp, ...dialogProps } = props;
   const generatedId = makeId('alert-dialog', useId());
-  const alertDialogId = idProp || generatedId;
+  const alertDialogId = generatedId;
```
benoitgrelard, line 54: I just removed that temporarily here because `id` was removed from `Dialog` but the changes to ids in `AlertDialog` will come in next PR.

jjenzz (approved):

PR #477, opened 2021-02-24
Update `createContext` to expose a Provider that accepts props

jjenzz:
Before our `Carousel` call earlier, I started applying this across our codebase. I'll PR each component in separate PRs.

### DIFF
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { getSelector } from '@radix-ui/utils';
 import {
   composeEventHandlers,
-  createContext,
+  createContextObj,
   useComposedRefs,
   useControlledState,
 } from '@radix-ui/react-utils';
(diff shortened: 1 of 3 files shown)


packages/react/accordion/src/Accordion.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { getSelector } from '@radix-ui/utils';
 import {
   composeEventHandlers,
-  createContext,
+  createContextObj,
```
jjenzz, line 5: I'll rename this back to `createContext` in the last PR. Unless we think this is more explicit? The context that the provider provides will always be an object.

packages/react/utils/src/createContext.tsx
```diff
@@ -12,3 +12,24 @@ export function createContext<ContextValueType>(displayName: string, rootCompone
   }
   return [Context, useContext] as const;
 }
+
+export function createContextObj<ContextValueType extends object>(rootComponentName: string) {
+  const Context = React.createContext<ContextValueType>(null as any);
+
+  function Provider(props: ContextValueType & { children: React.ReactNode }) {
+    const { children, ...providerProps } = props;
+    const value = React.useMemo(() => providerProps, [providerProps]) as ContextValueType;
```
jjenzz, line 21: wait, this looks like it would re-memoize every render lol.. 🙈  hmm, give me a sec

benoitgrelard (approved):

PR #479, opened 2021-02-24
Update context in `Avatar`

jjenzz:
------

------

### DIFF
--- a/packages/react/avatar/src/Avatar.tsx
+++ b/packages/react/avatar/src/Avatar.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { getSelector } from '@radix-ui/utils';
-import { createContext, useCallbackRef, useLayoutEffect } from '@radix-ui/react-utils';
+import { createContextObj, useCallbackRef, useLayoutEffect } from '@radix-ui/react-utils';
 import { Primitive } from '@radix-ui/react-primitive';
 
 import type * as Polymorphic from '@radix-ui/react-polymorphic';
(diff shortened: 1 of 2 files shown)


packages/react/avatar/src/Avatar.tsx
```diff
@@ -63,16 +63,16 @@ const AvatarImage = React.forwardRef((props, forwardedRef) => {
...
-  const [, setImageLoadingStatus] = useAvatarContext(IMAGE_NAME);
+  const { onImageLoadingStatusChange } = useAvatarContext(IMAGE_NAME);
   const imageLoadingStatus = useImageLoadingStatus(src);
   const onLoadingStatusChange = useCallbackRef(onLoadingStatusChangeProp);
 
   useLayoutEffect(() => {
     if (imageLoadingStatus !== 'idle') {
       onLoadingStatusChange(imageLoadingStatus);
-      setImageLoadingStatus(imageLoadingStatus);
+      onImageLoadingStatusChange(imageLoadingStatus);
```
benoitgrelard, line 73: I saw that rename in the other one too (was it `Accordion`? I think I've approved and merged that one).
Seeing it again now I'm not sure I'm fully onboard with it, why rename to and `onChange` callback, when really it's a state setter? I feel like it makes it less readable too for me, as it I can't see anymore that I'm doing `context.setSomething` which indicates I'm syncing something back up.

jjenzz: It felt weird to me to add methods to a component (`AccordionProvider`) that don't follow our callback naming convention. I see context like private props here and we don't have props `setOpen` props. 

We name our callbacks in a way that communicates what event occurred, not what the method does with the value (implementation details). It doesn't feel necessary for a child component to know the implementation detail of the context, it just needs to communicate to context that it changed and it's up to the context method to decide what to do with it. It might set state today, it might set a ref tomorrow but the API shouldn't need to change if that happens.

benoitgrelard (approved):

PR #482, opened 2021-02-24
Update context in `collection`

jjenzz:
----------

----------

### DIFF
--- a/packages/react/collection/src/collection.tsx
+++ b/packages/react/collection/src/collection.tsx
@@ -12,8 +12,8 @@ function createCollection<E extends React.ElementRef<any> = void, S = {}>(name: 
 
   type CollectionContextValue = {
     items: Item[];
-    addItem: (item: Item) => void;
-    removeItem: (ref: ItemElementRef) => void;
+    onItemAdd: (item: Item) => void;
+    onItemRemove: (ref: ItemElementRef) => void;
     // Tracks how many times `useCollectionItem` is called to get initial indexes for SSR.
     // We increment this synchronously during render in `useCollectionItem`
     // and reset it synchronously during render in `CollectionProvider`.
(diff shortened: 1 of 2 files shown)


packages/react/collection/src/collection.tsx
```diff
@@ -22,9 +22,8 @@ function createCollection<E extends React.ElementRef<any> = void, S = {}>(name:
     ssrSyncUseCollectionItemCountRef: React.MutableRefObject<number>;
   };
 
-  const [CollectionContext, useCollectionContext] = createContext<CollectionContextValue>(
-    name + 'CollectionContext',
-    providerName
+  const [CollectionProviderImpl, useCollectionContext] = createContextObj<CollectionContextValue>(
```
jjenzz, line 25: I wasn't sure what to name the `CollectionProviderImpl` part here because we already have a `CollectionProvider` in this file that wraps it.

benoitgrelard: Yeah I guess it makes sense. I've been thinking about it while reviewing these. I think the reason is because we've dropped the `Context` part in the name. When something was before `AccordionContext` and then used `<AccordionContext.Provider>`, that would translate now to `AccordionContextProvider` and here `CollectionContextProvider`.

So maybe we just need to make them all more explicit like that?

But at the same time feels kinda annoying… cause a provider is obviously a context thing…

benoitgrelard (approved):

packages/react/collection/src/collection.tsx
jjenzz: Haha yeah I dropped the `Context` part from the name because we don't do `SSRContextProvider` for example, it's just `SSRProvider`.

Are you happy for me to leave as is?

benoitgrelard: Yeah I'm happy with this.

benoitgrelard (approved):

packages/react/collection/src/collection.tsx
```diff
@@ -59,14 +58,14 @@ function createCollection<E extends React.ElementRef<any> = void, S = {}>(name:
...
     return (
-      <CollectionContext.Provider
-        value={React.useMemo(
-          () => ({ items, addItem, removeItem, ssrSyncUseCollectionItemCountRef }),
-          [items, addItem, removeItem]
-        )}
+      <CollectionProviderImpl
+        items={items}
+        addItem={addItem}
+        removeItem={removeItem}
```
jjenzz, line 64: I'll wait for #496 to merge and then update these because they should be `onItemAdd` and `onItemRemove` and then we can inline the handlers

benoitgrelard (approved):

PR #483, opened 2021-02-24
Update context in `ContextMenu`

jjenzz:
----------

----------

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { composeEventHandlers, createContext, extendComponent } from '@radix-ui/react-utils';
+import { composeEventHandlers, createContextObj, extendComponent } from '@radix-ui/react-utils';
 import { Primitive } from '@radix-ui/react-primitive';
 import { makeRect, getSelector } from '@radix-ui/utils';
 import * as MenuPrimitive from '@radix-ui/react-menu';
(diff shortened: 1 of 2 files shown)


packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -14,14 +14,13 @@ import type { Point, MeasurableElement } from '@radix-ui/utils';
 const CONTEXT_MENU_NAME = 'ContextMenu';
 
 type ContextMenuContextValue = {
-  open: boolean;
-  setOpen: React.Dispatch<React.SetStateAction<boolean>>;
   anchorPointRef: React.MutableRefObject<Point>;
   anchorRef: React.MutableRefObject<MeasurableElement | null>;
+  open: boolean;
+  onOpenChange(open: boolean): void;
```
benoitgrelard, line 20: Again, not sure how I feel here, even the change of signature, we're putting a setter in it, what if I actually want to use the `prevState => newState` syntax? I'm kinda confused why we're changing all these… I'm probably not seeing something.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -64,7 +71,7 @@ const ContextMenuTrigger = React.forwardRef((props, forwardedRef) => {
       onContextMenu={composeEventHandlers(props.onContextMenu, (event) => {
         event.preventDefault();
         const point = { x: event.clientX, y: event.clientY };
-        context.setOpen(true);
+        context.onOpenChange(true);
```
benoitgrelard, line 74: like, this doesn't read very well for me now.

packages/react/context-menu/src/ContextMenu.tsx
benoitgrelard, line 20: Again, not sure how I feel here, even the change of signature, we're putting a setter in it, what if I actually want to use the `prevState => newState` syntax? I'm kinda confused why we're changing all these… I'm probably not seeing something.

packages/react/context-menu/src/ContextMenu.tsx
benoitgrelard, line 74: This doesn't read very well to me anymore for example.

packages/react/context-menu/src/ContextMenu.tsx
jjenzz: I've explained why I renamed them [here]( https://github.com/radix-ui/primitives/pull/479#discussion_r582808283) but the reason for changing the signature was because we don't use the one you're mentioning so if we don't need it, I thought why complicate it? My thinking was that **when** we want that callback signature, we would update it then. It was one of those "only introduce complexity when inevitable" things for me.

I've actually always disliked the way we do `context.setOpen(prevOpen => !prevOpen)` and instead would have prefer nsomething like this:

```jsx
<ContextProvider onOpenToggle={() => setOpen(prevOpen => !prevOpen)} />
```

That way we're not spreading implementation detail across all the parts.

benoitgrelard (approved):

PR #488, opened 2021-02-25
[Label] Use SSR-safe ids

benoitgrelard:


packages/react/label/src/Label.tsx
```diff
@@ -29,8 +30,7 @@ const Label = React.forwardRef((props, forwardedRef) => {
   } = props;
   const labelRef = React.useRef<HTMLSpanElement>(null);
   const ref = useComposedRefs(forwardedRef, labelRef);
-  const generatedId = `label-${useId()}`;
-  const id = idProp || generatedId;
+  const id = useId(idProp);
```
benoitgrelard, line 33: Here it does mean that this one can't be overriden to `undefined`, so it's different from the other packages.

jjenzz (approved):

PR #489, opened 2021-02-25
[Tabs] Use SSR-safe ids

benoitgrelard:
----------

- [x] Change base branch to `main` once #487 is merged. 

----------

### DIFF
--- a/packages/react/tabs/src/Tabs.tsx
+++ b/packages/react/tabs/src/Tabs.tsx
@@ -1,13 +1,9 @@
 import * as React from 'react';
-import {
-  composeEventHandlers,
-  createContext,
-  useControlledState,
-  useId,
-} from '@radix-ui/react-utils';
+import { composeEventHandlers, createContext, useControlledState } from '@radix-ui/react-utils';
 import { Primitive } from '@radix-ui/react-primitive';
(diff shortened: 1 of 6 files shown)


packages/react/tabs/src/Tabs.tsx
```diff
@@ -79,14 +73,13 @@ const Tabs = React.forwardRef((props, forwardedRef) => {
...
-    () => ({ tabsId, value, setValue, orientation, activationMode }),
-    [activationMode, orientation, value, setValue, tabsId]
+    () => ({ baseId, value, setValue, orientation, activationMode }),
+    [baseId, value, setValue, orientation, activationMode]
   );
 
   return (
     <TabsContext.Provider value={ctx}>
       <Primitive
-        id={tabsId}
```
jjenzz, line 89: I'm guessing this one was redundant?

jjenzz (approved):
All goos assuming the above was intentional ☝️

packages/react/tabs/src/Tabs.tsx
benoitgrelard: yeah it's not really needed at all for accessibility.

PR #493, opened 2021-02-25
Update context in `Popper`

jjenzz:


packages/react/popper/src/Popper.tsx
```diff
@@ -12,23 +12,20 @@ import type { Side, Align, Size, MeasurableElement, Merge } from '@radix-ui/util
  * Root level context
  * -----------------------------------------------------------------------------------------------*/
 
-type PopperContextValue = {
```
benoitgrelard, line 15: Did you mean to leave the comment above?
Also, not sure I've paid attention to that in all the other ones 😬

jjenzz: No i meant to remove that, think i did in the others (just missed them in the ones I opened today 😬 )

benoitgrelard (approved):

PR #495, opened 2021-02-25
Update context in `RadioGroup`

jjenzz:


packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -95,9 +81,18 @@ const RadioGroup = React.forwardRef((props, forwardedRef) => {
...
-    <RadioGroupContext.Provider value={context}>
+    <RadioGroupProvider
+      name={name}
+      value={value}
+      required={required}
+      rovingFocus={rovingFocus}
+      onValueChange={React.useCallback(
+        composeEventHandlers(handleValueChange, (event) => setValue(event.target.value)),
+        [handleValueChange]
+      )}
```
jjenzz, line 92: This made me realise that we're passing the `event` to the consumer which doesn't seem to be inline with how we handle things elsewhere. I was expecting it to be `onValueChange(value: string): void;`. 

The docs say it has that signature but it actually doesn't 😬 

image

Shall I change it while I'm here?

benoitgrelard: hmm weird! Let's tackle separately.

benoitgrelard (approved):

jjenzz:
Before merging this, I'll wait for the context update PR to go in (#496) and then update this to remove the `React.useCallback`

jjenzz:
@benoitgrelard Happy for me to leave this as is and merge now that we've decided to close #496 ?

benoitgrelard (approved):

PR #496, opened 2021-02-25
Update `createContext` to memoize props internally

jjenzz:
You can test how this behaves here https://codesandbox.io/s/aged-darkness-wlkxr?file=/src/App.tsx

### DIFF
--- a/packages/react/utils/src/createContext.tsx
+++ b/packages/react/utils/src/createContext.tsx
@@ -18,12 +18,15 @@ export function createContextObj<ContextValueType extends object>(rootComponentN
 
   function Provider(props: ContextValueType & { children: React.ReactNode }) {
     const { children, ...providerProps } = props;
-    // Only re-memoize when prop values change
-    const value = React.useMemo(
-      () => providerProps,
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-      Object.values(providerProps)
-    ) as ContextValueType;
+    const providerPropsString = JSON.stringify(providerProps);
+    // Only re-memoize when stringified props change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    const value = React.useMemo(() => ({} as ContextValueType), [providerPropsString]);
+    // mutate properties to ensure non primitive values are up to date
+    Object.keys(providerProps).forEach((prop) => {
+      (value as any)[prop] = (providerProps as any)[prop];
+    });
+
     return <Context.Provider value={value}>{children}</Context.Provider>;
   }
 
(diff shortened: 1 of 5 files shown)


packages/react/utils/src/createContext.tsx
```diff
@@ -18,13 +18,13 @@ export function createContextObj<ContextValueType extends object>(rootComponentN
...
-      () => providerProps,
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-      Object.values(providerProps)
-    ) as ContextValueType;
-    return <Context.Provider value={value}>{children}</Context.Provider>;
+    const providerPropsString = JSON.stringify(providerProps);
+    // Only re-memoize when stringified props change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    const context = React.useMemo(() => ({} as ContextValueType), [providerPropsString]);
+    // mutate properties to ensure non primitive values are up to date
```
jjenzz, line 25:
```suggestion
    // mutate properties to ensure closures are up to date
```

packages/react/utils/src/createContext.tsx
```diff
@@ -18,13 +18,13 @@ export function createContextObj<ContextValueType extends object>(rootComponentN
...
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-      Object.values(providerProps)
-    ) as ContextValueType;
-    return <Context.Provider value={value}>{children}</Context.Provider>;
+    const providerPropsString = JSON.stringify(providerProps);
+    // Only re-memoize when stringified props change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    const context = React.useMemo(() => ({} as ContextValueType), [providerPropsString]);
+    // mutate properties to ensure closures are up to date
+    Object.entries(providerProps).forEach(([prop, value]) => ((context as any)[prop] = value));
```
benoitgrelard, line 26: I think I'm gonna need some help understanding how this actually works.

benoitgrelard (commented):
We have some build failures because some things passed in are sometimes react elements, so don't stringify.

I'm also keen to get on a call to go through this as I'm not sure I fully understand how this works.

packages/react/utils/src/createContext.tsx
jjenzz: Ah I think I need to do that `JSON.stringify` filter thing we did with Jonathan the other day. Perhaps there is another way to achieve this without `JSON.stringify` though. Let's have a call when you're here so I can explain what's needed here

jjenzz:
We've both agreed this is getting a bit out of hand, so we're gonna keep with the inline `React.useCallback` approach.

PR #503, opened 2021-02-26
[Utils refactor] new rect package

benoitgrelard:


packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -5,14 +5,17 @@ import { makeRect, getSelector } from '@radix-ui/utils';
...
+import type { Point } from '@radix-ui/utils';
 
 /* -------------------------------------------------------------------------------------------------
  * ContextMenu
  * -----------------------------------------------------------------------------------------------*/
 
 const CONTEXT_MENU_NAME = 'ContextMenu';
 
+type ExtractRef<T> = T extends React.RefObject<infer B> ? B : never;
+type MeasurableElement = ExtractRef<React.ComponentProps<typeof MenuPrimitive.Root>['anchorRef']>;
```
benoitgrelard, line 17: Is this weird? Should I name `MeasurableElement` something like `Anchor` here instead?

benoitgrelard: My thought process was basically that rather to add an explicit dependency on `@radix-ui/rect` just to get the `MeasurableElement` type, it's prob better to get it from `Menu`, and also carries more intent.

jjenzz: Yeah perhaps just `AnchorElement`?

jjenzz (approved):

packages/react/context-menu/src/ContextMenu.tsx
benoitgrelard: I went with `Anchor` and renamed the `MeasurableElement` type to `Measurable` as the whole point is that it's not always an element.

PR #505, opened 2021-02-26
Update context in `RovingFocusGroup`

jjenzz:


packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -19,15 +19,14 @@ type RovingFocusGroupOptions = {
 type RovingContextValue = {
   groupId: string;
   tabStopId: string | null;
-  setTabStopId: React.Dispatch<React.SetStateAction<string | null>>;
   reachable: boolean;
-  setReachable: React.Dispatch<React.SetStateAction<boolean | undefined>>;
+  onTabStopIdChange(itemId: string, active?: boolean): void;
```
benoitgrelard, line 23: tiny nit but I think it's nice to group the prop and change handlers together

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -46,15 +45,30 @@ function RovingFocusGroup(props: RovingFocusGroupProps) {
...
+  return (
+    <RovingFocusProvider
+      groupId={useId()}
+      tabStopId={tabStopId}
+      orientation={orientation}
+      dir={dir}
+      loop={loop}
+      reachable={reachable}
+      onReachableChange={setReachable}
+      onTabStopIdChange={handleTabStopIdChange}
```
benoitgrelard, line 67: Same for grouping here.

benoitgrelard (approved):
Approved with 2 tiny nits.

packages/react/roving-focus/src/RovingFocusGroup.tsx
jjenzz: Since you allowed my `onFoo(): void` changes from `onFoo: () => void` I'll happily oblige here 😉  I usually group all my handlers at the end but happy to see how the alternative feels

benoitgrelard: haha, I think the other places were left grouped also no?

jjenzz: coincidentally grouped perhaps haha, i tend not to look for it

benoitgrelard (approved):

PR #506, opened 2021-02-26
Update context in `ScrollArea`

jjenzz:


packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -487,9 +482,19 @@ const ScrollAreaImpl = React.forwardRef(function ScrollAreaImpl(props, forwarded
   };
 
   return (
-    <DispatchContext.Provider value={dispatch}>
-      <ScrollAreaRefsContext.Provider value={refs}>
-        <ScrollAreaContext.Provider value={context}>
+    <DispatchProvider dispatch={dispatch}>
+      <ScrollAreaRefsProvider {...refs}>
+        <ScrollAreaProvider
```
benoitgrelard, line 487: I can see the new provider here with all the props, but I can't see the `context` being deleted in the old one. What's up here?

jjenzz: Ah crap, I must have hastily pushed with `--no-verify` on Friday, slapped wrists 🙈

benoitgrelard (approved):

PR #510, opened 2021-02-26
Update context in `Toolbar`

jjenzz:


packages/react/toolbar/src/Toolbar.tsx
```diff
@@ -49,25 +49,14 @@ type ToolbarPrimitive = Polymorphic.ForwardRefComponent<
   ToolbarOwnProps
 >;
 
-type ToolbarContextValue = {
-  name: typeof TOOLBAR_CONTEXT_NAME;
```
benoitgrelard, line 53: what was this about? I see it was removed, but I can't see any usage removed.

jjenzz: It wasn't being used as far as I could tell. We must have just missed it in the original PR

benoitgrelard (approved):

Issue #26, opened 2020-07-31
Tooltip: add features and address bugs


chaance:
From Colm's notes in Notion:

- Support multi-line + children. Max-width prop?
- Prevent opening on focus when focus comes from clicks on other elements)

- Also the bug Jenna mentioned where it triggers on click sometime

kylemh:
One request I would add is the ability to have the tooltip render immediately on hover or the ability to control the delay in milliseconds!

benoitgrelard:
Closing in favour of #543

PR #513, opened 2021-02-27
Allow for global control of Tooltip mouse rest duration thresholds

kylemh:
- [X] Add or edit Storybook examples to reflect the change (run `yarn dev`).

This pull request:

- [X] Adds additional features/functionality to an existing package
- [X] Updates documentation or example code
- [ ] Other

# Summary

The ongoing [w3c discussion on tooltips](https://github.com/w3c/aria-practices/issues/128) suggests that a small delay isn't a requirement. A delay in hover-based tooltip rendering also has no negative affect on accessibility for the tooltip itself.

There is support for this feature via the unerring measurement of comment emoji count on my comment on #26

I believe there should _also_ be an ability to control both timeouts via a prop on the Trigger primitive; however, a window-based solution seems ideal for those who want a global change to the thresholds for the sake of continuing to emulate an Operating System-like environment within their app.

I would've implemented the prop-based one myself, but I am unfamiliar with state machines. I am unaware of whether or not prop-driven state machine initializations are viable (let alone possible).

## Failing Test

jest seems unaware of `DOMRect` and I found no solutions online for properly mocking it. I didn't want to try auto-mocking, since it seems like a required implementation detail to get the tooltip content actually rendering.

Perhaps something like Cypress is a better solution for tests like this 👀 . Willing to implement it for ya.

### DIFF
--- a/packages/react/tooltip/src/Tooltip.stories.tsx
+++ b/packages/react/tooltip/src/Tooltip.stories.tsx
@@ -635,7 +635,6 @@ export const Chromatic = () => (
             className={chromaticTriggerClass}
             style={{
               position: 'absolute',
-              side: 10,
               ...((side === 'right' || side === 'left') &&
                 (align === 'start'
                   ? { bottom: 10 }
--- a/packages/react/tooltip/src/Tooltip.stories.tsx
+++ b/packages/react/tooltip/src/Tooltip.stories.tsx
@@ -364,6 +364,51 @@ export const SlottableContent = () => (
   </Tooltip>
 );
 
+export const ControlledRestThresholdDuration = () => {
+  const [restThresholdDuration, setRestThresholdDuration] = React.useState(300);
+  const [skipRestThresholdDuration, setSkipRestThresholdDuration] = React.useState(300);
+
+  React.useEffect(() => {
+    window.RADIX_TOOLTIP_REST_THRESHOLD_DURATION = restThresholdDuration;
+    window.RADIX_TOOLTIP_SKIP_REST_THRESHOLD_DURATION = skipRestThresholdDuration;
+  }, [restThresholdDuration, skipRestThresholdDuration]);
+
+  const setRestThresholdsToZero = () => {
+    setRestThresholdDuration(0);
+    setSkipRestThresholdDuration(0);
+  };
+
+  const resetRestThresholds = () => {
+    setRestThresholdDuration(300);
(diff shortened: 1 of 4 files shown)


colmtuite:
> I believe there should also be an ability to control both timeouts via a prop

Can you talk more about this? I'm not sure I can think of a use case where a single tooltip should differ from other tooltips?

kylemh:
I guess I'd frame that question instead as: I can't think of why we'd limit that ability?

To answer your question anyways: we have a a design like the Slack team navigation bar. In this area, design wants tooltips to render immediately. We also have tooltips for column headers in a table view; however, these my design team wants to have render after a small delay.

jjenzz:
Thank you for raising this and contributing to the primitives! 

I agree it would be useful to configure but I'm not convinced that configuration via mutating the `window` object is the way to go.

Our components are very low-level with the intention that they would be wrapped for customisations like these. I would envisage we would add a prop add users can create their own preconfigured `Tooltip` from ours as follows:

```jsx
// components/Tooltip.tsx
import * as TooltipPrimitive from '@radix-ui/react-tooltip';
const Tooltip = TooltipPrimitive.Root;
const TooltipTrigger = TooltipPrimitive.Trigger;
const TooltipContent = (props) => {
  return <StyledTooltipContent delay={500} {...props} />
}
const StyledTooltipContent = styled(TooltipPrimitive.Content, {});
export { Tooltip, TooltipTrigger, TooltipContent };
```

That way all tooltips in your app will have your chosen consistent timing but can also be configured case-by-base by passing a different delay.

We have also spoken about potentially exporting a `TooltipProvider` for these sorts of "global" configurations in the past so that is also an option.

kylemh:
That a context provider seems way better! I also commented on a prop methodology too; however, I wouldn't know how to pass that prop along to the state machine properly 😞.

Any tips?

Feels like if I get that sorted, I could ship both of your ideas.

jjenzz:
I'm interested to hear @benoitgrelard's thoughts on which direction he thinks we should take here. The prop suggestion gets my vote but I recall some uncertainty surrounding that approach when we discussed it a while ago. 

Regarding passing it to the machine and tests, I'll try wrap up some bits I'm working on today and have a closer look at this for you later/tomorrow 🙂

kylemh:
If you want me to do something, I'm in the Discord server so just @ me.

PR #515, opened 2021-03-01
[Utils refactor] Rejig geometry utils

benoitgrelard:


packages/core/popper/src/popper.test.ts
```diff
@@ -1,113 +0,0 @@
-import { getSideAccountingForCollisions, getAlignAccountingForCollisions } from './popper';
```
benoitgrelard, line 1: I deleted this test file as it was testing internals rather than public API anyway and also, we're better covered with our visual tests on this.

packages/core/utils/src/geometry.ts
```diff
@@ -1,64 +0,0 @@
...
-
-type Point = { x: number; y: number };
-
-type Size = { width: number; height: number };
-
-/**
- * Creates a rect (`ClientRect`) based on a Size and and a position (x, y).
- * This is useful to compute the rect of an element without having to actually move it.
- */
-function makeRect({ width, height }: Size, { x, y }: Point): ClientRect {
```
benoitgrelard, line 18: Looking at this function, it was the only one reused so we said we'd move it into `@radix-ui/rect` but then I reconsidered given it's just using `DOMRect.fromRect`. I thought back then this API I exposed was nicer, but with spreading it's just as convenient, so you'll see I've updated all use cases accordingly.

packages/core/popper/src/popper.ts
```diff
@@ -105,7 +109,7 @@ export function getPlacementData({
   }
 
   // create a new rect as if element had been moved to new placement
-  const popperRect = makeRect(popperSize, popperPoint);
+  const popperRect = DOMRect.fromRect({ ...popperSize, ...popperPoint });
```
jjenzz, line 112: NIce :+1:

jjenzz (approved):
Do Repeat Yourself - I think we should make this a thing 😂

PR #517, opened 2021-03-01
[Utils refactor] delete utils package – remove selector prop 🔥 

benoitgrelard:
----------

- [x] Change base branch to `main` once #516 is merged. 

----------

I'll add versions file after the above is merged ☝️ 

----------

This PR complete the dismantling of `@radix-ui/utils`.
What remained in `domUtils` have been either inlined, or deleted.

A big part of this is the removal of all our `selector` stuff.

Obviously these changes flagged lots of things in Chromatic, I went through carefully and accepted the changes.

> Note: Sorry for the huge PR, I've left lots of comments on the relevant bits to bring to your attention. The rest is just churning through it. So perhaps you can start by just going through the comments.

PR #519, opened 2021-03-02
[React utils refactor] move composeEventHandlers to new @radix-ui/primitive package

benoitgrelard:
----------

- [x] Change base branch to `main` once #517 is merged. 

----------


packages/core/primitive/src/primitive.tsx
```diff
@@ -1,13 +1,15 @@
-export function composeEventHandlers<E extends React.SyntheticEvent | Event>(
+function composeEventHandlers<E>(
```
benoitgrelard, line 1: Removed the types to make it more generic as it's now a core package

jjenzz (approved):

Issue #500, opened 2021-02-26
[ScrollArea] Uses feature not present in Safari 13 — browser support (general)


benoitgrelard:
We have had [reports](https://twitter.com/emplums/status/1364681789227196422) of the documentation website breaking (because we use ScrollArea for the nav and other).
`ScrollArea` uses the media query list change event which isn't supported in Safari 13.

```tsx
export function onPrefersReducedMotionChange(
  callback: (prefers: boolean) => any,
  globalWindow: Window & typeof globalThis = window
) {
  try {
    const mediaQueryList = globalWindow.matchMedia(QUERY);
    mediaQueryList.addEventListener('change', listener);
    return function () {
      mediaQueryList.removeEventListener('change', listener);
    };
  } catch (e) {}

  return function () {};

  function listener(event: MediaQueryListEvent) {
    callback(!event.matches);
  }
}
```

Generally, we need to address how we're going to deal with the fact we don't include polyfills. What should the experience be, etc.

jjenzz:
A solution worth exploring might be to add a `browserlist` to our `package.json` files which should allow Parcel to handle bundling the necessary polyfills and transforms for us via `@babel/preset-env`:

https://babeljs.io/docs/en/babel-preset-env.html

braposo:
Hey! I was adding `ScrollArea` to one of our components and was getting some issue when testing with Safari 13.1 on Catalina through BrowserStack, which led me here.

Are there any news about the fix for this issue?

In case you need more info, this is the error I'm seeing in the console (page is blank).

![image](https://user-images.githubusercontent.com/38172/114110018-71829400-98ce-11eb-9b6d-1ee35a6e7ac7.png)

benoitgrelard:
@braposo I answered on Discord already but I'll put it here in case other people happen to read this:

- We technically don't officially support anything below Safari 13.3
- We are currently reworking `ScrollArea` so that bug will be dealt with anyway soon

jjenzz:
Just to clarify, we do support Safari 13.1 on MacOS but Safari 13.3 on iOS. As @benoitgrelard has mentioned, we're currently reworking the component to solve the `13.1` issues.

braposo:
Thanks for the explanation @jjenzz @benoitgrelard! 

I wasn't able to find a way to use a polyfill just for the `change` event so ended up taking a different approach and reverting to the native scrollbar whenever `ScrollArea` isn't supported.

[...]

PR #520, opened 2021-03-02
[React utils refactor] inline code that won't live in new packages

benoitgrelard:


packages/react/utils/src/index.ts
```diff
@@ -2,13 +2,10 @@ export * from './createContext';
 export * from './extendComponent';
 export * from './useCallbackRef';
 export * from './useComposedRefs';
-export * from './useConstant';
 export * from './useControlledState';
 export * from './useDisableBodyPointerEvents';
 export * from './useDocumentRef';
-export * from './useId';
```
benoitgrelard, line 9: fully replaced by `@radix-ui/react-id`

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -1865,6 +1863,41 @@ function useBorderBoxResizeObserver(
   }, [onResize, ref]);
 }
 
+function usePrefersReducedMotion(nodeRef: React.RefObject<Element>) {
```
benoitgrelard, line 1866: Simplified and inlined

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -1865,6 +1863,41 @@ function useBorderBoxResizeObserver(
...
+  }, [nodeRef]);
+
+  return prefersReducedMotion;
+}
+
+/**
+ * React hook for creating a value exactly once.
+ * @see https://github.com/Andarist/use-constant
+ */
+function useConstant<ValueType>(fn: () => ValueType): ValueType {
```
benoitgrelard, line 1893: inlined as is

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -1865,6 +1863,41 @@ function useBorderBoxResizeObserver(
...
+        setPrefersReducedMotion(!event.matches);
+      }
+      const mediaQueryList = globalWindow.matchMedia('(prefers-reduced-motion: no-preference)');
+      mediaQueryList.addEventListener('change', handleChange);
+      return function () {
+        mediaQueryList.removeEventListener('change', handleChange);
+      };
+    } catch (e) {}
+
+    return function () {};
```
jjenzz, line 1883: I know this wasn't you but if we're simplifying, why do we need the `try`/`catch` and the empty `return function () {}` here?

```suggestion
    function handleChange(event: MediaQueryListEvent) {
      setPrefersReducedMotion(!event.matches);
    }
    const mediaQueryList = globalWindow.matchMedia('(prefers-reduced-motion: no-preference)');
    mediaQueryList.addEventListener('change', handleChange);
    
    return () => {
      mediaQueryList.removeEventListener('change', handleChange);
    };
```

jjenzz (approved):

packages/react/scroll-area/src/ScrollArea.tsx
benoitgrelard: My guess is he was trying to get around the fact that some browsers don't support this. But given this issue #500 I'd say it doesn't really work as this is the error browsers that don't support this throw:

error

In fact, I don't really understand why the try catch doesn't catch that :think

PR #521, opened 2021-03-02
[React utils refactor] new react-use-layout-effect package

benoitgrelard:
Note: I'll move all the "utils" package inside the "react/utils" folder at the end once `react-utils` doesn't exist as it currently occupies that folder.


packages/react/use-layout-effect/src/useLayoutEffect.tsx
```diff
@@ -7,4 +7,6 @@ import * as React from 'react';
  *
  * See: https://reactjs.org/docs/hooks-reference.html#uselayouteffect
  */
-export const useLayoutEffect = Boolean(globalThis?.document) ? React.useLayoutEffect : () => {};
+const useLayoutEffect = Boolean(globalThis?.document) ? React.useLayoutEffect : () => {};
+
+export { useLayoutEffect };
```
jjenzz, line 12: Consistency king 👑 😁 :+1:

jjenzz (approved):

PR #523, opened 2021-03-02
[React utils refactor] new react-use-previous package

benoitgrelard:


packages/react/use-previous/src/usePrevious.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
 
-export function usePrevious<T>(value: T) {
+function usePrevious<T>(value: T) {
   // The ref object is a generic container whose current property is mutable ...
   // ... and can hold any value, similar to an instance property on a class
-  const ref = React.useRef<T>();
+  const ref = React.useRef<T>(value);
```
benoitgrelard, line 6: Added the initial value as we discussed

jjenzz (approved):

PR #527, opened 2021-03-02
[Accordion] Support multiple values

jjenzz:
- Adds support for multiple values
- Toggles items by default (clicking button again will close item)
- To prevent items from closing, consumer now needs to control it
- Updates tests
- Updates stories

### DIFF
--- a/packages/react/accordion/src/Accordion.stories.tsx
+++ b/packages/react/accordion/src/Accordion.stories.tsx
@@ -11,74 +11,235 @@ import {
 
 export default { title: 'Components/Accordion' };
 
-export const Styled = () => (
-  <Accordion className={rootClass}>
-    <AccordionItem className={itemClass} value="one">
-      <AccordionHeader className={headerClass}>
-        <AccordionButton className={buttonClass}>One</AccordionButton>
-      </AccordionHeader>
-      <AccordionPanel className={panelClass}>
-        Per erat orci nostra luctus sociosqu mus risus penatibus, duis elit vulputate viverra
-        integer ullamcorper congue curabitur sociis, nisi malesuada scelerisque quam suscipit
-        habitant sed.
-      </AccordionPanel>
-    </AccordionItem>
-    <AccordionItem className={itemClass} value="two">
-      <AccordionHeader className={headerClass}>
-        <AccordionButton className={buttonClass}>Two</AccordionButton>
-      </AccordionHeader>
-      <AccordionPanel className={panelClass}>
-        Cursus sed mattis commodo fermentum conubia ipsum pulvinar sagittis, diam eget bibendum
-        porta nascetur ac dictum, leo tellus dis integer platea ultrices mi.
-      </AccordionPanel>
-    </AccordionItem>
-    <AccordionItem className={itemClass} value="three" disabled>
-      <AccordionHeader className={headerClass}>
-        <AccordionButton className={buttonClass}>Three (disabled)</AccordionButton>
-      </AccordionHeader>
-      <AccordionPanel className={panelClass}>
-        Sociis hac sapien turpis conubia sagittis justo dui, inceptos penatibus feugiat himenaeos
-        euismod magna, nec tempor pulvinar eu etiam mattis.
-      </AccordionPanel>
-    </AccordionItem>
-    <AccordionItem className={itemClass} value="four">
-      <AccordionHeader className={headerClass}>
-        <AccordionButton className={buttonClass}>Four</AccordionButton>
-      </AccordionHeader>
-      <AccordionPanel className={panelClass}>
-        Odio placerat <a href="#">quisque</a> sapien sagittis non sociis ligula penatibus dignissim
-        vitae, enim vulputate nullam semper potenti etiam volutpat libero.
-        <button>Cool</button>
-      </AccordionPanel>
-    </AccordionItem>
-  </Accordion>
-);
-
-export const Controlled = () => {
-  const [value, setValue] = React.useState('one');
+export const Single = () => {
+  const [valueOne, setValueOne] = React.useState('one');
+  const [valueTwo, setValueTwo] = React.useState('two');
 
   return (
-    <Accordion className={rootClass} value={value} onValueChange={(value) => setValue(value)}>
-      <AccordionItem className={itemClass} value="one">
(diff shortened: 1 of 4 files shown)


packages/react/accordion/src/Accordion.tsx
```diff
@@ -37,46 +67,136 @@ type AccordionOwnProps = Polymorphic.Merge<
...
 
+  const handleItemOpen = React.useCallback(
+    (itemValue) => setValue((prevValue = []) => [...prevValue, itemValue]),
+    [setValue]
+  );
+
+  const handleItemClose = React.useCallback(
+    (itemValue) => setValue((prevValue = []) => prevValue.filter((value) => value !== itemValue)),
+    [setValue]
+  );
```
jjenzz, line 157: This is a good example of why it's nice to have the callback implementation in the component that provides the context. I have the same context API but two different handler implementations 👍

packages/react/accordion/src/Accordion.tsx
```diff
@@ -37,46 +67,136 @@ type AccordionOwnProps = Polymorphic.Merge<
...
+  Polymorphic.IntrinsicElement<typeof Primitive>,
+  AccordionImplOwnProps
+>;
+
+const AccordionImpl = React.forwardRef((props, forwardedRef) => {
+  const { disabled, ...accordionProps } = props;
+  const buttonNodesRef = React.useRef<Set<React.ElementRef<typeof AccordionButton>>>(new Set());
+  const accordionRef = React.useRef<React.ElementRef<typeof AccordionImpl>>(null);
+  const composedRefs = useComposedRefs(accordionRef, forwardedRef);
+
```
jjenzz, line 199: These aren't new https://github.com/radix-ui/primitives/pull/527/files#diff-4da10955329003056dec9c69a6f4f8b1fe4542d50f1747908130d6bc50a22cbaL70-L73

packages/react/accordion/src/Accordion.tsx
```diff
@@ -85,7 +205,7 @@ const Accordion = React.forwardRef((props, forwardedRef) => {
       return;
     }
 
-    const buttonNodes = [...buttonNodesRef.current].filter((node) => !(node && node.disabled));
+    const buttonNodes = [...buttonNodesRef.current].filter((node) => !node?.disabled);
```
jjenzz, line 208: Slipped this change in, forgive me 🙈

packages/react/accordion/src/Accordion.stories.tsx
```diff
@@ -11,74 +11,235 @@ import {
...
+          </AccordionHeader>
+          <AccordionPanel className={panelClass}>
+            Odio placerat <a href="#">quisque</a> sapien sagittis non sociis ligula penatibus
+            dignissim vitae, enim vulputate nullam semper potenti etiam volutpat libero.
+            <button>Cool</button>
+          </AccordionPanel>
+        </AccordionItem>
+      </Accordion>
+
+      <h1>Controlled (without close)</h1>
```
benoitgrelard, line 104: I got confused by this title as I was expected none to close, but I guess this is maybe more like 

```suggestion
      <h1>Controlled (at least one open)</h1>
```

?

packages/react/accordion/src/Accordion.test.tsx
```diff
@@ -60,44 +61,164 @@ describe('given a basic Accordion', () => {
...
+    let panelTwo: HTMLElement | null;
+
     beforeEach(async () => {
       button = rendered.getByText('Button One');
       fireEvent.click(button);
       await waitFor(() => {
-        panel = rendered.getByText('Panel One');
-        expect(panel).toBeVisible();
+        panelOne = rendered.getByText('Panel One');
+        return panelOne;
```
benoitgrelard, line 73: what does it do to return it here? Is that necessary considering it's not assigned?

packages/react/accordion/src/Accordion.test.tsx
```diff
@@ -60,44 +61,164 @@ describe('given a basic Accordion', () => {
...
+    let button: HTMLElement;
+    let panelOne: HTMLElement | null;
+    let panelTwo: HTMLElement | null;
+
+    beforeEach(async () => {
+      button = rendered.getByText('Button One');
+      fireEvent.click(button);
+      await waitFor(() => {
+        panelOne = rendered.getByText('Panel One');
+        return panelOne;
```
benoitgrelard, line 181: same question

packages/react/accordion/src/Accordion.tsx
```diff
@@ -37,46 +67,136 @@ type AccordionOwnProps = Polymorphic.Merge<
...
   const [value, setValue] = useControlledState({
     prop: valueProp,
     defaultProp: defaultValue,
-    onChange: (value) => value && onValueChange(value),
+    onChange: onValueChange,
+  });
+
+  return (
+    <AccordionValueProvider
+      value={value ? [value] : []}
```
benoitgrelard, line 98: Nice, so from the context perspective it's always an array?

packages/react/accordion/src/Accordion.tsx
benoitgrelard: Love that

benoitgrelard (commented):
This is great! Love the approach!
Left a few minor comments.

packages/react/accordion/src/Accordion.stories.tsx
jjenzz: Ah yes, that is exactly what it is, thanks :+1:

packages/react/accordion/src/Accordion.test.tsx
jjenzz: I thought `waitFor` expects a promise returned to is knows what to wait for... could be wrong

packages/react/accordion/src/Accordion.tsx
jjenzz: You got it :+1:

packages/react/accordion/src/Accordion.test.tsx
jjenzz: Yeah, `return`s are completely redundant... removed!

benoitgrelard (approved):

PR #536, opened 2021-03-03
Props/spread order

benoitgrelard:
This PR addresses a few issues brought up by some users whereby we had some cases of props that couldn't be overriden.

### DIFF
--- a/packages/react/radio-group/src/RadioGroup.tsx
+++ b/packages/react/radio-group/src/RadioGroup.tsx
@@ -124,10 +124,10 @@ const RadioGroupRovingFocusItem = React.forwardRef((props, forwardedRef) => {
   const rovingFocusProps = useRovingFocus({ disabled, active: isChecked });
   return (
     <RadioGroupItemImpl
+      disabled={disabled}
       {...itemProps}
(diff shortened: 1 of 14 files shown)


packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -157,20 +157,17 @@ type RadioGroupItemImplPrimitive = Polymorphic.ForwardRefComponent<
...
   const isChecked = context.value === props.value;
-  const handleChange = composeEventHandlers(itemProps.onCheckedChange, context.onValueChange);
 
   return (
     <Radio
-      {...itemProps}
+      required={context.required}
+      checked={isChecked}
+      {...props}
       name={context.name}
```
jjenzz, line 168: I'm guessing you left this one here because the prop it omitted from types?

jjenzz (approved):

packages/react/radio-group/src/RadioGroup.tsx
benoitgrelard: nope, missed! 👀

PR #537, opened 2021-03-04
[Presence][Collapsible][Accordion] Add height CSS property, remove transition support from Presence

jjenzz:
### Description

- Expose height CSS property on `CollapsibleContent`
- Expose height CSS property on `AccordionPanel`
- Removes transition support from `Presence` to make above possible (removes `requestAnimationFrame`s)
- Fix `Checkbox` Animated story

### DIFF
--- a/packages/react/accordion/src/Accordion.test.tsx
+++ b/packages/react/accordion/src/Accordion.test.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { axe } from 'jest-axe';
-import { render, fireEvent, RenderResult, waitFor } from '@testing-library/react';
+import { render, fireEvent, RenderResult } from '@testing-library/react';
 import * as Accordion from './Accordion';
 
 const ITEMS = ['One', 'Two', 'Three'];
@@ -18,13 +18,6 @@ describe('given a single Accordion', () => {
     expect(await axe(rendered.container)).toHaveNoViolations();
   });
 
-  it('should have no accessibility violations with an open panel', async () => {
-    const button = rendered.getByText('Button Two');
-    fireEvent.click(button);
-    await waitFor(() => rendered.getByText('Panel Two'));
-    expect(await axe(rendered.container)).toHaveNoViolations();
-  });
-
   describe('when navigating by keyboard', () => {
     beforeEach(() => {
       const button = rendered.getByText('Button One');
@@ -32,28 +25,28 @@ describe('given a single Accordion', () => {
     });
 
     describe('on `ArrowDown`', () => {
-      it('should move focus to the next button', async () => {
+      it('should move focus to the next button', () => {
         fireEvent.keyDown(document.activeElement!, { key: 'ArrowDown' });
         expect(rendered.getByText('Button Two')).toHaveFocus();
       });
     });
 
     describe('on `ArrowUp`', () => {
-      it('should move focus to the previous button', async () => {
+      it('should move focus to the previous button', () => {
         fireEvent.keyDown(document.activeElement!, { key: 'ArrowUp' });
         expect(rendered.getByText('Button Three')).toHaveFocus();
       });
     });
 
     describe('on `Home`', () => {
-      it('should move focus to the first button', async () => {
+      it('should move focus to the first button', () => {
         fireEvent.keyDown(document.activeElement!, { key: 'Home' });
         expect(rendered.getByText('Button One')).toHaveFocus();
       });
     });
 
     describe('on `End`', () => {
-      it('should move focus to the last button', async () => {
+      it('should move focus to the last button', () => {
         fireEvent.keyDown(document.activeElement!, { key: 'End' });
         expect(rendered.getByText('Button Three')).toHaveFocus();
       });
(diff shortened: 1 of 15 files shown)


packages/react/checkbox/src/Checkbox.stories.tsx
```diff
@@ -91,14 +91,34 @@ export const WithinForm = () => {
...
+          <CheckboxIndicator className={animatedIndicatorClass} />
+        </Checkbox>
+      </p>
+
+      <button
+        type="button"
+        onClick={() =>
+          setChecked((prevIsChecked) =>
+            prevIsChecked === 'indeterminate' ? false : 'indeterminate'
+          )
```
benoitgrelard, line 114: Is this actually a "fix" for what we saw?

packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -2,6 +2,8 @@ import * as React from 'react';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { createContext } from '@radix-ui/react-context';
 import { useControllableState } from '@radix-ui/react-use-controllable-state';
+import { useLayoutEffect } from '@radix-ui/react-use-layout-effect';
+import { useComposedRefs } from '@radix-ui/react-compose-refs';
```
benoitgrelard, line 6: Need to add these as dependencies

packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -120,34 +122,87 @@ type CollapsibleContentOwnProps = Polymorphic.Merge<
...
+const CollapsibleContentImpl = React.forwardRef((props, forwardedRef) => {
+  const { present, children, ...contentProps } = props;
+  const context = useCollapsibleContext(CONTENT_NAME);
+  const [isPresent, setIsPresent] = React.useState(present);
+  const ref = React.useRef<React.ElementRef<typeof Primitive>>(null);
+  const composedRefs = useComposedRefs(forwardedRef, ref);
+  const heightRef = React.useRef<number | undefined>(0);
+  const height = heightRef.current;
+  // when opening we want it to immediately open to retrieve dimensions
+  // when closing we delay `present` to retreive dimensions before closing
```
benoitgrelard, line 160: small typo

```suggestion
  // when closing we delay `present` to retrieve dimensions before closing
```

packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -120,34 +122,87 @@ type CollapsibleContentOwnProps = Polymorphic.Merge<
...
+      // kick off any animations/transitions that were originally set up
+      node.style.transition = originalTransition;
+      node.style.animation = originalAnimation;
+      setIsPresent(present);
+    }
+    /**
+     * depends on `context.open` because this will change to `false`
+     * immediately (before animations/transitions have run) which
+     * allows us to retrieve the dimensions before closing.
+     */
```
benoitgrelard, line 185: is this still relevant now that we don't have rafs?

benoitgrelard (commented):
Looks good! A few comments/questions.

packages/react/collapsible/src/Collapsible.tsx
jjenzz: Yeah still relevant because `context.open` will be `false` when a close is triggered but `present` will be `false` on animation end (so when close finishes).

packages/react/checkbox/src/Checkbox.stories.tsx
jjenzz: Nope, this is just a copy/paste of the `indeterminate` story so that I could show an animated example of all states.

benoitgrelard: So do we actually have an issue with checkbox animation or not? Or was it resolved by removing the rafs?

jjenzz: There wasn't an issue with the checkbox animation, it looked broken because when the checkbox was `unchecked` it had no height so it wasn't visible even though it *was* fading out.

I've added height to `&[data-state="unchecked"]` to fix that.

packages/react/accordion/src/Accordion.test.tsx
```diff
@@ -18,42 +18,35 @@ describe('given a single Accordion', () => {
     expect(await axe(rendered.container)).toHaveNoViolations();
   });
 
-  it('should have no accessibility violations with an open panel', async () => {
-    const button = rendered.getByText('Button Two');
-    fireEvent.click(button);
-    await waitFor(() => rendered.getByText('Panel Two'));
-    expect(await axe(rendered.container)).toHaveNoViolations();
-  });
```
jjenzz, line 26: moved this into the open panel `describe` block below

packages/react/accordion/src/Accordion.test.tsx
```diff
@@ -125,41 +124,34 @@ describe('given a multiple Accordion', () => {
     expect(await axe(rendered.container)).toHaveNoViolations();
   });
 
-  it('should have no accessibility violations with an open panel', async () => {
-    const button = rendered.getByText('Button Two');
-    fireEvent.click(button);
-    await waitFor(() => void rendered.getByText('Panel Two'));
-    expect(await axe(rendered.container)).toHaveNoViolations();
-  });
```
jjenzz, line 133: Same, moved into open panel `describe` block

benoitgrelard (approved):

PR #541, opened 2021-03-08
[Popper] Fix potential issues with non-measured positioning

### LINKED ISSUE #540: [DropdownMenu] first item is sometimes selected when opening menu
### Current Behavior

When using `side === 'bottom'` and `sideOffset === 0` and no padding anywhere (basically if the first item is flush with the trigger), the first item sometimes gets selected after clicking on the trigger towards the bottom of it.

### Expected behavior

It shouldn't get selected.

### Reproducible example

https://codesandbox.io/s/mystifying-saha-p031u?file=/src/App.js

### Suggested solution

We run our `onSelect` on `mouseup` in order to be able to open and select in one fell swoop like MacOS menu/select.
It seems that perhaps the menu shows up before it's completely in place or there's some kind of timing issue.

### Additional context

This issue was initially brought to our attention here: https://github.com/radix-ui/design-system/issues/305

benoitgrelard:
Closes #540 

The issue is kinda intricate but I will try and give a summary here…

- For a popper to position itself it needs a few measurements (anchor react, popper size)
- Until these measurements are available, the popper needs to be measurable, but not visible, nor impact the flow/layout
- For this, we return some default styles where it is positioned fixed, out of the flow, and also had `pointer-events: none` to avoid catching any events
- That said, we can see here that some events were still caught (including hover)
- This is because `DropdownMenu` and others use `DismissableLayer` and sometimes set `disabledOutsidePointerEvents: true`
- This in turn puts `pointerEvents: 'none'` on `document.body` and `pointerEvents: 'auto'` on the content part, so in this case it puts it on what ends up being the primitive wrapped by the outer popper div
- So even though in the popper we say prevent pointer events, it gets re-enabled

### DIFF
--- a/packages/core/popper/src/popper.ts
+++ b/packages/core/popper/src/popper.ts
@@ -359,7 +359,7 @@ const UNMEASURED_POPPER_STYLES: CSS.Properties = {
   top: 0,
   left: 0,
   opacity: 0,
-  pointerEvents: 'none',
+  transform: 'translate3d(0, -200%, 0)',
 };
 
 const UNMEASURED_ARROW_STYLES: CSS.Properties = {
--- a/.yarn/versions/b4dfb97b.yml
+++ b/.yarn/versions/b4dfb97b.yml
@@ -1,0 +1,11 @@
+releases:
+  "@radix-ui/popper": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-popper": patch
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives


packages/core/popper/src/popper.ts
```diff
@@ -359,7 +359,7 @@ const UNMEASURED_POPPER_STYLES: CSS.Properties = {
   top: 0,
   left: 0,
   opacity: 0,
-  pointerEvents: 'none',
+  transform: 'translate3d(0, -200%, 0)',
```
benoitgrelard, line 362: Whilst writing the recap in the issue description, I'm still wondering whether it makes more sense to actually apply the `pointerEvents` properly as we actually have what it takes to do so, and it seems like a more precise fix for the situation described.

It's that different from what we already do there:
https://github.com/radix-ui/primitives/blob/main/packages/react/popper/src/Popper.tsx#L100-L104

jjenzz: It seems we created a conflict of issues because the element is on the page during measuring phase when it shouldn't be. If it's off the page, it can't get pointer events so `pointerEvents: none` wouldn't really matter. 

I'm happy with either approach tho tbh 🙂

benoitgrelard: Yeah I suppose that's true haha
So I guess from that standpoint this fix is just as legitimate then. Let's roll with it.

jjenzz (approved):

PR #542, opened 2021-03-08
[Accordion] Move disabled prop after spread

jjenzz:
### Description

We moved props above spread to give consumers control but the variable here was already handling the prop value fallback. By moving it above the spread, the spread was overriding it completely because we do not destructure `disabled` from props.

```jsx
const disabled = accordionContext.disabled || props.disabled;
```

We had it the way it was because the expectation is:

- Root disabled = All items disabled
- **Root disabled with an item enabled = All items disabled**
- Root enabled = All items enabled
- Root enabled with an item disabled = Item disabled

I've updated the Chromatic stories to catch this in future.

### DIFF
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -301,9 +301,9 @@ const AccordionItem = React.forwardRef((props, forwardedRef) => {
     <AccordionItemProvider open={open} disabled={disabled} buttonId={buttonId}>
       <CollapsiblePrimitive.Root
         data-state={open ? 'open' : 'closed'}
-        disabled={disabled}
         {...accordionItemProps}
         ref={forwardedRef}
+        disabled={disabled}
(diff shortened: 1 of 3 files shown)


packages/react/accordion/src/Accordion.tsx
```diff
@@ -301,9 +301,9 @@ const AccordionItem = React.forwardRef((props, forwardedRef) => {
     <AccordionItemProvider open={open} disabled={disabled} buttonId={buttonId}>
       <CollapsiblePrimitive.Root
         data-state={open ? 'open' : 'closed'}
-        disabled={disabled}
         {...accordionItemProps}
         ref={forwardedRef}
+        disabled={disabled}
```
benoitgrelard, line 306: I think it might be more consistent if we stick to the way we do it everywhere else lie:

```tsx
disabled={context.disabled}
{...props}
```

PR #546, opened 2021-03-09
[Toggle] Rename `ToggleButton` to `Toggle` and `toggled` to `pressed` 🔥 

jjenzz:
### Description

- Renames package
- Updates some tsdoc comments for consistency. This actually made me wonder if the `toggled` prop should be renamed to `pressed`. We use `checked` for `Switch` instead of `toggled` so that might help differentiate the two more. One is "pressable" the other is "checkable" haha.


packages/react/toggle/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@radix-ui/react-toggle-button",
-  "version": "0.0.6",
+  "name": "@radix-ui/react-toggle",
+  "version": "0.0.0",
```
benoitgrelard, line 3: 👍

packages/react/toggle/src/Toggle.tsx
```diff
@@ -5,30 +5,32 @@ import { Primitive } from '@radix-ui/react-primitive';
...
+const NAME = 'Toggle';
 const DEFAULT_TAG = 'button';
 
-type ToggleButtonOwnProps = Polymorphic.Merge<
+type ToggleOwnProps = Polymorphic.Merge<
   Polymorphic.OwnProps<typeof Primitive>,
   {
-    /** Whether the button is toggled or not, if controlled */
+    /**
+     * The controlled pressed state of the toggle.
```
benoitgrelard, line 15: Just not sure about these renamed people don’t really need to know that it’s toggling `aria-pressed` under the hood. Personally I find pressed confusing anyway in aria because for me pressed is until I release.

packages/react/toggle/src/Toggle.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const {
+    as = DEFAULT_TAG,
+    pressed: pressedProp,
+    defaultPressed = false,
+    onClick,
+    onPressedChange,
+    ...buttonProps
+  } = props;
+
+  const [pressed = false, setToggled] = useControllableState({
```
benoitgrelard, line 43: should be setPressed

PR #550, opened 2021-03-11
[Tooltip] Improve state machine separation of concerns

benoitgrelard:
Sorry, it mostly looks like completely new code now because I moved stuff into 2 separate files 😢 
I was finding it hard to follow before, so now we have one file for the state machine implementation and one for the actual tooltip statechart.

---

- updated naming conventions to match `xstate` closer (I had `camelCase` vs. `UPPER_CASE` the wrong way around for state vs. events)
- decoupled context updating from events using similar assign concept from `xstate`
- I am still injecting `send` from entry/exit actions, which apparently is not really correct looking at `xstate`

---

> Note: overall this implementation is still far from the decoupling that `xstate` has because they distinguish the concept of a pure machine vs. an interpreter. Ours is conflated basically, but I'm not sure it's worth really pushing further. I experimented today but then I feel like I'm just re-implementing a fully featured state machine library and probably doing it wrong too…


packages/react/tooltip/src/createStateMachine.tsx
```diff
@@ -0,0 +1,135 @@
...
+    const transition: Transition<State, Event, Context> | undefined = stateDefinition.on?.[type];
+
+    if (transition === undefined) {
+      if (warnOnUnknownTransitions) {
+        console.warn(
+          `From state: "${CURRENT_STATE}", event "${type}" has no transition to any state`
+        );
+      }
+    } else {
+      PREVIOUS_STATE = CURRENT_STATE;
```
jjenzz, line 73: I know you haven't changed this but it really confused me seeing uppercase variables being reassigned like this. I would usually expect all uppercase to be a constant.

jjenzz (approved):

packages/react/tooltip/src/createStateMachine.tsx
benoitgrelard: Good point. Is it ok with you if I make this change once we've merged the stack of PRs to make conflict handling easier?

jjenzz: sure no prob 👍

PR #551, opened 2021-03-11
[Tooltip] Add custom duration

### LINKED ISSUE #543: [Tooltip] Add configurable timings

benoitgrelard:
----------

- [x] Change base branch to `main` once #550 is merged. 

----------

----------

Closes #543

### DIFF
--- a/packages/react/tooltip/src/tooltipStateChart.tsx
+++ b/packages/react/tooltip/src/tooltipStateChart.tsx
@@ -2,12 +2,6 @@ import { assign } from './createStateMachine';
 
 import type { StateChart } from './createStateMachine';
 
-// How long the mouse needs to stop moving for the tooltip to open
-const REST_THRESHOLD_DURATION = 300;
-
-// How much time does the user has to move from one tooltip to another without incurring the rest wait
-const SKIP_REST_THRESHOLD_DURATION = 300;
-
 type TooltipState =
   // tooltip is closed
   | 'closed'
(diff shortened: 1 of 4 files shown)


packages/react/tooltip/src/tooltipStateChart.tsx
```diff
@@ -71,7 +65,8 @@ const tooltipStateChart: TooltipStateChart = {
         (event, context, send) => {
           restTimerId = window.setTimeout(
             () => send({ type: 'REST_TIMER_ELAPSE', id: event.id }),
-            REST_THRESHOLD_DURATION
+            // @ts-ignore
+            event.restDuration
```
benoitgrelard, line 69: I don't know how to restrict the type further here so it knows which event it is…

jjenzz: I've been thinking about this and realising I wasn't fully understanding things when we spoke on the call. I'm now wondering why we need to pass the durations through every `MOUSE_ENTER` or `MOUSE_MOVE` event.

I would usually expect something like:

```jsx
const { restDuration, bypassRestDuration, ...restProps } = props;

React.useEffect(() => {
  send({ type: 'UPDATE_REST_DURATION', restDuration });
}, [send, restDuration]);

React.useEffect(() => {
  send({ type: 'UPDATE_BYPASS_REST_DURATION', bypassRestDuration });
}, [send, bypassRestDuration]);
```

Then the machine would have the following (these are a good example of events that don't transition):

```jsx
on: {
  UPDATE_REST_DURATION: { actions: [assignRestDuration] },
  UPDATE_BYPASS_REST_DURATION: { actions: [assignBypassRestDuration] },
}
```

And your `waitingForRest` would do:


```jsx
entry: [(_, context, send) => {
  restTimerId = window.setTimeout(() => send(), context.restDuration);
}]
```

This would alleviate type issues I believe.

benoitgrelard: It seems strange to have events just to update the context though I guess.

But also I don't think that would work, because if we did that, every single instance of tooltip would override each other because there's one shared state machine for all tooltips.

We do want these timings to change on the fly based on the tooltip currently being used.

jjenzz: > It seems strange to have events just to update the context though I guess.

The event could be `PROP_CHANGE`, or whatever you wanna call it. State machine events don't have to map to DOM events, they're just "things that happened".

> We do want these timings to change on the fly based on the tooltip currently being used.

Ah of course! That was the missing piece of the puzzle, cheers. Agree we should leave as is in that case 🙂

jjenzz (approved):
Approving now that you've built on top of this more in a separate one :+1:

PR #554, opened 2021-03-12
[Tooltip] Improve on custom durations + fixes

### LINKED ISSUE #543: [Tooltip] Add configurable timings

benoitgrelard:
----------

- [x] Change base branch to `main` once #551 is merged. 

----------

----------

Closes #543 

This PR builds on top of #550. After having had a chat with Colm, he raised good points that our implementation had a few issues. Those are unrelated to the custom duration stuff, but as I worked my way through fixing these, I ended up simplifying the tooltip statechart a lot.

The issues were:
- the delay shouldn't be based on resting (not moving) over the trigger, but just a delay from mouse enter
- escape should always work, even if the trigger is not focused

> Note: on that one I ended up copying over the `useEscapeKeydown` from `DismissableLayer`. If we go ahead with all this I guess we should make it a separate package.


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -73,44 +96,51 @@ const Tooltip: React.FC<TooltipOwnProps> = (props) => {
...
 
   // if we're controlling the component
   // put the state machine in the appropriate state
   useLayoutEffect(() => {
     if (openProp === true) {
-      stateMachine.send({ type: 'FOCUS', id: contentId });
+      handleFocus();
     }
-  }, [contentId, openProp]);
+  }, [handleFocus, openProp]);
```
jjenzz, line 133: This part was weird for me. Feels like we should be calling `OPEN` here but obviously we aren't because we don't want the delay. That reasoning is super implicit atm though. I'm wondering if we should do something like this instead:

```jsx
useLayoutEffect(() => {
	if (openProp === true) {
		stateMachine.send({ type: 'OPEN', id: contentId, delayDuration: 0 });
	}
}, [openProp, contentId]);
```

🤔  ?

jjenzz (approved):
One small question but otherwise :+1:

packages/react/tooltip/src/Tooltip.tsx
benoitgrelard: Yeah I don't love this bit either.
I'm down with your suggestion, feels more explicit to what we want to do. 👍

I'll update!

PR #558, opened 2021-03-15
[Tooltip] Add unmount animation support

### LINKED ISSUE #459: [Tooltip] Consider allowing `forceMount` and `Presence`
Currently, the `Tooltip` Primitive doesn't support `forceMount` or `Presence` making it impossible to add exit animations. 

This issue is to consider adding support for those.

From Vercel

> I tried adopting[ an animation example that Jenna made for Dropdown Menu](https://codesandbox.io/s/busy-heyrovsky-7t7qf?file=/src/styles.css), but couldn’t make it work. Is this not possible with Tooltip?
> Here’s a [sandbox](https://codesandbox.io/s/objective-cookies-nhr7l?file=/src/styles.css)

benoitgrelard:
----------

- [x] Change base branch to `main` once #554 is merged. 

----------

----------

Closes #459

### DIFF
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -7,6 +7,7 @@ import { useEscapeKeydown } from '@radix-ui/react-use-escape-keydown';
 import { useLayoutEffect } from '@radix-ui/react-use-layout-effect';
 import { usePrevious } from '@radix-ui/react-use-previous';
 import { useRect } from '@radix-ui/react-use-rect';
+import { Presence } from '@radix-ui/react-presence';
 import { Primitive, extendPrimitive } from '@radix-ui/react-primitive';
 import * as PopperPrimitive from '@radix-ui/react-popper';
 import { Portal } from '@radix-ui/react-portal';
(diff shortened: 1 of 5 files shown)


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -81,6 +82,8 @@ const Tooltip: React.FC<TooltipOwnProps> = (props) => {
         } else if (state === 'closed') {
           setStateAttribute('closed');
         }
+      } else {
+        setStateAttribute('closed');
```
benoitgrelard, line 86: The key fix here was this to ensure only one tooltip remain open at once because we had an issue where the state attribute was the same on all instances.

I thought I had fixed it in previous PRs by adding the check for the tooltip id, but we also want to make sure if it's not this particular tooltip, it should be considered closed.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -100,9 +101,11 @@ const Tooltip: React.FC<TooltipOwnProps> = (props) => {
       if (context.id === contentId) {
         if (state === 'open') {
           setStateAttribute(context.delayed ? 'delayed-open' : 'instant-open');
-        } else if (state === 'closed') {
+        } else {
```
benoitgrelard, line 104: Removed this explicit check, that's what was breaking the unmount animation. It was there because that PR was based on having only `open` and `closed` state, but now that we have more it should go (wasn't needed anyway).

jjenzz (approved):

PR #564, opened 2021-03-17
Upgrade parcel to latest nightly

benoitgrelard:
This parcel release contains a fix for an issue we had gone around with this PR: #538


jjenzz:
Do we care now that we have `import * as` everywhere now for single part components? Wondering if we should import those normally like they were before now?

jjenzz (approved):

benoitgrelard:
> Do we care that we have `import * as` everywhere now for single part components? Wondering if we should import those normally like they were before now?

Haha you read my mind, I was wondering the same too, that's why I linked up the PR where we made these changes in case we wanted to revert them. I'm not sure what to do to be honest, but I also feel like it's weird when we have just one part to import as.

jjenzz:
> I'm also wondering if it even makes sense for these multipart ones to export the `Root` one…

I had wondered this before too but it would mean consumers would need to do:

```jsx
import * as CarouselPrimitive from '@radix-ui/react-carousel';
import { Carousel } from '@radix-ui/react-carousel';
```

The `Root` is there so the first line is all they need.

jjenzz:
Good point on the docs stuff... maybe we should just leave as it is for now for consistency

benoitgrelard:
Yeah let's keep things as is for now I think, at least externally. As for internally for us what do you reckon? Would you rather use named one when we're using single part imports?

jjenzz:
I think it's fine for us to be consistent and just `import * as FooPrimitive` everywhere. What do you reckon?

benoitgrelard:
Yeah I think that's fine too, makes things easier.

PR #571, opened 2021-03-22
[Presence] Add support for multiple animations

### LINKED ISSUE #568: Popover does not close when an animation is present
## Bug report

The popover (v0.0.8) does not close when clicking away if the `Popover.Content` element has some css animation.

I've tested and it works fine in v0.0.7 and v0.0.6.

### Reproducible example

[CodeSandbox](https://codesandbox.io/s/exciting-lederberg-s0mfj?file=/src/App.js:0-415)

### Your environment

MacOS, latest Chrome

jjenzz:
### Description

Bit of an oversight here 🙈  

Took me a while to figure this all out but ended up being a few simple changes in the end.

Fixes #568.

### DIFF
--- a/packages/react/presence/src/Presence.tsx
+++ b/packages/react/presence/src/Presence.tsx
@@ -90,18 +90,25 @@ function usePresence(present: boolean) {
        */
       const handleAnimationEnd = (event: AnimationEvent) => {
         const currentAnimationName = getAnimationName(stylesRef.current);
-        const isCurrentAnimation = event.animationName === currentAnimationName;
+        const isCurrentAnimation = currentAnimationName.includes(event.animationName);
         if (event.target === node && isCurrentAnimation) {
-          // if animation occurred, store its name as the previous animation
-          prevAnimationNameRef.current = getAnimationName(stylesRef.current);
           send('ANIMATION_END');
         }
       };
 
+      const handleAnimationStart = (event: AnimationEvent) => {
+        if (event.target === node) {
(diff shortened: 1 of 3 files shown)


packages/react/presence/src/Presence.tsx
```diff
@@ -90,18 +90,25 @@ function usePresence(present: boolean) {
...
-          // if animation occurred, store its name as the previous animation
-          prevAnimationNameRef.current = getAnimationName(stylesRef.current);
           send('ANIMATION_END');
         }
       };
 
+      const handleAnimationStart = (event: AnimationEvent) => {
+        if (event.target === node) {
+          // if animation occurred, store its name as the previous animation.
+          prevAnimationNameRef.current = getAnimationName(stylesRef.current);
```
jjenzz, line 102: I had to move this to `animationstart` because if the animation in doesn't complete before the button is pressed again to close, the `animationend` wouldn't have fired so the `prevAnimationName` would be incorrect.

benoitgrelard: Just to confirm, this change is unrelated to having to support multiple animations right? Just an edge case you discovered whilst fixing this, in case you toggle the state quickly?

benoitgrelard (commented):
Looks good, just one question.

packages/react/presence/src/Presence.tsx
jjenzz: You got it :+1:

benoitgrelard (approved):

PR #572, opened 2021-03-23
[All] Add browserslist to fix babel transpilation

### LINKED ISSUE #549: ReferenceError: globalThis is not defined when using DropdownMenu
### Current Behavior
I get this exception when using DropdownMenu, (doesn't happen every time, and only happens to one of my coworkers, even thought we use the same env throught docker)
image
I'm using Nextjs with SSR.

### Expected behavior
Should not throw any error

### Reproducible example

Couldn't reproduce in a codesandbox

### Suggested solution

Check that globalThis exists and is not undefined.

### Additional context
I'm sorry for not being able to provide more info on this. But thought it might be better than not reporting at all.

### Your environment

| Radix Package(s) |      DropdownMenu  |   0.0.9      |
| React            | n/a     |    17.0.1     |
| Browser          |   Chrome      |    66.0.3359.181     |
| Assistive tech   |      N/A   |         |
| Node             |  n/a     |   12      |
| npm/yarn         |   yarn      |   1.8.21p2      |
| Operating System | Linux        |    5.4.5     |

jjenzz:
### Description

Adds `@babel/runtime-corejs3` to polyfil and transpile modern features to match our chosen browser matrix.

Fixes #549


benoitgrelard (approved):

itsMapleLeaf:
This PR may have introduced an unintended side effect... corejs imports are left behind in the ESM bundles, pretty much turning corejs into a peer dependency, but I _feel_ like that wasn't intended?

[See here](https://codesandbox.io/s/kind-pike-fdjlv?file=/src/App.tsx); I'm getting the same error in my project

peduarte:
@itsMapleLeaf thanks for that <3

benoitgrelard:
It's "not" not expected shall we say haha 😀 
We're working through adding polyfilling but obviously we've messed some stuff up.
Thanks for bearing with us…

jjenzz:
@itsMapleLeaf We've removed the `core-js` polyfills. The latest release should work error-free for you now but let us know if not.

itsMapleLeaf:
Awesome, the latest update seems to work without issues. Thanks for the quick update. ❤

packages/core/popper/package.json
```diff
@@ -13,9 +13,11 @@
   "sideEffects": false,
   "scripts": {
     "clean": "rm -rf dist",
-    "prepublish": "yarn clean"
+    "prepublish": "yarn clean",
+    "version": "yarn version"
   },
   "dependencies": {
+    "@babel/runtime-corejs3": "^7.13.10",
```
penx, line 20: As far as I can see, this is being used by your babel/parcel config prepublish - but not by the peer application.

=> Shouldn't this (now @babel/runtime) be in devDependencies and not dependencies?

penx: ah maybe they are needed as per this comment, never mind

> corejs imports are left behind in the ESM bundles

benoitgrelard: It needs to be a dependency because some imports to it may be added, for things like polyfills. So it is indeed a dependency of the package in the end.

PR #580, opened 2021-03-31
[Popover][DropdownMenu][Tooltip] Replace or remove `anchorRef` 🔥

### LINKED ISSUE #447: [Popper] Measurements of refs aren't always working properly
I discovered this whilst writing some more chromatic visual regression tests.
We have a `anchorRef` prop on `Popover` which can override the default internal `context.triggerRef`.
I noticed that in some situations (like when controlling and `open` by default) the popover wouldn't show.
It seems to be because our hooks `useRect`, `useSize` etc, aren't really aware of refs reassignment.
Therefore in some situations, we may end up with an initial `null` ref.

I then realised also, that the same thing happens by simply reordering `<Popover.Trigger>` and `<Popover.Content>`.

> Note: once we fix that issue, let's add back some more of these visual regression tests I have had to skip for now because of that.

jjenzz:
### Description

--------------

ℹ️ &nbsp;&nbsp;&nbsp;Easier to review with hidden whitespace changes.

--------------

We had refs as hook dependencies which meant the hook logic would never execute if refs weren't available on mount and wouldn't re-execute if refs were re-assigned.

~This adds a new `useNode` hook that will return an up to date `node` base on a given `ref` so that the `node` can be used as the hook dependency instead.~

This PR is the first part of a wider change. This part does the following: 

- Creates a new `Anchor` part in `Popper` that will allow us to attach a callback ref internally (next PR)
- Recomposes `Anchor` in the relevant components (`Tooltip`, `DropdownMenu`, `Popover` etc.)
- 🔥 Exposes the `Anchor` from `Popover` as an optional part
- 🔥 Removes `anchorRef` capability from `DropdownMenu` and `Tooltip` altogether as we believe it's not relevant there. We'll see what feedback we get here. 
- 🔥 `data-radix-popper-wrapper` => `data-radix-popper-content-wrapper`
- Adds new Chromatic stories to reflect fixes

These changes fix #447 including the part re-ordering issue mentioned there. Next PR will fix delayed mounting of things.

### DIFF
--- a/packages/react/alert-dialog/src/AlertDialog.stories.tsx
+++ b/packages/react/alert-dialog/src/AlertDialog.stories.tsx
@@ -69,7 +69,7 @@ export const Chromatic = () => (
   <div
     style={{
       display: 'grid',
-      gridTemplateColumns: 'repeat(2, 1fr)',
+      gridTemplateColumns: 'repeat(4, 1fr)',
       gridTemplateRows: 'repeat(2, 1fr)',
       height: '100vh',
     }}
@@ -91,13 +91,46 @@ export const Chromatic = () => (
       <h2>Open</h2>
       <AlertDialog defaultOpen>
         <AlertDialogTrigger className={triggerClass}>delete everything</AlertDialogTrigger>
-        <AlertDialogOverlay className={overlayClass} style={{ right: '50%', bottom: '50%' }} />
-        <AlertDialogContent className={chromaticContentClass} style={{ top: '25%', left: '25%' }}>
+        <AlertDialogOverlay
+          className={overlayClass}
+          style={{ left: 0, bottom: '50%', width: '25%' }}
+        />
+        <AlertDialogContent className={chromaticContentClass} style={{ top: '25%', left: '12%' }}>
           <AlertDialogTitle className={titleClass}>Title</AlertDialogTitle>
           <AlertDialogDescription className={descriptionClass}>Description</AlertDialogDescription>
           <AlertDialogAction className={actionClass}>Confirm</AlertDialogAction>
           <AlertDialogCancel className={cancelClass}>Cancel</AlertDialogCancel>
         </AlertDialogContent>
+      </AlertDialog>
+    </div>
+
+    <div>
+      <h1>Uncontrolled with reordered parts</h1>
+      <h2>Closed</h2>
+      <AlertDialog>
+        <AlertDialogOverlay className={overlayClass} />
+        <AlertDialogContent className={chromaticContentClass}>
+          <AlertDialogTitle className={titleClass}>Title</AlertDialogTitle>
+          <AlertDialogDescription className={descriptionClass}>Description</AlertDialogDescription>
+          <AlertDialogAction className={actionClass}>Confirm</AlertDialogAction>
+          <AlertDialogCancel className={cancelClass}>Cancel</AlertDialogCancel>
+        </AlertDialogContent>
+        <AlertDialogTrigger className={triggerClass}>delete everything</AlertDialogTrigger>
+      </AlertDialog>
+
+      <h2>Open</h2>
+      <AlertDialog defaultOpen>
+        <AlertDialogOverlay
+          className={overlayClass}
+          style={{ left: '25%', bottom: '50%', width: '25%' }}
+        />
+        <AlertDialogContent className={chromaticContentClass} style={{ top: '25%', left: '37%' }}>
+          <AlertDialogTitle className={titleClass}>Title</AlertDialogTitle>
+          <AlertDialogDescription className={descriptionClass}>Description</AlertDialogDescription>
+          <AlertDialogAction className={actionClass}>Confirm</AlertDialogAction>
+          <AlertDialogCancel className={cancelClass}>Cancel</AlertDialogCancel>
+        </AlertDialogContent>
+        <AlertDialogTrigger className={triggerClass}>delete everything</AlertDialogTrigger>
       </AlertDialog>
     </div>
 
@@ -118,13 +151,46 @@ export const Chromatic = () => (
       <h2>Open</h2>
       <AlertDialog open>
         <AlertDialogTrigger className={triggerClass}>delete everything</AlertDialogTrigger>
-        <AlertDialogOverlay className={overlayClass} style={{ left: '50%', bottom: '50%' }} />
-        <AlertDialogContent className={chromaticContentClass} style={{ top: '25%', left: '75%' }}>
+        <AlertDialogOverlay
+          className={overlayClass}
+          style={{ left: '50%', bottom: '50%', width: '25%' }}
+        />
+        <AlertDialogContent className={chromaticContentClass} style={{ top: '25%', left: '62%' }}>
           <AlertDialogTitle className={titleClass}>Title</AlertDialogTitle>
           <AlertDialogDescription className={descriptionClass}>Description</AlertDialogDescription>
           <AlertDialogAction className={actionClass}>Confirm</AlertDialogAction>
           <AlertDialogCancel className={cancelClass}>Cancel</AlertDialogCancel>
         </AlertDialogContent>
+      </AlertDialog>
+    </div>
+
+    <div>
+      <h1>Controlled with reordered parts</h1>
+      <h2>Closed</h2>
+      <AlertDialog open={false}>
+        <AlertDialogOverlay className={overlayClass} />
+        <AlertDialogContent className={chromaticContentClass}>
+          <AlertDialogTitle className={titleClass}>Title</AlertDialogTitle>
+          <AlertDialogDescription className={descriptionClass}>Description</AlertDialogDescription>
+          <AlertDialogAction className={actionClass}>Confirm</AlertDialogAction>
+          <AlertDialogCancel className={cancelClass}>Cancel</AlertDialogCancel>
+        </AlertDialogContent>
+        <AlertDialogTrigger className={triggerClass}>delete everything</AlertDialogTrigger>
+      </AlertDialog>
+
+      <h2>Open</h2>
+      <AlertDialog open>
+        <AlertDialogOverlay
+          className={overlayClass}
+          style={{ left: '75%', bottom: '50%', width: '25%' }}
+        />
+        <AlertDialogContent className={chromaticContentClass} style={{ top: '25%', left: '88%' }}>
+          <AlertDialogTitle className={titleClass}>Title</AlertDialogTitle>
+          <AlertDialogDescription className={descriptionClass}>Description</AlertDialogDescription>
+          <AlertDialogAction className={actionClass}>Confirm</AlertDialogAction>
+          <AlertDialogCancel className={cancelClass}>Cancel</AlertDialogCancel>
+        </AlertDialogContent>
+        <AlertDialogTrigger className={triggerClass}>delete everything</AlertDialogTrigger>
       </AlertDialog>
     </div>
 
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -14,12 +14,7 @@ type Point = { x: number; y: number };
 
 const CONTEXT_MENU_NAME = 'ContextMenu';
 
-type ExtractRef<T> = T extends React.RefObject<infer B> ? B : never;
-type Anchor = ExtractRef<React.ComponentProps<typeof MenuPrimitive.Root>['anchorRef']>;
-
 type ContextMenuContextValue = {
-  anchorPointRef: React.MutableRefObject<Point>;
-  anchorRef: React.MutableRefObject<Anchor | null>;
   open: boolean;
   onOpenChange(open: boolean): void;
 };
@@ -31,21 +26,13 @@ const [ContextMenuProvider, useContextMenuContext] = createContext<ContextMenuCo
 const ContextMenu: React.FC = (props) => {
   const { children } = props;
   const [open, setOpen] = React.useState(false);
-  const anchorPointRef = React.useRef<Point>({ x: 0, y: 0 });
-  const anchorRef = React.useRef({
-    getBoundingClientRect: () =>
-      DOMRect.fromRect({ width: 0, height: 0, ...anchorPointRef.current }),
-  });
 
   return (
-    <ContextMenuProvider
-      anchorPointRef={anchorPointRef}
-      anchorRef={anchorRef}
-      open={open}
-      onOpenChange={setOpen}
-    >
-      {children}
-    </ContextMenuProvider>
+    <MenuPrimitive.Root open={open} onOpenChange={setOpen}>
+      <ContextMenuProvider open={open} onOpenChange={setOpen}>
+        {children}
+      </ContextMenuProvider>
+    </MenuPrimitive.Root>
   );
 };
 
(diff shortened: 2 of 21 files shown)


packages/react/alert-dialog/src/AlertDialog.stories.tsx
```diff
@@ -99,6 +99,18 @@ export const Chromatic = () => (
...
+      <AlertDialog defaultOpen>
+        <AlertDialogOverlay className={overlayClass} style={{ right: '50%', bottom: '50%' }} />
+        <AlertDialogContent className={chromaticContentClass} style={{ top: '25%', left: '25%' }}>
+          <AlertDialogTitle className={titleClass}>Title</AlertDialogTitle>
+          <AlertDialogDescription className={descriptionClass}>Description</AlertDialogDescription>
+          <AlertDialogAction className={actionClass}>Confirm</AlertDialogAction>
+          <AlertDialogCancel className={cancelClass}>Cancel</AlertDialogCancel>
+        </AlertDialogContent>
+        <AlertDialogTrigger className={triggerClass}>delete everything</AlertDialogTrigger>
+      </AlertDialog>
```
jjenzz, line 113: I added these stories to all of the `Trigger` components because I thought it made sense that we would test this works for all in-case ref stuff changes internally.

packages/react/dismissable-layer/src/DismissableLayer.stories.tsx
```diff
@@ -529,7 +529,7 @@ function DummyDialog({ children, openLabel = 'Open', closeLabel = 'Close' }: Dum
                     {(focusScopeProps) => (
                       <div
                         {...dismissableLayerProps}
-                        ref={composeRefs(dismissableLayerProps.ref, focusScopeProps.ref)}
```
benoitgrelard, line 532: I believe `composeRefs` probably isn't used in the file anymore.

benoitgrelard (commented):
Code-wise it all looks good to me.

- I've left a few nits and suggestions.
- I've also approved the chromatic stories for now.
- I think we need to add the same stories to `Tooltip` as well. (`anchorRef` and swapped parts)

I'm not approving yet because I want to spend some time manually testing some of the changes.

packages/react/dismissable-layer/src/DismissableLayer.stories.tsx
jjenzz: You believe correct!

packages/react/use-node/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "prepublish": "yarn clean",
+    "version": "yarn version"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.13.10",
+    "@radix-ui/react-use-layout-effect": "workspace:*"
+  },
+  "devDependencies": {
+    "@types/resize-observer-browser": "^0.1.4"
+  },
```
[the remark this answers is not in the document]
jjenzz: ooops, bad copy/paste 🙈

packages/react/label/src/Label.tsx
```diff
@@ -80,17 +81,16 @@ Label.displayName = NAME;
 
 /* -----------------------------------------------------------------------------------------------*/
 
-const useLabelContext = <E extends HTMLElement>(ref?: React.RefObject<E>) => {
+const useLabelContext = <E extends HTMLElement>(ref: React.RefObject<E> = { current: null }) => {
   const context = React.useContext(LabelContext);
+  const node = useNode(ref);
```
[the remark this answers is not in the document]
jjenzz: I've changed it to `controlRef` and `control`

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -109,24 +102,20 @@ const ContextMenuContent = React.forwardRef((props, forwardedRef) => {
...
       style={{
         ...props.style,
         // re-namespace exposed content custom property
         ['--radix-context-menu-content-transform-origin' as any]: 'var(--radix-popper-transform-origin)',
       }}
-      anchorRef={context.anchorRef}
       trapFocus
       disableOutsideScroll
       portalled
-      onDismiss={() => context.onOpenChange(false)}
```
jjenzz, line 129: `Menu` does this internally now

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -164,7 +165,6 @@ const DropdownMenuContent = React.forwardRef((props, forwardedRef) => {
         },
         { checkForDefaultPrevented: false }
       )}
-      onDismiss={() => context.onOpenChange(false)}
```
jjenzz, line 167: `Menu` does this internally now

packages/react/menu/src/Menu.stories.tsx
```diff
@@ -13,9 +15,10 @@ import {
 import { css } from '../../../../stitches.config';
 import { foodGroups } from '../../../../test-data/foods';
 
-import type * as Polymorphic from '@radix-ui/react-polymorphic';
-
-export default { title: 'Components/Menu', excludeStories: ['TickIcon', 'styledComponents'] };
+export default {
+  title: 'Components/Menu',
+  excludeStories: ['TickIcon', 'styledComponents', 'classes'],
```
jjenzz, line 20: We had a `Classes` story because we export `classes`. Just fixed that while I was here.

packages/react/popover/src/Popover.tsx
```diff
@@ -44,27 +49,60 @@ type PopoverOwnProps = {
 const Popover: React.FC<PopoverOwnProps> = (props) => {
   const { children, open: openProp, defaultOpen, onOpenChange } = props;
   const triggerRef = React.useRef<HTMLButtonElement>(null);
+  const [hasCustomAnchor, setHasCustomAnchor] = React.useState(false);
```
jjenzz, line 52: I wasn't sure what to name this state ... ideas? 🤔

packages/react/popper/src/Popper.tsx
```diff
@@ -88,29 +143,29 @@ const Popper = React.forwardRef((props, forwardedRef) => {
 
   return (
     <div style={popperStyles} data-radix-popper-wrapper="">
```
jjenzz, line 145: This is actually `data-radix-popper-content-wrapper` now but I'm worried about renaming it given consumers are using it... thoughts?

packages/react/popper/src/Popper.tsx
```diff
@@ -18,17 +18,78 @@ import type { Measurable } from '@radix-ui/rect';
...
+>;
+
+const PopperAnchor = React.forwardRef((props, forwardedRef) => {
+  const { virtualRef, children, ...anchorProps } = props;
+  const context = usePopperContext(ANCHOR_NAME);
+  const ref = React.useRef<React.ElementRef<typeof Primitive>>(null);
+  const composedRefs = useComposedRefs(
+    forwardedRef,
+    ref,
+    context.anchorRef as React.MutableRefObject<React.ElementRef<typeof Primitive>>
```
benoitgrelard, line 54: Why do we need to cast here?

packages/react/popper/src/Popper.tsx
```diff
@@ -18,17 +18,78 @@ import type { Measurable } from '@radix-ui/rect';
...
+  );
+
+  React.useEffect(() => {
+    if (virtualRef?.current) {
+      context.anchorRef.current = virtualRef.current;
+    }
+  });
```
benoitgrelard, line 61: Might be a good idea to add a comment here as to why we do this.

packages/react/popper/src/Popper.tsx
```diff
@@ -18,17 +18,78 @@ import type { Measurable } from '@radix-ui/rect';
...
+
+  return virtualRef ? (
+    <>{children}</>
```
benoitgrelard, line 64: Do we need the fragment here?

benoitgrelard: Or actually, do we even care about children of the anchor if it's a virtual ref? I guess it's useful but when I read this I think, well if I have a virtual ref, I don't have an element, so I should be able to use the `Anchor` part as a sibling, I don't need to wrap some children.

packages/react/popover/src/Popover.tsx
benoitgrelard: That reads fine to me.

packages/react/popper/src/Popper.tsx
benoitgrelard: I think we should just update it and document the breaking change. I sometimes worry about these things too but I don't think we should shy away from it. I'd rather we make the changes to keep the codebase sane.

packages/react/popover/src/Popover.tsx
```diff
@@ -44,27 +49,60 @@ type PopoverOwnProps = {
...
+
+type PopoverAnchorOwnProps = Polymorphic.OwnProps<typeof PopperPrimitive.Anchor>;
+type PopoverAnchorPrimitive = Polymorphic.ForwardRefComponent<
+  Polymorphic.IntrinsicElement<typeof PopperPrimitive.Anchor>,
+  PopoverAnchorOwnProps
+>;
+
+const PopoverAnchor = React.forwardRef((props, forwardedRef) => {
+  const context = usePopoverContext(ANCHOR_NAME);
+  const handleCustomAnchorAdd = useCallbackRef(context.onCustomAnchorAdd);
```
benoitgrelard, line 93: I think usually we haven't used `useCallbackRef` for these, but rather used `useCallback` when passing into the provider no?

packages/react/popover/src/Popover.tsx
```diff
@@ -82,7 +120,8 @@ const PopoverTrigger = React.forwardRef((props, forwardedRef) => {
   const { as = TRIGGER_DEFAULT_TAG, ...triggerProps } = props;
   const context = usePopoverContext(TRIGGER_NAME);
   const composedTriggerRef = useComposedRefs(forwardedRef, context.triggerRef);
-  return (
+
+  const primitive = (
```
benoitgrelard, line 124: perhaps rename this to `trigger`?

packages/react/menu/src/Menu.tsx
```diff
@@ -31,17 +31,48 @@ const ALL_KEYS = [...FIRST_KEYS, ...LAST_KEYS];
...
-  {
-    open?: boolean;
+type MenuOwnProps = {
+  open?: boolean;
+  onOpenChange?(open: boolean): void;
+};
 
+const Menu: React.FC<MenuOwnProps> = (props) => {
+  const { open = false, children, onOpenChange } = props;
+  const handleOpenChange = useCallbackRef(onOpenChange);
```
benoitgrelard, line 47: Same on this one, I think we usually use `React.useCallback` before passing into providers no?

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -14,12 +14,7 @@ type Point = { x: number; y: number };
 
 const CONTEXT_MENU_NAME = 'ContextMenu';
 
-type ExtractRef<T> = T extends React.RefObject<infer B> ? B : never;
-type Anchor = ExtractRef<React.ComponentProps<typeof MenuPrimitive.Root>['anchorRef']>;
-
 type ContextMenuContextValue = {
-  anchorPointRef: React.MutableRefObject<Point>;
-  anchorRef: React.MutableRefObject<Anchor | null>;
```
benoitgrelard, line 22: Nice to see all this go!

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -67,18 +54,24 @@ type ContextMenuTriggerPrimitive = Polymorphic.ForwardRefComponent<
...
-      ref={forwardedRef}
-      onContextMenu={composeEventHandlers(props.onContextMenu, (event) => {
-        event.preventDefault();
-        const point = { x: event.clientX, y: event.clientY };
-        context.onOpenChange(true);
-        context.anchorPointRef.current = point;
-      })}
-    />
+    <MenuPrimitive.Anchor virtualRef={virtualRef}>
+      <Primitive
```
benoitgrelard, line 64: Yeah so this is the other side of the comment I left earlier. These don't really need to wrap each other right? They could be siblings I think.

benoitgrelard (commented):
Absolutely ♥️ this!

Left a few comments inline.

packages/react/toolbar/src/Toolbar.stories.tsx
```diff
@@ -18,7 +18,7 @@ import {
 } from '@radix-ui/react-dropdown-menu';
 
 import { classes } from '../../menu/src/Menu.stories';
-const { rootClass: dropdownMenuRootClass, itemClass: dropdownMenuItemClass } = classes;
+const { contentClass: dropdownMenuRootClass, itemClass: dropdownMenuItemClass } = classes;
```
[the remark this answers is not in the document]
jjenzz: oh yeah, good catch, my bad 🙈

packages/react/use-rect/src/useRect.tsx
```diff
@@ -1,19 +1,19 @@
...
 
 /**
  * Use this custom hook to get access to an element's rect (getBoundingClientRect)
  * and observe it along time.
  */
-function useRect(
+function useRect<T extends Measurable>(
   /** A reference to the element whose rect to observe */
-  refToObserve: React.RefObject<Measurable>
+  refToObserve: React.RefObject<T>
```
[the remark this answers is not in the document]
jjenzz: Ah, I don't think i need this now. This was left over from when I was returning a `ref`, I wanted consumer to be able to do `useSize<HTMLDivElement>()` so that the returned ref would have the correct type.

I'll see if I can clean this up now.

packages/react/popper/src/Popper.tsx
jjenzz: Because it is moaning that `Measurable` is not a `HTMLDivElement`:

> Types of parameters 'node' and 'instance' are incompatible.
>          Type 'HTMLDivElement | null' is not assignable to type 'Measurable'.

This was also why I had tried doing the `T extends Measurable` thing but it didn't work haha. I'll take another look at this but it's changed anyway in the most recent PR.

jjenzz: Oh wait, that's because `null` isn't assignable to `Measurable` but my `anchorRef` is `Measurable | null` 🤔 

I'll take a look, i'm confused now.

packages/react/popper/src/Popper.tsx
jjenzz: Yep, happy to do that, I did it that way too initially. Just means that `ContextMenu` would be like this instead:

```jsx
<>
  <Anchor virtualRef={} />
  <Primitive />
</>
```

I quite liked that it wrapped it though because I had meant to raise whether you think it would be nice to change `virtualRef` prop to something like:

```jsx
<Anchor toPointer>
  <Primitive />
</Anchor>
```

And then it would bind the `mousemove` stuff for you on the `Primitive`. That seems like a nice API but wasn't sure if this is the only real use-case for `virtualRef` atm?

jjenzz: Ah scrap that, I remember why I didn't raise that in the end... because `ContextMenu` is `onContextMenu`... 😅 

Yeah, okay, nevermind... I'll remove `children`.

packages/react/popover/src/Popover.tsx
jjenzz: These aren't passed into the provider, they're used in the `useEffect`.

packages/react/menu/src/Menu.tsx
jjenzz: Yeah we do that when passing an internal function, but this is a prop. If the consumer passes an inline function as the `onOpenChange` prop then the context consumers would all constantly re-render, even if I do `React.useCallback(onOpenChange, [onOpenChange])`.

packages/react/popper/src/Popper.tsx
jjenzz:
```
// Consumer can anchor the popper to something that isn't
// a DOM node e.g. pointer position, so we override the
// `anchorRef` with their virtual ref in this case.
```

Is this okay? I'll add it for now and we can tweak

packages/react/popper/src/Popper.tsx
jjenzz: I haven't been able to figure this out quickly and tbh it's all [replaced anyway in the next PR](https://github.com/radix-ui/primitives/pull/592/files#diff-0b55611a47427be3297d5c6da97844bb21baf24057718c14a15d277630450ac2R56). 

Do you mind if we leave this one?

packages/react/popper/src/Popper.tsx
benoitgrelard: Looks good to me.

packages/react/popover/src/Popover.tsx
benoitgrelard: I know but they come from the provider context and we usually do the `useCallback` there where we pass them in, so then we can pass them directly to dependencies of `useEffect`.

packages/react/popper/src/Popper.tsx
benoitgrelard: Yep that's fine, I saw the next PR.

benoitgrelard:
Looking at the stories, any ideas why this moved?

image

packages/react/popover/src/Popover.tsx
jjenzz: Oh yes sorry my bad!

jjenzz:
> Looking at the stories, any ideas why this moved?

I changed the margins so that the `Open with reordered parts` story was associated with the `Controlled` or `Uncontrolled` heading better. It was just sort of floating off on its own before.

benoitgrelard (approved):

Issue #581, opened 2021-04-01
[DropdownMenu] Items that are invisible cause focus issues


jportela:
### Current Behavior

When navigating with the arrow keys through a DropdownMenu that has hidden items (ie: with `display: none`), the focus selection stops when reaching the hidden item, and as such doesn't focus all items.

### Expected behavior

The visible items are focused when pressing the arrow keys, even if there are hidden items.

### Reproducible example

1. Open the Dropdown
2. Press the arrow down key repeatedly
3. Verify that the last element doesn't get focused

[CodeSandbox Example](https://codesandbox.io/s/relaxed-mestorf-ru0rd?file=/src/App.tsx)

### Suggested solution

Filter out the invisible items. I already have a PR ready for it (that includes a Storybook example).

### Additional context

N/A

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |   `@radix-ui/react-dropdown-menu`      |   0.0.17    |
| React            | n/a     |   17      |
| Browser          |         |     Brave    |
| Assistive tech   |         |         |
| Node             | n/a     |     14    |
| npm/yarn         |         |     yarn 1.x    |
| Operating System |         |    MacOS     |

benoitgrelard:
hi @jportela thanks for reporting this.
It does look like a bug.
Can I ask you what you are trying to do here though?

jportela:
Hey @benoitgrelard , I'm hiding certain items using media queries, so they are not displayed in mobile viewports, using `display: none` (they should not "exist" for those viewports).

jportela:
An alternative would be to render two dropdown menus, one hidden depending on the viewport, which could be a working interim solution.

benoitgrelard:
Another alternative would be to simply not render the items when on smaller viewports. All you would need would be a `useMediaQuery` hook (I'm sure there are some out there already) and you'd conditionally render the items. This is probably how I chose to do it.

jportela:
that's a good alternative @benoitgrelard, thanks! Since the menu dropdown is rendered closed on the first render (server-side), there wouldn't be a noticeable diference for users.

benoitgrelard:
Hey @jportela, the latest version has now been published and contain changes to take hidden items.
https://radix-ui.com/primitives/docs/overview/releases#03052021

jportela:
Awesome! Thanks!

PR #582, opened 2021-04-01
Support proper focus when having invisible menu items

jportela:
Fixes: https://github.com/radix-ui/primitives/issues/581
Reproduceable Issue: https://codesandbox.io/s/relaxed-mestorf-ru0rd?file=/src/App.tsx

Packages changed:
- `@radix-ui/react-menu`
- `@radix-ui/react-roving-focus`

Packages affected:
- `@radix-ui/react-context-menu`
- `@radix-ui/react-dropdown-menu`
- `@radix-ui/react-radio-group`
- `@radix-ui/react-tabs`
- `@radix-ui/react-toggle-group`
- `@radix-ui/react-toolbar`

I'm not sure we need to publish all of them, but I think they might all be impacted by the change.

### Description

When navigating with the arrow keys through a DropdownMenu that has hidden items (ie: with display: none), the focus selection stops when reaching the hidden item, and as such doesn't focus all items.

This PR fixes this issue. It does so by filtering invisible items on the RovingFocus and Menu (when opening it up with the arrow keys, in case the first/last items are hidden).

I've added a Storybook example.

### DIFF
--- a/packages/react/roving-focus/src/RovingFocusGroup.tsx
+++ b/packages/react/roving-focus/src/RovingFocusGroup.tsx
@@ -114,7 +114,7 @@ function useRovingFocus({ disabled, active }: UseRovingFocusItemOptions) {
       if (focusIntent !== undefined) {
         event.preventDefault();
 
-        const items = getRovingFocusItems(context.groupId);
+        const items = filterVisibleItems(getRovingFocusItems(context.groupId));
         const count = items.length;
         const currentItem = document.activeElement as HTMLElement | null;
         const currentIndex = currentItem ? items.indexOf(currentItem) : -1;
@@ -165,6 +165,11 @@ function getRovingFocusItems(groupId: string): HTMLElement[] {
   return Array.from(document.querySelectorAll(`[${GROUP_ID_DATA_ATTR}="${groupId}"]`));
 }
 
+// items that are not visible (ie: display: none) should not be cycled through
+function filterVisibleItems(elements: HTMLElement[]) {
+  return elements.filter((item) => Boolean(item.offsetParent));
+}
+
 const Root = RovingFocusGroup;
 
 export {
@@ -173,4 +178,6 @@ export {
   Root,
   //
   useRovingFocus,
+  //
+  filterVisibleItems,
 };
--- a/.yarn/versions/0ecaafde.yml
+++ b/.yarn/versions/0ecaafde.yml
@@ -1,0 +1,12 @@
+releases:
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-radio-group": patch
+  "@radix-ui/react-roving-focus": patch
+  "@radix-ui/react-tabs": patch
+  "@radix-ui/react-toggle-group": patch
+  "@radix-ui/react-toolbar": patch
+
+declined:
+  - primitives
--- a/packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
@@ -184,6 +184,24 @@ export const PreventClosing = () => (
   </div>
 );
 
+// Contains invisible (display: none) items
+export const InvisibleItems = () => (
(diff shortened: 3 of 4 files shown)


packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -173,4 +178,6 @@ export {
   Root,
   //
   useRovingFocus,
+  //
+  filterVisibleItems,
```
jportela, line 182: we could not export this, but if so, we need to copy this logic to the Menu (considering it's just a one liner, it should be fine, if you prefer it that way)

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -165,6 +165,11 @@ function getRovingFocusItems(groupId: string): HTMLElement[] {
   return Array.from(document.querySelectorAll(`[${GROUP_ID_DATA_ATTR}="${groupId}"]`));
 }
 
+// items that are not visible (ie: display: none) should not be cycled through
+function filterVisibleItems(elements: HTMLElement[]) {
+  return elements.filter((item) => Boolean(item.offsetParent));
```
benoitgrelard, line 170: My initial thoughts here are this is probably not a good idea as reading `.offsetParent` will cause layout thrashing every single time.

Have you been able to identify what actually causes this issue in our case?

When I test other libraries, they seem to be able to get past the hidden one (allbeit with an extra keydown).

benoitgrelard (commented):
Thanks for your contribution!
Just a few initial thoughts.

packages/react/roving-focus/src/RovingFocusGroup.tsx
jportela: that's a good point regarding `.offsetParent`. Since this is done as a consequence of a user interaction (arrow press), I think it should be fine performance wise, but I understand your reticence in adding this considering this is a low-level utility.
I can try and investigate what other libraries are doing, to see if I find a better solution

benoitgrelard: I'm going to look at it on my side too and try to first understand what is happening on our side.

benoitgrelard: Ok I think I know what's happening. Here are my findings:

- browsers won't focus something that is not visible, ie. `display: none`, `visibility: hidden`, `hidden` attribute, (however they will focus `opacity: 0`)
- as far as I can tell this results in the browser not moving the focus at all from where it currently is
- because our roving focus solution is always based on the `document.activeElement` as the "current" one it won't let us go further
- the reason reakit, react-aria, etc keep working (allbeit with the extra keydown, they have the same issue basically) is probably because they use an index to know what the current one is
- I did a quick test with an index and it does work the same way as theirs, meaning that when you keydown and the next item is not visible, it won't move, but the index is incremented, so the next keydown will jump over it this time

I think the issue actually probably goes beyond roving focus but is rather a general focus issue (ie. the same issues probably exist with our focus trapping solution for example).

I'm not too sure yet if/how we should address this. Accurately knowing when items are visible is not super cheap so I'm little wary about it. I'll discuss this with @jjenzz and we'll see what we can do.

Hopefully in the meantime you can get around it using the alternative I proposed here: https://github.com/radix-ui/primitives/issues/581#issuecomment-811796630

jportela: thanks for checking!

benoitgrelard:
@jportela just wanted to come back to you with an update.

@jjenzz and I have discussed the issue and we have a plan of attack.
I am currently reworking some of our internals around focus management so it's actually kinda great timing!

For roving focus, and most other places we would potentially have this issue, we will deal with it slightly differently:
- try to focus the item we think is the next (even though it might be hidden)
- check if focus has moved
- if it hasn't moved, try the next one
- repeat

This way we avoid having to check whether the element is hidden which is costly whichever way we decide to do it.

We do have one use case where we will likely have to check it with computed styles but we think we can minimize that one too.

If that's ok with you, I'll close your PR now and your issue should be addressed in some of my changes soon.

🙏

jportela:
that sounds good @benoitgrelard feel free to close this PR.

Issue #585, opened 2021-04-05
[Dialog] Consider adding `disableOutsidePointerEvents`


Avizura:
### Overview

I need to remove `disableOutsidePointerEvents` feature in `Dialog` component but it seems impossible  https://github.com/radix-ui/primitives/blob/main/packages/react/dialog/src/Dialog.tsx#L241. 

### Additional context

Such option is already exists in DropdownMenu.Content component, would be useful to have it in Dialog as well.

benoitgrelard:
Hi @Avizura,

I don't think it would make sense to allow that for our `Dialog` primitive as it's a modal dialog.
What is your use case?

Avizura:
Hi @benoitgrelard !
I need to able to click on other cards while dialog with details of current card is opened. But now I have doubts that sidebar on the right is actually a dialog 😅
![image](https://user-images.githubusercontent.com/11854101/113599710-d8186100-95f3-11eb-829b-5c5878fff712.png)

benoitgrelard:
Yeah so if you need to be able to interact with what's behind, that would probably be considered a non-modal dialog.
I don't think our `Dialog` primitive currently supports this well so there may be something there.

`Popover` lets you create non-modal things, but then here it's not really what you need as you don't need the whole positioning thing.

I'm currently re-writing bits of our focus scope logic so it's kinda good timing actually haha.
Let me consider your use case a little bit and discuss with @jjenzz as well and we'll figure something out.

Avizura:
I see, thanks a lot for this lib guys

benoitgrelard:
Hi @Avizura I just wanted to come back to you to say that we think we will be making changes to our components such as `Dialog`, `Popover`, etc to add the concept of modality. This should then resolve your issue here.

Not sure what timeframe that will be released but I will update you once that's done.

Avizura:
@benoitgrelard Hi, cool, thanks!

jjenzz:
@Avizura Sorry for the delay on an update here. 

We're hoping to release this week with a new `modal` prop for `Dialog`. Setting it to `false` would allow you to achieve the UI you posted 🙂

Avizura:
I see, thanks again for your amazing work @jjenzz

jjenzz:
@Avizura thanks but it was @benoitgrelard and @andy-hook who did this work so I cannot take credit I'm afraid 🙂

PR #589, opened 2021-04-09
[FocusScope] Rewrite internals

### LINKED ISSUE #499: [FocusScope] Dialog resets focus when leaving & returning to browser tab/window
### Current Behavior

Dialog resets focus to first focusable element when browser tab is blurred.

Gif with reproduction

### Expected behavior

Focus should be retained when leaving browser window/tab and returning to it.

### Reproducible example

[CodeSandbox](https://codesandbox.io/s/radix-ui-dialog-focus-issue-fddwi)

### Suggested solution

This might be because of `trapFocus` re-initializing on document focus, but I'm not 100% sure:
https://github.com/radix-ui/primitives/blob/main/packages/react/focus-scope/src/createFocusScope.tsx#L30

### Your environment

| Radix Package(s) | Dialog  | 0.06    |
| React            |         | 17.0.1  |
| Browser          | Chrome  | 88.0    |
| Assistive tech   | n/a     |         |
| npm/yarn         | n/a     |         |
| Operating System | MacOs   | 10.15.6 |

Reproducible pretty much everywhere except Safari (13.1.2).

benoitgrelard:
Closes #499 

This is in preparation for work coming down the pipe to fix modality issues like #585.
These changes shouldn't affect public API.

BTW @jjenzz this also fixes the weirdness you found with `DropdownMenu` 🙂


packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -26,44 +35,265 @@ type FocusScopeProps = {
...
+
+  const scopeRef = React.useRef({
+    paused: false,
+    pause() {
+      scopeRef.current = { ...scopeRef.current, paused: true };
+    },
+    resume() {
+      scopeRef.current = { ...scopeRef.current, paused: false };
+    },
+  });
```
jjenzz, line 54: Feel free to ignore but thought it might simplify things a bit if we did `.current` here and maybe use `this` to reduce some boilerplate:

```jsx
const focusScope = React.useRef({
	paused: false,
	pause() { this.paused = true },
	resume() { this.paused = false }
}).current;
```

Then everywhere else can just do 

- `focusScope.pause()`
- `focusScope.resume()`
- `if (focusScope.paused) {`
- `activeFocusStack.pause()` 

etc...

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -26,44 +35,265 @@ type FocusScopeProps = {
...
   React.useEffect(() => {
-    const container = containerRef.current;
     if (container) {
-      container.addEventListener(AUTOFOCUS_ON_CREATE, onMountAutoFocus);
-      container.addEventListener(AUTOFOCUS_ON_DESTROY, onUnmountAutoFocus);
-      focusScopeRef.current = createFocusScope(container);
+      focusScopesStack.add(scopeRef);
+      const PREVIOUSLY_FOCUSED_ELEMENT = document.activeElement as HTMLElement | null;
+
+      // we need to setup the listeners before we `dispatchEvent`
```
jjenzz, line 61: :+1: this is all much clearer now tbh

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -26,44 +35,265 @@ type FocusScopeProps = {
...
+  return false;
+}
+
+function isSelectableInput(element: any): element is FocusableTarget & { select: () => void } {
+  return element instanceof HTMLInputElement && 'select' in element;
+}
+
+function focus(element?: FocusableTarget | null, { select = false } = {}) {
+  // only focus if that element is focusable and not already focused
+  if (element && element.focus && element !== document.activeElement) {
```
jjenzz, line 253: curious why we need the `element !== document.activeElement` check here. AFAIK, browsers won't fire focus for something that is already focused. Is there another reason here that we should maybe document?

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -26,44 +35,265 @@ type FocusScopeProps = {
...
+      if (!wrapped && !contained) return;
+      if (scopeRef.current.paused) return;
+
+      if (event.key === 'Tab' && !event.altKey && !event.ctrlKey && !event.metaKey) {
+        const focusedElement = document.activeElement as HTMLElement | null;
+        if (focusedElement) {
+          const container = event.currentTarget as HTMLElement;
+          const [first, last] = getTabbableEdges(container);
+          if (!first || !last) {
+            event.preventDefault();
```
jjenzz, line 107: am i understanding this correctly that if there is no first or last element then we don't let them tab? if so, why not? is there an implicit assumption here that the `focusedElement` is the container?

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -26,44 +35,265 @@ type FocusScopeProps = {
...
+      return node.tabIndex >= 0 ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_SKIP;
+    },
+  });
+}
+
+/**
+ * Returns the first "real" tabbable in a list of candidates.
+ * NOTE: This takes into account if elements are hidden or not.
+ */
+function getFirstRealTabbable(candidates: HTMLElement[], container: HTMLElement) {
```
jjenzz, line 229: thoughts on renaming to `findVisibleTabbable` ? The word "real" didn't help much cos I don't immediately understand what a "fake" tabbable is haha. also, `find` always returns first in `Array.find` so that made sense too but not hugely fussed about that part.

The rename will prevent me from having to check implementation to understand.

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -26,44 +35,265 @@ type FocusScopeProps = {
...
+/**
+ * Returns a list of potential tabbable candidates.
+ * NOTE: They are considered candidates at this point, because it includes elements that are hidden.
+ */
+function getTabbableCandidates(container: HTMLElement) {
+  const nodes: HTMLElement[] = [];
+  const walker = createTabbableWalker(container);
+  while (walker.nextNode()) nodes.push(walker.currentNode as HTMLElement);
+  // we do not take into account the order of nodes with positive `tabIndex` as it is
+  // considered an accessibility issue.
```
jjenzz, line 199:
```suggestion
  // hinders accessibility to have tab order different from visual order.
```

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -26,44 +35,265 @@ type FocusScopeProps = {
...
+  return nodes;
+}
+
+/**
+ * Creates a tree that contains only tabbable elements.
+ * This is only a close aproximation. For example it doesn't take into account cases like when
```
jjenzz, line 205:
```suggestion
 * This is only a close approximation. For example it doesn't take into account cases like when
```

jjenzz (changes requested):

packages/react/focus-scope/src/FocusScope.tsx
jjenzz, line 199:
```suggestion
  // we do not take into account the order of nodes with positive `tabIndex` as it
  // hinders accessibility to have tab order different from visual order.
```

packages/react/focus-scope/src/FocusScope.tsx
benoitgrelard: I like that, will update 👍

packages/react/focus-scope/src/FocusScope.tsx
benoitgrelard: Hmm, yeah that's suspicious, although I clearly remember showing you that on a call and that it was indeed a small improvement on what we had before.

It might be for one of these edge cases where the whole page gets blurred and focus gets returned to the same place.
I will double-check that.

Edit: on second thoughts, I might also have added this to avoid "selection" as the browser would run this even if it was focused already. Again, I'll double check.

packages/react/focus-scope/src/FocusScope.tsx
benoitgrelard: Yeah, if there's not first and last, then it means nothing is tabbable so tabbing shouldn't do anything. It's already covered by the containment, but in the case of just wrapping, then I think something like this is needed anyway. But I agree it feels a bit implicit right now. I'll see what I can do.

packages/react/focus-scope/src/FocusScope.tsx
benoitgrelard: I like the name. Yeah I find the naming really tricky to get right because of the fact that we have tabbable that may not in fact be tabbable… annoying.
I'm tempted to call this one even just `findFirstTabbable`. Thoughts?

jjenzz (approved):

PR #592, opened 2021-04-12
[Popper][Label][Slider] Update hooks to receive node

jjenzz:
--------------

- [x] Add versions when #580 is merged.

--------------

### Description

Part 2 of the `ref` work. This PR updates our `useSize`, `useRect`, `useDirection` and `useLabelContext` hooks to receive a DOM node rather than a ref. This allows effects to re-execute if the node changes or mount is delayed.

### DIFF
--- a/packages/react/label/src/Label.tsx
+++ b/packages/react/label/src/Label.tsx
@@ -80,7 +80,7 @@ Label.displayName = NAME;
 
 /* -----------------------------------------------------------------------------------------------*/
 
-const useLabelContext = <E extends HTMLElement>(element?: E | null) => {
+const useLabelContext = (element?: HTMLElement | null) => {
   const context = React.useContext(LabelContext);
 
   React.useEffect(() => {
--- a/packages/react/label/src/Label.tsx
+++ b/packages/react/label/src/Label.tsx
@@ -80,17 +80,16 @@ Label.displayName = NAME;
 
 /* -----------------------------------------------------------------------------------------------*/
 
-const useLabelContext = <E extends HTMLElement>(ref?: React.RefObject<E>) => {
+const useLabelContext = <E extends HTMLElement>(element?: E | null) => {
   const context = React.useContext(LabelContext);
 
   React.useEffect(() => {
     const label = context?.ref.current;
-    const element = ref?.current;
 
     if (label && element) {
       return addLabelClickEventListener(label, element);
     }
-  }, [context, ref]);
+  }, [context, element]);
 
   return context?.id;
 };
--- a/packages/react/popper/src/Popper.tsx
+++ b/packages/react/popper/src/Popper.tsx
@@ -53,15 +53,13 @@ const PopperAnchor = React.forwardRef((props, forwardedRef) => {
   const { virtualRef, children, ...anchorProps } = props;
   const context = usePopperContext(ANCHOR_NAME);
   const ref = React.useRef<React.ElementRef<typeof Primitive>>(null);
-  const composedRefs = useComposedRefs(forwardedRef, ref, (node) => context.onAnchorChange(node));
+  const composedRefs = useComposedRefs(forwardedRef, ref);
 
   React.useEffect(() => {
     // Consumer can anchor the popper to something that isn't
     // a DOM node e.g. pointer position, so we override the
     // `anchorRef` with their virtual ref in this case.
-    if (virtualRef?.current) {
-      context.onAnchorChange(virtualRef.current);
-    }
+    context.onAnchorChange(virtualRef?.current || ref.current);
   });
 
   return virtualRef ? null : (
--- a/packages/react/popper/src/Popper.tsx
+++ b/packages/react/popper/src/Popper.tsx
@@ -18,13 +18,18 @@ import type { Measurable } from '@radix-ui/rect';
 const POPPER_NAME = 'Popper';
 
 type PopperContextValue = {
-  anchorRef: React.MutableRefObject<Measurable | null>;
+  anchor: Measurable | null;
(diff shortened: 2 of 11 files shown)


packages/react/label/src/Label.tsx
```diff
@@ -80,17 +80,16 @@ Label.displayName = NAME;
 
 /* -----------------------------------------------------------------------------------------------*/
 
-const useLabelContext = <E extends HTMLElement>(ref?: React.RefObject<E>) => {
+const useLabelContext = <E extends HTMLElement>(element?: E | null) => {
   const context = React.useContext(LabelContext);
 
   React.useEffect(() => {
     const label = context?.ref.current;
```
benoitgrelard, line 87: What about this ref? does it need to be reactive too?

packages/react/popper/src/Popper.tsx
```diff
@@ -48,15 +53,11 @@ const PopperAnchor = React.forwardRef((props, forwardedRef) => {
...
-    forwardedRef,
-    ref,
-    context.anchorRef as React.MutableRefObject<React.ElementRef<typeof Primitive>>
-  );
+  const composedRefs = useComposedRefs(forwardedRef, ref, (node) => context.onAnchorChange(node));
 
   React.useEffect(() => {
     if (virtualRef?.current) {
-      context.anchorRef.current = virtualRef.current;
+      context.onAnchorChange(virtualRef.current);
```
benoitgrelard, line 60: Am I reading this correctly here that if you passed a `virtualRef`, `onAnchorChange` will get  called twice potentially, first when the node mounts, set to the node, and then overriden in the effect here?

edit: actually after expanding the code I realised that the composedRef wouldn't be called if a virtualRef is passed obviously. It's a bit annoying how the code is split here. Not sure if we can improve this.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -33,10 +33,11 @@ const stateMachine = createStateMachine(tooltipStateChart);
 const TOOLTIP_NAME = 'Tooltip';
 
 type TooltipContextValue = {
-  triggerRef: React.RefObject<HTMLButtonElement>;
   contentId: string;
   open: boolean;
   stateAttribute: StateAttribute;
+  trigger: HTMLButtonElement | null;
+  onTriggerChange(trigger: HTMLButtonElement | null): void;
```
benoitgrelard, line 40: Probably doesn't need to be specifically a button element type?

packages/react/use-size/src/useSize.tsx
```diff
@@ -2,15 +2,11 @@
 
 import * as React from 'react';
 
-function useSize(
-  /** A reference to the element whose size to observe */
-  refToObserve: React.RefObject<HTMLElement | SVGElement>
-) {
+function useSize<T extends HTMLElement | SVGElement>(elementToObserve: T | null) {
```
benoitgrelard, line 5: I reckon we should simplify and call the param `element`, it's self-explanatory enough.

benoitgrelard (commented):
Looks great! Only left a few minor comments.

packages/react/label/src/Label.tsx
jjenzz: I don't think so because the consumer has to be inside the label for the Context to exist. So the `label` will always exist. We create this hook alongside the label component which we know isn't conditionally rendered or anything.

packages/react/popper/src/Popper.tsx
jjenzz: Oh wait, now that we're using nodes, I might be able to improve this again by doing:

```jsx
const ref = React.useRef();
const composedRefs = (forwardedRef, ref);

React.useEffect(() => {
	context.onAnchorChange(virtualRef ? virtualRef.current : ref.current);
});
```

Let me check...

jjenzz: Yep works... it broke the reordered parts stories before but all seems to be good now we're using nodes 👍

packages/react/tooltip/src/Tooltip.tsx
jjenzz: I just copied what was there before

```diff
- triggerRef: React.RefObject<HTMLButtonElement>;
+ trigger: HTMLButtonElement | null;
```

Would you prefer I change it to `HTMLElement` ?

packages/react/use-size/src/useSize.tsx
jjenzz: I agree... yay

packages/react/tooltip/src/Tooltip.tsx
benoitgrelard: In reality I don't think it really matters cause it's just internal…
Because I don't know if HTMLElement is better really as we know it should at least be an interactive element…

jjenzz: Oh, shall I change it back to `HTMLButtonElement` then?

benoitgrelard: Yeah I guess it probably carries the intent better considering that it matches the default tag for that part…

benoitgrelard: Sorry about that 😀

jjenzz: Maybe I should change it to `React.ElementRef<typeof TooltipTrigger>` like we do elsewhere?

benoitgrelard: Oh I like that yeah. I don't think we've done this consistently but yeah definitely the way to go.

benoitgrelard (approved):

benoitgrelard:
🎉

Issue #387, opened 2021-01-08
New component: HoverCard


chaance:
An example of a `HoverCard`.

https://user-images.githubusercontent.com/805073/114011235-eb7a3500-985c-11eb-9448-412b49479611.mov

This component would be primarily for sighted users. It should only be used on links with a href. It should only ever display content that is available on the href.

A rough feature-set for `HoverCard`:

- Popper positioning. I assume we can reuse our utility for this?
- Trigger delay (research whether to use same delay as tooltip. Potentially use the same code and link them globally? Gotta do some thinking there.)
- Support for Arrow or no Arrow.
- Close on ESC.
- Don't trigger on focus.
- Try to avoid what twitter does where you can reopen the HoverCard if you hover its container before the close animation has completed?
- Should work well with `Slot`.

No focus trapping. No focus management at all.

benoitgrelard:
At a glance I don't think it would need that logic that tooltips have where it's global and you can jump from one to the other quickly without having a delay. I think that might be distracting because they're usually much bigger.

andy-hook:
> It should only ever display content that is available on the href.

What would be an incorrect use case example? at a glance I'm unsure how this might relate to the component api, or are we just referring to usage guidelines?

benoitgrelard:
I think it is more about guidelines.
The reason we say this is that the content inside the hover card cannot really made accessible to keyboard or screen reader users. Therefore, it shouldn't be primary content, but rather content that can be accessed another way (like visiting a link with that same info behind it).

colmtuite:
It might just be usage guidelines yeah. This isn't an ARIA component, it's just kinda popped up everywhere around the web in the past few years. But it's not really possible to make it accessible, so we will just need to wing it, and do our best to document good guidelines.

I'm thinking, let's say someone puts it on a a button avatar like

```
<HoverCard as={Slot}>
  <Avatar as="button" />
</HoveCard>
```

(or however code works)

and it just displays log out links and stuff, that aren't on another page. Something like that.

andy-hook:
Makes sense! will do some quick research and report back

andy-hook:
I had a quick look at other uses of this pattern out in the wild

#### Dribbble
![image](https://user-images.githubusercontent.com/11708259/114035992-8f70da00-9877-11eb-8f17-ed092ff2aa4c.png)

#### Github
[...]

PR #595, opened 2021-04-13
Add `HoverCard` primitive

andy-hook:
Closes https://github.com/radix-ui/primitives/issues/387

### Description

This PR adds the `HoverCard` primitive, [see here for discussion](https://github.com/radix-ui/primitives/issues/387).

Simple, straightforward primitive heavily based on `Popover` and `Tooltip`, intended as a contextual enhancement for links via pointer interaction, no focus management.

https://user-images.githubusercontent.com/11708259/114609510-cb9ca400-9c96-11eb-8cfd-abb5bdda506c.mov

Basic usage:

```jsx
<HoverCard>
      <HoverCardTrigger href="/some-awesome-page">
        Link text
      </HoverCardTrigger>
      <HoverCardContent>
        <HoverCardArrow />
        Card content
      </HoverCardContent>
</HoverCard>
```

### DIFF
--- a/.yarn/versions/51d71892.yml
+++ b/.yarn/versions/51d71892.yml
@@ -1,0 +1,4 @@
+declined:
+  - primitives
+  - ssr-testing
+  - "@radix-ui/react-hover-card"
--- a/.yarn/versions/51d71892.yml
+++ b/.yarn/versions/51d71892.yml
@@ -1,4 +1,6 @@
+releases:
+  "@radix-ui/react-hover-card": patch
+
 declined:
   - primitives
   - ssr-testing
-  - "@radix-ui/react-hover-card"
--- a/packages/react/hover-card/README.md
+++ b/packages/react/hover-card/README.md
@@ -1,0 +1,15 @@
+# `react-hover-card`
+
+## Installation
+
+```sh
+$ yarn add @radix-ui/react-hover-card
+# or
+$ npm install @radix-ui/react-hover-card
+```
+
+## Usage
+
+TODO
+
+View docs ...
--- a/packages/react/hover-card/README.md
+++ b/packages/react/hover-card/README.md
@@ -10,6 +10,4 @@ $ npm install @radix-ui/react-hover-card
 
 ## Usage
 
-TODO
-
-View docs ...
+View docs [here](https://radix-ui.com/primitives/docs/components/hover-card).
--- a/packages/react/hover-card/package.json
+++ b/packages/react/hover-card/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@radix-ui/react-hover-card",
-  "version": "0.0.1",
+  "version": "0.0.0",
   "license": "MIT",
   "source": "src/index.ts",
   "main": "dist/index.js",
@@ -29,7 +29,8 @@
     "@radix-ui/react-use-controllable-state": "workspace:*"
   },
   "peerDependencies": {
-    "react": "^16.8 || ^17.0"
+    "react": "^16.8 || ^17.0",
+    "react-dom": "^16.8 || ^17.0"
   },
   "homepage": "https://radix-ui.com/primitives",
   "repository": {
--- a/packages/react/hover-card/package.json
+++ b/packages/react/hover-card/package.json
@@ -1,0 +1,43 @@
+{
+  "name": "@radix-ui/react-hover-card",
+  "version": "0.0.1",
+  "license": "MIT",
+  "source": "src/index.ts",
+  "main": "dist/index.js",
+  "module": "dist/index.module.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist",
+    "README.md"
+  ],
+  "sideEffects": false,
+  "scripts": {
+    "clean": "rm -rf dist",
+    "prepublish": "yarn clean",
+    "version": "yarn version"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.13.10",
+    "@radix-ui/primitive": "workspace:*",
+    "@radix-ui/react-compose-refs": "workspace:*",
+    "@radix-ui/react-context": "workspace:*",
+    "@radix-ui/react-id": "workspace:*",
+    "@radix-ui/react-polymorphic": "workspace:*",
+    "@radix-ui/react-popper": "workspace:*",
+    "@radix-ui/react-portal": "workspace:*",
+    "@radix-ui/react-presence": "workspace:*",
+    "@radix-ui/react-primitive": "workspace:*",
+    "@radix-ui/react-use-controllable-state": "workspace:*"
+  },
+  "peerDependencies": {
+    "react": "^16.8 || ^17.0"
+  },
+  "homepage": "https://radix-ui.com/primitives",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/radix-ui/primitives.git"
+  },
+  "bugs": {
+    "url": "https://github.com/radix-ui/primitives/issues"
+  }
+}
--- a/packages/react/hover-card/package.json
+++ b/packages/react/hover-card/package.json
@@ -19,9 +19,7 @@
   "dependencies": {
     "@babel/runtime": "^7.13.10",
     "@radix-ui/primitive": "workspace:*",
-    "@radix-ui/react-compose-refs": "workspace:*",
     "@radix-ui/react-context": "workspace:*",
-    "@radix-ui/react-id": "workspace:*",
     "@radix-ui/react-polymorphic": "workspace:*",
     "@radix-ui/react-popper": "workspace:*",
     "@radix-ui/react-portal": "workspace:*",
--- a/packages/react/hover-card/package.json
+++ b/packages/react/hover-card/package.json
@@ -25,7 +25,6 @@
     "@radix-ui/react-portal": "workspace:*",
     "@radix-ui/react-presence": "workspace:*",
     "@radix-ui/react-primitive": "workspace:*",
-    "@radix-ui/react-slot": "workspace:*",
     "@radix-ui/react-use-controllable-state": "workspace:*"
   },
   "peerDependencies": {
--- a/packages/react/hover-card/package.json
+++ b/packages/react/hover-card/package.json
@@ -27,6 +27,7 @@
     "@radix-ui/react-portal": "workspace:*",
     "@radix-ui/react-presence": "workspace:*",
     "@radix-ui/react-primitive": "workspace:*",
+    "@radix-ui/react-slot": "workspace:*",
     "@radix-ui/react-use-controllable-state": "workspace:*"
   },
   "peerDependencies": {
--- a/packages/react/hover-card/src/HoverCard.stories.tsx
+++ b/packages/react/hover-card/src/HoverCard.stories.tsx
@@ -1,0 +1,597 @@
+import * as React from 'react';
+import { HoverCard, HoverCardTrigger, HoverCardContent, HoverCardArrow } from './HoverCard';
+import { Slot } from '@radix-ui/react-slot';
+import { SIDE_OPTIONS, ALIGN_OPTIONS } from '@radix-ui/popper';
+import { css } from '../../../../stitches.config';
+
+export default { title: 'Components/HoverCard' };
+
+export const HoverInteractionTest = () => {
+  return (
+    <div
+      style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', height: '150vh' }}
+    >
+      <HoverCard>
+        <HoverCardTrigger className={triggerClass}>Trigger</HoverCardTrigger>
+        <HoverCardContent className={contentClass} sideOffset={5}>
+          <HoverCardArrow className={arrowClass} width={20} height={10} />
+          Lorem ipsum dolor sit amet
+          <br />
+          Lorem ipsum dolor sit amet
+          <br />
+          Lorem ipsum dolor sit amet
+          <br />
+          Lorem ipsum dolor sit amet
+        </HoverCardContent>
+      </HoverCard>
+    </div>
+  );
+};
+
+export const Styled = () => {
+  return (
+    <div
+      style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', height: '200vh' }}
+    >
+      <HoverCard>
+        <HoverCardTrigger className={triggerClass}>Trigger</HoverCardTrigger>
+        <HoverCardContent className={contentClass} sideOffset={5}>
+          <HoverCardArrow className={arrowClass} width={20} height={10} />
+        </HoverCardContent>
+      </HoverCard>
+    </div>
+  );
+};
+
+export const Controlled = () => {
+  const [open, setOpen] = React.useState(false);
+
+  return (
+    <div
+      style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', height: '50vh' }}
+    >
+      <HoverCard open={open} onOpenChange={setOpen}>
+        <HoverCardTrigger
+          className={triggerClass}
+          style={{ backgroundColor: 'red', padding: 50, display: 'block' }}
+        >
+          trigger
+        </HoverCardTrigger>
+        <HoverCardContent className={contentClass}>
+          <HoverCardArrow className={arrowClass} width={20} height={10} />
+        </HoverCardContent>
+      </HoverCard>
+    </div>
+  );
+};
+
+export const Animated = () => {
+  return (
+    <div
+      style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', height: '200vh' }}
+    >
+      <HoverCard>
+        <HoverCardTrigger className={triggerClass}>Trigger</HoverCardTrigger>
+        <HoverCardContent className={animatedContentClass} sideOffset={10}>
+          <HoverCardArrow className={arrowClass} width={20} height={10} />
+        </HoverCardContent>
+      </HoverCard>
+    </div>
+  );
+};
+
+export const ForcedMount = () => {
+  return (
+    <div
+      style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', height: '200vh' }}
+    >
+      <HoverCard>
+        <HoverCardTrigger className={triggerClass}>Trigger</HoverCardTrigger>
+        <HoverCardContent className={contentClass} sideOffset={10} forceMount>
+          <HoverCardArrow className={arrowClass} width={20} height={10} />
+        </HoverCardContent>
+      </HoverCard>
+    </div>
+  );
+};
+
+export const Nested = () => {
+  const anchorRef = React.useRef<HTMLAnchorElement>(null);
+
+  return (
+    <div
+      style={{
+        height: '300vh',
+        width: '300vw',
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+      }}
+    >
+      <a
+        href="/"
+        style={{ position: 'fixed', top: 10, left: 10 }}
+        onClick={() => anchorRef.current?.focus()}
+      >
+        Focus HoverCard button
+      </a>
+
+      <HoverCard>
+        <HoverCardTrigger className={triggerClass} ref={anchorRef}>
+          Open HoverCard
+        </HoverCardTrigger>
+
+        <HoverCardContent
+          className={contentClass}
+          sideOffset={5}
+          style={{ backgroundColor: 'crimson' }}
+        >
+          <HoverCard>
+            <HoverCardTrigger className={triggerClass}>Open nested HoverCard</HoverCardTrigger>
+            <HoverCardContent
+              className={contentClass}
+              side="top"
+              align="center"
+              sideOffset={5}
+              style={{ backgroundColor: 'green' }}
+            >
+              <HoverCardArrow
+                className={arrowClass}
+                width={20}
+                height={10}
+                offset={20}
+                style={{ fill: 'green' }}
+              />
+            </HoverCardContent>
+          </HoverCard>
+          dfsdf
+          <HoverCardArrow
+            className={arrowClass}
+            width={20}
+            height={10}
+            offset={20}
+            style={{ fill: 'crimson' }}
+          />
+        </HoverCardContent>
+      </HoverCard>
+    </div>
+  );
+};
+
+export const CustomAnchor = () => {
+  const itemBoxRef = React.useRef<HTMLDivElement>(null);
+
+  return (
+    <div
+      ref={itemBoxRef}
+      style={{
+        display: 'flex',
+        justifyContent: 'space-between',
+        alignItems: 'center',
+        width: 250,
+        padding: 20,
+        margin: 100,
+        backgroundColor: '#eee',
+      }}
+    >
+      Item
+      <HoverCard>
+        <HoverCardTrigger className={triggerClass}>hover</HoverCardTrigger>
+        <HoverCardContent
+          className={contentClass}
+          anchorRef={itemBoxRef}
+          side="right"
+          sideOffset={1}
+          align="start"
+          style={{ borderRadius: 0, width: 200, height: 100 }}
+        ></HoverCardContent>
+      </HoverCard>
+    </div>
+  );
+};
+
+export const NonPortal = () => {
+  return (
+    <HoverCard>
+      <HoverCardTrigger className={triggerClass}>hover</HoverCardTrigger>
+      <HoverCardContent portalled={false} className={contentClass} sideOffset={5}>
+        <HoverCardArrow className={arrowClass} width={20} height={10} offset={10} />
+      </HoverCardContent>
+    </HoverCard>
+  );
+};
+
+export const WithSlottedTrigger = () => {
+  return (
+    <HoverCard>
+      <HoverCardTrigger as={Slot}>
+        <button className={triggerClass} onClick={() => console.log('StyledTrigger click')}>
+          hover
+        </button>
+      </HoverCardTrigger>
+      <HoverCardContent className={contentClass} sideOffset={5}>
+        <HoverCardArrow className={arrowClass} width={20} height={10} offset={10} />
+      </HoverCardContent>
+    </HoverCard>
+  );
+};
+
+export const WithSlottedContent = () => (
+  <HoverCard>
+    <HoverCardTrigger className={triggerClass}>Hover or Focus me</HoverCardTrigger>
+    <HoverCardContent as={Slot} sideOffset={5}>
+      <div className={contentClass}>
+        Nicely done!
+        <HoverCardArrow className={arrowClass} offset={10} />
+      </div>
+    </HoverCardContent>
+  </HoverCard>
+);
+
+// change order slightly for more pleasing visual
+const SIDES = SIDE_OPTIONS.filter((side) => side !== 'bottom').concat(['bottom']);
+
+export const Chromatic = () => (
+  <div style={{ padding: 200 }}>
+    <h1>Uncontrolled</h1>
+    <h2>Closed</h2>
+    <HoverCard>
+      <HoverCardTrigger className={triggerClass}>trigger</HoverCardTrigger>
+      <HoverCardContent className={contentClass} sideOffset={5}>
+        <HoverCardArrow className={arrowClass} width={20} height={10} />
+      </HoverCardContent>
+    </HoverCard>
+
+    <h2>Open</h2>
+    <HoverCard defaultOpen>
+      <HoverCardTrigger className={triggerClass}>trigger</HoverCardTrigger>
+      <HoverCardContent className={contentClass} sideOffset={5}>
+        <HoverCardArrow className={arrowClass} width={20} height={10} />
+      </HoverCardContent>
+    </HoverCard>
+
+    <h1 style={{ marginTop: 100 }}>Controlled</h1>
+    <h2>Closed</h2>
+    <HoverCard open={false}>
+      <HoverCardTrigger className={triggerClass}>open</HoverCardTrigger>
+      <HoverCardContent className={contentClass} sideOffset={5}>
+        <HoverCardArrow className={arrowClass} width={20} height={10} />
+      </HoverCardContent>
+    </HoverCard>
+
+    <h2>Open</h2>
+    <HoverCard open>
+      <HoverCardTrigger className={triggerClass}>open</HoverCardTrigger>
+      <HoverCardContent className={contentClass} sideOffset={5}>
+        <HoverCardArrow className={arrowClass} width={20} height={10} />
+      </HoverCardContent>
+    </HoverCard>
+
+    <h1 style={{ marginTop: 100 }}>Force mounted content</h1>
+    <HoverCard>
+      <HoverCardTrigger className={triggerClass}>open</HoverCardTrigger>
+      <HoverCardContent className={contentClass} sideOffset={5} forceMount>
+        <HoverCardArrow className={arrowClass} width={20} height={10} />
+      </HoverCardContent>
+    </HoverCard>
+
+    <h1 style={{ marginTop: 100 }}>Positioning</h1>
+    <h2>No collisions</h2>
+    <h3>Side & Align</h3>
+    <div className={gridClass}>
+      {SIDES.map((side) =>
+        ALIGN_OPTIONS.map((align) => (
+          <HoverCard key={`${side}-${align}`} open>
+            <HoverCardTrigger className={chromaticTriggerClass} />
+            <HoverCardContent
+              className={chromaticContentClass}
+              side={side}
+              align={align}
+              avoidCollisions={false}
+            >
+              <p style={{ textAlign: 'center' }}>
+                {side}
+                <br />
+                {align}
+              </p>
+              <HoverCardArrow className={chromaticArrowClass} width={20} height={10} />
+            </HoverCardContent>
+          </HoverCard>
+        ))
+      )}
+    </div>
+
+    <h3>Arrow offset</h3>
+    <h4>Positive</h4>
+    <div className={gridClass}>
+      {SIDES.map((side) =>
+        ALIGN_OPTIONS.map((align) => (
+          <HoverCard key={`${side}-${align}`} open>
+            <HoverCardTrigger className={chromaticTriggerClass} />
+            <HoverCardContent
+              className={chromaticContentClass}
+              side={side}
+              align={align}
+              avoidCollisions={false}
+            >
+              <p style={{ textAlign: 'center' }}>
+                {side}
+                <br />
+                {align}
+              </p>
+              <HoverCardArrow className={chromaticArrowClass} width={20} height={10} offset={5} />
+            </HoverCardContent>
+          </HoverCard>
+        ))
+      )}
+    </div>
+    <h4>Negative</h4>
+    <div className={gridClass}>
+      {SIDES.map((side) =>
+        ALIGN_OPTIONS.map((align) => (
+          <HoverCard key={`${side}-${align}`} open>
+            <HoverCardTrigger className={chromaticTriggerClass} />
+            <HoverCardContent
+              className={chromaticContentClass}
+              side={side}
+              align={align}
+              avoidCollisions={false}
+            >
+              <p style={{ textAlign: 'center' }}>
+                {side}
+                <br />
+                {align}
+              </p>
+              <HoverCardArrow className={chromaticArrowClass} width={20} height={10} offset={-10} />
+            </HoverCardContent>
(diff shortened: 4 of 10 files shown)


.yarn/versions/51d71892.yml
```diff
@@ -0,0 +1,4 @@
+declined:
+  - primitives
+  - ssr-testing
+  - "@radix-ui/react-hover-card"
```
andy-hook, line 4: Unsure on release strategy, ran a `version check --interactive` to sidestep the pre-push

packages/react/hover-card/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+```sh
+$ yarn add @radix-ui/react-hover-card
+# or
+$ npm install @radix-ui/react-hover-card
+```
+
+## Usage
+
+View docs [here](https://radix-ui.com/primitives/docs/components/hover-card).
```
andy-hook, line 13: Preemptively adding this as my memory isn't what it used to be.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+};
+
+const HoverCard: React.FC<HoverCardOwnProps> = (props) => {
+  const {
+    children,
+    open: openProp,
+    defaultOpen,
+    onOpenChange,
+    enterDelayDuration = 700,
+    exitDelayDuration = 400,
```
andy-hook, line 51: These ended up feeling right to me based on my testing, entrance `700` matches `Tooltip` I believe.

packages/react/hover-card/src/HoverCard.stories.tsx
```diff
@@ -0,0 +1,696 @@
...
+        <CardContentPlaceholder />
+      </div>
+    </HoverCardContent>
+  </HoverCard>
+);
+
+// change order slightly for more pleasing visual
+const SIDES = SIDE_OPTIONS.filter((side) => side !== 'bottom').concat(['bottom']);
+
+export const Chromatic = () => (
```
andy-hook, line 303: Taken straight from the `Popover` visual tests, I felt like they covered this primitive equally well.

colmtuite (approved):

colmtuite:
Love it 🔥 🥳

.yarn/versions/51d71892.yml
benoitgrelard: That's the right command to run yep. In this case you just need to select patch for hover card so it gets released properly as 0.0.1 next time we do a release.

packages/react/hover-card/src/HoverCard.stories.tsx
```diff
@@ -0,0 +1,696 @@
...
+
+    <h2>Open</h2>
+    <HoverCard defaultOpen>
+      <HoverCardTrigger className={triggerClass}>trigger</HoverCardTrigger>
+      <HoverCardContent className={contentClass} sideOffset={5}>
+        <HoverCardArrow className={arrowClass} width={20} height={10} />
+        Some content
+      </HoverCardContent>
+    </HoverCard>
+
```
benoitgrelard, line 323: Can you also add the "open with reordered parts" stories in both controlled and uncontrolled like in popover, tooltip, etc? 🙏

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+/* -------------------------------------------------------------------------------------------------
+ * HoverCard
+ * -----------------------------------------------------------------------------------------------*/
+
+const HOVERCARD_NAME = 'HoverCard';
+
+type HoverCardContextValue = {
+  triggerRef: React.RefObject<HTMLAnchorElement>;
+  contentId: string;
```
benoitgrelard, line 22: I don't think we need to do anything with ids for this component as there is no accessibility built-in. This means you can get rid of `useId` and the dependency on ` react-id` too.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+import type * as Polymorphic from '@radix-ui/react-polymorphic';
+
+/* -------------------------------------------------------------------------------------------------
+ * HoverCard
+ * -----------------------------------------------------------------------------------------------*/
+
+const HOVERCARD_NAME = 'HoverCard';
+
+type HoverCardContextValue = {
+  triggerRef: React.RefObject<HTMLAnchorElement>;
```
benoitgrelard, line 21: It doesn't seem we use the `triggerRef` for anything internally so I don't think we need it.

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: I personally find 400 for `exitDelayDuration` a little too long for the default behaviour. 700 for `enterDelayDuration` feels right though I agree.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+const [HoverCardProvider, useHoverCardContext] = createContext<HoverCardContextValue>(
+  HOVERCARD_NAME
+);
+
+type HoverCardOwnProps = {
+  open?: boolean;
+  defaultOpen?: boolean;
+  enterDelayDuration?: number;
+  exitDelayDuration?: number;
+  onOpenChange?: (open: boolean) => void;
```
benoitgrelard, line 41: nit: can you bring `onOpenChange` just below `defaultOpen` so the 3 open props are grouped?

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+  const [open = false, setOpen] = useControllableState({
+    prop: openProp,
+    defaultProp: defaultOpen,
+    onChange: onOpenChange,
+  });
+
+  const clearEnterTimer = React.useCallback(() => {
+    clearTimeout(enterTimerRef.current);
+  }, [enterTimerRef]);
```
benoitgrelard, line 67: `enterTimerRef` doesn't need to be a dependency because ref references don't change over time.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+  const clearLeaveTimer = React.useCallback(() => {
+    clearTimeout(leaveTimerRef.current);
+  }, [leaveTimerRef]);
```
benoitgrelard, line 71: `leaveTimerRef` doesn't need to be a dependency because ref references don't change over time.

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: We could probably even do away from creating these 2 functions and just inline them as they do not necessitate any extra dependencies and arguably

`clearLeaveTimer` doesn't convey that much more intent that `clearTimeout(leaveTimerRef.current)`.

Just a suggestion, I'll let you test it out and see what you think.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+    children,
+    open: openProp,
+    defaultOpen,
+    onOpenChange,
+    enterDelayDuration = 700,
+    exitDelayDuration = 400,
+  } = props;
+  const triggerRef = React.useRef<HTMLAnchorElement>(null);
+  const [hasCustomAnchor, setHasCustomAnchor] = React.useState(false);
+  const enterEnvoked = React.useRef(false);
```
benoitgrelard, line 55: Small typo? Did you mean `enterInvoked`?

Also, if it's a ref, would be good to add `Ref` suffix like the others, we generally tend to do this.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+    enterTimerRef.current = window.setTimeout(() => {
+      setOpen(true);
+      enterEnvoked.current = true;
+    }, enterDelayDuration);
+  }, [clearLeaveTimer, enterDelayDuration, setOpen]);
+
+  const handelMouseLeave = React.useCallback(() => {
+    clearEnterTimer();
+    leaveTimerRef.current = window.setTimeout(() => {
+      if (enterEnvoked.current) {
```
benoitgrelard, line 84: I can see why you asked the question about the state machines.

This kind of logic here definitely makes me think that perhaps we'd have a clearer intent here if we had a state machine instead as this is trying to express the fact that the leave event should be ignored in some situations (ie. if you haven't entered).

With a state machine I believe this would be clearer, because you would only be able leave if you were in open state.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+>;
+
+const HoverCardTrigger = React.forwardRef((props, forwardedRef) => {
+  const { as = TRIGGER_DEFAULT_TAG, ...triggerProps } = props;
+  const context = useHoverCardContext(TRIGGER_NAME);
+  const composedTriggerRef = useComposedRefs(forwardedRef, context.triggerRef);
+
+  const trigger = (
+    <Primitive
+      type="a"
```
benoitgrelard, line 166: I think that's a leftover from copy pasting `type="button"`?

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+type HoverCardTriggerOwnProps = Polymorphic.OwnProps<typeof Primitive>;
+type HoverCardTriggerPrimitive = Polymorphic.ForwardRefComponent<
+  typeof TRIGGER_DEFAULT_TAG,
+  HoverCardTriggerOwnProps
+>;
+
+const HoverCardTrigger = React.forwardRef((props, forwardedRef) => {
+  const { as = TRIGGER_DEFAULT_TAG, ...triggerProps } = props;
+  const context = useHoverCardContext(TRIGGER_NAME);
+  const composedTriggerRef = useComposedRefs(forwardedRef, context.triggerRef);
```
benoitgrelard, line 162: You won't need to do this anymore as we don't need the `triggerRef` internally.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+const HoverCardContentImpl = React.forwardRef((props, forwardedRef) => {
+  const { portalled = true, ...contentProps } = props;
+  const context = useHoverCardContext(CONTENT_NAME);
+
+  const PortalWrapper = portalled ? Portal : React.Fragment;
+
+  return (
+    <PortalWrapper>
+      <PopperPrimitive.Content
+        id={context.contentId}
```
benoitgrelard, line 250: As I said above, this isn't needed in this component.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+        {...contentProps}
+        ref={forwardedRef}
+        style={{
+          ...contentProps.style,
+          // re-namespace exposed content custom property
+          ['--radix-hovercard-content-transform-origin' as any]: 'var(--radix-popper-transform-origin)',
```
benoitgrelard, line 256: I believe the casing would be

```suggestion
          ['--radix-hover-card-content-transform-origin' as any]: 'var(--radix-popper-transform-origin)',
```

?

benoitgrelard (commented):
This is awesome @andy-hook!

I genuinely ♥️ how little code was needed to make this work on top of what we have already built, kinda amazing, less than 300 lines of code (including lots of types).

I've left a few comments inline.

packages/react/hover-card/README.md
jjenzz: Haha, yep, we do the same :+1:

packages/react/hover-card/src/HoverCard.stories.tsx
```diff
@@ -0,0 +1,696 @@
...
+    }
+
+    setOpen(open);
+  }, []);
+
+  React.useEffect(() => {
+    return () => {
+      clearTimeout(timerRef.current);
+    };
+  }, []);
```
jjenzz, line 54: Well remembered :+1: I often forget to do this clean up haha.

packages/react/hover-card/src/HoverCard.tsx
andy-hook: Oh yes, nice! even slimmer code 😀

packages/react/hover-card/src/HoverCard.stories.tsx
jjenzz: Yep, sorry, we merged these extra ones yesterday which will be why they weren't included in your copy/paste 🙈

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+const [HoverCardProvider, useHoverCardContext] = createContext<HoverCardContextValue>(
+  HOVERCARD_NAME
+);
+
+type HoverCardOwnProps = {
+  open?: boolean;
+  defaultOpen?: boolean;
+  enterDelayDuration?: number;
+  exitDelayDuration?: number;
```
jjenzz, line 40: I know we've used `delayDuration` in `Tooltip` but I've since wondered why we need to include the word `duration` ... isn't a delay inherently a duration?

Thoughts on renaming these to `enterDelay` and `exitDelay`?

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+  const {
+    children,
+    open: openProp,
+    defaultOpen,
+    onOpenChange,
+    enterDelayDuration = 700,
+    exitDelayDuration = 400,
+  } = props;
+  const triggerRef = React.useRef<HTMLAnchorElement>(null);
+  const [hasCustomAnchor, setHasCustomAnchor] = React.useState(false);
```
jjenzz, line 54: I'm not sure if a `HoverCard` should have a custom anchor given that we're saying it's specifically to preview content behind a link. Would consumers ever need a link preview to be anchored to something else?

cc: @benoitgrelard

packages/react/hover-card/src/HoverCard.tsx
andy-hook: Nice spot on the deps, I agree with your second point too.

jjenzz: I agree with Benoit here, it would be nice to just inline the `clearTimeout()` where used to reduce some boilerplate since the abstraction doesn't convey additional meaning. It feels a bit like a premature application of the DRY principle, sowwie 🙈

packages/react/hover-card/src/HoverCard.tsx
andy-hook: Indeed a pasta error 😅

jjenzz (changes requested):
Yay for your first comp 🎉 🍻 

Benoit has already covered most of it, a couple bits here but also mostly just wanted to say that I **love** your attention to detail and how much effort you've put in to following existing standards. Reviewing was a breeeeze 💛

packages/react/hover-card/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "@radix-ui/react-polymorphic": "workspace:*",
+    "@radix-ui/react-popper": "workspace:*",
+    "@radix-ui/react-portal": "workspace:*",
+    "@radix-ui/react-presence": "workspace:*",
+    "@radix-ui/react-primitive": "workspace:*",
+    "@radix-ui/react-slot": "workspace:*",
+    "@radix-ui/react-use-controllable-state": "workspace:*"
+  },
+  "peerDependencies": {
+    "react": "^16.8 || ^17.0"
```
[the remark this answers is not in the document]
andy-hook: This is really helpful to know, thanks for mentioning it

packages/react/hover-card/src/HoverCard.tsx
andy-hook: My thought is that it’s hard to predict the layout challenges some users may have, and that a custom anchor does provide an escape hatch if they have some strange arrangement where the positioning props alone wouldn’t solve it. 🤔 

Depends how opinionated we want to be here.

packages/react/hover-card/src/HoverCard.tsx
andy-hook: I agree, I went with duration only for consistency.

packages/react/hover-card/src/HoverCard.tsx
jjenzz: We tend to prefer simplicity on our side for first releases and then add functionality as/when consumers request it if their use-case cannot be achieved another way since adding stuff prematurely and removing it later is usually harder.

benoitgrelard: I have been wondering about exactly this too…
I feel like we don't have a clear formed idea yet as to when to provide this functionality vs not.
But given that we've removed it from `DropdownMenu` and `Tooltip` and I feel like this is closer to `Tooltip` than it is from `Popover` perhaps we should not add it.
It's also our opportunity to start with less and only add if someone asks for it, this way we can see what the use cases are.

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: Yes you are right, but the reason we went with the added `duration` suffix in `Tooltip` is because of the second one `skipDelayDuration`, because otherwise `skipDelay` sounds like a boolean, or you end up with `skipDelayDelay`…

packages/react/hover-card/src/HoverCard.stories.tsx
andy-hook: S'all good, will take a look!

packages/react/hover-card/src/HoverCard.tsx
jjenzz: Seems the tooltip naming means we are now having to overcomplicate our other APIs for consistency-sake. Perhaps we should revisit the tooltip naming? Happy to keep these as they are for now till we do that if so.

packages/react/hover-card/src/HoverCard.tsx
jjenzz: Yep +1. My preference is always to avoid predicting what problems users may have as it can often cause unnecessary bloat in a codebase, and instead react to actual problems they report.

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: Happy to revisit the tooltip naming but Colm and I spent quite a while thinking about it and that's the least worse we could come up with… But again, I would be over the moon if we can find a better idea for sure!

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: But then we also don't know if people are using it in `Popover` so… 😀

packages/react/hover-card/src/HoverCard.tsx
jjenzz: I'm wondering if they should be renamed to something like `enterDelay` and `enterDelayThreshold`.

If I've not misunderstood, I believe the `skipDelayDuration` is an amount of time that the `enterDelay` will be ignored when hovering between tooltips? A threshold seems to fit that definition:

> the magnitude or intensity that must be exceeded for a certain reaction, phenomenon, result, or condition to occur or be manifested.

So, with a threshold of 400ms, it will take 400ms before the `enterDelay` will manifest between tooltips. 

If we can come up with something quick enough, maybe we can just make the change here to save increasing potential tech debt 🤷‍♀️

benoitgrelard: btw, I'm also wondering if this should have the same thing as tooltip with the skipping delay thing.
When I see UIs like this:
image
It's quite annoying that you get the delay everytime, and that's with them having made it as short as they could probably for that reason.

benoitgrelard: > I'm wondering if they should be renamed to something like `enterDelay` and `enterDelayThreshold`.
> 
> If I've not misunderstood, I believe the `skipDelayDuration` is an amount of time that the `enterDelay` will be ignored for when moving between tooltips? A threshold seems to fit that definition:
> 
> > the magnitude or intensity that must be exceeded for a certain reaction, phenomenon, result, or condition to occur or be manifested.
> 
> So, with a threshold of 400, it will take 400ms before the `enterDelay` will manifest between tooltips.
> 
> If we can come up with something quick enough, maybe we can just make the change here to save increasing potential tech debt 🤷‍♀️

That could work! I remember I played with "threshold" too.
I find that `enterDelayThreshold` doesn't communicate very well what it does, ie. it's hard to know if the threshold applies to the delay happening, or NOT happening.

Maybe `enterDelay` + `skipEnterDelayThreshold`?

I remember before internally it was called `SKIP_REST_THRESHOLD_DURATION`…

jjenzz: > I'm also wondering if this should have the same thing as tooltip with the skipping delay thing

Yeah, that's not a bad shout tbf. @andy-hook It might be worth copying over a bunch of the state machine stuff from `Tooltip` in that case 😅 

@colmtuite What are your thoughts on this one?

packages/react/hover-card/src/HoverCard.tsx
andy-hook: > It's also our opportunity to start with less and only add if someone asks for it, this way we can see what the use cases are.

> My preference is always to avoid predicting what problems users may have as it can often cause unnecessary bloat in a codebase, and instead react to actual problems they report.

I think this is a great approach, I am convinced 😁

packages/react/hover-card/src/HoverCard.tsx
jjenzz: > it's hard to know if the threshold applies to the delay happening, or NOT happening.

Given the definition of "threshold" above, a threshold in this case would be the time it takes **before** the `enterDelay` would manifest. The definition of threshold already encapsulates the "skip delay for {x} ms" part in my mind. The only part that's missing for me is the fact that this only applies _between_ tooltips. 

Something like `enterDelay` and `reEnterDelayThreshold` seems to cover it. Perhaps I'll create a ticket and we can discuss this all there and rename all later if we can come up with something fitting? It's proving to be a bit harder than I'd thought 😅

benoitgrelard: I actually really like `reEnter` that definitely brings more meaning to it :+1

benoitgrelard: I think regardless of if that's the best we can come up with `enterDelay` + `reEnterDelayThreshold` for me is definitely better than `delayDuration` + `skipDelayDuration`!

benoitgrelard: Ok, so @jjenzz and I continued that discussion on Slack and this is where we're at.

- `enterDelay` => `openDelay`
- `exitDelay` => `closeDelay`
> we generally use `open`/`close`, ie. css animations, `[data-state="open"]`
- `skipDelayDuration ` => `reEntryThreshold` (ie. when do we consider it a fresh re-entry, may or may not apply to `HoverCard` based on the decision we make about that feature)

So to summarise, as it stands:
- `Tooltip` would have `openDelay` + `reEntryThreshold`
- `HoverCard` would have `openDelay` + `closeDelay`

And we're considering whether `HoverCard` should have `reEntryThreshold` and its associated functionality.

Thoughts @andy-hook, @colmtuite?

andy-hook: Onboard with the naming 👍 

`reEntryThreshold` functionality does sound useful for the case mentioned above, interested to hear @colmtuite thoughts.

packages/react/hover-card/src/HoverCard.tsx
andy-hook: Tweaked down the `300` for now

packages/react/hover-card/src/HoverCard.tsx
jjenzz: > But then we also don't know if people are using it in `Popover` so…

@benoitgrelard Oh really? I thought we had it there because composer needed it 😬  does it not?

packages/react/hover-card/src/HoverCard.tsx
colmtuite: I love the `threshold` concept. I also love renaming to `open` and `closed`.

Not a fan of `reEntryThreshold`, we're getting into rocket science territory here 😓 I think `skipDelayThreshold` is more intuitive, though there's the issue with its meaning. We actually want an antonym for "skip", but `includeDelayThreshold` doesn't have the same ring to it. `skipDelayLimit` might work?

I think I'd vote to not include the skipDelay feature for HoverCard. In that exact GH example posted above, it would likely be beneficial. But I think in most cases, it could become quite annoying. These things can get pretty large, with many instances on the screen, like in a social feed for example. Having them open instantly as you move your mouse around could get very irritating.

jjenzz: > These things can get pretty large, with many instances on the screen, like in a social feed for example. Having them open instantly as you move your mouse around could get very irritating.

Oh yeah! Good point haha. I'm on board with leaving it out for `HoverCard` too in this case.

.yarn/versions/51d71892.yml
jjenzz: Looks like this is still awaiting the update to move `@radix-ui/react-hover-card` to patch.

Happy to have a quick call to show you how to do that if it's not clear! LMK 🙂

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,229 @@
...
+    <PopperPrimitive.Root>
+      <HoverCardProvider
+        open={open}
+        onOpenChange={setOpen}
+        onMouseEnter={handleMouseEnter}
+        onMouseLeave={handelMouseLeave}
+      >
+        {children}
+      </HoverCardProvider>
+    </PopperPrimitive.Root>
```
jjenzz, line 93: One thing worth mentioning for future reference. We used to put our provider inside like this so that it was only the children receiving the context, however, we since introduced a `Slot` component. For `Slot` to work, the component that wraps the `children` has to be the primitive part that the consumer can apply `as` to. 

In this case, it doesn't really matter because the `PopperPrimitive.Root` only renders a provider and not a node but this is good knowledge to have. Might be easier to go into more detail on a call on this one though if you're interested 🙂

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: Love it.

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: Interestingly I almost raised this on your PRs yesterday Jenna, but then I realised that they were root components which didn't render a node. I wonder if it's worth doing it on these too just for consistency as it caught my attention wrongly (especially when I saw `{children}`)

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: Great point on the size of them @colmtuite, let's leave that feature out of `HoverCard` then.
Regarding the name thing, I still prefer `reEntryThreshold` over `skipDelayDuration`.

colmtuite: `openDelayThreshold`?

.yarn/versions/51d71892.yml
andy-hook: Should be ok now 🙂

packages/react/hover-card/src/HoverCard.tsx
andy-hook: @jjenzz explained this well to me earlier but just so I'm totally clear, in this example I should switch the order of `PopperPrimitive.Root` with `HoverCardProvider` to maintain consistency?

jjenzz: Yeah, let's do that @andy-hook and we can do the same in our other components in another PR later to save confusion in future PRs.

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: we had this at some point, but we thought it doesn't convey the fact that it only applies when re-entering… I still think `reEntryThreshold` is clearer for that reason. ie. the time is about how do I have to re-enter.
The biggest issue with something like `openDelayThreshold` is that `delay` and `threshold` are kinda similar.

benoitgrelard: I suppose this discussion is now resolved from the POV of `HoverCard`. We can reopen when we look at aligning `Tooltip`'s API.

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: Hey @andy-hook, so for this, I think we can probably get away without a state machine as I think we don't need the extra `enterInvokedRef` potentially? Can we rely on the timer? As if there's a timer that kicked off, we know we entered the trigger.

andy-hook: There's no way to determine that a timer is running without a flag afaik, we could null the `timerRef` after clearing and check against that but it's arguably more complicated.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const context = useHoverCardContext(TRIGGER_NAME);
+
+  return (
+    <PopperPrimitive.Anchor
+      data-state={context.open ? 'open' : 'closed'}
+      {...triggerProps}
+      as={as}
+      ref={forwardedRef}
+      onPointerEnter={composeEventHandlers(props.onPointerEnter, excludeTouch(context.onOpen))}
+      onPointerLeave={composeEventHandlers(props.onPointerLeave, excludeTouch(context.onClose))}
```
andy-hook, line 116: Excluding internal handlers from `touch` to avoid jarring interactions on mobile devices.

packages/react/hover-card/src/HoverCard.tsx
andy-hook: Took a brief look at using a machine, if we really think it will help the clarity I can make the change but I didn't feel convinced by the extra lines required.

andy-hook:
@jjenzz @benoitgrelard  I've made the requested changes with the exclusion of open comments. Let me know if you'd like to discuss anything else.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,227 @@
...
+      setOpen(true);
+      openInvokedRef.current = true;
+    }, openDelay);
+  }, [openDelay, setOpen]);
+
+  const handleClose = React.useCallback(() => {
+    clearTimeout(openTimerRef.current);
+
+    closeTimerRef.current = window.setTimeout(() => {
+      if (openInvokedRef.current) {
```
benoitgrelard, line 68: I managed to confuse myself about this now, but remind me why we even need this check?

andy-hook: I see what you meant now 🤦 . We don't need this check.

colmtuite:
Here's a use case for skipping the `openDelay` when quickly hovering new triggers.

https://user-images.githubusercontent.com/805073/114895866-bf842400-9e07-11eb-9a60-cfd273ac7c4f.mov

I think this is just the wrong primitive here, since the popup is displaying key information that shouldn't be hidden, and its trigger is not a link or even a button. A combination of Popover and Tooltip would likely be a better option here.

Just sharing to further the research.

benoitgrelard:
> Here's a use case for skipping the `openDelay` when quickly hovering new triggers.
> 
>  Screen.Recording.2021-04-15.at.16.28.56.mov 
> I think this is just the wrong primitive here, since the popup is displaying key information that shouldn't be hidden, and its trigger is not a link or even a button. A combination of Popover and Tooltip would likely be a better option here.
> 
> Just sharing to further the research.

Yeah I think this is probably a `Tooltip` rather. If you wanted to make this accessible, I think you'd use roving tabindex and arrow keys to focus each graph line with the keyboard and the tooltip would enunciate correctly when focused.

packages/react/hover-card/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "dependencies": {
+    "@babel/runtime": "^7.13.10",
+    "@radix-ui/primitive": "workspace:*",
+    "@radix-ui/react-context": "workspace:*",
+    "@radix-ui/react-polymorphic": "workspace:*",
+    "@radix-ui/react-popper": "workspace:*",
+    "@radix-ui/react-portal": "workspace:*",
+    "@radix-ui/react-presence": "workspace:*",
+    "@radix-ui/react-primitive": "workspace:*",
+    "@radix-ui/react-slot": "workspace:*",
```
benoitgrelard, line 28: I think this one isn't used anymore, will need to re-run `yarn` for the lock file too.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const [open = false, setOpen] = useControllableState({
+    prop: openProp,
+    defaultProp: defaultOpen,
+    onChange: onOpenChange,
+  });
+
+  const handleOpen = React.useCallback(() => {
+    clearTimeout(closeTimerRef.current);
+
+    openTimerRef.current = window.setTimeout(() => {
```
benoitgrelard, line 57: quick nit if I may ask, can we remove the curly brackets around `setOpen` to clean things up even more?

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,222 @@
...
+      setOpen(true);
+    }, openDelay);
+  }, [openDelay, setOpen]);
+
+  const handleClose = React.useCallback(() => {
+    clearTimeout(openTimerRef.current);
+
+    closeTimerRef.current = window.setTimeout(() => {
```
benoitgrelard, line 65: quick nit if I may ask, can we remove the curly brackets around `setOpen` to clean things up even more?

packages/react/hover-card/src/HoverCard.stories.tsx
```diff
@@ -0,0 +1,676 @@
...
+            marginTop: 10,
+          }}
+        />
+      </div>
+    </div>
+  );
+}
+
+const RECOMMENDED_CSS__HOVERCARD__CONTENT = {
+  transformOrigin: 'var(--radix-hovercard-content-transform-origin)',
```
benoitgrelard, line 598:
```suggestion
  transformOrigin: 'var(--radix-hover-card-content-transform-origin)',
```

packages/react/hover-card/src/HoverCard.stories.tsx
```diff
@@ -0,0 +1,676 @@
...
+  to: { transform: 'scale(1)', opacity: 1 },
+});
+
+const fadeOut = css.keyframes({
+  from: { transform: 'scale(1)', opacity: 1 },
+  to: { transform: 'scale(0.9)', opacity: 0 },
+});
+
+const animatedContentClass = css(contentClass, {
+  transformOrigin: 'top center',
```
benoitgrelard, line 623: We should remove this, as it is actually overriding the more accurate one from line `598`.

The reason it's a CSS custom property is because it's calculated dynamically from the our popper solution to be the perfect point at the arrow tip. This will update correctly also based on all side, align, and collisions (when it swaps side, etc).

benoitgrelard (commented):
Last few rounds of comments from me and I think for me it's good to go!

packages/react/hover-card/src/HoverCard.stories.tsx
andy-hook: The `Popper` prim is just great

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+  const [open = false, setOpen] = useControllableState({
+    prop: openProp,
+    defaultProp: defaultOpen,
+    onChange: onOpenChange,
+  });
+
+  const handleOpen = React.useCallback(() => {
+    clearTimeout(closeTimerRef.current);
+
```
benoitgrelard, line 56: Sorry to be a pain, but can we remove the line in between here as it doesn't really add anything now that it's so small. (same for the close one below too).

andy-hook: Keep up the nits, it's good brain training for aligning me on project style.

benoitgrelard (approved):
All good from me! ♥️
I'll let @jjenzz do her last pass and then we can merge 🎉

jjenzz (approved):

PR #604, opened 2021-04-14
[ContextMenu] Add `onOpenChange` callback

jjenzz:
### Description

We were missing this prop so that people can use the state to apply styles or whatever fancy stuffs they're up to 🙂


benoitgrelard (approved):
I've just modified the highlight to set it on the trigger as it was a bit weird to set it on the entire wrapper, otherwise all good!

PR #613, opened 2021-04-20
[Polymorphic] Improve perf and fix `as={Comp}` props

jjenzz:
### Description

Does what it says on the 🥫 

Pedro reported an issue today with this. When trying to pass a component to the `as` prop we would lose prop autocompletion for the custom props on the passed component.

I've fiddled about with this for a while and it seems I've managed to improve perf too as an unintended side effect 😅  bonus.

I've also added some new tests for the `ref` prop to make sure that still works!

https://user-images.githubusercontent.com/175330/115457283-37918600-a21c-11eb-9445-fd6862a7da14.mov

### DIFF
--- a/packages/react/polymorphic/src/polymorphic.test.tsx
+++ b/packages/react/polymorphic/src/polymorphic.test.tsx
@@ -99,8 +99,21 @@ const Anchor = React.forwardRef((props, forwardedRef) => {
 /* -----------------------------------------------------------------------------------------------*/
 
 export function Test() {
+  const buttonRef = React.useRef<React.ElementRef<typeof Button>>(null);
+  const buttonAsDivRef = React.useRef<HTMLDivElement>(null);
+  const buttonAsLinkRef = React.useRef<React.ElementRef<typeof Link>>(null);
+
   return (
     <>
+      {/* Button accepts ref */}
+      <Button ref={buttonRef} />
+
+      {/* Button as "div" accepts ref */}
+      <Button as="div" ref={buttonAsDivRef} />
+
+      {/* Button as Link accepts ref */}
+      <Button as={Link} ref={buttonAsLinkRef} />
+
       {/* Link accepts onToggle prop */}
       <Link onToggle={(open) => console.log(open)} />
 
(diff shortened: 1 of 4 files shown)


packages/react/polymorphic/src/polymorphic.test.tsx
```diff
@@ -40,7 +40,7 @@ export function ExtendedButtonUsingReactUtilsWithInternalInlineAs(
   props: React.ComponentProps<typeof Button>
 ) {
   /* Should not error with inline `as` component */
-  return <Button as={(props) => <button {...props} />} {...props} />;
+  return <Button as={(prop: any) => <button {...props} />} {...props} />;
```
jjenzz, line 43: I spent a while trying to type these props for the consumer but it wasn't as easy as it sounds so parked it. Before they were typed as `any` but now they consumer needs explicitly type them.

Inline `as` prop functions are pretty rare so I'm not too worried about this. Hoping we can wait for feedback from consumers before we invest more time on it.

packages/react/polymorphic/src/polymorphic.ts
```diff
@@ -32,30 +31,26 @@ interface ForwardRefComponent<
...
+  <
+    As extends
+      | keyof JSX.IntrinsicElements
+      | React.JSXElementConstructor<any> = NarrowIntrinsic<IntrinsicElementString>
+  >(
+    props: As extends keyof JSX.IntrinsicElements
+      ? Merge<JSX.IntrinsicElements[As], OwnProps & { as: As }>
+      : As extends React.JSXElementConstructor<infer P>
+      ? Merge<P, OwnProps & { as: As }>
+      : never
```
jjenzz, line 53: Instead of having two separate overloads, I've merged them into one overload for intrinsic elements and components. I narrow the types based on which type is passed to ensure we maintain the event types for intrinsics.

packages/react/polymorphic/src/polymorphic.test.tsx
jjenzz: I've just noticed I've done `prop` instead of `props` 🙈 I'll fix that later

peduarte:
🔥

benoitgrelard (approved):
Love it, this is noticeably faster 💨

PR #618, opened 2021-04-23
[RovingFocusGroup] Refactor internals to take into account hidden items

### LINKED ISSUE #581: [DropdownMenu] Items that are invisible cause focus issues

benoitgrelard:
Closes #581 

---
> Note: Better view with whitespace turned off
---

This PR primarily refactors the internals of `RovingFocusGroup` so we can take into account the fact that hidden items don't get focused by the browser.

The crux of the implementation here is that we've introduced a wrapper as part of that API which will act as a proxy to send focus programatically to the correct place. We have a reference to all items in the group and so, we can programatically and progressively attempt to focus the "first" item, and eventually land on the correct one (the first visible one) without having to resort to calculating computed styles to know if an item is indeed hidden.

Once that was done, all of the dependent packages needed updated accordingly.

This has made the overall implementation(s) more declarative, and more the React way.

> Note: There are now technically new changes to document for some packages because they now get the underlying props from `RovingFocusGroup`
- ✨ `RadioGroup` now gets new `orientation`, `dir` and `loop` props
- ✨ `ToggleGroup` now gets new `orientation`, `dir` and `loop` props 

cc @andy-hook, may be good for you to take a quick look through this to see if there's any impact on what you've been working on.

### DIFF
--- a/packages/react/collection/src/Collection.tsx
+++ b/packages/react/collection/src/Collection.tsx
@@ -1,62 +1,68 @@
 import React from 'react';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
-import { Primitive } from '@radix-ui/react-primitive';
+import { Slot } from '@radix-ui/react-slot';
 
-import type * as Polymorphic from '@radix-ui/react-polymorphic';
+type SlotProps = React.ComponentProps<typeof Slot>;
 
-function createCollection<ItemProps>() {
+function createCollection<ItemElement extends HTMLElement, ItemData>() {
   /* -----------------------------------------------------------------------------------------------
    * Collection
    * ---------------------------------------------------------------------------------------------*/
 
-  type Ref = React.RefObject<HTMLElement>;
-  type ContextValue = { collectionRef: Ref; itemMap: Map<Ref, ItemProps & { ref: Ref }> };
+  type CollectionElement = HTMLElement;
+
+  type ContextValue = {
+    collectionRef: React.RefObject<CollectionElement>;
+    itemMap: Map<React.RefObject<ItemElement>, { ref: React.RefObject<ItemElement> } & ItemData>;
+  };
   const Context = React.createContext<ContextValue>({} as any);
 
-  type CollectionOwnProps = Polymorphic.OwnProps<typeof Primitive>;
-  type CollectionPrimitive = Polymorphic.ForwardRefComponent<
-    Polymorphic.IntrinsicElement<typeof Primitive>,
-    CollectionOwnProps
-  >;
+  const COLLECTION_SLOT_NAME = 'CollectionSlot';
 
-  const Collection = React.forwardRef((props, forwardedRef) => {
-    const ref = React.useRef<React.ElementRef<typeof Primitive>>(null);
+  const CollectionSlot = React.forwardRef<CollectionElement, SlotProps>((props, forwardedRef) => {
+    const { children } = props;
+    const ref = React.useRef<CollectionElement>(null);
     const composedRefs = useComposedRefs(forwardedRef, ref);
     const itemMap = React.useRef<ContextValue['itemMap']>(new Map()).current;
     return (
       <Context.Provider value={React.useMemo(() => ({ itemMap, collectionRef: ref }), [itemMap])}>
-        <Primitive {...props} ref={composedRefs} />
+        <Slot ref={composedRefs}>{children}</Slot>
       </Context.Provider>
     );
-  }) as CollectionPrimitive;
+  });
+
+  CollectionSlot.displayName = COLLECTION_SLOT_NAME;
 
   /* -----------------------------------------------------------------------------------------------
    * CollectionItem
    * ---------------------------------------------------------------------------------------------*/
 
-  type CollectionItemOwnProps = Polymorphic.Merge<
-    Polymorphic.OwnProps<typeof Primitive>,
-    ItemProps
-  >;
-  type CollectionItemPrimitive = Polymorphic.ForwardRefComponent<
-    Polymorphic.IntrinsicElement<typeof Primitive>,
-    CollectionItemOwnProps
-  >;
-
+  const ITEM_SLOT_NAME = 'CollectionItemSlot';
   const ITEM_DATA_ATTR = 'data-radix-collection-item';
 
-  const CollectionItem = React.forwardRef((props, forwardedRef) => {
-    const ref = React.useRef<HTMLElement>(null);
-    const composedRefs = useComposedRefs(forwardedRef, ref);
-    const context = React.useContext(Context);
+  type CollectionItemSlotProps = { children: React.ReactNode } & ItemData;
 
-    React.useEffect(() => {
-      context.itemMap.set(ref, { ref, ...props });
-      return () => void context.itemMap.delete(ref);
-    });
+  const CollectionItemSlot = React.forwardRef<ItemElement, CollectionItemSlotProps>(
+    (props, forwardedRef) => {
+      const { children, ...itemData } = props;
+      const ref = React.useRef<ItemElement>(null);
+      const composedRefs = useComposedRefs(forwardedRef, ref);
+      const context = React.useContext(Context);
 
-    return <Primitive {...{ [ITEM_DATA_ATTR]: '' }} {...props} ref={composedRefs} />;
-  }) as CollectionItemPrimitive;
+      React.useEffect(() => {
+        context.itemMap.set(ref, { ref, ...((itemData as unknown) as ItemData) });
+        return () => void context.itemMap.delete(ref);
+      });
+
+      return (
+        <Slot {...{ [ITEM_DATA_ATTR]: '' }} ref={composedRefs}>
+          {children}
+        </Slot>
+      );
+    }
+  );
+
+  CollectionItemSlot.displayName = ITEM_SLOT_NAME;
 
   /* -----------------------------------------------------------------------------------------------
    * useCollection
@@ -78,7 +84,7 @@ function createCollection<ItemProps>() {
     };
   }
 
-  return [Collection, CollectionItem, useCollection] as const;
+  return [CollectionSlot, CollectionItemSlot, useCollection] as const;
 }
 
 export { createCollection };
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -414,9 +414,8 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
   };
 
   const handleLeave = () => {
-    // on leave, we reset the content (we focus it)
+    // on leave, we reset (the content gets focused, and the current item tab stop gets reset)
     contentContext.onReset();
-    if (ref.current) ref.current.tabIndex = -1;
   };
 
   React.useEffect(() => {
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -67,7 +67,7 @@ const CONTENT_NAME = 'MenuContent';
 
 type ItemData = { disabled: boolean };
 const [CollectionSlot, CollectionItemSlot, useCollection] = createCollection<
-  HTMLSpanElement,
+  React.ElementRef<typeof MenuItem>,
   ItemData
 >();
 
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { RemoveScroll } from 'react-remove-scroll';
 import { hideOthers } from 'aria-hidden';
 import { composeEventHandlers } from '@radix-ui/primitive';
+import { createCollection } from '@radix-ui/react-collection';
 import { composeRefs, useComposedRefs } from '@radix-ui/react-compose-refs';
 import { createContext } from '@radix-ui/react-context';
 import { DismissableLayer } from '@radix-ui/react-dismissable-layer';
@@ -64,11 +65,13 @@ Menu.displayName = MENU_NAME;
 
 const CONTENT_NAME = 'MenuContent';
 
-type MenuItemRef = React.RefObject<React.ElementRef<typeof MenuItem>>;
-type MenuContentContextValue = {
-  itemMap: Map<MenuItemRef, { ref: MenuItemRef; disabled: boolean }>;
-  onItemLeave(): void;
-};
+type ItemData = { disabled: boolean };
+const [CollectionSlot, CollectionItemSlot, useCollection] = createCollection<
+  HTMLSpanElement,
+  ItemData
+>();
+
+type MenuContentContextValue = { onItemLeave(): void };
 const [MenuContentProvider, useMenuContentContext] = createContext<MenuContentContextValue>(
   CONTENT_NAME
 );
@@ -94,11 +97,13 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
   const context = useMenuContext(CONTENT_NAME);
   return (
     <Presence present={forceMount || context.open}>
-      <MenuContentImpl
-        data-state={getOpenState(context.open)}
-        {...contentProps}
-        ref={forwardedRef}
-      />
+      <CollectionSlot>
+        <MenuContentImpl
+          data-state={getOpenState(context.open)}
+          {...contentProps}
+          ref={forwardedRef}
+        />
+      </CollectionSlot>
     </Presence>
   );
 }) as MenuContentPrimitive;
@@ -206,8 +211,8 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
   const context = useMenuContext(CONTENT_NAME);
   const contentRef = React.useRef<HTMLDivElement>(null);
   const typeaheadProps = useMenuTypeahead();
+  const { getItems } = useCollection();
 
-  const itemMap = React.useRef<MenuContentContextValue['itemMap']>(new Map()).current;
   const [currentItemId, setCurrentItemId] = React.useState<string | null>(null);
   const [
     isPermittedPointerDownOutsideEvent,
@@ -231,7 +236,6 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
     <PortalWrapper>
       <ScrollLockWrapper>
         <MenuContentProvider
-          itemMap={itemMap}
           onItemLeave={React.useCallback(() => {
             contentRef.current?.focus();
             setCurrentItemId(null);
@@ -337,9 +341,7 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
                           if (event.target !== content) return;
                           if (!ALL_KEYS.includes(event.key)) return;
                           event.preventDefault();
-                          const items = Array.from(itemMap.values()).filter(
-                            (item) => !item.disabled
-                          );
+                          const items = getItems().filter((item) => !item.disabled);
                           const candidateNodes = items.map((item) => item.ref.current!);
                           if (LAST_KEYS.includes(event.key)) candidateNodes.reverse();
                           focusFirst(candidateNodes);
@@ -385,13 +387,6 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
   const context = useMenuContext(ITEM_NAME);
   const contentContext = useMenuContentContext(ITEM_NAME);
 
-  // We keep an up to date map of every item. We do this on every render
-  // to make sure the map insertion order reflects the DOM order.
-  React.useEffect(() => {
-    contentContext.itemMap.set(ref, { ref, disabled });
-    return () => void contentContext.itemMap.delete(ref);
-  });
-
   // get the item's `.textContent` as default strategy for typeahead `textValue`
   const [textContent, setTextContent] = React.useState('');
   React.useEffect(() => {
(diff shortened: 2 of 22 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -645,6 +652,16 @@ function getCheckedState(checked: boolean) {
...
+function focusFirst(candidates: HTMLElement[]) {
+  const PREVIOUSLY_FOCUSED_ELEMENT = document.activeElement;
+  for (const candidate of candidates) {
+    // if focus is already where we want to go, we don't want to keep going through the candidates
+    if (candidate === PREVIOUSLY_FOCUSED_ELEMENT) return;
+    candidate.focus();
+    if (document.activeElement !== PREVIOUSLY_FOCUSED_ELEMENT) return;
+  }
+}
+
```
benoitgrelard, line 664: This is duplicated from `RovingFocusGroup`, let me know if you think we should do something about that.

packages/react/menu/src/Menu.tsx
```diff
@@ -348,36 +359,40 @@ MenuContent.displayName = CONTENT_NAME;
...
+  const ref = React.useRef<HTMLDivElement>(null);
+  const composedRefs = useComposedRefs(forwardedRef, ref);
   const context = useMenuContext(ITEM_NAME);
   const contentContext = useMenuContentContext(ITEM_NAME);
-  const rovingFocusProps = useRovingFocus({ disabled });
+
+  // We keep an up to date map of every item. We do this on every render
+  // to make sure the map insertion order reflects the DOM order.
+  React.useEffect(() => {
+    if (!disabled) {
```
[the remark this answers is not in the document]
benoitgrelard: Also, this is analogous (duplicated) to what we do in `RovingFocusGroup`. Let me know what you think.

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -1,43 +1,46 @@
...
+  Omit<
+    Polymorphic.OwnProps<typeof RovingFocusGroup>,
+    'currentTabStopId' | 'defaultCurrentTabStopId' | 'onCurrentTabStopIdChange' | 'onEntryFocus'
+  >,
   {
     name?: string;
     value?: string;
     defaultValue?: string;
     required?: React.ComponentProps<typeof Radio>['required'];
-    rovingFocus?: boolean;
```
benoitgrelard, line 26: Removed all the dead code about handling no-roving case as it was a leftover from its use in `Toolbar` which has now been replaced by `ToggleGroup`.

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -47,12 +50,15 @@ const [RadioGroupProvider, useRadioGroupContext] = createContext<RadioGroupConte
...
+    as = RADIO_GROUP_DEFAULT_TAG,
     name,
     'aria-labelledby': ariaLabelledby,
     defaultValue,
     value: valueProp,
     required,
-    rovingFocus = true,
+    orientation,
+    dir = 'ltr',
+    loop = true,
```
benoitgrelard, line 61: These are now new props for `RadioGroup` and will need documenting.

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -92,47 +102,37 @@ RadioGroup.displayName = RADIO_GROUP_NAME;
 
 const ITEM_NAME = 'RadioGroupItem';
 
-type RadioGroupItemOwnProps = Polymorphic.OwnProps<typeof RadioGroupItemImpl>;
+type RadioGroupItemOwnProps = Polymorphic.Merge<
+  Polymorphic.Merge<
+    Omit<Polymorphic.OwnProps<typeof RovingFocusItem>, 'focusable' | 'active'>,
+    Polymorphic.OwnProps<typeof Radio>
+  >,
+  { value: string; name?: never }
```
benoitgrelard, line 110: Fixed the remove `name` prop as it's a DOM attribute it would get re-added otherwise.

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -1,144 +1,237 @@
...
+        tabIndex={isTabbingBackOut ? -1 : 0}
+        aria-orientation={orientation}
+        data-orientation={orientation}
+        {...groupProps}
+        as={as}
+        ref={composedRefs}
+        onFocus={composeEventHandlers(props.onFocus, (event) => {
+          if (!isTabbingBackOut && event.target === event.currentTarget) {
+            const entryFocusEvent = new Event(ENTRY_FOCUS, EVENT_OPTIONS);
+            event.currentTarget.dispatchEvent(entryFocusEvent);
```
benoitgrelard, line 124: Like I said, not a huge fan of the name here, would gladly take other suggestions :)

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -165,18 +258,31 @@ function getFocusIntent(event: React.KeyboardEvent, orientation?: Orientation, d
...
 
-function getRovingFocusItems(groupId: string): HTMLElement[] {
-  return Array.from(document.querySelectorAll(`[${GROUP_ID_DATA_ATTR}="${groupId}"]`));
+/**
+ * Wraps an array around itself at a given start index
+ * Example: `wrapArray(['a', 'b', 'c', 'd'], 2) === ['c', 'd', 'a', 'b']`
+ */
+function wrapArray<T>(array: T[], startIndex: number) {
+  return array.map((_, index) => array[(startIndex + index) % array.length]);
 }
```
benoitgrelard, line 277: This is duplicated from `menuTypeahead.tsx`. Let me know if you think we should do something about it.

packages/react/toggle-group/src/ToggleGroup.tsx
```diff
@@ -178,34 +179,54 @@ const [ToggleGroupContext, useToggleGroupContext] = createContext<ToggleGroupCon
...
 const ToggleGroupImpl = React.forwardRef((props, forwardedRef) => {
-  const { disabled = false, rovingFocus = true, ...toggleGroupProps } = props;
-  const primitive = <Primitive role="group" {...toggleGroupProps} ref={forwardedRef} />;
+  const {
+    as = TOGGLE_GROUP_DEFAULT_TAG,
+    disabled = false,
+    rovingFocus = true,
+    orientation,
+    dir = 'ltr',
+    loop = true,
```
benoitgrelard, line 212: These are new props for `ToggleGroup` which will need documenting.

packages/react/toggle-group/src/ToggleGroup.tsx
```diff
@@ -227,48 +250,41 @@ type ToggleGroupItemPrimitive = Polymorphic.ForwardRefComponent<
...
 const ToggleGroupItem = React.forwardRef((props, forwardedRef) => {
+  const valueContext = useToggleGroupValueContext(ITEM_NAME);
   const context = useToggleGroupContext(ITEM_NAME);
-  return context.rovingFocus ? (
-    <ToggleGroupRovingFocusItem {...props} ref={forwardedRef} />
-  ) : (
-    <ToggleGroupItemImpl {...props} ref={forwardedRef} />
-  );
+  const pressed = valueContext.value.includes(props.value);
+  const disabled = context.disabled || props.disabled;
```
benoitgrelard, line 256: I changed this to calculate `pressed` and `disabled` in one place, and pass them down to both implementation.

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -1,144 +1,237 @@
...
+
+const RovingFocusItem = React.forwardRef((props, forwardedRef) => {
+  const { as = ITEM_DEFAULT_TAG, focusable = true, active = false, ...itemProps } = props;
+  const id = useId();
+  const ref = React.useRef<HTMLElement>(null);
+  const composedRefs = useComposedRefs(forwardedRef, ref);
   const context = useRovingFocusContext(ITEM_NAME);
-  const isTabStop = itemId === context.tabStopId;
+  const isCurrentTabStop = context.currentTabStopId === id;
+  const isCurrentTabStopRef = React.useRef(isCurrentTabStop);
```
jjenzz, line 173: Do we still need this ref? I can't see it being used.

packages/react/tabs/src/Tabs.tsx
```diff
@@ -140,77 +135,60 @@ TabsList.displayName = TAB_LIST_NAME;
...
       role="tab"
       aria-selected={isSelected}
       aria-controls={tabPanelId}
       aria-disabled={disabled || undefined}
       data-state={isSelected ? 'active' : 'inactive'}
       data-disabled={disabled ? '' : undefined}
-      data-orientation={context.orientation}
       id={tabId}
+      focusable={!disabled}
+      active={isSelected}
```
jjenzz, line 168: Do we want to move these below `...tabProps` like we did with `name` in `RadioGroup` because we are omitting them from the types?

packages/react/toggle-group/src/ToggleGroup.tsx
```diff
@@ -227,48 +250,41 @@ type ToggleGroupItemPrimitive = Polymorphic.ForwardRefComponent<
...
+  const Comp = context.rovingFocus ? ToggleGroupRovingFocusItem : ToggleGroupItemImpl;
```
jjenzz, line 257: We've avoided naming these sorts of conditionals `Comp` in the past I believe because it clashes with when we do `as: Comp = DEFAULT_TAG`. 

For a second I thought that this wasn't using `Primitive` but it is.

packages/react/toolbar/src/Toolbar.tsx
```diff
@@ -14,57 +14,40 @@ import type * as Polymorphic from '@radix-ui/react-polymorphic';
...
-     */
-    loop?: RovingFocusGroupProps['loop'];
-  }
+type ToolbarOwnProps = Omit<
+  Polymorphic.OwnProps<typeof RovingFocusGroup>,
+  'currentTabStopId' | 'defaultCurrentTabStopId' | 'onCurrentTabStopIdChange' | 'onEntryFocus'
 >;
-
 type ToolbarPrimitive = Polymorphic.ForwardRefComponent<
   Polymorphic.IntrinsicElement<typeof Primitive>,
```
jjenzz, line 24: It's not rendering `Primitive` anymore so `typeof TOOLBAR_DEFAULT_TAG`

jjenzz (changes requested):
Super excited about this :tada:

I noticed one little thing when testing that felt a bit odd to me. If the items have been interacted with, and then the focused one is hidden/removed, it feels like we need to reset `hasInteracted`.

CleanShot 2021-04-26 at 12 34 44

packages/react/menu/src/Menu.tsx
```diff
@@ -398,8 +413,13 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
     }
   };
 
+  const handleLeave = () => {
+    // on leave, we reset (the content gets focused, and the current item tab stop gets reset)
+    contentContext.onReset();
```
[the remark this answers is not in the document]
benoitgrelard: It's not always on mouse leave though, as you'll see below, the same callback is called when mouse moves over a disabled item for example.

packages/react/roving-focus/src/RovingFocusGroup.tsx
benoitgrelard: Great catch, no needed anymore, forgot to delete when we removed all the other code.

packages/react/tabs/src/Tabs.tsx
benoitgrelard: Yeah, can do. Do you think we'd need a similar comment?

packages/react/toggle-group/src/ToggleGroup.tsx
benoitgrelard: Would you rather I keep the conditional inline and duplicate the ref/pressed/disabled props?

packages/react/toolbar/src/Toolbar.tsx
benoitgrelard: Yep, missed that one!

packages/react/menu/src/Menu.tsx
```diff
@@ -62,9 +63,10 @@ Menu.displayName = MENU_NAME;
 
 const CONTENT_NAME = 'MenuContent';
 
+type MenuItemRef = React.RefObject<React.ElementRef<typeof MenuItem>>;
 type MenuContentContextValue = {
-  contentRef: React.RefObject<HTMLDivElement>;
-  onItemsReachableChange(open: boolean): void;
+  itemMap: Map<MenuItemRef, MenuItemRef>;
```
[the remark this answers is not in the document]
benoitgrelard: I haven't no, it's just something that came to my mind when I was trying out that sandbox. I'll dig.

packages/react/menu/src/Menu.tsx
```diff
@@ -101,10 +103,18 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
+    Omit<
+      Polymorphic.OwnProps<typeof RovingFocusGroup>,
+      | 'orientation'
+      | 'currentTabStopId'
+      | 'defaultCurrentTabStopId'
+      | 'onCurrentTabStopIdChange'
+      | 'onEntryFocus'
+    >,
+    Polymorphic.OwnProps<typeof PopperPrimitive.Content>
+  >,
```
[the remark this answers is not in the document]
benoitgrelard: I'm not sure I follow here. We still need to "declare" the props this component takes no?

packages/react/menu/src/Menu.tsx
```diff
@@ -220,67 +227,76 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+                {(dismissableLayerProps) => (
+                  <RovingFocusGroup
+                    as={Slot}
+                    dir={dir}
+                    orientation="vertical"
+                    loop={loop}
+                    currentTabStopId={currentItemId}
+                    onCurrentTabStopIdChange={setCurrentItemId}
+                    // we override the default behaviour which automatically focuses the first item
+                    onEntryFocus={(event) => event.preventDefault()}
```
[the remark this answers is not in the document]
benoitgrelard: We have that name in `FocusScope` yes, but that's because it's truly on mount.
This here, doesn't have anything to do with mount or unmount, it's the refocus that happens when the wrapper gets focused.

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
benoitgrelard: Great. I agree btw 🙂

benoitgrelard:
> Super excited about this 🎉
> 
> I noticed one little thing when testing that felt a bit odd to me. If the items have been interacted with, and then the focused one is hidden/removed, it feels like we need to reset `hasInteracted`.
> 
> CleanShot 2021-04-26 at 12 34 44
> 
> Happy to have a call about this one if i'm not making sense 🙂

What I see in the gif seems to make sense to me, so yeah perhaps we should have a call on that particular one!

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
jjenzz: Yeah, but we only extend the `OwnProps` like this when the props are being spread onto that element, which they aren't here. 

You can see below in this file you have done `onEscapeKeyDown?: DismissableLayerProps['onEscapeKeyDown'];` to pull out the props from wrapping components but you haven't merged all their props in.

Might be easier to call on this one?

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
jjenzz: So is it just `onFocusEnter` like `onMouseEnter` ? When focus enters the roving focus group we want to prevent it?

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
jjenzz: I'm confused by that one. I'm thinking that if a `disabled` item is hovered, we either have no items hovered yet or an item's mouse leave was triggered. I think I need you to help me understand this one on our call but either way, we can have another callback for that if needed. My point here is more about changing this to events and not commands.

packages/react/tabs/src/Tabs.tsx
jjenzz: I don't think so personally, because i think that when things are moved below props spread, they are there for a reason. But I know you like to keep things consistently ordered so might be more of an issue for you looking at your own code later?

packages/react/toggle-group/src/ToggleGroup.tsx
jjenzz: Yeah I'm okay with that duplication in this instance because it's super simple, nothing complex going on there.

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -1,144 +1,237 @@
...
+      onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
+        if (event.key === 'Tab' && event.shiftKey) {
+          context.onItemShiftTab();
+          return;
+        }
+
+        if (event.target !== event.currentTarget) return;
+
+        // stop key events from propagating to parent in case we're in a nested roving focus group
+        if (KEYS.includes(event.key)) event.stopPropagation();
```
andy-hook, line 206: I have a case where I'd like to handle `keydown` in a parent element rather than binding directly to the roving items, has this come up before? wondering whether it's worth having an option for this.

benoitgrelard: You mean perhaps it's not a good idea to stop propagation early here?

This made me realise that it's not needed anymore actually because I now check on the line above that it's an even happening on the current target, and not a bubbling event.

So I can remove it, and I think that solves your question too then?

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
benoitgrelard: Heads up, after checking this intuition, it turned out to be true, and if the users memoizes an item, it absolutely destroys our solution and makes the order unpredictable… So we're back to the drawing board on the tracking aspect. After discussing with @jjenzz, we think we're going to go back to the DOM again but perhaps in something a bit more of a general solution as it's needed in a few places, and we realised that other components are broken because of the same issue.

packages/react/roving-focus/src/RovingFocusGroup.tsx
andy-hook: Yes exactly, this solves it.

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
benoitgrelard: Well that's the thing, `onFocusEnter` just sounds like `onFocus`, which is the event we hook onto. But because this is a custom event, I'd rather want to name it according to what it does, rather than what event it hooks on so.
If I describe what the event does (with its associated behaviour that you can prevent) is:

> when the wrapper itself gets focused, move focus to intended item

That's what the name should be based on.

packages/react/collection/src/Collection.tsx
```diff
@@ -1,62 +1,68 @@
...
+
+  const CollectionItemSlot = React.forwardRef<ItemElement, CollectionItemSlotProps>(
+    (props, forwardedRef) => {
+      const { children, ...itemData } = props;
+      const ref = React.useRef<ItemElement>(null);
+      const composedRefs = useComposedRefs(forwardedRef, ref);
+      const context = React.useContext(Context);
+
+      React.useEffect(() => {
+        context.itemMap.set(ref, { ref, ...((itemData as unknown) as ItemData) });
```
benoitgrelard, line 53: had to do `as unknow as` for some reason once I added `React.forwardRef`

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -60,12 +82,12 @@ type RovingFocusGroupOwnProps = Polymorphic.Merge<
...
 
-type RovingFocusGroupPrimitive = Polymorphic.ForwardRefComponent<
+type RovingFocusGroupImplPrimitive = Polymorphic.ForwardRefComponent<
   typeof GROUP_DEFAULT_TAG,
-  RovingFocusGroupOwnProps
+  RovingFocusGroupImplOwnProps
 >;
 
-const RovingFocusGroup = React.forwardRef((props, forwardedRef) => {
+const RovingFocusGroupImpl = React.forwardRef((props, forwardedRef) => {
```
benoitgrelard, line 90: Added extra `Impl` component so we can get access to the `useCollection` hook.

packages/react/menu/src/Menu.tsx
```diff
@@ -64,11 +65,13 @@ Menu.displayName = MENU_NAME;
...
 const CONTENT_NAME = 'MenuContent';
 
-type MenuItemRef = React.RefObject<React.ElementRef<typeof MenuItem>>;
-type MenuContentContextValue = {
-  itemMap: Map<MenuItemRef, { ref: MenuItemRef; disabled: boolean }>;
-  onItemLeave(): void;
-};
+type ItemData = { disabled: boolean };
+const [CollectionSlot, CollectionItemSlot, useCollection] = createCollection<
+  HTMLSpanElement,
```
benoitgrelard, line 70: Is there a way to get this dynamically from the type of element actually defined in the menu item rather than duplicating here with the risk of getting out of sync?

benoitgrelard (commented):
@jjenzz This is now ready for review again. The last commit contains the changes to `@radix-ui/react-collection` to use `@radix-ui/react-slot`.

I've left a few new comments/questions inline for you below.

Also, I'm pretty sure that we'll be able to reuse the collection we've just put in place in `Menu` and its items for the typeahead stuff instead of the custom hook. I think it'll help also to bring it inside the main file as part of the process, all in one place.

packages/react/menu/src/Menu.tsx
jjenzz: Do you mean `React.ElementRef<typeof MenuItem>` ?

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
jjenzz: `onFocusEnter` isn't `onFocus` in my mind because React `onFocus` is actually `onFocusIn` 🙈  I'm imagining `onMouseEnter` which doesn't bubble, so it is only triggered when mouse first enters the element and not its children. I was applying that same thinking here, so `onFocusEnter` would be when focus first enters the wrapper and not when it's focusing children. Isn't that what this is?

jjenzz (approved):
I've left a comment above about the `onEntryFocus` naming but just something to consider, I'll leave it up to you as I don't want to hold this up because of it.

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
benoitgrelard: I get that, but my previous comment was more that I think the event should reflect the specific "what", rather than the "when", seeing as that's the behaviour we want to cancel by "preventDefaulting"ing it.

I know one could say the opposite for example, that on a `select`, arrow down to open isn't an event called `onArrowDownOpen` that you prevent, but you'd prevent the keydown when the key is arrow (ie. it's more about the "when" than the "why") but then if we want to go for that, then I'm not sure why we need a custom event at all, we would just stay on what I had originally which was to `preventDefault` in `onFocus`.

What do you think?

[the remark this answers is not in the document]
jjenzz: Interesting. Didn't we move away from `onFocus` because that would mean preventing browser behaviour too? I'm good to stick with `onEntryFocus` :+1:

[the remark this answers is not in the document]
benoitgrelard: Well, there isn't anything to prevent natively `onFocus` but the reason we moved away was rather that it seemed like implicit knowledge to have the `Menu` component prevent default in `onFocus` just because we happen to know that the `RovingFocusGroup` is doing something in the `onFocus`.

But yeah I think it's enough bikeshedding here considering that it's not really public API anyway.

jjenzz:
All good from me except this one is still there if that's the answer you were looking for?  https://github.com/radix-ui/primitives/pull/618#discussion_r623906613

packages/react/menu/src/Menu.tsx
benoitgrelard: of course! why do I keep forgetting about this?? 😀

PR #625, opened 2021-04-28
[Collection] New package to track items in a collection

benoitgrelard:
This PR is preliminary work to fix some issues we discovered with the new `RovingFocusGroup` solution at #618 (and other packages using the same render-based mechanism to track items).

What it does:
- bring the old collection code into `Slider` (as it's the only package using it) in view of replacing it inside `Slider` by something else, until we see if we need a more generalised solution for index based things. @jjenzz will look at what to do in `Slider` in the meantime
- adds a new implementation inside the `react-collection` package which is not render dependent (for the order)

Details:
- Tracks the items via a `Map`: source of truth of the items tracked
- On demand (`getItems`): uses the DOM as the source of truth for items order in a "just in time" fashion
- This is different from the previous collection package which intention was mostly to provide an index for an item

> Notes

Because of typing, I can't find another way to do it without a factory. I've tried with JSX generics but I don't think it can work because:
- When creating a context with `React.createContext`, there isn't a way to provide a generic
- With our technique for the polymorphism component, we can't pass a generic inside either via inference


jjenzz (approved):
Nice and simple, 💛  &nbsp;&nbsp;it

Issue #383, opened 2021-01-08
New component: Submenu


chaance:
See https://github.com/radix-ui/primitives/discussions/375

andy-hook:
Some research from earlier this week

## MacOS

https://user-images.githubusercontent.com/11708259/115904712-0358d800-a45d-11eb-874b-a2fedfb8e9c9.mov

Notable interactions
- Closes entire menu on `Esc` press

## Chrome

https://user-images.githubusercontent.com/11708259/115904775-179cd500-a45d-11eb-865d-b2e10c786273.mov

Notable interactions
- Closes individual sub menus with each `Esc` press
- Provides a debounced, timer based closing mechanism when browsing the parent menu. After interacting with the sub menu it will remain open until moving back to the parent.

## Figma

https://user-images.githubusercontent.com/11708259/115904831-2be0d200-a45d-11eb-89c3-92670fcff9de.mov

Notable interactions
- Closes individual sub menus with each `Esc` press
- Menus remain open until interacting with the parent menu

### Baseline requirements

- Must be keyboard navigable
- Must be able to open and close sub menus with `ArrowLeft` / `ArrowRight` and `Esc` / `Enter`
- Must focus first in list when opening via keyboard (saves a keypress)
- Must support multiple levels ("infinitely" nest-able)
- Must highlight item in parent list from which the sub menu originates
- Must close entire menu when clicking outside

### Additional thoughts for our implementation

- Should close sub menus individually for each `Esc` press per the chrome / figma example
- Should offer some amount of pointer grace at the contact edge for easier motor control (will need to test if this is necessary on average)

### Initial api thoughts

I'm thinking we add a couple more primitives to our [base Menu](https://github.com/radix-ui/primitives/blob/main/packages/react/menu/src/Menu.tsx) which provide uniform behaviour for the keyboard and pointer interactions. How that is exposed publicly will depend on the component, but they will likely be [name spaced and exported verbatim](https://github.com/radix-ui/primitives/blob/b2807bc610e04fe4088a36a279ffd9c1b2fb3a6c/packages/react/dropdown-menu/src/DropdownMenu.tsx#L176-L200)

Will follow up with an api proposal shortly.

PR #627, opened 2021-04-30
[Menu] Add submenu primitives

andy-hook:
Closes https://github.com/radix-ui/primitives/issues/383

### Description

This PR adds submenu functionality to our base `Menu`

#### Approach

Additional primitives have been added to `Menu` to provide consistent submenu support across several consuming components.

Here is an example of how we can use the pieces to provide submenu support inside our components

```jsx
<Menu.Root>
  <Menu.Anchor>Open</Menu.Anchor>
  <Menu.Content>
    <Menu.Item>Item 1</Menu.Item>
    <Menu.Item>Item 2</Menu.Item>
    
    <Menu.Sub>
      <Menu.SubTrigger>
         Open Sub Menu
      </Menu.SubTrigger>

      <Menu.Content>
        <Menu.Item>Sub Item 1</Menu.Item>
        <Menu.Item>Sub Item 2</Menu.Item>
      </Menu.Content>
    </Menu.Sub>
  </Menu.Content>
</Menu.Root>
```

#### TODO
- [x] Adjust on top of https://github.com/radix-ui/primitives/pull/618 once merged
- [x] Add `rtl` support
- [x] Update versions
- [ ] Expose via nesting api in `ContextMenu` and `DropdownMenu` (see https://github.com/radix-ui/primitives/pull/629)
- [ ] Adjust typeahead to support nesting (see https://github.com/radix-ui/primitives/pull/662)
- [ ] Apply pointer grace on transit to sub menu (will raise a separate improvement)

### DIFF
--- a/packages/react/menu/src/Menu.stories.tsx
+++ b/packages/react/menu/src/Menu.stories.tsx
@@ -1,7 +1,9 @@
 import * as React from 'react';
 import {
-  Menu as MenuPrimitive,
+  Menu,
+  MenuSub,
   MenuAnchor,
+  MenuSubTrigger,
   MenuContent,
   MenuGroup,
   MenuLabel,
@@ -11,6 +13,7 @@ import {
   MenuRadioItem,
   MenuItemIndicator,
   MenuSeparator,
+  MenuArrow,
 } from './Menu';
 import { css } from '../../../../stitches.config';
 import { foodGroups } from '../../../../test-data/foods';
@@ -21,7 +24,7 @@ export default {
 };
 
 export const Styled = () => (
-  <Menu>
+  <MenuWithAnchor>
     <MenuItem className={itemClass} onSelect={() => window.alert('undo')}>
       Undo
     </MenuItem>
@@ -38,11 +41,113 @@ export const Styled = () => (
     <MenuItem className={itemClass} onSelect={() => window.alert('paste')}>
       Paste
     </MenuItem>
-  </Menu>
+  </MenuWithAnchor>
 );
 
+export const Submenus = () => {
+  const [open1, setOpen1] = React.useState(false);
+  const [open2, setOpen2] = React.useState(false);
+  const [open3, setOpen3] = React.useState(false);
+  const [open4, setOpen4] = React.useState(false);
+  const [rtl, setRtl] = React.useState(false);
+  const [animated, setAnimated] = React.useState(false);
+
+  React.useEffect(() => {
+    if (rtl) {
+      document.documentElement.setAttribute('dir', 'rtl');
+      return () => document.documentElement.removeAttribute('dir');
+    }
+  }, [rtl]);
+
+  return (
+    <>
+      <div style={{ marginBottom: 8, display: 'grid', gridAutoFlow: 'row', gridGap: 4 }}>
+        <label>
+          <input
+            type="checkbox"
+            checked={rtl}
+            onChange={(event) => setRtl(event.currentTarget.checked)}
+          />
+          rtl
+        </label>
+        <label>
+          <input
+            type="checkbox"
+            checked={animated}
+            onChange={(event) => setAnimated(event.currentTarget.checked)}
+          />
+          Animated
+        </label>
+      </div>
+      <MenuWithAnchor>
+        <MenuItem className={itemClass} onSelect={() => window.alert('undo')}>
+          Undo
+        </MenuItem>
+        <Submenu open={open1} onOpenChange={setOpen1} animated={animated}>
+          <MenuItem className={itemClass} disabled>
+            Disabled
+          </MenuItem>
+          <MenuItem className={itemClass} onSelect={() => window.alert('one')}>
+            One
+          </MenuItem>
+          <Submenu open={open2} onOpenChange={setOpen2} animated={animated}>
+            <MenuItem className={itemClass} onSelect={() => window.alert('one')}>
+              One
+            </MenuItem>
+            <MenuItem className={itemClass} onSelect={() => window.alert('two')}>
+              Two
+            </MenuItem>
+            <MenuItem className={itemClass} onSelect={() => window.alert('three')}>
+              Three
+            </MenuItem>
+          </Submenu>
+          <Submenu open={open3} onOpenChange={setOpen3} animated={animated}>
+            <MenuItem className={itemClass} onSelect={() => window.alert('one')}>
+              One
+            </MenuItem>
+            <MenuItem className={itemClass} onSelect={() => window.alert('two')}>
+              Two
+            </MenuItem>
+            <MenuItem className={itemClass} onSelect={() => window.alert('three')}>
+              Three
+            </MenuItem>
+          </Submenu>
+          <MenuItem className={itemClass} onSelect={() => window.alert('two')}>
+            Two
+          </MenuItem>
+          <Submenu open={open4} onOpenChange={setOpen4} animated={animated} disabled>
+            <MenuItem className={itemClass} onSelect={() => window.alert('one')}>
+              One
+            </MenuItem>
+            <MenuItem className={itemClass} onSelect={() => window.alert('two')}>
+              Two
+            </MenuItem>
+            <MenuItem className={itemClass} onSelect={() => window.alert('three')}>
+              Three
+            </MenuItem>
+          </Submenu>
+          <MenuItem className={itemClass} onSelect={() => window.alert('three')}>
+            Three
+          </MenuItem>
+        </Submenu>
+
+        <MenuSeparator className={separatorClass} />
+        <MenuItem className={itemClass} disabled onSelect={() => window.alert('cut')}>
+          Cut
+        </MenuItem>
+        <MenuItem className={itemClass} onSelect={() => window.alert('copy')}>
+          Copy
+        </MenuItem>
+        <MenuItem className={itemClass} onSelect={() => window.alert('paste')}>
+          Paste
+        </MenuItem>
+      </MenuWithAnchor>
+    </>
+  );
+};
+
 export const WithLabels = () => (
-  <Menu>
+  <MenuWithAnchor>
     {foodGroups.map((foodGroup, index) => (
       <MenuGroup key={index}>
         {foodGroup.label && (
@@ -63,7 +168,7 @@ export const WithLabels = () => (
         {index < foodGroups.length - 1 && <MenuSeparator className={separatorClass} />}
       </MenuGroup>
     ))}
-  </Menu>
+  </MenuWithAnchor>
 );
 
 const suits = [
@@ -104,7 +209,7 @@ export const Typeahead = () => (
       <div>
         <h2>Complex children</h2>
         <p>(relying on `.textContent` — default)</p>
-        <Menu>
+        <MenuWithAnchor>
           {suits.map((suit) => (
             <MenuItem key={suit.emoji} className={itemClass}>
               {suit.label}
@@ -113,13 +218,13 @@ export const Typeahead = () => (
               </span>
             </MenuItem>
           ))}
-        </Menu>
+        </MenuWithAnchor>
       </div>
 
       <div>
         <h2>Complex children</h2>
         <p>(with explicit `textValue` prop)</p>
-        <Menu>
+        <MenuWithAnchor>
           {suits.map((suit) => (
             <MenuItem key={suit.emoji} className={itemClass} textValue={suit.label}>
               <span role="img" aria-label={suit.label}>
@@ -128,7 +233,7 @@ export const Typeahead = () => (
               {suit.label}
             </MenuItem>
           ))}
-        </Menu>
+        </MenuWithAnchor>
       </div>
     </div>
   </>
@@ -143,7 +248,7 @@ export const CheckboxItems = () => {
   ];
 
   return (
-    <Menu>
+    <MenuWithAnchor>
       <MenuItem className={itemClass} onSelect={() => window.alert('show')}>
         Show fonts
       </MenuItem>
@@ -168,7 +273,7 @@ export const CheckboxItems = () => {
           </MenuItemIndicator>
         </MenuCheckboxItem>
       ))}
-    </Menu>
+    </MenuWithAnchor>
   );
 };
 
@@ -177,7 +282,7 @@ export const RadioItems = () => {
   const [file, setFile] = React.useState(files[1]);
 
   return (
-    <Menu>
+    <MenuWithAnchor>
       <MenuItem className={itemClass} onSelect={() => window.alert('minimize')}>
         Minimize window
       </MenuItem>
@@ -198,7 +303,7 @@ export const RadioItems = () => {
           </MenuRadioItem>
         ))}
       </MenuRadioGroup>
-    </Menu>
+    </MenuWithAnchor>
   );
 };
 
@@ -221,7 +326,7 @@ export const Animated = () => {
       </label>
       <br />
       <br />
-      <Menu className={animatedRootClass} open={open}>
+      <MenuWithAnchor className={animatedContentClass} open={open}>
         {checkboxItems.map(({ label, state: [checked, setChecked], disabled }) => (
           <MenuCheckboxItem
             key={label}
@@ -246,14 +351,13 @@ export const Animated = () => {
             </MenuRadioItem>
           ))}
         </MenuRadioGroup>
-      </Menu>
+      </MenuWithAnchor>
     </>
   );
 };
 
 type MenuOwnProps = Omit<
-  React.ComponentProps<typeof MenuPrimitive> & React.ComponentProps<typeof MenuContent>,
-  | 'onOpenChange'
+  React.ComponentProps<typeof Menu> & React.ComponentProps<typeof MenuContent>,
   | 'portalled'
   | 'trapFocus'
   | 'onOpenAutoFocus'
@@ -262,11 +366,12 @@ type MenuOwnProps = Omit<
   | 'disableOutsideScroll'
 >;
 
-const Menu: React.FC<MenuOwnProps> = (props) => {
+const MenuWithAnchor: React.FC<MenuOwnProps> = (props) => {
   const { open = true, children, ...contentProps } = props;
   return (
-    <MenuPrimitive open={open} onOpenChange={() => {}}>
-      <MenuAnchor />
+    <Menu open={open} onOpenChange={() => {}}>
+      {/* inline-block allows anchor to move when rtl changes on document */}
+      <MenuAnchor style={{ display: 'inline-block' }} />
       <MenuContent
         className={contentClass}
         portalled
@@ -280,7 +385,26 @@ const Menu: React.FC<MenuOwnProps> = (props) => {
       >
         {children}
       </MenuContent>
-    </MenuPrimitive>
+    </Menu>
+  );
+};
+
+const Submenu: React.FC<MenuOwnProps & { animated: boolean; disabled?: boolean }> = (props) => {
+  const { open = true, onOpenChange, children, animated, disabled, ...contentProps } = props;
+  return (
+    <MenuSub open={open} onOpenChange={onOpenChange}>
+      <MenuSubTrigger className={subTriggerClass} disabled={disabled}>
+        Submenu →
+      </MenuSubTrigger>
+      <MenuContent
+        className={animated ? animatedContentClass : contentClass}
+        sideOffset={12}
+        {...contentProps}
+      >
+        {children}
+        <MenuArrow offset={8} />
+      </MenuContent>
+    </MenuSub>
   );
 };
 
@@ -333,37 +457,44 @@ const itemClass = css({
   },
 });
 
+const subTriggerClass = css(itemClass, {
+  '&[data-state="open"]': {
+    backgroundColor: '$gray100',
+    color: '$black',
+  },
+});
+
 const separatorClass = css({
   height: 1,
   margin: '5px 10px',
   backgroundColor: '$gray100',
 });
 
-const fadeIn = css.keyframes({
-  from: { opacity: 0 },
-  to: { opacity: 1 },
+const animateIn = css.keyframes({
+  from: { transform: 'scale(0.95)', opacity: 0 },
+  to: { transform: 'scale(1)', opacity: 1 },
 });
 
-const fadeOut = css.keyframes({
-  from: { opacity: 1 },
-  to: { opacity: 0 },
+const animateOut = css.keyframes({
+  from: { transform: 'scale(1)', opacity: 1 },
+  to: { transform: 'scale(0.95)', opacity: 0 },
 });
 
-const animatedRootClass = css(contentClass, {
+const animatedContentClass = css(contentClass, {
   '&[data-state="open"]': {
-    animation: `${fadeIn} 300ms ease-out`,
+    animation: `${animateIn} 300ms ease`,
   },
   '&[data-state="closed"]': {
-    animation: `${fadeOut} 300ms ease-in`,
+    animation: `${animateOut} 300ms ease`,
   },
 });
 
 const animatedItemIndicatorClass = css({
   '&[data-state="checked"]': {
-    animation: `${fadeIn} 300ms ease-out`,
+    animation: `${animateIn} 300ms ease`,
   },
   '&[data-state="unchecked"]': {
-    animation: `${fadeOut} 300ms ease-in`,
+    animation: `${animateOut} 300ms ease`,
   },
 });
 
--- a/packages/react/menu/src/Menu.stories.tsx
+++ b/packages/react/menu/src/Menu.stories.tsx
@@ -68,7 +68,7 @@ export const Submenus = () => {
             checked={rtl}
             onChange={(event) => setRtl(event.currentTarget.checked)}
           />
-          rtl
+          Right-to-left
         </label>
         <label>
           <input
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -132,9 +132,7 @@ const MenuSub: React.FC<MenuOwnProps> = (props) => {
   React.useEffect(() => {
     const parentMenuContent = context.content;
     const handleParentMenuItemEnter = (event: Event) => {
-      const isItemSubMenuTrigger = event.target === trigger;
-      if (!isItemSubMenuTrigger) handleOpenChange(false);
-      event.stopPropagation();
+      if (event.target !== trigger) handleOpenChange(false);
     };
     if (parentMenuContent) {
       parentMenuContent.addEventListener(ITEM_ENTER, handleParentMenuItemEnter);
@@ -347,13 +345,22 @@ const MenuSubContent = React.forwardRef((props, forwardedRef) => {
       // don't want it to refocus the trigger in that case so we handle trigger focus ourselves.
       onCloseAutoFocus={(event) => event.preventDefault()}
       onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => context.trigger?.focus())}
-      onFocusLeave={composeEventHandlers(props.onFocusLeave, () => context.onOpenChange(false))}
+      onFocusLeave={composeEventHandlers(props.onFocusLeave, (event) => {
+        // We prevent closing when the trigger is focused to avoid triggering a re-open animation
+        // on pointer interaction.
+        if (event.relatedTarget !== context.trigger) context.onOpenChange(false);
+      })}
       onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
         const element = event.target as HTMLElement;
         // Submenu key events bubble through portals. We only care about keys in this menu.
         const isKeyDownInside = event.currentTarget.contains(element);
         const isCloseKey = SUB_CLOSE_KEYS[context.dir].includes(event.key);
-        if (isKeyDownInside && isCloseKey) context.trigger?.focus();
+        if (isKeyDownInside && isCloseKey) {
+          // We need to explicitly close when keyboard focuses trigger because we prevented this
+          // in onFocusLeave for pointers.
+          context.onOpenChange(false);
+          context.trigger?.focus();
+        }
       })}
     />
   ) : null;
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -349,11 +349,15 @@ const MenuSubContent = React.forwardRef((props, forwardedRef) => {
       disableOutsidePointerEvents={false}
       disableOutsideScroll={false}
       trapFocus={false}
+      onEntryFocus={context.onEntryFocus}
       onOpenAutoFocus={(event) => event.preventDefault()}
       // The menu might close because of focusing another menu item in the parent menu. We
       // don't want it to refocus the trigger in that case so we handle trigger focus ourselves.
       onCloseAutoFocus={(event) => event.preventDefault()}
-      onEntryFocus={context.onEntryFocus}
+      onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => {
+        context.onOpenChange(false);
+        context.trigger?.focus();
+      })}
       onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
         const element = event.target as HTMLElement;
         // Submenu key events bubble through portals. We only care about keys in this menu.
@@ -364,9 +368,8 @@ const MenuSubContent = React.forwardRef((props, forwardedRef) => {
           context.trigger?.focus();
         }
       })}
-      onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => {
+      onPointerDownOutside={composeEventHandlers(props.onPointerDownOutside, () => {
         context.onOpenChange(false);
-        context.trigger?.focus();
       })}
     />
   ) : null;
@@ -537,13 +540,13 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
                   style={{ outline: 'none', ...contentProps.style }}
                   onKeyDownCapture={composeEventHandlers(
                     contentProps.onKeyDownCapture,
-                    composeEventHandlers(typeaheadProps.onKeyDownCapture, (event) => {
-                      if (event.key === 'Tab') event.preventDefault();
-                    })
+                    typeaheadProps.onKeyDownCapture
                   )}
                   // focus first/last item based on key pressed
                   onKeyDown={composeEventHandlers(contentProps.onKeyDown, (event) => {
                     const content = contentRef.current;
+                    // menus should not be navigated using tab key so we prevent it
+                    if (event.key === 'Tab') event.preventDefault();
                     if (event.target !== content) return;
                     if (!FIRST_LAST_KEYS.includes(event.key)) return;
                     event.preventDefault();
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -3,7 +3,7 @@ import { RemoveScroll } from 'react-remove-scroll';
 import { hideOthers } from 'aria-hidden';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { createCollection } from '@radix-ui/react-collection';
-import { useComposedRefs } from '@radix-ui/react-compose-refs';
+import { useComposedRefs, composeRefs } from '@radix-ui/react-compose-refs';
 import { createContext } from '@radix-ui/react-context';
 import { DismissableLayer } from '@radix-ui/react-dismissable-layer';
 import { FocusScope } from '@radix-ui/react-focus-scope';
@@ -13,15 +13,30 @@ import * as PopperPrimitive from '@radix-ui/react-popper';
 import { Portal } from '@radix-ui/react-portal';
 import { RovingFocusGroup, RovingFocusItem } from '@radix-ui/react-roving-focus';
 import { Slot } from '@radix-ui/react-slot';
+import { useDirection } from '@radix-ui/react-use-direction';
 import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
 import { useFocusGuards } from '@radix-ui/react-focus-guards';
+import { useId } from '@radix-ui/react-id';
 import { useMenuTypeahead, useMenuTypeaheadItem } from './useMenuTypeahead';
 
 import type * as Polymorphic from '@radix-ui/react-polymorphic';
 
+type MenuContentElement = React.ElementRef<typeof MenuContent>;
+type MenuSubTriggerElement = React.ElementRef<typeof MenuSubTrigger>;
+type Direction = 'ltr' | 'rtl';
+
+const SELECTION_KEYS = ['Enter', ' '];
 const FIRST_KEYS = ['ArrowDown', 'PageUp', 'Home'];
 const LAST_KEYS = ['ArrowUp', 'PageDown', 'End'];
-const ALL_KEYS = [...FIRST_KEYS, ...LAST_KEYS];
+const FIRST_LAST_KEYS = [...FIRST_KEYS, ...LAST_KEYS];
+const SUB_OPEN_KEYS: Record<Direction, string[]> = {
+  ltr: [...SELECTION_KEYS, 'ArrowRight'],
+  rtl: [...SELECTION_KEYS, 'ArrowLeft'],
+};
+const SUB_CLOSE_KEYS: Record<Direction, string[]> = {
+  ltr: ['ArrowLeft'],
+  rtl: ['ArrowRight'],
+};
 
 /* -------------------------------------------------------------------------------------------------
  * Menu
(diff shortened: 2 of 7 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -45,6 +49,17 @@ type MenuOwnProps = {
...
+
+  React.useEffect(() => {
+    const handleMenuClose = (event: Event) => {
+      if (event.defaultPrevented) return;
+      handleOpenChange(false);
+    };
+
+    document.addEventListener(ITEM_SELECT, handleMenuClose);
+    return () => document.removeEventListener(ITEM_SELECT, handleMenuClose);
+  }, [handleOpenChange]);
```
andy-hook, line 61: Raised this up to take advantage of bubbling and close all open menus on selection.

packages/react/menu/src/Menu.tsx
```diff
@@ -621,6 +636,206 @@ const MenuItemIndicator = React.forwardRef((props, forwardedRef) => {
...
+
+const [SubMenuProvider, useSubMenuContext] = createContext<MenuSubMenuContextValue>(SUB_MENU_NAME);
+
+type MenuSubMenuOwnProps = {
+  open?: boolean;
+  onOpenChange?(open: boolean): void;
+  defaultOpen?: boolean;
+};
+
+const MenuSubMenu: React.FC<MenuSubMenuOwnProps> = (props) => {
```
andy-hook, line 661: Adhering to our naming convention though it reads a bit weird. Any thoughts on something better?

packages/react/menu/src/Menu.tsx
```diff
@@ -621,6 +636,206 @@ const MenuItemIndicator = React.forwardRef((props, forwardedRef) => {
...
+
+/* -------------------------------------------------------------------------------------------------
+ * MenuSubMenuContent
+ * -----------------------------------------------------------------------------------------------*/
+
+const SUB_MENU_CONTENT_NAME = 'MenuSubMenuContent';
+
+type MenuSubMenuContentOwnProps = Omit<
+  Polymorphic.OwnProps<typeof MenuContent>,
+  'portalled' | 'disableOutsidePointerEvents' | 'disableOutsideScroll'
```
andy-hook, line 713: Narrowing here because:
- `popper` doesn't currently support relative position afaik, so submenus must always be `portalled`
- The way pointer events interact here is fundamental to submenus working as expected, I can't think of a case for users wanting to change this.

packages/react/menu/src/Menu.tsx
```diff
@@ -621,6 +636,206 @@ const MenuItemIndicator = React.forwardRef((props, forwardedRef) => {
...
+      align="start"
+      {...contentProps}
+      portalled
+      ref={forwardedRef}
+      onOpenAutoFocus={composeEventHandlers(contentProps.onOpenAutoFocus, (event) => {
+        const content = event.target as HTMLElement;
+        if (content && subMenuContext.focusFirstItem) {
+          const items = Array.from(content.querySelectorAll(ENABLED_ITEM_SELECTOR));
+          (items[0] as HTMLElement | undefined)?.focus();
+        }
```
andy-hook, line 738: I don't think this is going to work correctly with hidden items. Was planning to adjust after https://github.com/radix-ui/primitives/pull/618

packages/react/menu/src/Menu.tsx
```diff
@@ -621,6 +636,206 @@ const MenuItemIndicator = React.forwardRef((props, forwardedRef) => {
...
+
+const MenuSubMenuTrigger = React.forwardRef((props, forwardedRef) => {
+  const { disabled, ...triggerProps } = props;
+  const context = useMenuContext(SUB_MENU_TRIGGER_NAME);
+  const subMenuContext = useSubMenuContext(SUB_MENU_TRIGGER_NAME);
+  const [mouseInteracting, setMouseInteracting] = React.useState(false);
+
+  return (
+    <MenuAnchor as={Slot}>
+      <MenuItem
```
andy-hook, line 794: Composing via `Slot` as it felt simpler, the TS event types also seem to get lost when using `as`. Looking at this now perhaps `SubMenuTrigger` should default to `button`?

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -114,8 +114,7 @@ function useRovingFocus({ disabled, active = false }: UseRovingFocusItemOptions)
     onKeyDown: (event: React.KeyboardEvent) => {
       const focusIntent = getFocusIntent(event, context.orientation, context.dir);
 
-      // stop key events from propagating to parent in case we're in a nested roving focus group
-      if (KEYS.includes(event.key)) event.stopPropagation();
+      if (event.target !== event.currentTarget) return;
```
andy-hook, line 117: This change can go away after https://github.com/radix-ui/primitives/pull/618

packages/react/menu/src/Menu.tsx
```diff
@@ -621,6 +636,206 @@ const MenuItemIndicator = React.forwardRef((props, forwardedRef) => {
...
+         * We hook into focus to control menu visibility rather than mouse events
+         * This solves a sticky open state problem when certain browser controls (e.g. devtools inspect overlay) are given focus
+         * This is because mouse events continue to fire while `onDismiss` in `DismissableLayer` won't get called due to it being a focus outside check
+         */
+        onFocus={composeEventHandlers(triggerProps.onFocus, (event) => {
+          if (mouseInteracting) subMenuContext.onMouseOpen();
+        })}
+        onKeyDown={composeEventHandlers(triggerProps.onKeyDown, (event) => {
+          setMouseInteracting(false);
+          if (['Enter', ' ', 'ArrowRight'].includes(event.key)) subMenuContext.onKeyboardOpen();
```
andy-hook, line 830: Should we provide `rtl` support for flipping the `ArrowRight / ArrowLeft` bindings and `side` prop default?

packages/react/menu/src/Menu.tsx
jjenzz: Haha yeah this does read a bit weird but tbh, I'm thinking it would be better DX not to expose these parts to consumers and then in that case the naming doesn't matter too much.

My thinking is that we would render these conditionally internally based on whether the menu is nested inside a menu in the React Tree... something like:

```jsx
const IsSubMenuContext = React.createContext(false);

const Menu = React.forwardRef((props, forwardedRef) => {
  const isSubMenu = React.useContext(IsSubMenuContext);
  
  if (isSubMenu) {
    return <MenuSubMenu {...props} ref={forwardedRef} />
  } else {
    return (
		<IsSubMenuContext.Provider value={true}>
			<MenuImpl {...props} ref={forwardedRef} />
		</IsSubMenuContext.Provider>
	);
  }
});

const MenuSubMenu = React.forwardRef((props, forwardedRef) => {
  // any logic specific to submenu and then
  return <MenuImpl {...props} ref={forwardedRef} />
});

const MenuImpl = React.forwardRef((props, forwardedRef) => {
  // common logic across menu and submenu types
  return <Primitive {...props} ref={forwardedRef} />
});
```

And the same for the other parts. That way the consumers can just nest `Menu`s inside `Menu`s.

jjenzz (changes requested):
Some initial feedback here that will mean moving things around a bit so I'll hold off reviewing in more detail till we've discussed the outcome of this bit.

packages/react/menu/src/Menu.tsx
andy-hook: We discussed this on a call but for posterity – Totally agree 💯 , this is the way I would expect it to work as well. Originally wasn't sure on the user expectation hence the explicitness but our discussion cleared up doubts.

packages/react/menu/src/Menu.tsx
```diff
@@ -361,6 +471,89 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+          onKeyDown={composeEventHandlers(contentProps.onKeyDown, (event) => {
+            const tabNext = event.key === 'Tab' && !event.shiftKey;
+            if (SUB_CLOSE_KEYS[direction].includes(event.key) || tabNext) {
+              // close a single level unless it's a forward tab, in which case close the entire stack
+              if (!tabNext) {
+                event.stopPropagation();
+              }
+              timerRef.current = window.setTimeout(() => {
+                subMenuContext.triggerElement?.focus();
+              }, 0);
```
andy-hook, line 532: I need the `timeout` for when the user tabs forward, this is so I can catch focus and redirect to the root menu via bubble.

packages/react/menu/src/Menu.tsx
```diff
@@ -428,44 +681,80 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
...
          */
-        onMouseMove={composeEventHandlers(props.onMouseMove, (event) => {
-          if (!disabled) {
-            const item = event.currentTarget;
-            item.focus();
-          } else {
-            contentContext.onItemLeave();
+        onFocus={composeEventHandlers(triggerProps.onFocus, (event) => {
+          if (mouseInteracting) {
+            subMenuContext.onMenuOpen(event);
```
andy-hook, line 744: The `mouseInteracting` dance reads a little strange but the inline comment should clarify. I'm basically qualifying the interaction behind `focus` because of the way the layering is designed. Without doing this, when using inspect tools the menus will appear but not dismiss which is annoying AF.

andy-hook:
@jjenzz @benoitgrelard – Would appreciate if you could take a look at this and [its sister](https://github.com/radix-ui/primitives/pull/629) when you have time next week. I still have an outstanding improvement to make but will raise that separately.

jjenzz:
I'm still reviewing this but just posting this potential bug I spotted for you to look into in the meantime. Pressing down on a hovered item in the top-level menu moves down an item in the submenu.

https://user-images.githubusercontent.com/175330/118322760-8e2a7100-b4f7-11eb-969e-8262ed4a0a28.mp4

This is how MacOS behaves:

https://user-images.githubusercontent.com/175330/118322907-c2059680-b4f7-11eb-8b8a-8ef99d07d6ec.mp4

jjenzz:
I think I might have spotted another one?

https://user-images.githubusercontent.com/175330/118330414-a69e8980-b4ff-11eb-8fef-27ff85231311.mp4

Versus:

https://user-images.githubusercontent.com/175330/118330764-bddd7700-b4ff-11eb-9c71-eb6403ecc66f.mp4

packages/react/menu/src/Menu.tsx
jjenzz: Would you mind showing me what the issue was to help me understand it a bit better?

packages/react/menu/src/Menu.tsx
```diff
@@ -636,7 +925,7 @@ const MenuGroup = extendPrimitive(Primitive, {
 const MenuLabel = extendPrimitive(Primitive, { displayName: 'MenuLabel' });
 const MenuSeparator = extendPrimitive(Primitive, {
   defaultProps: { role: 'separator', 'aria-orientation': 'horizontal' },
-  displayName: 'MenuSeparator ',
+  displayName: 'MenuSeparator',
```
jjenzz, line 928: :+1: Good spot!

packages/react/menu/src/Menu.tsx
```diff
@@ -361,6 +471,89 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+const MenuSubContent = React.forwardRef((props, forwardedRef) => {
+  const { forceMount, dir, ...contentProps } = props;
+  const context = useMenuContext(SUB_CONTENT_NAME);
+  const subMenuContext = useMenuSubContext(SUB_CONTENT_NAME);
+  const direction = dir || context.dir;
+  const timerRef = React.useRef(0);
+
+  React.useEffect(() => {
+    return () => window.clearTimeout(timerRef.current);
+  }, []);
```
jjenzz, line 507: When we have a part that renders the `Presence` component, we move all the logic that isn't associated with visibility into a `Impl` part so that none of the logic is running in the background unnecessarily when `present` is false.

This part tripped me up because I thought you were expecting it to clear the timeout when it closes, which wouldn't be the case here as this part would always be mounted when it's closed. But reading more, I don't _think_ it matters in this particular case but it would be good to follow our convention just to prevent that mental mapping in future.

Usually our `<Presence />` parts look pretty bare:

image

jjenzz (changes requested):
Love the way you've created parts that recompose existing parts and build on top of them 😍 

Here's a first pass review but there a few bits I want to take a closer look at on Monday with you if that's okay so I can review them properly 🙏

packages/react/menu/src/Menu.tsx
```diff
@@ -296,14 +404,16 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
-                    // we override the default behaviour which automatically focuses the first item
-                    onEntryFocus={(event) => event.preventDefault()}
+                    onEntryFocus={(event) => {
+                      if (!focusFirstItem) event.preventDefault();
+                    }}
                   >
                     <PopperPrimitive.Content
                       role="menu"
                       {...focusScopeProps}
                       {...contentProps}
```
jjenzz, line 414: Oh yeah, while testing i noticed that we don't pass the `dir` prop to the DOM node here which seems like a bug. Would be good to pass that from context while you're here if you wouldn't mind otherwise `dir` on the document behaves differently to `dir` as a prop.

packages/react/menu/src/Menu.tsx
jjenzz, line 532: I was just trying to test this and couldn't seem to get it working: 

https://user-images.githubusercontent.com/175330/118336449-a0ada600-b509-11eb-978b-860ddfe17f89.mp4

I realised after it works if you never use the mouse though. Feels like a bug?

jjenzz:
Possibly related to the one i reported earlier where pressing down navigates the submenu. This one doesn't allow me to enter the submenu when pressing right arrow:

https://user-images.githubusercontent.com/175330/118336722-25002900-b50a-11eb-9cd3-4d10f2ef6b77.mp4

MacOS:

https://user-images.githubusercontent.com/175330/118336798-4e20b980-b50a-11eb-9194-45e8594e922c.mp4

jjenzz:
Looks like the submenu doesn't close when I move my mouse from its trigger onto a disabled item:

https://user-images.githubusercontent.com/175330/118338146-51697480-b50d-11eb-94ea-6f4c8e83e9d6.mp4

MacOS:

https://user-images.githubusercontent.com/175330/118338209-778f1480-b50d-11eb-986d-f6baf7c4f73f.mp4


I hope I'm not making you feel despondent by raising these btw. These things are super hard to get right 🙈

packages/react/menu/src/Menu.tsx
```diff
@@ -361,6 +471,89 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+          portalled
+          disableOutsidePointerEvents={false}
+          disableOutsideScroll={false}
+          trapFocus={false}
+          ref={composeRefs(forwardedRef, subMenuContext.contentRef)}
+          dir={direction}
+          onKeyDown={composeEventHandlers(contentProps.onKeyDown, (event) => {
+            const tabNext = event.key === 'Tab' && !event.shiftKey;
+            if (SUB_CLOSE_KEYS[direction].includes(event.key) || tabNext) {
+              // close a single level unless it's a forward tab, in which case close the entire stack
```
jjenzz, line 526: Curious where you got the idea for this behaviour from? I'm testing a nested menu on MacOS and it seems `shift+tab` and `tab` always closes the entire stack.

jjenzz (changes requested):
Few more bits below as I had some time to make up this week.

I also had an ah-ha moment this morning. I was thinking about your `root` prop idea and realised that we probably wouldn't need all of the extra exposed parts for submenus if we applied that same thinking and just have a `submenu` prop and utilised the `as` prop for the `Menu.Anchor`.

This lead me to realise that maybe all we really need is a `Menu.Sub` part.

```jsx
<Menu.Root open={openRoot} onOpenChange={setOpenRoot}>
	<Menu.Anchor />
	<Menu.Content>
		<Menu.Item>One</Menu.Item>
		<Menu.Item>Two</Menu.Item>
		<Menu.Item>Three</Menu.Item>
		<Menu.Sub open={openSub} onOpenChange={setOpenSub}>
			<Menu.Anchor as={Menu.Item}>Sub menu</Menu.Anchor>
			<Menu.Content>
				<Menu.Item>One</Menu.Item>
				<Menu.Item>Two</Menu.Item>
				<Menu.Item>Three</Menu.Item>
			</Menu.Content>
		</Menu.Sub>
	</Menu.Content>
</Menu.Root>
```

I had a go at hacking this together locally to verify my thinking first (didn't wanna waste more of your time) and it seems to work well. 

I didn't implement everything you have here ofc (and there are some edge-case bugs) so the concept isn't 100% proven, but so far I've only written about 140 lines of PoC code and this is how it's looking:

https://user-images.githubusercontent.com/175330/118371301-e9b23880-b5a3-11eb-9bc7-7499d27b0195.mp4

It's a bit flickery but you probably have a better idea why that is happening than me at this stage. Thoughts on exploring this concept further?

jjenzz:
@benoitgrelard While reviewing this and fiddling with the above, I noticed that we now have `onMountAutoFocus` prop from `FocusScope` and an `onEntryFocus` prop from `RovingFocusScope` that are both seemingly doing the same thing. However, under closer inspection it appears that `onMountAutoFocus` focuses the content area while `onEntryFocus` focuses the first item. Is that the correct distinction between the two?

jjenzz:
Spotted another potential bug because I noticed I have the same issue in my PoC above 😅  Pressing right arrow after select an item with mouse doesn't enter sub menu.

https://user-images.githubusercontent.com/175330/118372665-b7f0a000-b5aa-11eb-8739-abace518ce1c.mp4

Versus:

https://user-images.githubusercontent.com/175330/118372714-e0789a00-b5aa-11eb-85b4-ae6598516e96.mp4

packages/react/menu/src/Menu.tsx
```diff
@@ -361,6 +471,89 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+/* -------------------------------------------------------------------------------------------------
+ * MenuSubContent
+ * -----------------------------------------------------------------------------------------------*/
+
+const SUB_CONTENT_NAME = 'MenuSubContent';
+
+type MenuSubContentOwnProps = Polymorphic.Merge<
+  Omit<
+    Polymorphic.OwnProps<typeof MenuContentImpl>,
+    'portalled' | 'disableOutsidePointerEvents' | 'disableOutsideScroll' | 'side' | 'trapFocus'
```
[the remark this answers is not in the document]
jjenzz: Thought about this some more and in the [lil PoC](https://github.com/radix-ui/primitives/pull/627#pullrequestreview-660341222) I did below I wound up just doing this:

image


I pass that `isPointer` state down through context from the `Menu.Sub` component and the `MenuSubContent` prevents default if it's true.

packages/react/menu/src/Menu.tsx
```diff
@@ -36,29 +52,117 @@ const MENU_NAME = 'Menu';
...
+          /**
+           * We need to programatically focus the menu due to animation cancellation in `Presence` not triggering `FocusScope` lifecycle effects (by design)
+           * As a result, `FocusScope` won't re-focus the menu if a cancellation event occurs during an exit animation
+           */
+          contentRef.current?.focus();
+        },
+        [setOpen]
+      )}
+    >
+      <Menu open={open} onOpenChange={setOpen} dir={dir || context.dir}>
```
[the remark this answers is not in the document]
benoitgrelard: I would think the same, perhaps we're missing something?

packages/react/menu/src/Menu.tsx
```diff
@@ -361,6 +471,89 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+
+  React.useEffect(() => {
+    return () => window.clearTimeout(timerRef.current);
+  }, []);
+
+  return (
+    <Presence present={forceMount || context.open}>
+      <CollectionSlot>
+        <MenuContentImpl
+          id={subMenuContext.contentId}
```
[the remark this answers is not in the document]
benoitgrelard: Our stance is we let them override anything and we don't make any assumptions. so this is correct.

packages/react/menu/src/Menu.tsx
```diff
@@ -236,10 +346,8 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
     <PortalWrapper>
```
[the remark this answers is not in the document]
benoitgrelard: I get the feeling that once the modality stuff is in, sub-menus will become non-modal, so this wouldn't be true anymore. And it kinda makes sense, if you're able to focus or hover the parent menu, then screen readers should also be able to "see" them. So I would say for now, no need to do anything specific for it in this branch.

benoitgrelard:
> ~@benoitgrelard While reviewing this and fiddling with the above, I noticed that since the `RovingFocusGroup` work we did, we now have an `onMountAutoFocus` prop from `FocusScope` and an `onEntryFocus` prop from `RovingFocusGroup` that are both seemingly doing the same thing but they're handled differently by `Menu`.~
> 
> ~Currently we do `event.preventDefault()` in `onEntryFocus` and we don't expose it to consumers, but we expose `onMountAutoFocus` (as `onOpenAutoFocus`) without calling `event.preventDefault()` internally.~
> 
> ~With this setup, it seems that `onMountAutoFocus` only focuses the content area atm (I'm guessing because `RovingFocusScope` hasn't made anything focusable yet at the time it runs) but is it supposed to focus the first item?~
> 
> ~I'm worried there is a race condition or a happy accident here.~
> 
> I think I've realised the mistake I've made here. I remember struggling with this when we were coming up with the prop name 🙈 . They're different...
> 

haha, yep I remember we riffed on a few names for that one but we couldn't find anything better and thought let's just keep that cause it's not exposed to public components.

> * `onMountAutoFocus` fires when the component mounts. It will attempt to focus the first item and fallback to focusing the wrapper if it can't find a focusable item
> * `onEntryFocus` fires when the wrapper is focused. It will attempt to focus the first item.
> 

Yes they are different, and the description above is correct.

> If the above is correct then `onEntryFocus` will always happen after `onMountAutoFocus` but we've always prevented the first item from focusing. I'm now wondering a few things:
> 
> * Should we remove `onOpenAutoFocus` from `MenuContent` API if we're always doing `event.preventDefault()` in `onEntryFocus`?

What's the thinking here? not sure I'm following

> * Should we rename `onEntryFocus` to `onFocusAutoFocus` to clear up the difference between the two? 🤔

Not sure that improves much…

> * Should we `preventDefault` in `onMountAutoFocus` also and explicitly focus the wrapper on mount? Currently it appears it is only focusing the wrapper because it can't find a focusable item (yet) which is super implicit and feels like a bug waiting to happen.

Yeah I have been wondering this a few times. Although I do like that it lets us check that nothing should in fact be focusable yet…

packages/react/menu/src/Menu.stories.tsx
```diff
@@ -311,7 +473,15 @@ const itemStyles: any = {
...
   color: '$black',
+  width: '100%',
   borderRadius: 3,
+
+  // Reset button styles on sub menu triggers
+  border: 0,
+  fontFamily: 'inherit',
+  backgroundColor: 'transparent',
+  boxSizing: 'border-box',
+  fontSize: 'inherit',
```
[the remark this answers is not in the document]
andy-hook: Yes exactly, It was a quick way to test across components but the tidy up slipped me. Will update!

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
andy-hook: I agree

packages/react/menu/src/Menu.tsx
benoitgrelard: Yeah it looks like the tab behaviour is more linked to `MenuBar`.

andy-hook: I was a little unsure after looking at the spec plus some different implementations. In the context of `DropdownMenu` what would the expected behaviour be when we're inside a submenu? should tabbing close all subs and refocus the root menu? or close the entire menu stack (including root) and focus the button?

The focus trapping in `DropdownMenu` threw me because it doesn't let you tab out.

benoitgrelard: I feel like generally these widgets are supposed to function as one tab stop. So once your widget is "focused" you use other keyboard navigation means (arrow keys, space, enter, escape, etc). So I feel like pressing tab or shift tab just shouldn't do anything.

It seems to be what MacOS does:

https://user-images.githubusercontent.com/1539897/118475196-d918d400-b703-11eb-9654-74d80c07cf94.mov

> Note: I'm pressing tab and shift-tab at the end.

jjenzz:
> What's the thinking here? not sure I'm following

`onOpenAutoFocus` will never focus the first item because `RovingFocusGroup` won't have made the items focusable yet at the time it fires and we also do `onEntryFocus={event => event.preventDefault()}` which means we're always preventing the first item from focusing anyway.

So, why does the consumer need an `onOpenAutoFocus` event if the first item is never auto focused? If it's because we want them to be able to prevent the fact that `onMountAutoFocus` falls back to focusing the content, I'm thinking we should definitely focus the content more explicitly internally instead of relying on the fact that the content is only focused because `RovingFocusScope` hasn't initialised yet.

> > * Should we rename `onEntryFocus` to `onFocusAutoFocus` to clear up the difference between the two? 🤔
> 
> Not sure that improves much…

I think it does... because we're saying `onMountAutoFocus` means it will try to auto focus the first item when the component mounts, whereas `onFocusAutoFocus` means it will try to auto focus the first item when the component is focused. 

I remember we debated the naming and agreed to leave it because it's internal but this really tripped me up when trying to work with the two seemingly similar events. It wasn't obvious that the difference is one applies on mount and one applies on focus, so I'm thinking we could just put that in the name to clear that up?

benoitgrelard:
> onOpenAutoFocus will never focus the first item because RovingFocusGroup won't have made the items focusable yet at the time it fires and we do onEntryFocus={event => event.preventDefault()} on RovingFocusGroup anyway which means we're always preventing the first item from focusing.

> So, why does the consumer need an onOpenAutoFocus event if the first item is never auto focused? If it's because we want them to be able to prevent the fact that onMountAutoFocus falls back to focusing the content, I'm thinking we should focus the content more explicitly instead of relying on the fact that the content is only focused because RovingFocusScope hasn't initialised yet.

To me `onOpenAutoFocus` means it will autofocus the component, and focusing the container if nothing else is possible is part of that. So I'm still comfortable with relying on that. That said I can also see why you'd want to be more explicit, but it's just more code to effectively get the same result so I'm not really sure about it.

> I think it does... because we're saying onMountAutoFocus means it will try to auto focus the first item when the component mounts, whereas onFocusAutoFocus means it will try to auto focus the first item when the component is focused.

> I remember we debated the naming and agreed to leave it because it's internal but this really tripped me up when trying to work with the two seemingly similar events. It wasn't obvious that the difference is that one applies on mount and one applies on focus. Took me an embarrassing amount of time to work that out so I'm thinking we could just put that in the name to clear that up?

What I meant is I don't think the name clarifies a huge deal, but I'm not fussed, happy to rename if it makes things easier for you! I think with these things honestly, if you come back to it in 2 months, you'll still be confused by the 2 even with the new names. It's one of these things where perhaps leaving a comment might help our future selves more.

jjenzz:
> focusing the container if nothing else is possible is part of that

This is the part that feels like a happy accident with `Menu` in particular tho. 

If `RovingFocusGroup` finishes its setup before `FocusScope` then `onOpenAutoFocus` would focus the first item and not the container. The fact that it is focusing the container atm is purely because `RovingFocusGroup` hasn't added `tabIndex={0}` **yet** when `FocusScope` initialises 😬

In this video below, I've just changed `tabIndex={isCurrentTabStop ? 0 : -1}` in `RovingFocusGroup` to `tabIndex={0}` to simulate that it initialises first and this is what happens:

CleanShot 2021-05-21 at 13 33 54

The fact that the `Menu` doesn't do this currently is because it is relying on two unrelated components initialising at the right time. It's a race condition waiting to happen. Changes to `RovingFocusGroup` could change the behaviour of the `onOpenAutoFocus` handler which is breaking encapsulation IMO.

Happy to run through this in more detail on a call if I'm not making any sense here.

> happy to rename if it makes things easier for you

Cool, gonna go with that for now because yeah, I think it would help, it certainly won't make it worse anyway.

benoitgrelard:
No it does make sense, I'm fine with being more explicit as focusing the `Content` really is what we would always want (at least for the first level menu).

packages/react/menu/src/Menu.tsx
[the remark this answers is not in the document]
jjenzz: > if you're able to focus or hover the parent menu, then screen readers should also be able to "see" them

I'm saying that currently they can't "see" them. If a sub menu opens, the parent will become hidden even though it's still open because this effect will run in the submenu causing everything outside of it to be hidden.

I agree non-modal stuff might solve some of this but that's not ready yet and we need to release with an accessible thing so should we need to prevent this block from running in sub menus?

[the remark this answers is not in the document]
benoitgrelard: Yes exactly, that's why I said sub-menus would be non-modal.
Non-modal means that it everything is still visible to SR users.
So yeah in the meantime you just have to guard the effect in sub-menus for now 👍

packages/react/menu/src/Menu.tsx
andy-hook: Thanks for this btw, it's a good habit and useful to remember!

packages/react/menu/src/Menu.tsx
```diff
@@ -73,7 +227,7 @@ const [MenuContentProvider, useMenuContentContext] = createContext<MenuContentCo
 );
 
 type MenuContentOwnProps = Polymorphic.Merge<
-  Polymorphic.OwnProps<typeof MenuContentImpl>,
+  Polymorphic.OwnProps<typeof MenuRootContent>,
```
jjenzz, line 230: We purposefully don't union `MenuRootContent` and `MenuSubContent` props here because they have conflicting prop types. We discussed previously that we would just allow the `MenuSubContent` to accept props that it would ignore.

packages/react/menu/src/Menu.tsx
```diff
@@ -55,6 +98,106 @@ const Menu: React.FC<MenuOwnProps> = (props) => {
...
+          if (!props.disabled) context.onOpenChange(true);
+        })}
+        onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
+          if (!props.disabled && SUB_OPEN_KEYS[context.dir].includes(event.key)) {
+            context.onKeyOpen();
+          }
+        })}
+      />
+    </MenuAnchor>
+  ) : null;
```
benoitgrelard, line 196: Are we doing this in case someone was to add a sub trigger where it's not supposed to be?
I wonder if we need to bother guarding against that considering it's a private component for us to use in `DropdownMenu` and `ContextMenu`.

packages/react/menu/src/Menu.tsx
```diff
@@ -55,6 +98,106 @@ const Menu: React.FC<MenuOwnProps> = (props) => {
...
+    <PopperPrimitive.Root>
+      <MenuProvider
+        isSubmenu={true}
+        dir={context.dir}
+        open={open}
+        onOpenChange={handleOpenChange}
+        trigger={trigger}
+        onTriggerChange={setTrigger}
+        content={content}
+        contentId={contentId}
```
benoitgrelard, line 140: nit: can inline that one.

packages/react/menu/src/Menu.tsx
benoitgrelard: Should we add this as a comment in the code for our future selves?

packages/react/menu/src/Menu.tsx
```diff
@@ -94,65 +237,179 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
-  Polymorphic.OwnProps<typeof PopperPrimitive.Content>,
-  Omit<DismissableLayerOwnProps, 'onDismiss'> & {
-    /**
-     * Whether focus should be trapped within the `MenuContent`
-     * (default: false)
-     */
-    trapFocus?: FocusScopeOwnProps['trapped'];
+type MenuRootContentOwnProps = Omit<
+  Polymorphic.OwnProps<typeof MenuContentImpl>,
+  keyof MenuContentImplPrivateProps
```
benoitgrelard, line 254: I like this pattern, makes me think we can probably do that in other places 👍

packages/react/menu/src/Menu.tsx
```diff
@@ -94,65 +237,179 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
+    <MenuContentImpl
+      {...props}
+      ref={composedRefs}
+      onDismiss={() => context.onOpenChange(false)}
+      onEntryFocus={(event) => event.preventDefault()}
+      onOpenAutoFocus={composeEventHandlers(props.onOpenAutoFocus, (event) => {
+        // explicitly focus the content area only when opening
+        event.preventDefault();
+        ref.current?.focus();
+      })}
```
benoitgrelard, line 282: 👍 yeah it's nice to see these 2 grouped now, reads really well.

packages/react/menu/src/Menu.tsx
```diff
@@ -94,65 +237,179 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
+  MenuSubContentOwnProps
+>;
+
+const MenuSubContent = React.forwardRef((props, forwardedRef) => {
+  const context = useMenuContext(CONTENT_NAME);
+  const ref = React.useRef<MenuContentElement>(null);
+  const composedRefs = useComposedRefs(forwardedRef, ref);
+  return context.isSubmenu ? (
+    <MenuContentImpl
+      align="start"
```
benoitgrelard, line 312: This threw me off a bit. If it's just to set a default value, we would normally do:
`const { align = 'start', ...subContentProps } = props` no?

packages/react/menu/src/Menu.tsx
```diff
@@ -94,65 +237,179 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
+      side={context.dir === 'rtl' ? 'left' : 'right'}
+      portalled
+      disableOutsidePointerEvents={false}
+      disableOutsideScroll={false}
+      trapFocus={false}
+      onEntryFocus={context.onEntryFocus}
+      onOpenAutoFocus={(event) => event.preventDefault()}
+      // The menu might close because of focusing another menu item in the parent menu. We
+      // don't want it to refocus the trigger in that case so we handle trigger focus ourselves.
+      onCloseAutoFocus={(event) => event.preventDefault()}
```
benoitgrelard, line 325: Do we still want to compose these as I think we used to no?

packages/react/menu/src/Menu.tsx
```diff
@@ -94,65 +237,179 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
+      onFocusOutside={composeEventHandlers(props.onFocusOutside, (event) => {
+        // We prevent closing when the trigger is focused to avoid triggering a re-open animation
+        // on pointer interaction.
+        if (event.target !== context.trigger) context.onOpenChange(false);
+      })}
+      onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => {
+        // We need to explicitly close when escape key focuses trigger because we prevented this
+        // in on focus outside for pointers.
```
benoitgrelard, line 333: I find the comment reads weird here.

```suggestion
        // We need to explicitly focus the trigger when pressing escape because we prevented this
        // in `onCloseAutoFocus` so it doesn't happen with pointers.
```

Does this make sense?

packages/react/menu/src/Menu.tsx
```diff
@@ -94,65 +237,179 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
+        context.onOpenChange(false);
+        context.trigger?.focus();
+      })}
+      onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
+        const element = event.target as HTMLElement;
```
benoitgrelard, line 338: Would it help to name this more specifically? `subContentElement` or `subContent` maybe?

packages/react/menu/src/Menu.tsx
```diff
@@ -94,65 +237,179 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
+        // Submenu key events bubble through portals. We only care about keys in this menu.
+        const isKeyDownInside = event.currentTarget.contains(element);
+        const isCloseKey = SUB_CLOSE_KEYS[context.dir].includes(event.key);
+        if (isKeyDownInside && isCloseKey) {
+          // We need to explicitly close when keyboard focuses trigger because we prevented this
+          // in on focus outside for pointers.
```
benoitgrelard, line 344: This is similar to the other one, I think I see why you wrote it like that now.
But I still find it odd to read.

Really, it's 2 things:
- we have to close manually because we're not wiring onto `onDismiss` in `onPointerDownOutside` and `onEscapeKeydown` (but crucially not in `onFocusOutside`)
- we need to focus manually because we have prevented it in `onCloseAutoFocus`

packages/react/menu/src/Menu.tsx
```diff
@@ -94,65 +237,179 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
+type MenuSubContentPrimitive = Polymorphic.ForwardRefComponent<
+  Polymorphic.IntrinsicElement<typeof MenuContentImpl>,
+  MenuSubContentOwnProps
+>;
+
+const MenuSubContent = React.forwardRef((props, forwardedRef) => {
+  const context = useMenuContext(CONTENT_NAME);
+  const ref = React.useRef<MenuContentElement>(null);
+  const composedRefs = useComposedRefs(forwardedRef, ref);
+  return context.isSubmenu ? (
```
benoitgrelard, line 310: Again, do we need to guard here?

packages/react/use-direction/src/useDirection.tsx
```diff
@@ -8,8 +8,11 @@ export function useDirection(element: HTMLElement | null, directionProp?: Direct
   const rAFRef = React.useRef<number>(0);
 
   React.useEffect(() => {
-    if (element && directionProp === undefined) {
-      const computedStyle = getComputedStyle(element);
+    // We check inherited direction of parent instead of `element` so that computed styles are
+    // not overridden by dir attribute on element if inherited direction changes. The `dir`
+    // attribute should always sync with direction prop OR its inherited direction.
```
benoitgrelard, line 13: Does this work better?

```suggestion
    // We check inherited direction of the parent instead of the `element` itself.
    // This is because we internally set the computed `dir` on the element and so we wouldn't be
    // able to react to changes to inherit direction. The `dir` attribute we set should always sync
    // with the direction prop OR its inherited direction.
```

benoitgrelard (commented):
This is spectacular stuff guys!
So happy with it so far ♥️ 

I've left a few minor comments.

packages/react/menu/src/Menu.tsx
jjenzz: We're doing it because we need to access properties on the context that only exist in `MenuSubContextValue`. So this helps TS narrow the union and an added bonus is that it prevents it from being rendered where it's not supposed to.

packages/react/menu/src/Menu.tsx
jjenzz: Yeah, I was wondering that too...👀 👍

packages/react/menu/src/Menu.tsx
jjenzz: oh yeah..... hmmm... do we think consumers should be able to override this one?

packages/react/menu/src/Menu.tsx
jjenzz: We never auto focus on close so why would they need to bind to it? We didn't want them to be able to prevent it because it completely breaks the menus if it just keeps trying to focus the sub menu trigger every time it closes.

packages/react/menu/src/Menu.tsx
jjenzz: `element` isn't the sub content, it's `event.target` which could be an item. We're checking here if the thing that was clicked is inside `subContent`.

packages/react/menu/src/Menu.tsx
benoitgrelard: ha ok cool

packages/react/menu/src/Menu.tsx
benoitgrelard: it's not the conventional positioning but I guess we could, although we could just simplify things by removing the prop from sub-menus until people ask for something custom, I'd be happy with that.
Especially considering that it might go away from the the root dropdown menu too once we do a more custom positioning for menus.

packages/react/menu/src/Menu.tsx
jjenzz: I've simplified this one and the `onEscapeKeyDown` one to read just:

```
// We focus manually because we prevented it in `onCloseAutoFocus`.
```

We already explain why we prevent `onCloseAutoFocus` alongside that event so seems unnecessary to explain it again each time.

packages/react/menu/src/Menu.tsx
benoitgrelard: Ok makes sense.

packages/react/menu/src/Menu.tsx
benoitgrelard: Ah yeah of course haha, that's the whole point of the check 🤦‍♂️

benoitgrelard: That said it might be better to then call it `target`, then I wouldn't have been confused I feel 😀 
I'd just inline it as `const isKeyDownInside = event.currentTarget.contains(event.target);` to be honest.

packages/react/menu/src/Menu.tsx
jjenzz: Ah, it is actually already removed from the types haha... so i'll move this below the prop spread

benoitgrelard (approved):

jjenzz (approved):

PR #629, opened 2021-04-30
[ContextMenu][DropdownMenu] Integrate new submenu parts

andy-hook:
Builds on https://github.com/radix-ui/primitives/pull/627

[See discussion](https://github.com/radix-ui/primitives/issues/383)

### Description

This PR uses the new submenu primitives from `Menu` to provide submenu support in `ContextMenu` and `DropdownMenu`

#### Approach

Menus can be nested to create sub menus.

#### Basic example

```jsx
<ContextMenu>
  <ContextMenuTrigger/>
  <ContextMenuContent>
    <ContextMenuItem>Item 1</ContextMenuItem>
    <ContextMenuItem>Item 2</ContextMenuItem>
    
    <ContextMenu>
      <ContextMenuTrigger>
         Open Sub Menu
      </ContextMenuTrigger>

      <ContextMenuContent>
        <ContextMenuItem>Sub Item 1</ContextMenuItem>
        <ContextMenuItem>Sub Item 2</ContextMenuItem>
      </ContextMenuContent>
    </ContextMenu>
  </ContextMenuContent>
</ContextMenu>
```

https://user-images.githubusercontent.com/11708259/116690178-ed916880-a9b0-11eb-969d-644e4237beb9.mov

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -28,7 +28,7 @@ const [ContextMenuProvider, useContextMenuContext] = createContext<ContextMenuCo
   CONTEXT_MENU_NAME
 );
 
-type ContextMenuOwnProps = ContextMenuRootOwnProps & ContextMenuSubOwnProps;
+type ContextMenuOwnProps = React.ComponentProps<typeof ContextMenuImpl>;
 
 const ContextMenu: React.FC<ContextMenuOwnProps> = (props) => {
   const parentSubmenuContext = React.useContext(SubmenuContext);
@@ -42,11 +42,11 @@ const ContextMenu: React.FC<ContextMenuOwnProps> = (props) => {
   );
 };
 
-const ContextMenuImpl: React.FC<ContextMenuOwnProps> = (props) => {
-  const isSubmenu = React.useContext(SubmenuContext);
-  const ContextMenu = isSubmenu ? ContextMenuSub : ContextMenuRoot;
+type ContextMenuImplOwnProps = ContextMenuRootOwnProps & ContextMenuSubOwnProps;
 
-  return <ContextMenu {...props} />;
+const ContextMenuImpl: React.FC<ContextMenuImplOwnProps> = (props) => {
+  const isSubmenu = React.useContext(SubmenuContext);
+  return isSubmenu ? <ContextMenuSub {...props} /> : <ContextMenuRoot {...props} />;
 };
 
 /* -----------------------------------------------------------------------------------------------*/
@@ -87,14 +87,18 @@ type ContextMenuSubOwnProps = {
 };
 
 const ContextMenuSub: React.FC<ContextMenuSubOwnProps> = (props) => {
-  const { open: openProp, defaultOpen, onOpenChange } = props;
+  const { children, open: openProp, defaultOpen, onOpenChange } = props;
   const [open = false, setOpen] = useControllableState({
     prop: openProp,
     defaultProp: defaultOpen,
     onChange: onOpenChange,
   });
 
-  return <MenuPrimitive.Sub open={open} onOpenChange={setOpen} />;
+  return (
+    <MenuPrimitive.Sub open={open} onOpenChange={setOpen}>
+      {children}
+    </MenuPrimitive.Sub>
+  );
 };
 
 ContextMenu.displayName = CONTEXT_MENU_NAME;
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -200,27 +200,33 @@ const ContextMenuContent = React.forwardRef((props, forwardedRef) => {
   const isSubmenu = React.useContext(SubmenuContext);
   const context = useContextMenuContext(CONTENT_NAME);
 
+  const commonProps = {
+    ...contentProps,
+    sideOffset: offset,
+    style: {
+      ...props.style,
+      // re-namespace exposed content custom property
+      ['--radix-context-menu-content-transform-origin' as any]: 'var(--radix-popper-transform-origin)',
+    },
+  };
+
   return (
     <InsideContentContext.Provider value={true}>
-      <MenuPrimitive.Content
-        {...contentProps}
-        {...(!isSubmenu && {
-          disableOutsidePointerEvents: context.open ? disableOutsidePointerEvents : false,
-          trapFocus: true,
-          disableOutsideScroll: true,
-          portalled: true,
-          side: 'bottom' as const,
-          align: 'start' as const,
-          alignOffset: 2,
-        })}
-        sideOffset={offset}
-        ref={forwardedRef}
-        style={{
-          ...props.style,
-          // re-namespace exposed content custom property
-          ['--radix-context-menu-content-transform-origin' as any]: 'var(--radix-popper-transform-origin)',
-        }}
-      />
+      {isSubmenu ? (
+        <MenuPrimitive.Content {...commonProps} ref={forwardedRef} />
+      ) : (
+        <MenuPrimitive.Content
+          {...commonProps}
+          ref={forwardedRef}
+          disableOutsidePointerEvents={context.open ? disableOutsidePointerEvents : false}
+          trapFocus
+          disableOutsideScroll
+          portalled
+          side="bottom"
+          align="start"
+          alignOffset={2}
+        />
+      )}
     </InsideContentContext.Provider>
   );
 }) as ContextMenuContentPrimitive;
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -122,14 +122,18 @@ const ContextMenuTrigger = React.forwardRef((props, forwardedRef) => {
   const { as, ...triggerProps } = props;
   const isSubmenu = React.useContext(SubmenuContext);
 
-  return isSubmenu ? (
-    <MenuPrimitive.SubTrigger
-      {...triggerProps}
-      as={as as Polymorphic.IntrinsicElement<typeof ContextMenuRootTrigger>}
-      ref={forwardedRef}
-    />
-  ) : (
-    <ContextMenuRootTrigger {...triggerProps} as={as} ref={forwardedRef} />
+  return (
+    <InsideContentContext.Provider value={false}>
+      {isSubmenu ? (
+        <MenuPrimitive.SubTrigger
+          {...triggerProps}
+          as={as as Polymorphic.IntrinsicElement<typeof ContextMenuRootTrigger>}
+          ref={forwardedRef}
+        />
+      ) : (
+        <ContextMenuRootTrigger {...triggerProps} as={as} ref={forwardedRef} />
+      )}
+    </InsideContentContext.Provider>
   );
 }) as ContextMenuTriggerPrimitive;
 
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -23,7 +23,7 @@ type ContextMenuContextValue = {
 };
 
 const SubmenuContext = React.createContext<boolean | undefined>(undefined);
-const InsideContentContext = React.createContext<boolean>(false);
+const InsideContentContext = React.createContext(false);
 const [ContextMenuProvider, useContextMenuContext] = createContext<ContextMenuContextValue>(
   CONTEXT_MENU_NAME
 );
@@ -84,18 +84,17 @@ type ContextMenuSubOwnProps = {
   open?: boolean;
   onOpenChange?(open: boolean): void;
   defaultOpen?: boolean;
-  dir?: Direction;
 };
 
 const ContextMenuSub: React.FC<ContextMenuSubOwnProps> = (props) => {
-  const { open: openProp, defaultOpen, onOpenChange, ...menuSubProps } = props;
+  const { open: openProp, defaultOpen, onOpenChange } = props;
   const [open = false, setOpen] = useControllableState({
     prop: openProp,
     defaultProp: defaultOpen,
     onChange: onOpenChange,
   });
 
-  return <MenuPrimitive.Sub {...menuSubProps} open={open} onOpenChange={setOpen} />;
+  return <MenuPrimitive.Sub open={open} onOpenChange={setOpen} />;
 };
 
 ContextMenu.displayName = CONTEXT_MENU_NAME;
@@ -105,7 +104,6 @@ ContextMenu.displayName = CONTEXT_MENU_NAME;
  * -----------------------------------------------------------------------------------------------*/
 
 const TRIGGER_NAME = 'ContextMenuTrigger';
-const ROOT_TRIGGER_DEFAULT_TAG = 'span';
 
 type ContextMenuTriggerOwnProps = Polymorphic.Merge<
   Polymorphic.OwnProps<typeof ContextMenuRootTrigger>,
@@ -135,14 +133,16 @@ ContextMenuTrigger.displayName = TRIGGER_NAME;
 
 /* ---------------------------------------------------------------------------------------------- */
 
+const TRIGGER_DEFAULT_TAG = 'span';
+
 type ContextMenuRootTriggerOwnProps = Polymorphic.OwnProps<typeof Primitive>;
 type ContextMenuRootTriggerPrimitive = Polymorphic.ForwardRefComponent<
-  typeof ROOT_TRIGGER_DEFAULT_TAG,
+  typeof TRIGGER_DEFAULT_TAG,
   ContextMenuRootTriggerOwnProps
 >;
 
 const ContextMenuRootTrigger = React.forwardRef((props, forwardedRef) => {
-  const { as = ROOT_TRIGGER_DEFAULT_TAG, ...triggerProps } = props;
+  const { as = TRIGGER_DEFAULT_TAG, ...triggerProps } = props;
   const context = useContextMenuContext(TRIGGER_NAME);
   const pointRef = React.useRef<Point>({ x: 0, y: 0 });
   const virtualRef = React.useRef({
@@ -175,7 +175,13 @@ const CONTENT_NAME = 'ContextMenuContent';
 type ContextMenuContentOwnProps = Polymorphic.Merge<
   Omit<
     Polymorphic.OwnProps<typeof MenuPrimitive.Content>,
-    'trapFocus' | 'disableOutsideScroll' | 'portalled' | 'onOpenAutoFocus' | 'side' | 'sideOffset'
+    | 'trapFocus'
+    | 'disableOutsideScroll'
+    | 'portalled'
+    | 'onOpenAutoFocus'
+    | 'side'
+    | 'sideOffset'
+    | 'align'
   >,
   { offset?: number }
 >;
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -1,12 +1,14 @@
 import * as React from 'react';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { createContext } from '@radix-ui/react-context';
+import { useControllableState } from '@radix-ui/react-use-controllable-state';
 import { Primitive, extendPrimitive } from '@radix-ui/react-primitive';
 import * as MenuPrimitive from '@radix-ui/react-menu';
 import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
 
 import type * as Polymorphic from '@radix-ui/react-polymorphic';
 
+type Direction = 'ltr' | 'rtl';
 type Point = { x: number; y: number };
 
 /* -------------------------------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -18,17 +25,53 @@ const CONTEXT_MENU_NAME = 'ContextMenu';
...
+  return isSubmenu ? <MenuPrimitive.Sub {...props} /> : <ContextMenuRoot {...props} />;
+};
+
+const ContextMenuRoot: React.FC<ContextMenuCommonOwnProps> = (props) => {
   const { children, onOpenChange } = props;
   const [open, setOpen] = React.useState(false);
-  const handleOpenChangeProp = useCallbackRef(onOpenChange);
+  const [rootContentElement, setRootContentElement] = React.useState<HTMLDivElement | null>(null);
+  const directionContext = React.useContext(DirectionContext);
+  const direction = useDirection(rootContentElement, directionContext);
```
andy-hook, line 72: Performing the style query for `direction` at the root level only and threading it down via context. I couldn't find a good enough reason to take a perf hit and do it JIT for every sub menu.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -18,17 +25,53 @@ const CONTEXT_MENU_NAME = 'ContextMenu';
...
 const [ContextMenuProvider, useContextMenuContext] = createContext<ContextMenuContextValue>(
   CONTEXT_MENU_NAME
 );
 
-const ContextMenu: React.FC<{ onOpenChange?(open: boolean): void }> = (props) => {
+type ContextMenuCommonOwnProps = {
+  open?: boolean;
+  defaultOpen?: boolean;
+  onOpenChange?(open: boolean): void;
+};
```
andy-hook, line 39: As we discussed – these props won't accurately reflect the interface as it varies during runtime depending on the nested level of the menu.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -53,16 +100,43 @@ ContextMenu.displayName = CONTEXT_MENU_NAME;
...
 
 const ContextMenuTrigger = React.forwardRef((props, forwardedRef) => {
-  const { as = TRIGGER_DEFAULT_TAG, ...triggerProps } = props;
+  const { as, ...triggerProps } = props;
+  const isSubmenu = React.useContext(NestedMenuContext);
+
+  return isSubmenu ? (
+    <MenuPrimitive.SubTrigger
+      {...triggerProps}
+      as={as as Polymorphic.IntrinsicElement<typeof ContextMenuAreaTrigger>}
```
andy-hook, line 120: Perhaps I'm missing a trick but I'm having to cast because of the merge between the differing IntrinsicElements (`span`, `button`) is there a better way?

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -38,31 +70,79 @@ const ContextMenu: React.FC<{ onOpenChange?(open: boolean): void }> = (props) =>
...
 
 const ContextMenuTrigger = React.forwardRef((props, forwardedRef) => {
-  const { as = TRIGGER_DEFAULT_TAG, ...triggerProps } = props;
+  const { as, ...triggerProps } = props;
+  const isSubmenu = React.useContext(SubmenuContext);
+
+  return isSubmenu ? (
+    <MenuPrimitive.SubTrigger
+      {...triggerProps}
+      as={as as Polymorphic.IntrinsicElement<typeof ContextMenuRootTrigger>}
```
andy-hook, line 126: Not sure on the best way to merge these two `IntrinsicElement` types so decided cast here. Is there a better approach?

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -23,27 +25,43 @@ type DropdownMenuContextValue = {
...
 };
 
 const DropdownMenu: React.FC<DropdownMenuOwnProps> = (props) => {
-  const { children, open: openProp, defaultOpen, onOpenChange } = props;
+  const parentSubmenuContext = React.useContext(SubmenuContext);
+  const isInsideContent = React.useContext(InsideContentContext);
+  const isRootMenu = parentSubmenuContext === undefined;
+
+  return (
+    <SubmenuContext.Provider value={!isRootMenu && isInsideContent}>
```
andy-hook, line 47: I added the `isInsideContent` requirement to `DropdownMenu` as well for completeness. While I don't expect anyone needing to nest these within triggers, it at least doesn't break the code if accidentally doing so.

andy-hook:
@jjenzz @benoitgrelard – This one is also ready. Might be easier to review after we've worked through https://github.com/radix-ui/primitives/pull/627 though!

jjenzz (changes requested):
Initial bits. Love how simple these changes were! Although, I'm a bit worried about the TS issues with the trigger parts, need to think about that some 😬

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -20,12 +22,42 @@ type ContextMenuContextValue = {
...
+      <ContextMenuImpl {...props} />
+    </SubmenuContext.Provider>
+  );
+};
+
+const ContextMenuImpl: React.FC<ContextMenuOwnProps> = (props) => {
+  const isSubmenu = React.useContext(SubmenuContext);
+  const ContextMenu = isSubmenu ? ContextMenuSub : ContextMenuRoot;
+
+  return <ContextMenu {...props} />;
```
benoitgrelard, line 49: It seems we tend to prefer to do the ternary inline to see both components being used.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -97,14 +175,7 @@ const CONTENT_NAME = 'ContextMenuContent';
...
   Omit<
     Polymorphic.OwnProps<typeof MenuPrimitive.Content>,
-    | 'trapFocus'
-    | 'disableOutsideScroll'
-    | 'portalled'
-    | 'onOpenAutoFocus'
-    | 'side'
-    | 'sideOffset'
-    | 'align'
-    | 'alignOffset'
```
benoitgrelard, line 107: we're not removing these 2 anymore?

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -116,25 +187,31 @@ type ContextMenuContentPrimitive = Polymorphic.ForwardRefComponent<
...
+        {...contentProps}
+        {...(!isSubmenu && {
+          disableOutsidePointerEvents: context.open ? disableOutsidePointerEvents : false,
+          trapFocus: true,
+          disableOutsideScroll: true,
+          portalled: true,
+          side: 'bottom' as const,
+          align: 'start' as const,
+          alignOffset: 2,
+        })}
```
benoitgrelard, line 205: I don't love the inline conditional spread like that, I find it a bit hard to read or even see that this is about the submenu.
I wonder if we'd be better off doing something like:

```tsx
const subMenuContentProps = isSubmenu ? { … } : {};
…
<MenuPrimitive.Content
  {...contentProps}
  {...submenuProps}
   …
/>
```

Not great either though. Anybody got other ideas to make this clearer?
Or we could duplicate the whole jsx block with a big ternary around `isSubmenu`.
I'm actually wondering if this means we did the inner swap too high up in `Menu`.

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: Yeah I agree on this one, I don't think we should add code that isn't necessary.
I can imagine we return to this in a few months and now we wonder what it's all about, whether it actually is important or not, etc and then are afraid to remove it.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -23,27 +25,43 @@ type DropdownMenuContextValue = {
...
+  const { children, open: openProp, defaultOpen, onOpenChange, dir } = props;
+  const isSubmenu = React.useContext(SubmenuContext);
   const triggerRef = React.useRef<HTMLButtonElement>(null);
   const [open = false, setOpen] = useControllableState({
     prop: openProp,
     defaultProp: defaultOpen,
     onChange: onOpenChange,
   });
-
+  const DropdownMenu = isSubmenu ? MenuPrimitive.Sub : MenuPrimitive.Root;
```
benoitgrelard, line 62: Same with this one we usually tend to do the ternary inline and when there's a common child like that, that's the one we pull out instead in a const.

benoitgrelard (changes requested):

packages/react/context-menu/src/ContextMenu.tsx
andy-hook: We need them exposed for the submenu

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -20,12 +22,42 @@ type ContextMenuContextValue = {
...
   CONTEXT_MENU_NAME
 );
 
-const ContextMenu: React.FC<{ onOpenChange?(open: boolean): void }> = (props) => {
-  const { children, onOpenChange } = props;
+type ContextMenuOwnProps = ContextMenuRootOwnProps & ContextMenuSubOwnProps;
+
+const ContextMenu: React.FC<ContextMenuOwnProps> = (props) => {
+  const parentSubmenuContext = React.useContext(SubmenuContext);
+  const isInsideContent = React.useContext(InsideContentContext);
```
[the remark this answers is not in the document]
andy-hook: Whoopsie, this is a great spot! not tested but I think you're right

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -38,31 +70,79 @@ const ContextMenu: React.FC<{ onOpenChange?(open: boolean): void }> = (props) =>
...
+
 ContextMenu.displayName = CONTEXT_MENU_NAME;
 
 /* -------------------------------------------------------------------------------------------------
  * ContextMenuTrigger
  * -----------------------------------------------------------------------------------------------*/
 
 const TRIGGER_NAME = 'ContextMenuTrigger';
-const TRIGGER_DEFAULT_TAG = 'span';
+const ROOT_TRIGGER_DEFAULT_TAG = 'span';
```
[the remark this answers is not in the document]
andy-hook: It only applies to the root trigger so figured it was more appropriate.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -38,31 +70,79 @@ const ContextMenu: React.FC<{ onOpenChange?(open: boolean): void }> = (props) =>
...
   );
 };
 
+/* -----------------------------------------------------------------------------------------------*/
+
+type ContextMenuSubOwnProps = {
+  open?: boolean;
+  onOpenChange?(open: boolean): void;
+  defaultOpen?: boolean;
+  dir?: Direction;
```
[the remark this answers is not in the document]
andy-hook: Yes, having it in this type is redundant 👍  though it will still appear in user land after the prop merge.

packages/react/context-menu/src/ContextMenu.tsx
benoitgrelard: I think we've taken out `align` in `Menu` already and set it to `start` always so only `alignOffset` would remain I guess then.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -38,31 +70,79 @@ const ContextMenu: React.FC<{ onOpenChange?(open: boolean): void }> = (props) =>
...
       </ContextMenuProvider>
     </MenuPrimitive.Root>
   );
 };
 
+/* -----------------------------------------------------------------------------------------------*/
+
+type ContextMenuSubOwnProps = {
+  open?: boolean;
+  onOpenChange?(open: boolean): void;
```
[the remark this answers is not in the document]
andy-hook: hmmm perhaps we shouldn't be spreading on to the component here? I worry about these chains of prop inheritance. If a required prop is added to `MenuPrimitive.Sub` for an internal reason, it would unintentionally make its way up to `ContextMenu`.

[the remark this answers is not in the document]
jjenzz: > hmmm perhaps we shouldn't be spreading on to the component here?

Ah yes, I think you're right here. It's just a provider so doesn't need to spread props.

> I worry about these chains of prop inheritance. If a required prop is added to `MenuPrimitive.Sub` for an internal reason, it would unintentionally make its way up to `ContextMenu`.

I think that would be intentional if we're inheriting. That's the expectation in my mind. I'm curious why you think that would be an issue? I find it's usually a lot less overhead to inherit and omit than to create manual duplications of a component's prop types at every level.

Interested in your reasoning on this one though. @benoitgrelard what are your thoughts here? Obviously one benefit of avoiding inheritance is TS perf 😅

packages/react/context-menu/src/ContextMenu.tsx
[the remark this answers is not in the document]
jjenzz: I'm not sure what we usually do here, sometimes I've left the name as it was and just used it in the other part, or I've renamed and moved it alongside the other part.

@benoitgrelard what should we do here? I'm thinking move it but I've never been sure myself on this one haha.

packages/react/context-menu/src/ContextMenu.tsx
jjenzz: Haha yeah we've done a bit of both on this one. For a bit of backstory...

I think we used to do it this way but we got tripped up a few times trying to read the code. In `Slider` for example we have `const SliderOrientation = condition ? SliderVertical : SliderHorizontal;` and I always read the render and see `<SliderOrientation />` and then try to find a `SliderOrientation` part that doesn't exist 😅 

We started to favour slight duplication to ease this confusion.

packages/react/context-menu/src/ContextMenu.tsx
jjenzz: I tripped up on this one too but at the time didn't have any better ideas. Just thinking now though, maybe we could do this (which I did for `ScrollArea`):

```jsx
const commonProps = { ...contentProps, sideOffset: offset, style: { ... } }

return isSubMenu ? (
  <MenuPrimitive.Content
    {...commonProps}
    ref={forwardedRef}
    disableOutsidePointerEvents={context.open ? disableOutsidePointerEvents : false}
    trapFocus={true}
    disableOutsideScroll={true}
    portalled={true}
    side="bottom"
    align="start"
    alignOffset={2}
  />
) : (
  <MenuPrimitive.Content {...commonProps} ref={forwardedRef} />
);
```

packages/react/context-menu/src/ContextMenu.tsx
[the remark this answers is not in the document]
andy-hook: I actually have an inconsistency here because I moved it in `DropdownMenu` but not here 😅 I'm happy to keep the previous naming if our standard is to use proximity to define the relationship 👍

packages/react/context-menu/src/ContextMenu.tsx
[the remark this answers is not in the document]
benoitgrelard: I think inheriting in general makes sense when we spread props onto another component, it makes sense to me to describe the props of the new based on that, and omit the ones that we hardcode.

packages/react/context-menu/src/ContextMenu.tsx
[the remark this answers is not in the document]
benoitgrelard: I'm not sure I follow what you mean by "move" it.

packages/react/context-menu/src/ContextMenu.tsx
benoitgrelard: Yep, I like that much better. Great idea! 👍

packages/react/context-menu/src/ContextMenu.tsx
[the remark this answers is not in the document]
andy-hook: Moved the `const` to sit next to the component using it.

packages/react/context-menu/src/ContextMenu.tsx
andy-hook: Love it 👍

packages/react/context-menu/src/ContextMenu.tsx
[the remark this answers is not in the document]
benoitgrelard: ohhh I see, yeah generally I like to have these just above their own part.

packages/react/context-menu/src/ContextMenu.tsx
andy-hook: We all discussed this and decided try and address this `IntrinsicElement ` conundrum via a separate `TriggerItem` component. Will raise as a separate improvement.

packages/react/context-menu/src/ContextMenu.tsx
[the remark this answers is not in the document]
andy-hook: Agree on the inheritance when spreading. My original bad feeling was that we would merge just because the types happen to line up. I don't think that was the nature of the feedback though 👍 

I'll remove the spread from here.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -20,12 +22,42 @@ type ContextMenuContextValue = {
...
+  const isRootMenu = parentSubmenuContext === undefined;
+
+  return (
+    <SubmenuContext.Provider value={!isRootMenu && isInsideContent}>
+      <ContextMenuImpl {...props} />
+    </SubmenuContext.Provider>
+  );
+};
+
+type ContextMenuImplOwnProps = ContextMenuRootOwnProps & ContextMenuSubOwnProps;
```
benoitgrelard, line 45: Why is it `&` rather than `|`?

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -23,18 +25,32 @@ type DropdownMenuContextValue = {
...
+const SubmenuContext = React.createContext<boolean | undefined>(undefined);
 const [DropdownMenuProvider, useDropdownMenuContext] = createContext<DropdownMenuContextValue>(
   DROPDOWN_MENU_NAME
 );
 
-type DropdownMenuOwnProps = {
+type DropdownMenuOwnProps = React.ComponentProps<typeof DropdownMenuImpl>;
+
+const DropdownMenu: React.FC<DropdownMenuOwnProps> = (props) => {
+  const parentSubmenuContext = React.useContext(SubmenuContext);
```
jjenzz, line 36: Bit of a nit I guess but the `parentSubmenuContext` name here keeps tripping me up... because in my mind a menu is a submenu if it is in another menu, not just when it is inside a submenu. Feels like `parentMenuContext` would be enough?

benoitgrelard: Oh yeah I agree.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -23,18 +25,32 @@ type DropdownMenuContextValue = {
...
+  return (
+    <SubmenuContext.Provider value={parentSubmenuContext === undefined ? false : true}>
```
jjenzz, line 38: Also.... 💭

```jsx
<SubmenuContext.Provider value={Boolean(parentMenuContext)}>
```

I remember we did this `undefined` check for `ContextMenu` because we needed to check if it was `undefined` to determine `isRoot` but we don't need that in `DropdownMenu` right?

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -20,12 +22,42 @@ type ContextMenuContextValue = {
...
 const [ContextMenuProvider, useContextMenuContext] = createContext<ContextMenuContextValue>(
   CONTEXT_MENU_NAME
 );
 
-const ContextMenu: React.FC<{ onOpenChange?(open: boolean): void }> = (props) => {
-  const { children, onOpenChange } = props;
+type ContextMenuOwnProps = React.ComponentProps<typeof ContextMenuImpl>;
+
+const ContextMenu: React.FC<ContextMenuOwnProps> = (props) => {
+  const parentSubmenuContext = React.useContext(SubmenuContext);
```
jjenzz, line 34: Same here with the `parentSubmenuContext` naming if you wouldn't mindddd 🙏

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -20,12 +22,42 @@ type ContextMenuContextValue = {
...
+  const isInsideContent = React.useContext(InsideContentContext);
+  const isRootMenu = parentSubmenuContext === undefined;
+
+  return (
+    <SubmenuContext.Provider value={!isRootMenu && isInsideContent}>
```
jjenzz, line 39: Actually wait... do we even need the `isRootMenu` stuff in `ContextMenu` either now? If we're always checking `isInsideContent` then it's inherently not a root menu if it's inside content right? I feel like I've forgotten something here 🙈😅

jjenzz, line 39: Actually wait.... 💭

Do we still need the `isRootMenu` check here if we're checking `isInsideContent`? Isn't it inherently not a root menu if we're inside content?

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: The same applies here, we need `undefined` to determine the root, else it'll always fall one way or the other.

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: yes indeedy

packages/react/context-menu/src/ContextMenu.tsx
andy-hook: GOOD POINT. No, I don't think we do!

benoitgrelard (approved):

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -23,18 +25,32 @@ type DropdownMenuContextValue = {
...
   DROPDOWN_MENU_NAME
 );
 
-type DropdownMenuOwnProps = {
+type DropdownMenuOwnProps = React.ComponentProps<typeof DropdownMenuImpl>;
+
+const DropdownMenu: React.FC<DropdownMenuOwnProps> = (props) => {
+  const parentMenuContext = React.useContext(SubmenuContext);
+  return (
+    <SubmenuContext.Provider value={parentMenuContext === undefined ? false : true}>
```
jjenzz, line 38: Why don't we do the same here then and just move this context to `DropdownMenuContent`... that way we're only a submenu if rendered inside content?

```jsx
<SubmenuContext.Provider value={true}>
```

I can't see an `const isRoot = parentContext === undefined` check anywhere in dropdown menu hence my confusion as to why we need `parentMenuContext` to be `undefined`.

andy-hook: In the previous config we needed to have an additional state available for the first render to know if the context even exists, hence the `undefined`, but wrapping the content means we can do it after that point, so we implicitly know it's root. https://github.com/radix-ui/primitives/pull/629/commits/f9061af8d8f45b4224c005daa0212a22a66b061f

jjenzz (approved):

PR #630, opened 2021-04-30
Update @radix-ui/react-[id|polymorphic] READMEs

djm:
### Description

Updates both @radix-ui/react-id & @radix-ui/react-polymorphic READMEs.

To ensure these utilities are correctly documented as public and that they have a link to their respective docs page.

### DIFF
--- a/packages/react/polymorphic/README.md
+++ b/packages/react/polymorphic/README.md
@@ -10,4 +10,4 @@ $ npm install @radix-ui/react-polymorphic
 
 ## Usage
 
-This is an internal utility, not intended for public usage.
+View docs [here](https://radix-ui.com/primitives/docs/utilities/polymorphic).
--- a/.yarn/versions/6423be2b.yml
+++ b/.yarn/versions/6423be2b.yml
@@ -1,0 +1,38 @@
+releases:
+  "@radix-ui/react-id": patch
+  "@radix-ui/react-polymorphic": patch
+
+declined:
+  - primitives
+  - "@radix-ui/react-accordion"
+  - "@radix-ui/react-alert-dialog"
+  - "@radix-ui/react-announce"
+  - "@radix-ui/react-arrow"
+  - "@radix-ui/react-aspect-ratio"
+  - "@radix-ui/react-avatar"
+  - "@radix-ui/react-checkbox"
+  - "@radix-ui/react-collapsible"
+  - "@radix-ui/react-collection"
+  - "@radix-ui/react-context-menu"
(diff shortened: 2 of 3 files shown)


packages/react/polymorphic/README.md
```diff
@@ -10,4 +13,4 @@ $ npm install @radix-ui/react-polymorphic
 
 ## Usage
 
-This is an internal utility, not intended for public usage.
+See [the docs](https://radix-ui.com/primitives/docs/utilities/polymorphic).
```
jjenzz, line 16: For consistency with our other packages, we usual write the following here:

```suggestion
View docs [here](https://radix-ui.com/primitives/docs/utilities/polymorphic).
```

Just in case we ever decide to do a project wide find & replace 🙏

jjenzz (changes requested):

djm:
@jjenzz Thanks for the review!

I made some changes based on your feedback:

1) The other packages don't have descriptions, just the docs link - so I also made that the case here so that all the READMEs match that pattern.

2) In doing that I noticed that `@radix-ui/react-id` also had public docs, so I've updated the README to fit this pattern too.

It's just a two-line diff now :+1:

Thanks!

packages/react/polymorphic/README.md
djm: Resolved locally, closing this :)

jjenzz (approved):
🙌 thanks for these!

benoitgrelard (approved):
Cheers Darian! 🙂

benoitgrelard:
cc @jjenzz I bypassed the branch security here so I could merge quickly, but we should perhaps think about how to make our contributors (and our) life easier here.

There are 2 main barriers:
- the version stuff, although this is easy enough for us to quickly jump in and perform (that's what I did here)
- Chromatic won't run on forked repos because the token is not public, Chromatic say if that's what you want, you can make the token public from a script (https://www.chromatic.com/docs/github-actions#forked-repositories) but I'm not sure what the impact of this is generally

jjenzz:
For small branches like these where they do not impact Chromatic tests, I think what you did is fine. For any component changes/contributions, like `Toolbar`, I changed the base branch for the `Toolbar` PR to an internal toolbar branch and then created a PR from that branch into `main`.

It doesn't seem like something we would need to do too often and I'm not sure exposing our Chromatic token publicly is a good idea 😬

djm:
> I'm not sure exposing our Chromatic token publicly is a good idea 😬

Not at their prices 😂

PR #641, opened 2021-05-06
[FocusScope] Fix type definition issues

### LINKED ISSUE #636: Cannot find name 'FocusScopeProps'.
### Current Behavior

When I try to build the types for my component library that uses `@radix-ui`, the TypeScript compiler screams that it can't find the `FocusScopeProps`:

image

https://github.com/kaisermann/radix-repro-1

### Expected behavior

Build the types without an issue 🙈 

### Reproducible example

https://github.com/kaisermann/radix-repro-1

```shell
$ npm install
$ npm run build
```

### Suggested solution

Didn't investigate it yet, but it seems that it's just missing an import on the generated typing file:

image

## System:
 - OS: macOS 11.2.1
 - CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
 - Memory: 189.97 MB / 16.00 GB
 - Shell: 3.1.2 - /usr/local/bin/fish
## Binaries:
 - Node: 14.15.4 - ~/.volta/tools/image/node/14.15.4/bin/node
 - Yarn: 1.22.10 - ~/.volta/tools/image/yarn/1.22.10/bin/yarn
 - npm: 7.5.4 - ~/.volta/tools/image/npm/7.5.4/bin/npm
 - Watchman: 4.9.0 - /usr/local/bin/watchman
## npmPackages:
 - @radix-ui/react-dialog: ^0.0.16 => 0.0.16

benoitgrelard:
Fixes #636

### DIFF
--- a/.yarn/versions/4c0325da.yml
+++ b/.yarn/versions/4c0325da.yml
@@ -1,0 +1,11 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-focus-scope": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-popover": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


andy-hook:
What was happening under the hood that this solves?

jjenzz (approved):

benoitgrelard:
> What was happening under the hood that this solves?

The type name was the same as the component name, so it was getting confused and not outputing the right thing in the type definition file. We think it's an issue with parcel's optimisation cause looking at just TS seems ok with it:

image

Although you can't export both anyway:

image

And admittedly it's weird and confusing to have 2 things called the same, hence the change.

PR #642, opened 2021-05-06
[Primitive][Slot] Fix slot event composition for custom handlers

cappuc:
### Description

Right now `SlotClone` components replace child handlers (props starting with 'on') with a composed event with child handler and slot handler.
If the child handler is a custom handler that should receive more than 1 argument, it will receive only the first argument because the `composeEventHandlers` only pass the first argument to the handlers. 

With this change it will pass all received arguments to the handlers.

### DIFF
--- a/packages/core/primitive/src/primitive.tsx
+++ b/packages/core/primitive/src/primitive.tsx
@@ -1,10 +1,10 @@
 function composeEventHandlers<E>(
-  originalEventHandler?: (event: E) => void,
+  originalEventHandler?: (...args: any[]) => void,
   ourEventHandler?: (event: E) => void,
   { checkForDefaultPrevented = true } = {}
 ) {
   return function handleEvent(event: E) {
-    originalEventHandler?.(event);
+    originalEventHandler?.(...arguments);
 
     if (checkForDefaultPrevented === false || !((event as unknown) as Event).defaultPrevented) {
       return ourEventHandler?.(event);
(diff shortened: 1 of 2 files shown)


cappuc:
I added a check on type of `event` param to prevent errors if it is not an object

benoitgrelard:
Hi @cappuc, this code is meant for events so at first glance it seems a bit odd to accept multiple args, or check for object, etc. Are you able to provide us with an example where things don't work as you expected?

cappuc:
Hi,
I built a custom `Select` component using your `@radix-ui/react-collection` (which uses Slot under the hood) to integrate with `SelectOption` childs.

The `SelectOption` component accept an optional payload and the onChange callback on my select component is like this:

```ts
interface SelectProps {
    onChange?: (newValue: string, payload?: unknown) => void;
}
```

and with the current implementation only the first parameter is passed to the original callback.

I made the change in the `composeEventHandlers` function but it can also be done specifically for the `Slot` component. 

The problem is in the `mergeProps` function ([here](https://github.com/radix-ui/primitives/blob/974e678ae40709e959a2f4f340754a5dc0975e39/packages/react/slot/src/Slot.tsx#L74-L90))
which calls `composeEventHandlers` if the prop name starts with `on` and this works only for native-like callbacks with only an event argument.

jjenzz:
@benoitgrelard yeah, looks like we made a bad assumption in `Slot` that all `on*` props would be DOM event handlers, but they're not necessarily. Some of our own APIs are called with a boolean and not an event that can be prevented, e.g. `onOpenChange`.

I'd say it should be solved in `Slot` but not sure how best to do this atm.

benoitgrelard:
Thanks for this @cappuc, I'll now merge in a branch our side and will take it from there to run our checks, version, etc.
Thanks for your contribution! 🙏

PR #643, opened 2021-05-08
Switch to Preconstruct for building packages

Andarist:
### Description

The PR has been prompted by my conversation on Twitter with @jjenzz: https://twitter.com/jjenzz/status/1390973850716884994

Basically... with Preconstruct you don't need to run builds in topological order because it first just outputs dummy `.d.ts` that redirects requests to the respective `src/index.ts`. Thanks to that all builds can just happen in parallel.


.yarnrc.yml
```diff
@@ -17,4 +17,4 @@ plugins:
   - path: .yarn/plugins/@yarnpkg/plugin-version.cjs
     spec: '@yarnpkg/plugin-version'
 
-yarnPath: .yarn/releases/yarn-2.1.1.cjs
+yarnPath: .yarn/releases/yarn-2.4.1.cjs
```
Andarist, line 20: For some random reason, I couldn't install Preconstruct on a previous version because installs were failing with:
```
➤ YN0001: │ Error: fsevents@patch:fsevents@npm%3A2.3.2#builtin<compat/fsevents>::version=2.3.2&hash=495457: Cannot apply hunk #1
    at d (/radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:190674)
    at async /radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:189912
    at async h (/radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:188524)
    at async g (/radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:189885)
    at async q.fetchers.patchPackage (/radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:204417)
    at async B (/radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:226253)
    at async Q (/radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:226716)
    at async v (/radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:227192)
    at async C.fetchPackageFromCache (/radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:227431)
    at async q.fetchers.fetch (/radix-ui-primitives/.yarn/releases/yarn-2.1.1.cjs:2:203286)
```

Upgrading Yarn has helped so I didn't dig deeper into this 🤷‍♂️

package.json
```diff
@@ -15,16 +15,13 @@
     "storybook": "start-storybook -p 9009 --ci",
     "dev": "yarn storybook",
     "build-storybook": "build-storybook",
-    "// build": "For context on tsconfig replacements in build scripts, see https://github.com/radix-ui/primitives/pull/361#discussion_r555004944",
-    "build": "yarn build:config && yarn build:packages && yarn build:cleanup",
-    "build:config": "mv tsconfig.json tsconfig.tmp.json && mv tsconfig.production.json tsconfig.json",
```
Andarist, line 20: to the best of my knowledge - you don't need this tsconfig dance anymore

package.json
```diff
@@ -15,16 +15,13 @@
...
-    "build:packages": "parcel build 'packages/*/*/' --no-cache",
-    "build:cleanup": "mv tsconfig.json tsconfig.production.json && mv tsconfig.tmp.json tsconfig.json",
+    "build": "preconstruct build",
     "publish": "yarn bump && yarn build && yarn workspaces foreach -pv --exclude primitives --exclude ssr-testing npm publish --tolerate-republish --access public",
     "clean": "yarn workspaces foreach -pv --exclude primitives --exclude ssr-testing run clean",
     "reset": "yarn clean && rm -rf node_modules .yarn/cache .parcel-cache",
     "bump": "yarn version apply --all",
-    "bump:check": "yarn version check"
+    "bump:check": "yarn version check",
+    "_postinstall": "preconstruct dev"
```
Andarist, line 24: The current version of Yarn fails executing this `postinstall` with:
```
➤ YN0006: │ primitives@workspace:. lists build scripts, but is referenced through a soft link. Soft links don't support build scripts, so they'll be ignored.
➤ YN0007: │ primitives@workspace:. must be built because it never did before or the last one failed
➤ YN0009: │ primitives@workspace:. couldn't be built successfully (exit code 1, logs can be found here: /private/var/folders/ws/tlwxqvs5407f21nxy7w8bgr80000gp/T/xfs-e80184bd/build.log)
➤ YN0009: │ primitives@workspace:. couldn't be built successfully (exit code 1, logs can be found here: /private/var/folders/ws/tlwxqvs5407f21nxy7w8bgr80000gp/T/xfs-e80184bd/build.log)
```

From what I recall - at least this error message is a bug and it has already been fixed in Yarn so you could try switching to the version from their master/main branch.

packages/core/number/package.json
```diff
@@ -3,9 +3,9 @@
   "version": "0.0.6",
   "license": "MIT",
   "source": "src/index.ts",
-  "main": "dist/index.js",
-  "module": "dist/index.module.js",
-  "types": "dist/index.d.ts",
+  "main": "dist/radix-ui-number.cjs.js",
+  "module": "dist/radix-ui-number.esm.js",
+  "types": "dist/radix-ui-number.cjs.d.ts",
```
Andarist, line 8: Technically the `"types"` field is not required so it could be removed but there are still some reasons to keep it around. See https://github.com/preconstruct/preconstruct/issues/259

Andarist:
@jjenzz could you describe what you are doing to get yourself into this situation on this branch? What commands do u run and what files are you inspecting?

benoitgrelard:
- Start with a clean slate, so: `yarn clean`
- `yarn build`
- open `packages/react/menu/dist/index.d.ts`
- look for `MenuAnchor`

Andarist:
Okay - so the problem is not as much that this doesn't work but that the generated declarations contain references to the `src` directory which in turn would force you to publish those, right? 

I think I understand the issue better now and I agree it's a slight problem. I have an idea how this could be fixed in Preconstruct - gonna be consulting that with Mitchell and I will let u know if we can fix this for you.

jjenzz:
> the generated declarations contain references to the `src` directory

Yep. We would expect these imports to be `import("@radix-ui/rect")` which we get if we build the types in topological order.

Thank you for taking a look @Andarist 🙂 In the meantime, do you mind if we close this PR until preconstruct has a way to handle it?

PR #651, opened 2021-05-14
[Collapsible][Accordion] Rename parts 🔥 + add `collapsible` prop for `type="single"` 🔥 

benoitgrelard:
This PR contains a few things that came from discussing with @peduarte whilst he was writing the article for `Accordion`.

- 🐛 Omit the `type` prop on `Accordion` so it doesn't appear in the DOM
- 🔥 Rename `Collapsible.Button` to `Collapsible.Trigger`
- 🔥 Rename `Accordion.Button` to `Accordion.Trigger`
- 🔥 Rename `Accordion.Panel` to `Accordion.Content`
  - 🔥 Rename `--radix-accordion-panel-height` to `--radix-accordion-content-height`
- ✨ ~**Add new `preventAllClosed` prop to `Accordion`**~
- :fire: **add new `collapsible` prop  to `<Accordion type="single">`**

Now, for a bit of context on that last one:

> Add new `preventAllClosed` prop to `Accordion`

Originally I was against that, as we had discussed a few times when working on `Accordion` this is just state so people can control it and do what they want to not set the state in some circumstances. However, Pedro felt like it was a common enough feature to be a top-level config prop. At that point I was still pushing back because I didn't see much advantage to it and just saw the API surface growing for not much reason.

Later on, Pedro discovered something in the spec that we probably misinterpreted.

> If the accordion panel associated with an accordion header **is visible**, and **if the accordion does not permit the panel to be collapsed**, the header button element has aria-disabled set to true.

In our code, we had it that whenever any panel is open, the `button` has `aria-disabled="true"` which was wrong. To be able to achieve what the spec says, suddenly there was an actual justification for having a `preventAllClosed` prop so we decided to go with that.

> **Note:** Not a fan of the name of the prop, would gladly take suggestions 🙂 

---

**UPDATE**

We've now got only a `collapsible` prop for `type="single"` as that behaviour doesn't really make sense with `type="multiple"`. It's `false` by default, meaning that the default behaviour has now changed and it closing all items isn't possible by default. 🔥

### DIFF
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -61,11 +61,11 @@ type AccordionSingleOwnProps = Polymorphic.Merge<
      */
     type: 'single';
     /**
-     * The controlled stateful value of the accordion item whose panel is expanded.
+     * The controlled stateful value of the accordion item whose content is expanded.
      */
     value?: string;
     /**
-     * The value of the item whose panel is expanded when the accordion is initially rendered. Use
+     * The value of the item whose content is expanded when the accordion is initially rendered. Use
      * `defaultValue` if you do not need to control the state of an accordion.
      */
     defaultValue?: string;
@@ -114,11 +114,11 @@ type AccordionMultipleOwnProps = Polymorphic.Merge<
      */
     type: 'multiple';
     /**
-     * The controlled stateful value of the accordion items whose panels are expanded.
+     * The controlled stateful value of the accordion items whose contents are expanded.
      */
     value?: string[];
     /**
-     * The value of the items whose panels are expanded when the accordion is initially rendered. Use
+     * The value of the items whose contents are expanded when the accordion is initially rendered. Use
      * `defaultValue` if you do not need to control the state of an accordion.
      */
     defaultValue?: string[];
@@ -401,36 +401,36 @@ const AccordionTrigger = React.forwardRef((props, forwardedRef) => {
 AccordionTrigger.displayName = TRIGGER_NAME;
 
 /* -------------------------------------------------------------------------------------------------
- * AccordionPanel
+ * AccordionContent
  * -----------------------------------------------------------------------------------------------*/
 
-const PANEL_NAME = 'AccordionPanel';
+const CONTENT_NAME = 'AccordionContent';
 
-type AccordionPanelOwnProps = Polymorphic.OwnProps<typeof CollapsiblePrimitive.Content>;
-type AccordionPanelPrimitive = Polymorphic.ForwardRefComponent<
+type AccordionContentOwnProps = Polymorphic.OwnProps<typeof CollapsiblePrimitive.Content>;
+type AccordionContentPrimitive = Polymorphic.ForwardRefComponent<
   Polymorphic.IntrinsicElement<typeof CollapsiblePrimitive.Content>,
-  AccordionPanelOwnProps
+  AccordionContentOwnProps
 >;
 /**
- * `AccordionPanel` contains the collapsible content for an `AccordionItem`.
+ * `AccordionContent` contains the collapsible content for an `AccordionItem`.
  */
-const AccordionPanel = React.forwardRef((props, forwardedRef) => {
-  const itemContext = useAccordionItemContext(PANEL_NAME);
+const AccordionContent = React.forwardRef((props, forwardedRef) => {
+  const itemContext = useAccordionItemContext(CONTENT_NAME);
   return (
     <CollapsiblePrimitive.Content
       role="region"
       aria-labelledby={itemContext.triggerId}
       {...props}
       style={{
-        ['--radix-accordion-panel-height' as any]: 'var(--radix-collapsible-content-height)',
+        ['--radix-accordion-content-height' as any]: 'var(--radix-collapsible-content-height)',
         ...props.style,
       }}
       ref={forwardedRef}
     />
   );
-}) as AccordionPanelPrimitive;
+}) as AccordionContentPrimitive;
 
-AccordionPanel.displayName = PANEL_NAME;
+AccordionContent.displayName = CONTENT_NAME;
 
 /* -----------------------------------------------------------------------------------------------*/
 
@@ -446,25 +446,25 @@ const Root = Accordion;
 const Item = AccordionItem;
 const Header = AccordionHeader;
 const Trigger = AccordionTrigger;
-const Panel = AccordionPanel;
+const Content = AccordionContent;
 
 export {
   Accordion,
   AccordionItem,
   AccordionHeader,
   AccordionTrigger,
-  AccordionPanel,
+  AccordionContent,
   //
   Root,
   Item,
   Header,
   Trigger,
-  Panel,
+  Content,
 };
 export type {
   AccordionPrimitive,
   AccordionItemPrimitive,
   AccordionHeaderPrimitive,
   AccordionTriggerPrimitive,
-  AccordionPanelPrimitive,
+  AccordionContentPrimitive,
 };
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -17,8 +17,8 @@ const ACCORDION_NAME = 'Accordion';
 const ACCORDION_KEYS = ['Home', 'End', 'ArrowDown', 'ArrowUp'];
 
 type AccordionOwnProps =
-  | Polymorphic.OwnProps<typeof AccordionSingle>
-  | Polymorphic.OwnProps<typeof AccordionMultiple>;
+  | ({ type: 'single' } & Polymorphic.OwnProps<typeof AccordionSingle>)
+  | ({ type: 'multiple' } & Polymorphic.OwnProps<typeof AccordionMultiple>);
 
 type AccordionPrimitive = Polymorphic.ForwardRefComponent<
   | Polymorphic.IntrinsicElement<typeof AccordionSingle>
@@ -28,14 +28,14 @@ type AccordionPrimitive = Polymorphic.ForwardRefComponent<
 
 const Accordion = React.forwardRef((props, forwardedRef) => {
   const { type, ...accordionProps } = props;
-  const singleProps = accordionProps as React.ComponentProps<typeof AccordionSingle>;
-  const multipleProps = accordionProps as React.ComponentProps<typeof AccordionMultiple>;
 
   if (type === 'single') {
+    const singleProps = accordionProps as React.ComponentProps<typeof AccordionSingle>;
     return <AccordionSingle {...singleProps} ref={forwardedRef} />;
   }
 
   if (type === 'multiple') {
+    const multipleProps = accordionProps as React.ComponentProps<typeof AccordionMultiple>;
     return <AccordionMultiple {...multipleProps} ref={forwardedRef} />;
   }
 
@@ -62,10 +62,6 @@ const AccordionCollapsibleContext = React.createContext(false);
 type AccordionSingleOwnProps = Polymorphic.Merge<
   Polymorphic.OwnProps<typeof AccordionImpl>,
   {
-    /**
-     * Allow only one item to be open at a time.
-     */
-    type: 'single';
     /**
      * The controlled stateful value of the accordion item whose content is expanded.
      */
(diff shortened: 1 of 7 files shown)


packages/react/accordion/src/Accordion.tsx
```diff
@@ -99,9 +101,13 @@ const AccordionSingle = React.forwardRef((props, forwardedRef) => {
     <AccordionValueProvider
       value={value ? [value] : []}
       onItemOpen={setValue}
-      onItemClose={() => setValue('')}
+      onItemClose={() => !preventAllClosed && setValue('')}
```
benoitgrelard, line 104: This is the bit where the `preventAllClosed` is implemented for `type="single"`.

packages/react/accordion/src/Accordion.tsx
```diff
@@ -154,22 +162,32 @@ const AccordionMultiple = React.forwardRef((props, forwardedRef) => {
   );
 
   const handleItemClose = React.useCallback(
-    (itemValue) => setValue((prevValue = []) => prevValue.filter((value) => value !== itemValue)),
-    [setValue]
+    (itemValue) =>
+      setValue((prevValue = []) => {
+        if (preventAllClosed && prevValue.length === 1 && prevValue[0] === itemValue)
+          return prevValue;
```
benoitgrelard, line 168: This is the bit where the `preventAllClosed` is implemented for `type="multiple"`.

packages/react/accordion/src/Accordion.tsx
```diff
@@ -353,84 +380,89 @@ const AccordionHeader = React.forwardRef((props, forwardedRef) => {
...
-        buttonNodes.delete(buttonNode);
+        triggerNodes.delete(triggerNode);
       };
     }
     return;
-  }, [buttonNodesRef]);
+  }, [triggerNodesRef]);
+
+  const ariaDisabled =
+    itemContext.open && valueContext.value.length === 1 && accordionContext.preventAllClosed;
```
benoitgrelard, line 421: This is now the new condition for `aria-disable`ing a trigger.

peduarte:
Excellent stuff thank you 🙏

jjenzz:
> In our code, we had it that whenever any panel is open, the button has aria-disabled="true" which was wrong

This will have been because we originally never allowed accordion items to be closed. Sounds like a regression was created when we allowed them to close 🙈 oops.

I'm not a fan of props that reverse things tho tbh. I prefer additive APIS so, it wouldn't close by default and if people want them to close then they add a `toggleable` prop or something.

colmtuite:
>  it wouldn't close by default and if people want them to close then they add a toggleable prop

+1

packages/react/accordion/src/Accordion.tsx
```diff
@@ -61,18 +63,23 @@ type AccordionSingleOwnProps = Polymorphic.Merge<
...
     defaultValue?: string;
     /**
      * The callback that fires when the state of the accordion changes.
      */
     onValueChange?(value: string): void;
+    /**
+     * Whether the accordion can be completely collapsed.
+     * @default false
+     */
+    collapsible?: boolean;
```
benoitgrelard, line 82: Prop only on `AccordionSingleOwnProps`.

packages/react/accordion/src/Accordion.tsx
```diff
@@ -99,9 +108,11 @@ const AccordionSingle = React.forwardRef((props, forwardedRef) => {
     <AccordionValueProvider
       value={value ? [value] : []}
       onItemOpen={setValue}
-      onItemClose={() => setValue('')}
+      onItemClose={() => collapsible && setValue('')}
     >
-      <AccordionImpl {...accordionSingleProps} ref={forwardedRef} />
+      <AccordionCollapsibleContext.Provider value={collapsible}>
```
benoitgrelard, line 113: Using a new context just for this as it's optional (only for type single)

packages/react/accordion/src/Accordion.tsx
```diff
@@ -353,84 +365,88 @@ const AccordionHeader = React.forwardRef((props, forwardedRef) => {
...
-  }, [buttonNodesRef]);
+  }, [triggerNodesRef]);
 
   return (
-    <CollapsiblePrimitive.Button
-      aria-disabled={itemContext.open || undefined}
-      id={itemContext.buttonId}
+    <CollapsiblePrimitive.Trigger
+      aria-disabled={
+        (itemContext.open && valueContext.value.length === 1 && !collapsible) || undefined
```
benoitgrelard, line 408: reading that context

packages/react/accordion/src/Accordion.tsx
```diff
@@ -61,18 +63,23 @@ type AccordionSingleOwnProps = Polymorphic.Merge<
...
+     * The value of the item whose content is expanded when the accordion is initially rendered. Use
      * `defaultValue` if you do not need to control the state of an accordion.
      */
     defaultValue?: string;
     /**
      * The callback that fires when the state of the accordion changes.
      */
     onValueChange?(value: string): void;
+    /**
+     * Whether the accordion can be completely collapsed.
```
jjenzz, line 79: Read this and thought if someone passes `collapsible={false}` it will still be collapsed by default which could seem like a bug given this description. Maybe we should make it more explicit?

```suggestion
     * Whether an accordion item can be collapsed after it has been opened.
```

packages/react/accordion/src/Accordion.tsx
```diff
@@ -86,6 +93,8 @@ const AccordionSingle = React.forwardRef((props, forwardedRef) => {
     value: valueProp,
     defaultValue,
     onValueChange = () => {},
+    collapsible = false,
+    type, // pulling it out as we don't need it and don't want it to end up on the DOM element
```
jjenzz, line 97: I'm wondering if this should be moved up a level so it wouldn't even exist in these props and then wouldn't need omitting. It's only the parent that needs this prop to determine which one to render.

```jsx
type AccordionOwnProps = 
	| { type: 'single' } & Polymorphic.OwnProps<typeof AccordionSingle>
	| { type: 'multiple' } & Polymorphic.OwnProps<typeof AccordionMultiple>; 

const Accordion = React.forwardRef((props, forwardedRef) => {
	const { type, ...accordionProps } = props;

	if (type === 'single') {
		// ...
	}

	if (type === 'multiple') {
		// ...
	}
});
```

packages/react/accordion/src/Accordion.tsx
```diff
@@ -338,7 +350,7 @@ type AccordionHeaderPrimitive = Polymorphic.ForwardRefComponent<
  */
 const AccordionHeader = React.forwardRef((props, forwardedRef) => {
   const { as = HEADER_DEFAULT_TAG, ...headerProps } = props;
-  const itemContext = useAccordionItemContext(BUTTON_NAME);
+  const itemContext = useAccordionItemContext(HEADER_NAME);
```
jjenzz, line 353: 👀 👍

packages/react/accordion/src/Accordion.tsx
jjenzz: How about if we add the `AccordionCollapsibleContext.Provider` to both types and in the `AccordionSingle` its value will be the `collapsible` prop and in `AccordionMultiple` its value will always be `false`. 

This can then be simplified to `itemContext.open && !collapsible` so the trigger doesn't have implicit logic that assumes 1 value means single type.

jjenzz (changes requested):

packages/react/accordion/src/Accordion.tsx
benoitgrelard: I tried that but it didn't let me do it, because of the types.

packages/react/accordion/src/Accordion.tsx
benoitgrelard: > How about if we add the AccordionCollapsibleContext.Provider to both types and in the AccordionSingle its value will be the collapsible prop and in AccordionMultiple its value will always be true.

Not sure we need to do it for multiple, I like that currently it's explicitly for single.

> This can then be simplified to itemContext.open && !collapsible so the trigger doesn't have implicit logic that assumes 1 value means single type.

Good catch, that was there because we were handling both before, but we're not anymore, so we can remove.

packages/react/accordion/src/Accordion.tsx
jjenzz: it didn't let you do which part?

benoitgrelard: image

Won't work because obviously they are supposed to be parts on it.

packages/react/accordion/src/Accordion.tsx
jjenzz: I think we need it because otherwise `collapsible` will be undefined in the `AccordionMultiple` which is falsey. That's why I suggested doing `<AccordionCollapsibleContext.Provider value={true}>` in the `AccordionMultiple` component, because collapsible does exist for multiple, it's just always collapsible.

benoitgrelard: ok, I don't have an issue with the falsey thing but I can see what you're saying, will do now.

packages/react/accordion/src/Accordion.tsx
benoitgrelard: Right, casting yeah but that's just going around it, seems weird to me, like, we're removing it, but then telling TS it's still there. I'm not sure at this point that it's better than where we're doing it at the moment.

jjenzz: That's not weird at all, that's what you have to do with TS when you have union types like this. It doesn't know what the spread props are anymore so you need to hold its hand. That't the way it has always been. 

I'd prefer to not pass the prop down if it's not needed, so the code does what it is supposed to rather than omitting it later with a comment that basically says "we're passing this down unecessarily", just don't pass it no?

benoitgrelard: > It doesn't know what the spread props

That's only half the problem here really, you've removed the `type` so yes, it can't figure out which part of the union it is, you've also removed it, so more importantly because that prop is mandatory it's obviously not in the props anymore so that's really the problem here.

As in, `AccordionSingle` expects `type="single` and `AccordionMultiple` expects `type="multiple"`, it just so happens that we don't use that value, but from a types standpoint, they are required.

Casting fixes both issues at once obviously, but to me I'm thinking the main issue is that we've removed it and now it's missing (from a types standpoint).

jjenzz: > we're removing it, but then telling TS it's still there

I'm not telling TS it is still there btw? Maybe there was a misunderstanding somewhere. You would remove it from the Single/Multiple component prop types because they don't need it.

benoitgrelard: I've done it now as I don't think there's any other way to do it in a cleaner fashion.

packages/react/accordion/src/Accordion.tsx
jjenzz: It's not about having a personal issue with the falsey thing, it's a bug is what I'm saying. If there is no `collapsible` context in `AccordionMultiple` then 

```
const collapsible = useCollapsibleContext() // = `undefined` which is == `false`;
```

So, where you have done `aria-disabled={context.open && !collapsible}` that would be `true` in `AccordionMultiple` which is wrong. It's a bug. That's why I'm saying we need to pass `collapsible` through context as `true`.

packages/react/accordion/src/Accordion.tsx
benoitgrelard: > I'm not telling TS it is still there btw? Maybe there was a misunderstanding somewhere. We would remove it from the Single/Multiple component prop types because they don't need it.

Maybe I'm missing something then…

benoitgrelard: Oh, are you saying we'd do this?

```
type AccordionOwnProps =
  | ({ type: 'single' } & Polymorphic.OwnProps<typeof AccordionSingle>)
  | ({ type: 'multiple' } & Polymorphic.OwnProps<typeof AccordionMultiple>);
```

?

packages/react/accordion/src/Accordion.tsx
```diff
@@ -61,18 +67,23 @@ type AccordionSingleOwnProps = Polymorphic.Merge<
      */
     type: 'single';
```
jjenzz, line 68: Can we remove these from here? This is the part I don't think you understood from my original feedback.

> I'm wondering if this should be moved up a level so it wouldn't even exist in these props [...]. It's only the parent that needs this prop

image


I was saying there that we would declare the `type` prop in the `AccordionOwnProps` and not in the lower level components that don't need it. Only the parent needs them to determine the switch.

jjenzz (changes requested):

packages/react/accordion/src/Accordion.tsx
jjenzz: Yep, [that's what I said](https://github.com/radix-ui/primitives/pull/651#discussion_r632558843) 😛

benoitgrelard: oh man, I'm freaking blind…

jjenzz (approved):
bellissimo 👌

PR #661, opened 2021-05-24
[FocusScope][DismissableLayer] Use `Primitive` and `Slot`

benoitgrelard:
> Note: Easier to review without whitespace.

Another preliminary PR for the upcoming modality changes.

@jjenzz and I had discussed this at some point, and it's easier to do it now, rather than in the midst of everything else.
It rejigs `FocusScope` and `DismissableLayer` as self-contained primitives using our usual paradigm.
This has the effect of simplifying how they are composed externally as previously render props made it look almost like leaky abstractions.

### DIFF
--- a/packages/react/slot/src/Slot.tsx
+++ b/packages/react/slot/src/Slot.tsx
@@ -74,14 +74,18 @@ function mergeProps(slotProps: AnyProps, childProps: AnyProps) {
   // all child props should override
   const overrideProps = { ...childProps };
 
-  // if it's a handler, modify the override by composing the base handler
   for (const propName in childProps) {
     const slotPropValue = slotProps[propName];
     const childPropValue = childProps[propName];
-    const isHandler = /^on[A-Z]/.test(propName);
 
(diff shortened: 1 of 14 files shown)


packages/react/slot/src/Slot.tsx
```diff
@@ -74,15 +74,19 @@ function mergeProps(slotProps: AnyProps, childProps: AnyProps) {
...
 
+    const isHandler = /^on[A-Z]/.test(propName);
+    // if it's a handler, modify the override by composing the base handler
     if (isHandler) {
       overrideProps[propName] = composeHandlers(childPropValue, slotPropValue);
     }
+    // if it's `style`, we merge them
+    else if (propName === 'style') {
+      overrideProps[propName] = { ...slotPropValue, ...childPropValue };
+    }
```
benoitgrelard, line 89: This is probably the most notable change in this PR, had to make sure we compose `style` in `Slot` as well because we now have components that get slotted which have internal styles (`DismissableLayer`).

jjenzz (approved):

PR #662, opened 2021-05-24
[Menu] Provide predictable typeahead scoping

andy-hook:
Builds on https://github.com/radix-ui/primitives/pull/627

This change allows typeahead to reliably scope its search lookups to a single menu level, this is a prerequisite for making submenus function correctly with our typeahead functionality.

It also adds an explicit way to fire a handler at the leading edge of a search, useful for exclusively triggering interactions on items without inadvertently mixing them into the typeahead bindings.

While `querySelectorAll` is not a problem with portals, it will cause issues if the submenus become inline. Hence I'm preemptively making a collection the source of truth.


packages/react/menu/src/MenuTypeahead.tsx
```diff
@@ -0,0 +1,189 @@
...
+  return (
+    <TypeaheadProvider searching={searching}>
+      <Primitive
+        {...props}
+        ref={composeRefs(forwardedRef, containerRef)}
+        onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
+          const modifierKey = event.ctrlKey || event.altKey || event.metaKey;
+          if (event.key.length === 1 && !modifierKey) {
+            // Perform search on a single menu level only
+            event.stopPropagation();
```
andy-hook, line 86: I don't love `stopPropagation`, but here it's the more performant option vs triggering n handlers and excluding them.

packages/react/menu/src/MenuTypeahead.tsx
```diff
@@ -0,0 +1,189 @@
...
+type MenuTypeaheadItemOwnProps = {
+  disabled?: boolean;
+  textValue?: string;
+
+  /**
+   * Callback that fires at the leading edge of a typeahead.
+   * Helpful for preventing a handler from firing part way through
+   * a lookup e.g. if a search includes a character followed by a space
+   */
+  onSearchingExclusiveKeyDown?(event: React.KeyboardEvent): void;
```
andy-hook, line 114: Naming this is really hard 🙈 , if you have suggestions let me know.

packages/react/menu/src/MenuTypeahead.tsx
```diff
@@ -0,0 +1,189 @@
...
+import { createCollection } from '@radix-ui/react-collection';
+import { composeEventHandlers } from '@radix-ui/primitive';
+import { composeRefs } from '@radix-ui/react-compose-refs';
+import { Primitive } from '@radix-ui/react-primitive';
+
+import type * as Polymorphic from '@radix-ui/react-polymorphic';
+
+type ItemElement = React.ElementRef<typeof MenuTypeaheadItem>;
+
+const [CollectionSlot, CollectionItemSlot, useCollection] = createCollection<ItemElement, {}>();
```
andy-hook, line 12: There is some overlap here as `Menu` does already handle a collection but I found having this makes the separation much cleaner.

benoitgrelard:
Ha! Interesting! When I made the new `collection` package and then updated `Menu` to use it I remember telling @jjenzz that we could then make use of it for typeahead (https://github.com/radix-ui/primitives/pull/618#pullrequestreview-648680190).

So this is more of an architecture review for now, but my plan was to simply break away from the hook based abstraction and instead inline the state and `onKeyDownCapture` handler inside `MenuContentImpl` directly as the abstraction is kinda unnecessary as it's not reused anywhere, and now it should be easier than before because we can reuse the `getItems` which is already available in `MenuContentImpl` as items needed to be tracked there anyway.

Let me know if that makes sense or if perhaps you're one step ahead of me in terms of requirements due to sub-menus.

jjenzz:
@benoitgrelard we had discussed inlining this all but it would mean littering boolean checks, like:

```jsx
if (!disabled && event.key === ' ' && !isTypingAhead)
```

It started to feel boolean-heavy with the inline approach so we discussed various solutions and in the end decided that if there was a custom event that fired from the typeahead stuff when a non-search specific keydown occurred then these code blocks could bind to that instead.

**However**, I've just suddenly thought, "why didn't we use a state machine?". It was a long week going back and forth on various `Menu` issues we were having so perhaps we just overlooked that as an option 😬

I'm trying to remember the various problems this solves now @andy-hook, can you remind me? I think they were:

- Avoiding having `!isTypingAhead` checks littered about
- Removing `querySelectorAll` so that the collection is scoped 

Anything else? I'm wondering how hard it would be to inline what you have here with a state machine instead if these were the only two issues.

benoitgrelard:
Yeah I think we should see start by inlining and see if we can get this to feel good. Then if we can't we can explore other options.

jjenzz:
Maybe we can keep this PR for now but open a separate one with the inlined stuff so we can compare.

packages/react/menu/src/MenuTypeahead.tsx
jjenzz: I know we're potentially getting rid of the separate component idea but just wanted to say I agree with this decision haha, I'd have done the same :+1:

andy-hook:
> I'm trying to remember the various problems this solves now @andy-hook, can you remind me?

- Moving away from `keyDownCapture` due to the nesting requirement
- Ability to provide different `keyDown` handlers depending on context (for a standard item vs a submenu trigger) that are also excluded during the “typing ahead” interaction
- Scoping values via collection

All while:
- Avoiding excessive boolean logic
- Duplicating too many values across the two contexts we will have with the submenu implementation

I don’t love the duplicated collection work here and I’m not opposed to inlining, this PR is simply the result of our conversations last week.

It’s difficult to reason here with only half the picture, I can test an inline approach and see how it looks.

packages/react/menu/src/MenuTypeahead.tsx
jjenzz: For future reference, I'm thinking we should probably dispatch custom events (that don't bubble) when we want things to `stopPropagation`. It makes me uneasy that we'd be messing with native event behaviour for consumers otherwise.

andy-hook: The rub here is that we want to be able to trigger a "typeahead" search when focused on the container as well as the items.

andy-hook:
It might be more useful to show this relative to the submenu work. Let's leave this here for now and diff it against the submenu PR once that's ready.

packages/react/menu/src/MenuTypeahead.tsx
benoitgrelard: We could still rely on bubbling with a custom event and make it cancellable, cause then you're just stopping propagation of a custom event so it's not disrupting native events.

benoitgrelard:
Regardless of where we end up with this, I don't think we would need the data attribute at all or use the DOM anymore if we use the collection right? We should be able to add the `textValue` to the tracked data for each item, allowing us to then get a ref to an element by text value.

andy-hook:
> I don't think we would need the data attribute at all or use the DOM anymore if we use the collection right? We should be able to add the `textValue` to the tracked data for each item, allowing us to then get a ref to an element by text value.

Yes! this would be much better.

packages/react/menu/src/MenuTypeahead.tsx
andy-hook: oh yeah I see what you both mean. I hadn't thought about it that way, will take a look!

andy-hook:
@benoitgrelard @jjenzz – This one is ready for review, I've inlined and switched to a custom event as discussed.

packages/react/menu/src/Menu.tsx
```diff
@@ -186,11 +188,9 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
         onMouseMove={composeEventHandlers(props.onMouseMove, () => {
           if (!props.disabled) context.onOpenChange(true);
         })}
-        onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
-          if (!props.disabled && SUB_OPEN_KEYS[context.dir].includes(event.key)) {
-            context.onKeyOpen();
-          }
-        })}
+        onSearchExclusiveKeyDown={(event) => {
```
jjenzz, line 191: Looking at this `onSearchExclusiveKeyDown` implementation now, I'm starting to think we probably don't need it. It seems we can use `onKeyDownCapture` on `content` to `stopPropagation` when we're searching instead... something like this: https://codesandbox.io/s/heuristic-albattani-p1u0v?file=/src/App.js

packages/react/menu/src/Menu.tsx
```diff
@@ -186,11 +188,9 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
...
+          if (SUB_OPEN_KEYS[context.dir].includes(event.key)) context.onKeyOpen();
```
jjenzz, line 192: Curious why we don't need the `!props.disabled` check here anymore?

packages/react/menu/src/Menu.tsx
```diff
@@ -516,15 +565,21 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
-                    typeaheadProps.onKeyDownCapture
-                  )}
-                  // focus first/last item based on key pressed
                   onKeyDown={composeEventHandlers(contentProps.onKeyDown, (event) => {
-                    const content = contentRef.current;
+                    // perform typeahead search
+                    const modifierKey = event.ctrlKey || event.altKey || event.metaKey;
+                    if (event.key.length === 1 && !modifierKey) {
+                      const typeaheadKeyDownEvent = new CustomEvent(TYPEAHEAD_SEARCH, {
+                        bubbles: true,
```
jjenzz, line 573: Do we need this to `bubble` if we're always calling `event.stopPropagation`? Looks like we're dispatching and listening on `content` so seems we can remove the latter if we set this to `false`?

packages/react/menu/src/Menu.tsx
```diff
@@ -435,16 +437,62 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+    },
+    [getItems, setSearch]
+  ) as EventListener;
+
+  React.useEffect(() => {
+    const container = contentRef.current;
+    if (container) {
+      container.addEventListener(TYPEAHEAD_SEARCH, handleTypeaheadSearch);
+      return () => container.removeEventListener(TYPEAHEAD_SEARCH, handleTypeaheadSearch);
+    }
```
jjenzz, line 493: nit: 🙈

```suggestion
    const content = contentRef.current;
    if (content) {
      content.addEventListener(TYPEAHEAD_SEARCH, handleTypeaheadSearch);
      return () => content.removeEventListener(TYPEAHEAD_SEARCH, handleTypeaheadSearch);
    }
```

packages/react/menu/src/Menu.tsx
```diff
@@ -435,16 +437,62 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+    timerRef.current = window.setTimeout(() => {
+      setSearch('');
+    }, 1000);
+  }, []);
+
+  React.useEffect(() => {
+    return () => window.clearTimeout(timerRef.current);
+  }, []);
+
+  const handleTypeaheadSearch = React.useCallback(
```
jjenzz, line 466: Since this method is only used in the effect, we tend to just declare it in the effect so we don't need `React.useCallback`. Makes it clearer it's only used there too that way.

packages/react/menu/src/Menu.tsx
```diff
@@ -516,15 +565,21 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
                   style={{ outline: 'none', ...contentProps.style }}
-                  onKeyDownCapture={composeEventHandlers(
-                    contentProps.onKeyDownCapture,
-                    typeaheadProps.onKeyDownCapture
-                  )}
-                  // focus first/last item based on key pressed
                   onKeyDown={composeEventHandlers(contentProps.onKeyDown, (event) => {
-                    const content = contentRef.current;
+                    // perform typeahead search
+                    const modifierKey = event.ctrlKey || event.altKey || event.metaKey;
```
jjenzz, line 570: ❤️ love that you abstracted this into its own variable

packages/react/menu/src/Menu.tsx
```diff
@@ -553,9 +608,12 @@ const ITEM_NAME = 'MenuItem';
...
-type MenuItemOwnProps = Polymorphic.Merge<
-  Polymorphic.OwnProps<typeof MenuItemImpl>,
-  { onSelect?: (event: Event) => void }
+type MenuItemOwnProps = Omit<
+  Polymorphic.Merge<
+    Polymorphic.OwnProps<typeof MenuItemImpl>,
+    { onSelect?: (event: Event) => void }
+  >,
+  keyof MenuItemImplPrivateProps
 >;
```
jjenzz, line 617:
```suggestion
type MenuItemOwnProps = Polymorphic.Merge<
  Omit<Polymorphic.OwnProps<typeof MenuItemImpl>, keyof MenuItemImplPrivateProps>,
  { onSelect?: (event: Event) => void }
>;
```

Just means we're only omitting from the bit where they exist. Reduces some of the wrapping too hahaha, and you know how i feel about wrapping 😅 

Although, we probably don't need this anymore if we do the `onKeyCapture` thing.

packages/react/menu/src/Menu.tsx
```diff
@@ -435,16 +437,62 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
   const isPointerDownOutsideRef = React.useRef(false);
+  const timerRef = React.useRef(0);
+  const searchRef = React.useRef('');
+  const [isSearching, setIsSearching] = React.useState(false);
 
   const PortalWrapper = portalled ? Portal : React.Fragment;
   const ScrollLockWrapper = disableOutsideScroll ? RemoveScroll : React.Fragment;
 
+  // Reset `searchRef` 1 second after it was last updated
+  const setSearch = React.useCallback((search: string) => {
```
jjenzz, line 453: I know you copied this from before but do you mind if we rename this to `handleSearchChange` or something. I thought that when we were calling `setSearch` on line 469 that we were calling a state setter and I couldn't see that on existed, I got lost haha.

jjenzz (changes requested):
This was a nice and simple change :+1: realised we might actually be able to delete some code here tho after all that:

packages/react/menu/src/Menu.tsx
```diff
@@ -435,16 +437,62 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+    searchRef.current = search;
+    setIsSearching(search.length > 0);
+    window.clearTimeout(timerRef.current);
+    timerRef.current = window.setTimeout(() => {
+      setSearch('');
+    }, 1000);
```
jjenzz, line 459: nit 😛 

```suggestion
    timerRef.current = window.setTimeout(() => setSearch(''), 1000);
```

haha, me and my wrapping woes!

packages/react/menu/src/Menu.tsx
andy-hook: Haha, wrapping or not I think that just reads better! I'll apply it for the time being.

packages/react/menu/src/Menu.tsx
andy-hook: I did wonder if this was explicit enough. Will leave as is for now.

packages/react/menu/src/Menu.tsx
andy-hook: Yeah I thought this too, but we need to dispatch on `event.target` (menu item) and bubble it up to `content`

packages/react/menu/src/Menu.tsx
jjenzz: Okay ignore this, you're right, we did need this because we needed the custom event so we don't prevent consumer events... keep forgetting that part 🤯

packages/react/menu/src/Menu.tsx
jjenzz: Ah but why do we need to dispatch on `event.target`? I can't see `event.target` being referenced in `handleTypeaheadSearch`. Would dispatching on `event.currentTarget` suffice? 💭

jjenzz: In fact, if we can dispatch on `event.currentTarget` then we wouldn't need `event.stopPropagation` and looks like we wouldn't need the custom event then either? I feel like I must be missing something 🤔

packages/react/menu/src/Menu.tsx
jjenzz, line 191: Since we are probably going to keep this... I was going to suggest renaming this to just `onExclusiveKeyDown` since it is only fired when we're not searching so the `Search` part in the name was a bit confusing.

packages/react/menu/src/Menu.tsx
andy-hook: It needs to be `event.target` so that it is scoped correctly to the menu of origin. We can’t use standard `onKeyDown` events on `content` without stopping the source events propagation (which we try not to do) else they bubble all the way up the stack. Instead we are using a `customEvent` to give us the relationship between the item and menu of origin so that the `newItem` is taken from the single, correct collection.

andy-hook: It was confusing to think it through again. Because we've decided to not stop prop in the traditional way, what's actually happening is that the handlers will fire for every menu in the stack because `onKeyDown` will still bubble. So `event.target` is our way to qualify against that.

jjenzz: oh my, this is hurting my brain... can we get on a call quick so you can walk me through this?

jjenzz: Just trying to get my head around this one again and this is what my thinking was:

image

This appears to work okay but I have no idea if I'm catching all the edge-cases here. Does this cover it?

packages/react/menu/src/Menu.tsx
```diff
@@ -438,16 +439,57 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+  }, []);
+
+  React.useEffect(() => {
+    return () => window.clearTimeout(timerRef.current);
+  }, []);
+
+  React.useEffect(() => {
+    const handleTypeaheadSearch = ((event: CustomEvent<{ key: string }>) => {
+      event.stopPropagation();
+      handleSearchChange(searchRef.current + event.detail.key);
```
jjenzz, line 469: So this part has been bothering me a bit. I initially was going to suggest calling this after all the `const`s but then realised we can't do that because there is an implicit dependency on `searchRef.current` being updated first here so the `const nextMatch` can reference it.

The `handleSearchChange` seems to only exist so that it can call itself from the timeout and isn't being re-used anywhere so I was wondering what thoughts are on consolidating these two methods?

```jsx
const handleTypeaheadSearch = (key: string) => {
  const search = searchRef.current + key;
  const items = getItems().filter((item) => !item.disabled);
  const currentItem = document.activeElement;
  const currentMatch = items.find((item) => item.ref.current === currentItem)?.textValue;
  const values = items.map((item) => item.textValue);
  const nextMatch = getNextMatch(values, search, currentMatch);
  const newItem = items.find((item) => item.textValue === nextMatch)?.ref.current;

  (function updateSearch(search: string) {
    searchRef.current = search;
    setIsSearching(search.length > 0);
    window.clearTimeout(timerRef.current);
    timerRef.current = window.setTimeout(() => updateSearch(''), 1000);
  })(search);

  if (newItem) {
    /**
      * Imperative focus during keydown is risky so we prevent React's batching updates
      * to avoid potential bugs. See: https://github.com/facebook/react/issues/20332
      */
    setTimeout(() => (newItem as HTMLElement).focus());
  }
};
```

I ran this by @andy-hook already and we just wanted to check what your thoughts are @benoitgrelard really because we don't use IIFEs much.

benoitgrelard (changes requested):
First initial play with it, the biggest issue I found is this:

when there's currently a search in buffer, pressing right key or enter on a subtrigger should open instantly (because they're not characters) but right now it doesn't.

It seems to do the right thing for every other navigation keys (up/down/page up/page down/ home/end) just not the right key (left in RTL too I'm guessing) and enter key.

packages/react/menu/src/Menu.tsx
andy-hook: Tested and works great, such a simpler solution.

packages/react/menu/src/Menu.tsx
andy-hook: Went ahead and applied this – https://github.com/radix-ui/primitives/pull/662/commits/bb2f5f8f3026e811527eac47e3534e842c0b5be7 I personally prefer it

packages/react/menu/src/Menu.tsx
```diff
@@ -438,16 +437,49 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+    const nextMatch = getNextMatch(values, search, currentMatch);
+    const newItem = items.find((item) => item.textValue === nextMatch)?.ref.current;
+
+    (function updateSearch(value: string) {
+      searchRef.current = value;
+      setIsSearching(value.length > 0);
+      window.clearTimeout(timerRef.current);
+      if (searchRef.current !== '') {
+        timerRef.current = window.setTimeout(() => updateSearch(''), 1000);
+      }
```
jjenzz, line 467: Looks like we can use `value.length > 0` **or** `value !== ''` for `setIsSearching` and the timeout guard so they're consistent. I don't mind which one but having the two different approaches made me wonder if there was a difference between the two some how... Is there?

```suggestion
      const isSearching = value !== '';
      setIsSearching(isSearching);
      searchRef.current = value;
      window.clearTimeout(timerRef.current);
      if (isSearching) timerRef.current = window.setTimeout(() => updateSearch(''), 100);

Comment (andy-hook) [reply packages/react/menu/src/Menu.tsx:467]: I had it this way but am thinking of introducing a difference in the morning so wanted to remind myself! I also need to remember the single line if statements (I personally don't like them so it's a hard habit for me to break haha)

File: packages/react/menu/src/Menu.tsx [hunk 15]
```diff
@@ -601,7 +604,9 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
       onMouseUp={composeEventHandlers(props.onMouseUp, handleSelect)}
       onMouseLeave={composeEventHandlers(props.onMouseLeave, () => contentContext.onItemLeave())}
       onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
-        if (!disabled && SELECTION_KEYS.includes(event.key)) {
+        const isTypingAhead = contentContext.searchRef.current !== '' && event.key === ' ';
+        if (disabled || isTypingAhead) return;
```
benoitgrelard, line 608: I still find it odd to read this, this would make more sense to me to read:

```suggestion
        const isTypingAhead = contentContext.searchRef.current !== '';
        if (disabled || (isTypingAhead && event.key === ' ')) return;
```

What do you think?

packages/react/menu/src/Menu.tsx
```diff
@@ -683,6 +682,46 @@ const MenuItemImpl = React.forwardRef((props, forwardedRef) => {
...
+        aria-controls={context.contentId}
+        data-state={getOpenState(context.open)}
+        {...props}
+        ref={composeRefs(forwardedRef, context.onTriggerChange)}
+        onMouseMove={composeEventHandlers(props.onMouseMove, () => {
+          if (!props.disabled) context.onOpenChange(true);
+        })}
+        onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
+          const isTypingAhead = contentContext.searchRef.current !== '' && event.key === ' ';
+          if (props.disabled || isTypingAhead) return;
```
benoitgrelard, line 715: Same on this one if you agree.

benoitgrelard (commented):
Works really well now! 🎉 
Just a minor change I'd like, but I'll leave it up to you.

packages/react/menu/src/Menu.tsx
andy-hook: Agreed

benoitgrelard (approved):

PR #666, opened 2021-05-27
Add heuristics to contributing docs

jjenzz:
@andy-hook had a good idea that we should add something like this to help remind ourselves what our architectural goals are.

These are heuristics that Benoit, Pedro and I all followed while at a previous role together and they have worked well so I'm suggesting these as a starting point.

Feel free to suggest alternatives/extras 🙂


benoitgrelard (commented):
♥️

benoitgrelard (approved):

PR #669, opened 2021-05-28
[DismissableLayer] Remove `onInteractOutside` and rename `onFocusOutside` to `onFocusLeave` 🔥 

benoitgrelard:
Something I had discussed with @jjenzz, I regretted adding `onInteractOutside` as a convenience for the already 2 existing callbacks, as it's mostly unnecessary and gives a weird handler signature (both types of events) so I've decided to remove it.

Additionally, it dawned on me that `onFocusOutside` isn't actually accurate as it's not a focus event happening outside, but rather a focus event leaving a particular subtree so I'm renaming it to `onFocusLeave` which is much more accurate.

> Note for docs updates:

- 🔥 Remove `onInteractOutside` from all concerned components
- 🔥 Rename `onFocusOutside` to `onFocusLeave` on all concerned components

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -170,8 +170,7 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
     disableOutsidePointerEvents,
     onEscapeKeyDown,
     onPointerDownOutside,
-    onFocusOutside,
-    onInteractOutside,
+    onFocusLeave,
     disableOutsideScroll,
     portalled,
     ...contentProps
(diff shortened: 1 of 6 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -236,16 +235,15 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
                 },
                 { checkForDefaultPrevented: false }
               )}
-              onFocusOutside={composeEventHandlers(
-                onFocusOutside,
+              onFocusLeave={composeEventHandlers(
+                onFocusLeave,
                 (event) => {
-                  // When focus is trapped, a focusout event may still happen.
+                  // When focus is trapped, the focus may still leave temporarily.
```
andy-hook, line 241: 👌  👌

andy-hook (approved):
Appreciate these changes! I initially found the distinction a little confusing.

jjenzz (approved):

PR #671, opened 2021-05-28
[DismissableLayer] Re-add `onInteractOutside` and reimplement `onFocusOutside`

jjenzz:
Benoit and I had a chat about these and realised that the `onFocusOutside` we've added to `Menu` is probably all we really need in `DismissableLayer` so the name has returned as it fires whenever a focus outside occurs now and not just when focus leaves.

We also agreed to re-add the `onInteractOutside` handler as I've brought the two event types for `onPointerDownOutside` and `onFocusOutside` inline with each other now and the `onInteractOutside` is a useful convenience when consumers need to prevent closing on both of these.


packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -207,64 +219,63 @@ type PointerDownOutsideEvent = CustomEvent<{ originalEvent: MouseEvent | TouchEv
...
+
+function usePointerDownOutside(onPointerDownOutside?: (event: PointerDownOutsideEvent) => void) {
+  const handlePointerDownOutside = useCallbackRef(onPointerDownOutside) as EventListener;
+  const isPointerInsideReactTreeRef = React.useRef(false);
 
   React.useEffect(() => {
-    const handlePointerDown = (event: MouseEvent | TouchEvent) => {
-      const target = event.target as HTMLElement | null;
-      if (target && !isEventInside.current) {
+    const handlePointerDown = (event: PointerEvent) => {
```
benoitgrelard, line 228: 👀

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -207,64 +219,63 @@ type PointerDownOutsideEvent = CustomEvent<{ originalEvent: MouseEvent | TouchEv
...
  */
-function usePointerDownOutside(
-  onPointerDownOutsideProp?: (event: PointerDownOutsideEvent) => void
-) {
-  const onPointerDownOutside = useCallbackRef(onPointerDownOutsideProp);
-  const isEventInside = React.useRef(false);
+
+function usePointerDownOutside(onPointerDownOutside?: (event: PointerDownOutsideEvent) => void) {
+  const handlePointerDownOutside = useCallbackRef(onPointerDownOutside) as EventListener;
+  const isPointerInsideReactTreeRef = React.useRef(false);
```
benoitgrelard, line 225: 👍

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -207,64 +219,63 @@ type PointerDownOutsideEvent = CustomEvent<{ originalEvent: MouseEvent | TouchEv
...
 /**
- * Listens for when focus leaves a react subtree.
+ * Listens for when focus happens outside a react subtree.
  * Returns props to pass to the root (node) of the subtree we want to check.
  */
-function useFocusLeave(onFocusLeave?: (event: React.FocusEvent) => void) {
-  const timerRef = React.useRef<number>(0);
 
-  // Cleanup timer if an unmount occurs before onFocusCapture fires
+function useFocusOutside(onFocusOutside?: (event: FocusOutsideEvent) => void) {
```
benoitgrelard, line 255: I like the parallelism between the 2 hooks now, they're very similar.

benoitgrelard (approved):

PR #672, opened 2021-05-28
[Checkbox] Bug fixes 🔥 

### LINKED ISSUE #512: [Checkbox] Inconsistent behavior with plain label
### Current Behavior

When the `Checkbox` primitive is used with controlled state, is wrapped in a plain `<label>`, and has an `Indicator` with conditionally-rendered children, the checkbox's click handler competes with the label's native browser behavior and eventually becomes "stuck" when interacting with the indicator. 

### Expected behavior

Checkbox should not get stuck when interacting with the indicator.

### Reproducible example

https://codesandbox.io/s/crazy-sutherland-vxqoj?file=/src/App.js

### Suggested solution

N/A

### Additional context

This doesn't happen when using the `Label` primitive from radix-ui.

### Your environment

(any)

### LINKED ISSUE #609: [Label] `onClick` `stopPropagation()` not working inside `Label`
### Current Behavior

Calling `stopPropagation()` on an `onClick` event on a child of a `Label` still clicks the label and toggles the associated `Checkbox`.
EDIT: this also applies to calling `preventDefault()` on the `click` event.

### Expected behavior

For the `Label` to not receive a click event when a child stops propagation.

### Reproducible example

https://codesandbox.io/s/radix-label-stop-propagation-bug-2loxf

### Suggested solution

It seems like `Label` is not using the React event system for click events, so using the React event system should fix this issue.

### Additional context

Use case: a Label that contains an info link. When the link is clicked I want to stop propagation so the Checkbox isn't toggled.

### Your environment

| Radix Package(s) |     Label    |     0.0.11    |
| React            | n/a     |     17    |

### LINKED ISSUE #667: [Checkbox] onClick stopPropagation() not working inside Checkbox
### Current Behavior

Calling stopPropagation() on an onClick event on a Checkbox element does not actually stop event propagation.

### Expected behavior

Calling stopPropagation() should stop event propagation.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/checkbox-stoppropagation-doesnt-work-7xn4u?file=/src/App.js)

### Your environment

react 17.0.2 and @radix-ui/react-checkbox 0.0.15

jjenzz:
Cleans this up a bit to use the same `BubbleInput` approach that `Slider` uses and fixes a bunch of bugs in the process.

- Fixes #667
- Fixes #512 
- Removes additional tab stop on checkbox that was a regression from https://github.com/radix-ui/primitives/pull/650
- Improves accessibility when wrapped in a native `label` 🎉
- Updates `onCheckedChange(event)` to `onCheckedChange(checked: CheckedState)` (this follows the shape of all our other handlers, e.g. `onOpenChange`) 🔥  

https://user-images.githubusercontent.com/175330/120019306-30f6eb00-bfe0-11eb-900e-e38969dd4d11.mp4

### Accessibility

https://user-images.githubusercontent.com/175330/120191038-51f25280-c211-11eb-93df-e7bdd1093977.mov

Interesting that our custom checkbox with custom label also improves cursor position accuracy in VoiceOver.


benoitgrelard (changes requested):
There seems to be some issues with this.

If I have:
```tsx
<div onClick={() => alert('click bubble')}>
      <Checkbox className={rootClass}>
        <CheckboxIndicator className={indicatorClass} />
      </Checkbox>
    </div>
```

It happens straight on page load, and then also it doesn't seem to have resolved the double event that we see in #667

jjenzz:
@benoitgrelard Should all be working now 🙈

jjenzz:
@benoitgrelard This should be ready for a review now 🙏

packages/react/label/src/Label.tsx
```diff
@@ -94,16 +94,20 @@ const useLabelContext = (element?: HTMLElement | null) => {
   return context?.id;
 };
 
-function addLabelClickEventListener(label: HTMLSpanElement, element: HTMLElement) {
+function addLabelClickEventListener(label: HTMLSpanElement, control: HTMLElement) {
   const handleClick = (event: MouseEvent) => {
+    const isClickingControl = control.contains(event.target as Node);
+    // Ensure event was generated by a user action
+    // https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted
+    const isUserClick = event.isTrusted === true;
```
benoitgrelard, line 102: so AFAICT this is the only real addition right?
In this case, how does this case help us?
I would have thought that the check for whether we're already clicking on the control itself would be enough.

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -168,6 +158,33 @@ CheckboxIndicator.displayName = INDICATOR_NAME;
 
 /* ---------------------------------------------------------------------------------------------- */
 
+type BubbleCheckedProps = Omit<React.ComponentProps<'input'>, 'checked'> & {
```
benoitgrelard, line 161: nit: should this be?
```suggestion
type BubbleInputProps = Omit<React.ComponentProps<'input'>, 'checked'> & {
```

packages/react/label/src/Label.tsx
benoitgrelard: Is it specifically so we don't catch the one that is dispatched from `BubbleInput`?

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -159,6 +148,7 @@ const CheckboxIndicator = React.forwardRef((props, forwardedRef) => {
         {...indicatorProps}
         as={as}
         ref={forwardedRef}
+        style={{ pointerEvents: 'none', ...props.style }}
```
benoitgrelard, line 151: This is the fix for #512 right?

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -54,72 +55,60 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
...
+        checked={checked}
         required={required}
         disabled={disabled}
-        value={value}
         style={{
           position: 'absolute',
           pointerEvents: 'none',
           opacity: 0,
           margin: 0,
+          marginLeft: -(buttonSize?.width || 0),
```
benoitgrelard, line 104: This is to pull it aligned left-wise without having to rely on `position: relative` on the parent and `left: 0` on the child?

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -54,72 +55,60 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
...
+        data-state={getState(checked)}
+        data-disabled={disabled ? '' : undefined}
+        disabled={disabled}
+        value={value}
+        {...checkboxProps}
+        as={as}
+        ref={composedRefs}
+        onClick={composeEventHandlers(props.onClick, (event) => {
+          setChecked((prevChecked) => (prevChecked === 'indeterminate' ? true : !prevChecked));
+          hasConsumerStoppedPropagationRef.current = event.isPropagationStopped();
```
benoitgrelard, line 87: Oh I've never seen this before! Is this a React thing, I'm not a aware of a native method like that.

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -54,72 +55,60 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
...
+          // stop propagation from the button so that we only propagate one click event (from the input)
+          event.stopPropagation();
```
benoitgrelard, line 89: You may need to run me through this routine here, because I don't think I'm following, it seems we always stop propagation…

packages/react/label/src/Label.tsx
jjenzz: Yep, we only want to make sure that when it's a user clicking the label, we click the button for them.

packages/react/checkbox/src/Checkbox.tsx
jjenzz: Oops, initially called the component `BubbleChecked` and then renamed for consistency with `Slider`. Which do you prefer actually?

packages/react/checkbox/src/Checkbox.tsx
jjenzz: Yep, so... I moved the `input` **after** the button so that the label references the `button` state with screen readers. This is what gave us the accessibility improvements with native label.

Moving it after made it sit next to the button though so this pulls it back without the need for any `position: relative` stuff.

packages/react/checkbox/src/Checkbox.tsx
benoitgrelard: I like `BubbleInput` and then it's also the same as in `Slider`

packages/react/checkbox/src/Checkbox.tsx
benoitgrelard: yeah that's what I thought, nice! 👍

packages/react/checkbox/src/Checkbox.tsx
jjenzz: Yeah, it's a React thing :+1:

benoitgrelard (approved):

PR #678, opened 2021-06-01
[Slider] Prevent form change event on initial render

jjenzz:
- Adds `usePrevious` to prevent `BubbleInput` from bubbling change on initial render
- Adds an `isFormControl` check to `Slider` and `Checkbox` so that `BubbleInput` only renders if these controls are used in forms
- Added stories to show form change event working (multiple thumb version was strangely satisfying to play with 😄 ).

https://user-images.githubusercontent.com/175330/120327568-cdbbd000-c2e1-11eb-8ef6-998e1cf7746b.mp4

### DIFF
--- a/packages/react/checkbox/src/Checkbox.tsx
+++ b/packages/react/checkbox/src/Checkbox.tsx
@@ -57,10 +57,11 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
   } = props;
   const [button, setButton] = React.useState<HTMLButtonElement | null>(null);
   const composedRefs = useComposedRefs(forwardedRef, (node) => setButton(node));
-  const buttonSize = useSize(button);
   const labelId = useLabelContext(button);
   const labelledBy = ariaLabelledby || labelId;
   const hasConsumerStoppedPropagationRef = React.useRef(false);
+  // We set this to true by default so that events bubble to forms without JS (SSR)
+  const isFormControl = button ? button.closest('form') : true;
   const [checked = false, setChecked] = useControllableState({
     prop: checkedProp,
     defaultProp: defaultChecked,
@@ -84,28 +85,27 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
         ref={composedRefs}
         onClick={composeEventHandlers(props.onClick, (event) => {
           setChecked((prevChecked) => (prevChecked === 'indeterminate' ? true : !prevChecked));
-          hasConsumerStoppedPropagationRef.current = event.isPropagationStopped();
-          // Stop propagation from the button so that we only propagate one click event (from the input).
-          // We propagate changes from an input so that form events reflect checkbox updates.
-          if (!hasConsumerStoppedPropagationRef.current) event.stopPropagation();
+          if (isFormControl) {
+            hasConsumerStoppedPropagationRef.current = event.isPropagationStopped();
+            // if checkxbox is in a form, stop propagation from the button so that we only propagate
+            // one click event (from the input). We propagate changes from an input so that native
+            // form validation works and form events reflect checkbox updates.
+            if (!hasConsumerStoppedPropagationRef.current) event.stopPropagation();
+          }
         })}
       />
-      <BubbleInput
-        stoppedPropagation={hasConsumerStoppedPropagationRef.current}
-        name={name}
-        value={value}
-        checked={checked}
-        required={required}
-        disabled={disabled}
-        style={{
-          position: 'absolute',
-          pointerEvents: 'none',
-          opacity: 0,
-          margin: 0,
-          transform: 'translateX(-100%)',
-          ...buttonSize,
-        }}
-      />
+      {isFormControl && (
+        <BubbleInput
+          control={button}
+          stoppedPropagation={hasConsumerStoppedPropagationRef.current}
+          name={name}
+          value={value}
+          checked={checked}
+          required={required}
+          disabled={disabled}
+          style={{ transform: 'translateX(-100%)' }}
+        />
+      )}
     </CheckboxProvider>
   );
 }) as CheckboxPrimitive;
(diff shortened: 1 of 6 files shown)


packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -84,28 +85,27 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
         ref={composedRefs}
         onClick={composeEventHandlers(props.onClick, (event) => {
           setChecked((prevChecked) => (prevChecked === 'indeterminate' ? true : !prevChecked));
-          hasConsumerStoppedPropagationRef.current = event.isPropagationStopped();
-          // Stop propagation from the button so that we only propagate one click event (from the input).
-          // We propagate changes from an input so that form events reflect checkbox updates.
-          if (!hasConsumerStoppedPropagationRef.current) event.stopPropagation();
+          if (isFormControl) {
+            hasConsumerStoppedPropagationRef.current = event.isPropagationStopped();
+            // if checkxbox is in a form, stop propagation from the button so that we only propagate
```
jjenzz, line 90:
```suggestion
            // if checkbox is in a form, stop propagation from the button so that we only propagate
```

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -84,28 +85,27 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
...
+      {isFormControl && (
+        <BubbleInput
+          control={button}
+          stoppedPropagation={hasConsumerStoppedPropagationRef.current}
+          name={name}
+          value={value}
+          checked={checked}
+          required={required}
+          disabled={disabled}
+          style={{ transform: 'translateX(-100%)' }}
```
jjenzz, line 106: I kept this style up here because it is only needed when `BubbleInput` is rendered **after** the button. It seems this level should be in control of that style in case we move it before the button later.

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -57,10 +57,11 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
   } = props;
   const [button, setButton] = React.useState<HTMLButtonElement | null>(null);
   const composedRefs = useComposedRefs(forwardedRef, (node) => setButton(node));
-  const buttonSize = useSize(button);
```
jjenzz, line 60: Moved this into `BubbleInput` because it's only needed if we render `BubbleInput`

packages/react/slider/src/Slider.tsx
```diff
@@ -676,7 +679,7 @@ const BubbleInput = (props: React.ComponentProps<'input'>) => {
    * Adding the `value` will cause React to consider the programatic
    * dispatch a duplicate and it will get swallowed.
    */
-  return <input style={{ display: 'none' }} {...inputProps} ref={ref} />;
+  return <input style={{ display: 'none' }} {...inputProps} ref={ref} defaultValue={value} />;
```
jjenzz, line 682: Added `defaultValue` here to fix this issue:


https://user-images.githubusercontent.com/175330/120328686-fc867600-c2e2-11eb-8c39-db8b109ac472.mp4

The form change event was firing but these inputs had no initial value so our `legend` gets updated with an undefined value.

yarn.lock
```diff
@@ -3579,7 +3579,6 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "@radix-ui/react-polymorphic@workspace:packages/react/polymorphic"
   dependencies:
-    "@babel/runtime": ^7.13.10
```
jjenzz, line 3582: Looks like we forgot to `yarn` in this PR https://github.com/radix-ui/primitives/pull/670

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -57,10 +57,11 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
...
   const labelId = useLabelContext(button);
   const labelledBy = ariaLabelledby || labelId;
   const hasConsumerStoppedPropagationRef = React.useRef(false);
+  // We set this to true by default so that events bubble to forms without JS (SSR)
+  const isFormControl = button ? button.closest('form') : true;
```
benoitgrelard, line 64: We could even do:
```suggestion
  const isFormControl = button ? Boolean(button.form) : true;
```
because controls will have a `form` property associated to them when inside a form.

packages/react/checkbox/src/Checkbox.tsx
benoitgrelard: I agree with the place the code lives in but I think it would be nice to add a comment perhaps now that it's disconnected from the rest of the styles.

packages/react/slider/src/Slider.tsx
```diff
@@ -79,11 +80,13 @@ const Slider = React.forwardRef((props, forwardedRef) => {
...
   } = props;
-  const sliderRef = React.useRef<HTMLSpanElement>(null);
-  const composedRefs = useComposedRefs(forwardedRef, sliderRef);
+  const [slider, setSlider] = React.useState<HTMLSpanElement | null>(null);
+  const composedRefs = useComposedRefs(forwardedRef, (node) => setSlider(node));
   const thumbRefs = React.useRef<SliderContextValue['thumbs']>(new Set());
   const valueIndexToChangeRef = React.useRef<number>(0);
   const isHorizontal = orientation === 'horizontal';
+  // We set this to true by default so that events bubble to forms without JS (SSR)
+  const isFormControl = slider ? slider.closest('form') : true;
```
benoitgrelard, line 89: I wonder if what I was suggesting above would work here though, cause our `slider` element here is not a regular control right?

packages/react/checkbox/src/Checkbox.tsx
jjenzz: I did exactly this initially, but `Slider` isn't a form control, it's a `span` so I went with this to keep the two approaches consistent. Should I do it differently here?

packages/react/slider/src/Slider.tsx
jjenzz: Haha exactly...

benoitgrelard (approved):

PR #682, opened 2021-06-03
[ContextMenu][DropdownMenu] Add submenu support

andy-hook:
Closes https://github.com/radix-ui/primitives/issues/383

This PR adds submenu support to `ContextMenu` and `DropdownMenu `

While there are variations in submenu behaviour across desktop and web, we've generally attempted to stick close to the native macos implementation. [See original discussion here](https://github.com/radix-ui/primitives/issues/383)

### Demo

https://user-images.githubusercontent.com/11708259/120844660-a37b4400-c567-11eb-9b08-927b43f95857.mp4

#### TODO
- [X] Add underlying support to `Menu` (https://github.com/radix-ui/primitives/pull/627)
- [x] Adjust typeahead to support nesting (https://github.com/radix-ui/primitives/pull/662)
- [x] Expose in `ContextMenu` and `DropdownMenu` (https://github.com/radix-ui/primitives/pull/629)
- [x] Add new `TriggerItem` component to resolve static type issues (https://github.com/radix-ui/primitives/pull/683)
- [x] Prevent submenu dismissing when clicking on trigger (#684)
- [x] Apply pointer timeout grace on transit to submenu (#688)
- [x] Update versions
- [x] Improve screen reader accessibility (#692 #694)
- [x] Sticky focus when entering and leaving a sub trigger quickly https://github.com/radix-ui/primitives/pull/682#issuecomment-854983610 (see #691)
- [x] Improve pointer grace (#691)
- [x] Regression test
- [x] Screen reader test
- [x] PO review


andy-hook:
I've added two issues to the TODOs

Prevent submenu dismissing when clicking on trigger:

https://user-images.githubusercontent.com/11708259/120673735-588bfe80-c48b-11eb-983d-c625f262ee3c.mp4

Fix sticky open state when dev tools have focus:

https://user-images.githubusercontent.com/11708259/120673765-60e43980-c48b-11eb-980f-2d55f1cb05c5.mp4

andy-hook:
I'm so pleased with how this feels now, especially after https://github.com/radix-ui/primitives/pull/688 👏 , I did notice one very minor detail and that's when we enter and leave within the 100ms window of the pointer grace, the focus sticks:

https://user-images.githubusercontent.com/11708259/120859541-42aa3680-c57c-11eb-86e1-18fed4ad12a4.mp4

benoitgrelard:
> I'm so pleased with how this feels now, especially after #688 👏 , I did notice one very minor detail and that's when we enter and leave within the 100ms window of the pointer grace, the focus sticks:
> 
>  CleanShot.2021-06-04.at.21.26.14.mp4

ah yep good catch.

andy-hook:
> > I'm so pleased with how this feels now, especially after #688 👏 , I did notice one very minor detail and that's when we enter and leave within the 100ms window of the pointer grace, the focus sticks:
> > CleanShot.2021-06-04.at.21.26.14.mp4
> 
> ah yep good catch.

I don't think it was introduced in that change, but just exposed by this new state it can be in. Will take a look during a full reg test and cross browser bizzo.

andy-hook:
Did a final round of browser testing, here is what I found.


You cannot enter a submenu with keyboard if it was originally opened via the pointer

https://user-images.githubusercontent.com/11708259/121864377-bd99ec80-ccf4-11eb-9b87-b2437a392460.mp4


If you navigate to a submenu > two levels deep and click the root trigger, the open state sticks

https://user-images.githubusercontent.com/11708259/121864401-c68abe00-ccf4-11eb-82ae-4ae9f3650a0d.mp4


With touch inputs there is some odd behaviour when layered click targets overlap e.g. after a collision adjustment

https://user-images.githubusercontent.com/11708259/121864338-b4108480-ccf4-11eb-8e7d-c914ae84de56.mp4

packages/react/menu/src/Menu.tsx
```diff
@@ -299,78 +591,229 @@ const ITEM_DEFAULT_TAG = 'div';
...
-    disabled,
-  });
+  const composedRefs = useComposedRefs(forwardedRef, ref);
+  const handleSelectProp = useCallbackRef(onSelect);
 
   const handleSelect = () => {
     const menuItem = ref.current;
     if (!disabled && menuItem) {
       const itemSelectEvent = new Event(ITEM_SELECT, { bubbles: true, cancelable: true });
       menuItem.dispatchEvent(itemSelectEvent);
```
benoitgrelard, line 615: I think we can simplify this here by doing this:

```suggestion
      const itemSelectEvent = new Event(ITEM_SELECT, { bubbles: true, cancelable: true });
      menuItem.addEventListener(ITEM_SELECT, handleSelectProp, { once: true });
      menuItem.dispatchEvent(itemSelectEvent);
```

packages/react/menu/src/Menu.tsx
```diff
@@ -299,78 +591,229 @@ const ITEM_DEFAULT_TAG = 'div';
...
   React.useEffect(() => {
     const menuItem = ref.current;
     if (menuItem) {
-      const handleItemSelect = (event: Event) => onSelect?.(event);
-      menuItem.addEventListener(ITEM_SELECT, handleItemSelect);
-      return () => menuItem.removeEventListener(ITEM_SELECT, handleItemSelect);
+      menuItem.addEventListener(ITEM_SELECT, handleSelectProp);
+      return () => menuItem.removeEventListener(ITEM_SELECT, handleSelectProp);
+    }
+  }, [handleSelectProp]);
```
benoitgrelard, line 627: And removing this entire effect.

benoitgrelard (commented):
I've done one more thorough review of the whole change overall and I'm very very happy with it!
I've found just one thing we could do to simplify things a little.

👏

packages/react/menu/src/Menu.tsx
andy-hook: Yeah def, this is much cleaner.

packages/react/menu/src/Menu.tsx
```diff
@@ -94,65 +211,157 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
 
-    /**
-     * Whether keyboard navigation should loop around
-     * @defaultValue false
-     */
-    loop?: RovingFocusGroupOwnProps['loop'];
+type MenuSubContentOwnProps = Omit<
+  Polymorphic.OwnProps<typeof MenuContentImpl>,
+  | keyof MenuContentImplPrivateProps
+  | 'align'
```
andy-hook, line 256: Small detail but are we sure we should be omitting `align` here? this should be user adjustable for submenus right?

benoitgrelard: No I think that's correct, because we hardcode `start` a bit below as you can see.
That's because that's how sub-menus should be positioned, that said it's also temporary until we redo more specific positioning for menus in general.

But yeah all in all that's correct I think.

PR #683, opened 2021-06-03
[ContextMenu][DropdownMenu] Add `TriggerItem` part

jjenzz:
We realised that our initial approach of merging the sub trigger part with the trigger part for these would cause `ref` and TS issues. See original discussion here https://github.com/radix-ui/primitives/pull/629#discussion_r643042075

This exposes a new `TriggerItem` part for each component that should be used when rendering an item as a sub menu trigger.

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -106,48 +106,15 @@ ContextMenu.displayName = CONTEXT_MENU_NAME;
  * -----------------------------------------------------------------------------------------------*/
 
 const TRIGGER_NAME = 'ContextMenuTrigger';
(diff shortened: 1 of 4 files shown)


packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -168,7 +139,9 @@ const ContextMenuRootTrigger = React.forwardRef((props, forwardedRef) => {
       />
     </>
   );
-}) as ContextMenuRootTriggerPrimitive;
+}) as ContextMenuTriggerPrimitive;
+
+ContextMenuTrigger.displayName = TRIGGER_NAME;
```
benoitgrelard, line 144: 👀

benoitgrelard (approved):

andy-hook (approved):

PR #685, opened 2021-06-03
[RadioGroup] Add `BubbleInput` 🔥 

jjenzz:
- Fixes `event.stopPropagation`
- Prevents duplicate click events from firing
- Changes `onValueChange` shape to `onValueChange?(value: string): void` - Finally matches docs 🔥
- Converts `Radio` to a controlled component only because we only use it in `RadioGroup` which controls it. This allowed me to remove `onCheckedChange` from `Radio` because `onCheckedChange` never actually fires with `false` so the name confused me. We just `setValue(props.value)` when a radio is clicked.

https://user-images.githubusercontent.com/175330/120681102-efa88480-c492-11eb-9eae-c2f94555f34d.mp4

### DIFF
--- a/packages/react/radio-group/src/Radio.tsx
+++ b/packages/react/radio-group/src/Radio.tsx
@@ -4,6 +4,7 @@ import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { createContext } from '@radix-ui/react-context';
 import { useControllableState } from '@radix-ui/react-use-controllable-state';
 import { useSize } from '@radix-ui/react-use-size';
+import { usePrevious } from '@radix-ui/react-use-previous';
 import { Presence } from '@radix-ui/react-presence';
 import { Primitive } from '@radix-ui/react-primitive';
 import { useLabelContext } from '@radix-ui/react-label';
@@ -22,9 +23,7 @@ type RadioOwnProps = Polymorphic.Merge<
   Polymorphic.OwnProps<typeof Primitive>,
   {
     checked?: boolean;
-    defaultChecked?: boolean;
     required?: InputDOMProps['required'];
-    onCheckedChange?: InputDOMProps['onChange'];
   }
 >;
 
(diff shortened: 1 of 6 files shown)


packages/react/radio-group/src/Radio.tsx
```diff
@@ -155,6 +140,49 @@ RadioIndicator.displayName = INDICATOR_NAME;
 
 /* ---------------------------------------------------------------------------------------------- */
 
+type BubbleInputProps = Omit<React.ComponentProps<'input'>, 'checked'> & {
+  checked: boolean;
+  control: HTMLElement | null;
+  stoppedPropagation: boolean;
+};
+
+const BubbleInput = (props: BubbleInputProps) => {
```
jjenzz, line 149: This is exactly the same as `Switch` and mostly the same as `Checkbox` which just handles `indeterminate` as well. I dunno if we wanna start thinking about a re-useable thing for this in a new PR?

benoitgrelard: Yeah I think we should probably look at it next.

packages/react/radio-group/src/Radio.tsx
```diff
@@ -39,74 +38,60 @@ const Radio = React.forwardRef((props, forwardedRef) => {
...
+        ref={composedRefs}
+        onClick={composeEventHandlers(props.onClick, (event) => {
+          if (isFormControl) {
+            hasConsumerStoppedPropagationRef.current = event.isPropagationStopped();
+            // if radio is in a form, stop propagation from the button so that we only propagate
+            // one click event (from the input). We propagate changes from an input so that native
+            // form validation works and form events reflect radio updates.
+            if (!hasConsumerStoppedPropagationRef.current) event.stopPropagation();
+          }
         })}
```
benoitgrelard, line 85: It is a little odd how this is not the source of truth anymore no?

jjenzz: you mean because it doesn't call `onCheckedChange` anymore? I thought the same and was waiting to hear your thoughts here because this would literally just do `onCheckedChange(true)`. So, I thought about calling it `onCheck` or something, I dunno... how do you think we should handle this? It's weird.

benoitgrelard: Yeah exactly, I mean when compared with for example `Checkbox` which is driven from the change event on the actual input.

benoitgrelard (approved):
Much nicer 👍

PR #686, opened 2021-06-03
[Menu] Disable mouse open when focused in dev tool

andy-hook:
Leans on the `focus` event to prevent the open state when focused inside dev tool.

Not sure how I feel about it, or whether it's really necessary to do so.

Before:

https://user-images.githubusercontent.com/11708259/120682680-a3f6da80-c494-11eb-8598-20f574eda4f9.mp4

After:

https://user-images.githubusercontent.com/11708259/120682704-abb67f00-c494-11eb-87b7-4ea14665a86b.mp4


benoitgrelard (commented):
Yeah I'm not sure we should really do anything about it.
I don't love the idea of seeing some defensive code on our side to take care of it.
This is a common problem with devtools, that's why they even have a command to get around it: `> Emulate a focused page`

andy-hook:
> This is a common problem with devtools, that's why they even have a command to get around it: `> Emulate a focused page`

Oh that's awesome, yeah if the tooling resolves this then there's totally no need for it in our code.

PR #687, opened 2021-06-04
[Checkbox][Switch][RadioGroup] Always emit `BubbleInput` event

jjenzz:
Makes sure we always emit the `BubbleInput` event even if propagation has been stopped. Without this the underlying checkbox/radio wouldn't become checked. We want it to check, we just don't want the event to propagate.

Also adds `defaultChecked` attributes to ensure the default state is submitted with forms.

### DIFF
--- a/packages/react/checkbox/src/Checkbox.tsx
+++ b/packages/react/checkbox/src/Checkbox.tsx
@@ -73,7 +73,7 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
       <Primitive
         type="button"
         role="checkbox"
-        aria-checked={checked === 'indeterminate' ? 'mixed' : checked}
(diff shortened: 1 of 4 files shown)


packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -97,7 +97,7 @@ const Checkbox = React.forwardRef((props, forwardedRef) => {
       {isFormControl && (
         <BubbleInput
           control={button}
-          stoppedPropagation={hasConsumerStoppedPropagationRef.current}
+          bubbles={!hasConsumerStoppedPropagationRef.current}
```
jjenzz, line 100: I renamed this prop again because it was bothering me that the prop name described what was happening externally. The component shouldn't care what makes it bubble or not, we just want to tell it whether to bubble.

benoitgrelard (approved):

PR #688, opened 2021-06-04
[ContextMenu][DropdownMenu] Add pointer grace for navigation to sub-menus

benoitgrelard:
This PR adds 2 things:
- adds 100ms delay to before opening a submenu so skimming over them isn't jarring (100ms is what native seems to do here, we've measured in a video recording)
- adds pointer grace for sub-menu navigation. This is a combination of 2 things:
  - a grace duration of 300ms where we check for a specific area (see below)
  - an area duration where we check the mouse pointer is in a specific triangle area

https://user-images.githubusercontent.com/1539897/120825726-95222d80-c551-11eb-84cc-588df78c561d.mov

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
@@ -473,7 +473,7 @@ export const Chromatic = () => {
           </DropdownMenuItem>
           <DropdownMenuSeparator className={separatorClass} />
           <DropdownMenu open>
-            <DropdownMenuTrigger className={subTriggerClass}>Submenu →</DropdownMenuTrigger>
+            <DropdownMenuTriggerItem className={subTriggerClass}>Submenu →</DropdownMenuTriggerItem>
             <DropdownMenuContent className={contentClass} sideOffset={12} avoidCollisions={false}>
               <DropdownMenuItem className={itemClass} onSelect={() => console.log('one')}>
                 One
@@ -484,7 +484,9 @@ export const Chromatic = () => {
               </DropdownMenuItem>
               <DropdownMenuSeparator className={separatorClass} />
               <DropdownMenu open>
-                <DropdownMenuTrigger className={subTriggerClass}>Submenu →</DropdownMenuTrigger>
+                <DropdownMenuTriggerItem className={subTriggerClass}>
+                  Submenu →
+                </DropdownMenuTriggerItem>
                 <DropdownMenuContent
                   className={contentClass}
                   sideOffset={12}
@@ -540,7 +542,7 @@ export const Chromatic = () => {
           </DropdownMenuItem>
           <DropdownMenuSeparator className={separatorClass} />
           <DropdownMenu open>
-            <DropdownMenuTrigger className={subTriggerClass}>Submenu →</DropdownMenuTrigger>
+            <DropdownMenuTriggerItem className={subTriggerClass}>Submenu →</DropdownMenuTriggerItem>
             <DropdownMenuContent
               className={contentClass}
               sideOffset={12}
@@ -556,7 +558,9 @@ export const Chromatic = () => {
               </DropdownMenuItem>
               <DropdownMenuSeparator className={separatorClass} />
               <DropdownMenu open>
-                <DropdownMenuTrigger className={subTriggerClass}>Submenu →</DropdownMenuTrigger>
+                <DropdownMenuTriggerItem className={subTriggerClass}>
+                  Submenu →
+                </DropdownMenuTriggerItem>
                 <DropdownMenuContent
                   className={contentClass}
                   sideOffset={12}
@@ -615,7 +619,9 @@ export const Chromatic = () => {
             </DropdownMenuItem>
             <DropdownMenuSeparator className={separatorClass} />
             <DropdownMenu open>
-              <DropdownMenuTrigger className={subTriggerClass}>Submenu →</DropdownMenuTrigger>
+              <DropdownMenuTriggerItem className={subTriggerClass}>
+                Submenu →
+              </DropdownMenuTriggerItem>
               <DropdownMenuContent className={contentClass} sideOffset={12} avoidCollisions={false}>
                 <DropdownMenuItem className={itemClass} onSelect={() => console.log('one')}>
                   One
@@ -626,7 +632,9 @@ export const Chromatic = () => {
                 </DropdownMenuItem>
                 <DropdownMenuSeparator className={separatorClass} />
                 <DropdownMenu open>
-                  <DropdownMenuTrigger className={subTriggerClass}>Submenu →</DropdownMenuTrigger>
+                  <DropdownMenuTriggerItem className={subTriggerClass}>
+                    Submenu →
+                  </DropdownMenuTriggerItem>
                   <DropdownMenuContent
                     className={contentClass}
                     sideOffset={12}
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -717,13 +717,13 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
   const context = useMenuContext(SUB_TRIGGER_NAME);
   const contentContext = useMenuContentContext(SUB_TRIGGER_NAME);
   const contentRectRef = React.useRef<ClientRect | undefined>();
-  const openTimerRef = React.useRef<number | undefined>(undefined);
+  const openTimerRef = React.useRef<number | null>(null);
   const { onPointerGraceAreaChange } = contentContext;
   const pointerGraceDurationTimer = React.useRef(0);
 
   const clearOpenTimer = React.useCallback(() => {
-    window.clearTimeout(openTimerRef.current);
-    openTimerRef.current = undefined;
+    if (openTimerRef.current) window.clearTimeout(openTimerRef.current);
+    openTimerRef.current = null;
   }, []);
 
   React.useEffect(() => clearOpenTimer, [clearOpenTimer]);
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -756,7 +756,7 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
           const contentRect = context.content?.getBoundingClientRect();
           if (contentRect) {
             contentContext.onPointerGraceAreaChange([
-              { x: event.pageX, y: event.pageY },
+              { x: event.clientX, y: event.clientY },
               { x: contentRect.left, y: contentRect.top },
               { x: contentRect.left, y: contentRect.bottom },
             ]);
@@ -1029,7 +1029,7 @@ function isPointInTriangle(point: Point, a: Point, b: Point, c: Point) {
 
 function isMouseMovingToSubmenu(event: React.MouseEvent, triangle: Triangle | null) {
   if (triangle) {
-    const cursorPos = { x: event.pageX, y: event.pageY };
+    const cursorPos = { x: event.clientX, y: event.clientY };
     const isCursorInTriangle = isPointInTriangle(cursorPos, ...triangle);
     if (isCursorInTriangle) return true;
   }
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -701,6 +701,16 @@ type MenuSubTriggerPrimitive = Polymorphic.ForwardRefComponent<
 const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
   const context = useMenuContext(SUB_TRIGGER_NAME);
   const contentContext = useMenuContentContext(SUB_TRIGGER_NAME);
+  const openTimersRef = React.useRef<number[]>([]);
+
+  React.useEffect(() => {
+    const timers = openTimersRef.current;
+    return () => {
+      timers.map(window.clearTimeout);
+      openTimersRef.current = [];
+    };
+  }, []);
+
   return context.isSubmenu ? (
     <MenuAnchor as={Slot}>
       <MenuItemImpl
@@ -711,7 +721,13 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
         {...props}
         ref={composeRefs(forwardedRef, context.onTriggerChange)}
         onMouseMove={composeEventHandlers(props.onMouseMove, () => {
-          if (!props.disabled) context.onOpenChange(true);
+          if (!props.disabled && !context.open) {
+            openTimersRef.current.push(window.setTimeout(() => context.onOpenChange(true), 100));
+          }
+        })}
+        onMouseLeave={composeEventHandlers(props.onMouseLeave, () => {
+          openTimersRef.current.map(window.clearTimeout);
+          openTimersRef.current = [];
         })}
         onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
           const isTypingAhead = contentContext.searchRef.current !== '';
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -170,10 +170,10 @@ const [CollectionSlot, CollectionItemSlot, useCollection] = createCollection<
 >();
 
 type MenuContentContextValue = {
-  onItemLeave(): void;
+  onItemEnter(event: React.MouseEvent): void;
+  onItemLeave(event: React.MouseEvent): void;
   searchRef: React.RefObject<string>;
   onPointerGraceAreaChange(area: Triangle | null): void;
-  isMouseMovingToSubmenu(event: React.MouseEvent): boolean;
 };
 const [MenuContentProvider, useMenuContentContext] = createContext<MenuContentContextValue>(
   CONTENT_NAME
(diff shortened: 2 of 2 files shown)


packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
```diff
@@ -473,7 +473,7 @@ export const Chromatic = () => {
           </DropdownMenuItem>
           <DropdownMenuSeparator className={separatorClass} />
           <DropdownMenu open>
-            <DropdownMenuTrigger className={subTriggerClass}>Submenu →</DropdownMenuTrigger>
+            <DropdownMenuTriggerItem className={subTriggerClass}>Submenu →</DropdownMenuTriggerItem>
```
benoitgrelard, line 476: A bunch of these were not updated, I just realised because Chromatic flagged it.

andy-hook: Ah nice, I just sat down to look at it but you already did 👍

packages/react/menu/src/Menu.tsx
```diff
@@ -456,6 +459,13 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
             contentRef.current?.focus();
             setCurrentItemId(null);
           }, [])}
+          onPointerGraceAreaChange={React.useCallback(
+            (area) => (pointerGraceAreaRef.current = area),
+            []
+          )}
+          isMouseMovingToSubmenu={(event) =>
```
jjenzz, line 466: I was expecting this to be named something like `onMouseMoveToSubmenu` and then I saw it returns a boolean. We don't usually do this so I tried to work out why we're suddenly needing to now...

It feels like another case where we might be better off just passing the ref through context and calling the `isMouseMovingToSubMenu` utility inline. Or preferably something like this:

```jsx
// new handler fired from `onMouseMove`
onItemEnter={React.useCallback((event) => {
	if (isMouseMovingToSubmenu(event, pointerGraceAreaRef.current)) {
		event.preventDefault();
	}
}, []))
onItemLeave={React.useCallback((event) => {
	if (isMouseMovingToSubmenu(event, pointerGraceAreaRef.current)) {
		event.preventDefault();
	}
}, [])}
```

This reads more like English to me, like, when an item is entered, if we're moving to submenu then do nothing and vice versa.

I'm not sure if that would work though. Thoughts?

packages/react/menu/src/Menu.tsx
```diff
@@ -710,8 +743,31 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
         data-state={getOpenState(context.open)}
         {...props}
         ref={composeRefs(forwardedRef, context.onTriggerChange)}
-        onMouseMove={composeEventHandlers(props.onMouseMove, () => {
-          if (!props.disabled) context.onOpenChange(true);
+        onMouseMove={composeEventHandlers(props.onMouseMove, (event) => {
+          if (contentContext.isMouseMovingToSubmenu(event)) return;
+          if (!props.disabled && !context.open) {
+            openTimersRef.current.push(window.setTimeout(() => context.onOpenChange(true), 100));
```
jjenzz, line 749: I'm really struggling to work out why we need an array of open timers, what's the reasoning?

packages/react/menu/src/Menu.tsx
```diff
@@ -956,6 +1012,30 @@ function getNextMatch(values: string[], search: string, currentMatch?: string) {
...
+
+function isPointInTriangle(point: Point, a: Point, b: Point, c: Point) {
+  const A = getTriangleArea(a, b, c);
+  const A1 = getTriangleArea(point, b, c);
+  const A2 = getTriangleArea(a, point, c);
+  const A3 = getTriangleArea(a, b, point);
+  return A === A1 + A2 + A3;
+}
+
+function isMouseMovingToSubmenu(event: React.MouseEvent, triangle: Triangle | null) {
```
jjenzz, line 1030: In my mind the name `isMouseMovingToSubmenu` could be `true` under broader terms like just leaving the sub trigger and not hovering another sibling item so I was going to suggest renaming it to `isPointerWithinGraceArea` and then I saw we already have that.

So now, I'm actually wondering why we need this abstraction at all... can't we just call `isPointInTriangle` (maybe rename to `isPointInGraceArea` for consistency) instead?

```jsx
const pointerPos = { x: event.clientX, y: event.clientY };
if (isPointInGraceArea(pointerPos, ...pointerGraceAreaRef.current)) {}
```

packages/react/menu/src/Menu.tsx
```diff
@@ -710,8 +743,31 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
...
+          if (contentContext.isMouseMovingToSubmenu(event)) return;
+          if (!props.disabled && !context.open) {
+            openTimersRef.current.push(window.setTimeout(() => context.onOpenChange(true), 100));
+          }
+        })}
+        onMouseLeave={composeEventHandlers(props.onMouseLeave, (event) => {
+          openTimersRef.current.map(window.clearTimeout);
+          openTimersRef.current = [];
+
+          const contentRect = context.content?.getBoundingClientRect();
```
jjenzz, line 756: I'm wondering if we can store the bounding client rect in a ref somewhere when the content opens instead of every time the mouse leaves. 

Atm, this looks like we'll be getting a new bounding rect every time user does this:

https://user-images.githubusercontent.com/175330/120836251-42e70980-c55d-11eb-8833-b7b2b81d0844.mp4

Is that correct? Seems like unnecessary reflow.

packages/react/menu/src/Menu.tsx
```diff
@@ -710,8 +743,31 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
...
+              { x: contentRect.left, y: contentRect.top },
+              { x: contentRect.left, y: contentRect.bottom },
+            ]);
+
+            pointerGraceDurationTimer.current = window.setTimeout(
+              () => contentContext.onPointerGraceAreaChange(null),
+              300
+            );
+          } else {
+            contentContext.onPointerGraceAreaChange(null);
```
jjenzz, line 769: I'm curious when we fall into this branch? If the mouse is leaving the submenu trigger then isn't the content always open at that point?

packages/react/menu/src/Menu.tsx
```diff
@@ -710,8 +743,31 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
...
+        onMouseLeave={composeEventHandlers(props.onMouseLeave, (event) => {
+          openTimersRef.current.map(window.clearTimeout);
+          openTimersRef.current = [];
+
+          const contentRect = context.content?.getBoundingClientRect();
+          if (contentRect) {
+            contentContext.onPointerGraceAreaChange([
+              { x: event.clientX, y: event.clientY },
+              { x: contentRect.left, y: contentRect.top },
+              { x: contentRect.left, y: contentRect.bottom },
```
andy-hook, line 761: Are we sure we don't need to flip `x` when RTL? I know we tested earlier today but having another look things don't feel quite right in that configuration.

packages/react/menu/src/Menu.tsx
```diff
@@ -956,6 +1012,30 @@ function getNextMatch(values: string[], search: string, currentMatch?: string) {
...
+  return Math.abs((a.x * (b.y - c.y) + b.x * (c.y - a.y) + c.x * (a.y - b.y)) / 2);
+}
+
+function isPointInTriangle(point: Point, a: Point, b: Point, c: Point) {
+  const A = getTriangleArea(a, b, c);
+  const A1 = getTriangleArea(point, b, c);
+  const A2 = getTriangleArea(a, point, c);
+  const A3 = getTriangleArea(a, b, point);
+  return A === A1 + A2 + A3;
+}
```
jjenzz, line 1028: 🤯  obviously just going to have to trust you on this one, way over my head haha

jjenzz (changes requested):

packages/react/menu/src/Menu.tsx
jjenzz: Ah good spot, yeah, looks like we would need `contentRect.right` here somewhere in rtl

benoitgrelard: I think you're right, I think they should x should be `contentRect.right`, oh and actually now that I think about it, it's more complicated than that, it depends on the actual side it gets set on (collisions)

packages/react/menu/src/Menu.tsx
benoitgrelard: That's exactly what we had before we changed it. But it was named differently. I'm easy on that one, with the new name for `pointerGraceAreaRef` it does read better.

packages/react/menu/src/Menu.tsx
benoitgrelard: It's because it's in `onMouseMove`, so in `onMouseLeave` when we want to clear the timeout, we'd be clearing only the last one, but there's 100ms where new timeouts would override it and thus old ones would be lost and un-clearable. Let me know if that makes sense, we can go on a call otherwise.

packages/react/menu/src/Menu.tsx
benoitgrelard: We can, but we put it in a util as it's used 3 times cause otherwise each time you'd need to construct a point, handle the null case, and spread the ref.

packages/react/menu/src/Menu.tsx
benoitgrelard: We could cache it, when do we clear the cache? on unmount? what happens if new items get added dynamically? Maybe we don't worry about it?

packages/react/menu/src/Menu.tsx
jjenzz: Which did you have before, the `event.preventDefault()` stuff or passed the ref through context? I think I prefer the `event.preventDefault()` approach personally but happy with either or.

packages/react/menu/src/Menu.tsx
benoitgrelard: There's 100ms where you could leave without opening a submenu.

packages/react/menu/src/Menu.tsx
benoitgrelard: Don't worry we copied it from reakit 😀

benoitgrelard: The overall formula makes sense though when you think about it, adding the area of each triangle formed using the point would be exactly the area of the triangle only if it's inside, otherwise it would be bigger.

packages/react/menu/src/Menu.tsx
benoitgrelard: Oh didn't realise you had new handlers, I meant the ref through context. I don't really get why we'd need to re-create `onMouseEnter`/`onMouseLeave`…

packages/react/menu/src/Menu.tsx
benoitgrelard: Bit of a hack for now because it's deep down, but we should be able to use the `content` and read the `data-side` on it to get the final placed side.

benoitgrelard: Not too bothered about the hack because all the positioning logic will get swapped soon.

andy-hook: I'd be ok with that considering it's changing soon, my biggest worry though is remembering to come back to make the change. How do we usually track short term fix regressions? I guess more tests? 😅

packages/react/menu/src/Menu.tsx
jjenzz: I personally don't feel that re-constructing a point and spreading a ref warrant this entire abstraction. In fact, we wouldn't even need to spread and the `null` case would just be this no?

```jsx
function isPointInGraceArea(pointerPoint, graceArea: Triangle | null) {
  if (!graceArea) return false;
  const [a, b, c] = graceArea;
  const A = getTriangleArea(a, b, c);
  const A1 = getTriangleArea(pointerPoint, b, c);
  const A2 = getTriangleArea(a, pointerPoint, c);
  const A3 = getTriangleArea(a, b, pointerPoint);
  return A === A1 + A2 + A3;
}
```

packages/react/menu/src/Menu.tsx
jjenzz: We already have an `onItemLeave` so that's not a new thing, only `onItemEnter` would exist which would replace the strange `isMouseMovingToSubmenu` method. It's not new code, just a rename and it would `event.preventDefault()` instead of returning a boolean.

packages/react/menu/src/Menu.tsx
benoitgrelard: I'll remember 😀

packages/react/menu/src/Menu.tsx
benoitgrelard: I'd still keep both utils to make things clean rather than flattening, but otherwise I'm ok with it.

```tsx
function isPointerInGraceArea(pointerPos: Point, graceArea: Triangle | null) {
  if (!graceArea) return false;
  return isPointInTriangle(pointerPos, ...graceArea);
}
```

packages/react/menu/src/Menu.tsx
jjenzz: I dunno, you asked me to store this stuff in `ScrollArea` so I thought we'd be consistent. Given the perf issues we had on the design-system page lately, it seems this would be pretty heavy there and we're best off trying to avoid reflows where possible.

> what happens if new items get added dynamically

Could maybe worry about that when someone reports an issue?

packages/react/menu/src/Menu.tsx
benoitgrelard: Yeah but that's quite different, if I follow correctly, you're talking about the current `onItemLeave` on context?
If so, that's just a void callback, not a custom event. We'd have to implement 2 new custom events to do the prevent default thing.

packages/react/menu/src/Menu.tsx
benoitgrelard: Happy to go with simple caching for now for the life of the open submenu.

packages/react/menu/src/Menu.tsx
jjenzz: Yep makes sense, can we add that as a comment 🙏 ?

packages/react/menu/src/Menu.tsx
jjenzz: Can we add that as a comment as well please? 🙏

packages/react/menu/src/Menu.tsx
jjenzz: > you're talking about the current `onItemLeave` on context

yes, on the same context where that `isMouseMovingToSubmenu` method is.

We wouldn't have to add custom events... `onItemLeave` is called from existing events. It's just a case of passing that event to it:

```jsx
onItemEnter(event: React.MouseEvent): void;
onItemLeave(event: React.MouseEvent): void;
```

image

Seems to work okay but you'll know better than me 😅

jjenzz: This made me realise we could even do this then which feels more "complete":

image

So our two `onItemEnter` and `onItemLeave` events handle the side effects:

image

packages/react/menu/src/Menu.tsx
```diff
@@ -715,18 +715,18 @@ type MenuSubTriggerPrimitive = Polymorphic.ForwardRefComponent<
 const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
   const context = useMenuContext(SUB_TRIGGER_NAME);
   const contentContext = useMenuContentContext(SUB_TRIGGER_NAME);
-  const openTimersRef = React.useRef<number[]>([]);
+  const contentRectRef = React.useRef<ClientRect | undefined>();
+  const openTimerRef = React.useRef<number | undefined>(undefined);
```
benoitgrelard, line 720: I've managed to achieve this a different way using just one timer but checking if it's set before setting another one.

packages/react/menu/src/Menu.tsx
```diff
@@ -745,27 +745,35 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
...
-
-          const contentRect = context.content?.getBoundingClientRect();
-          if (contentRect) {
+          clearOpenTimer();
+
+          contentRectRef.current =
+            contentRectRef.current || context.content?.getBoundingClientRect();
+          if (contentRectRef.current) {
+            // TODO: make sure to update this when we change positioning logic
+            const side = context.content?.dataset.side;
```
benoitgrelard, line 764: Checking the side now

packages/react/menu/src/Menu.tsx
```diff
@@ -673,11 +685,14 @@ const MenuItemImpl = React.forwardRef((props, forwardedRef) => {
...
-            item.focus();
+          if (disabled) {
+            contentContext.onItemLeave(event);
           } else {
-            contentContext.onItemLeave();
+            contentContext.onItemEnter(event);
+            if (!event.defaultPrevented) {
+              const item = event.currentTarget;
+              item.focus();
+            }
```
jjenzz, line 695: 😍

packages/react/menu/src/Menu.tsx
andy-hook: Oh nice! 👍  could we be explicit with a `null` instead?

benoitgrelard: I was being lazy because otherwise it moans when doing clearTimeout and I have to null-check it 😀 I'll update.

jjenzz (approved):

PR #690, opened 2021-06-07
[ContextMenu][DropdownMenu] Fix sticky submenu trigger focus

andy-hook:
See https://github.com/radix-ui/primitives/pull/682#issuecomment-854983610

This change resolves stuck focus when quickly leaving submenu triggers, the behaviour now matches that of normal items by simply refocusing `content`.

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -54,6 +54,7 @@ type MenuContextValue = {
 type MenuSubContextValue = Omit<MenuContextValue, 'isSubmenu'> & {
   isSubmenu: true;
   contentId: string;
+  parentContent: MenuContentElement | null;
   trigger: MenuSubTriggerElement | null;
   onTriggerChange(trigger: MenuSubTriggerElement | null): void;
   onKeyOpen(): void;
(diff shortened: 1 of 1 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -707,6 +709,7 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
           } else {
             // There's 100ms where the user may leave an item before the submenu was opened.
             contentContext.onPointerGraceAreaChange(null);
+            context.parentContent?.focus();
```
jjenzz, line 712: Would `contentContext.onItemLeave()` work here? I think it does the same thing 💭

My thinking is that we wouldn't need the `context.parentContent` then because `contentContext` should be the same thing from this part.

jjenzz (changes requested):

packages/react/menu/src/Menu.tsx
andy-hook: Good point, original thinking was it referenced the submenu content but as that’s raised should have worked here.

andy-hook:
Closing as the changes we intend to make for pointer intent should supersede.

PR #691, opened 2021-06-08
[ContextMenu][DropdownMenu] Improve pointer grace

benoitgrelard:


andy-hook:
@jjenzz @benoitgrelard  I've updated pointer grace to be a bit more robust, the triggers also now align better with native by closing when leaving away from the target menu. I was fiddling around for a while on this and found that using a polygon to map the entire grace area was the simplest and most robust option compared to others.

https://user-images.githubusercontent.com/11708259/121558803-e167e480-ca0d-11eb-8d24-a57a6b65f00b.mp4

The polygon sidesteps the issue we had with event `clientX / clientY` accuracy, we can now snap our pointer very quickly and the menu will remain open.

https://user-images.githubusercontent.com/11708259/121558836-e7f65c00-ca0d-11eb-87a4-9b9ef23e1049.mp4

As we are now tracking pointer `x` we can bail early on the grace area if we suddenly move in the opposite direction

https://user-images.githubusercontent.com/11708259/121558825-e5940200-ca0d-11eb-8d1e-b4bde7a4c4d6.mp4

packages/react/menu/src/Menu.tsx
```diff
@@ -554,6 +584,15 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
                     }
                   })}
+                  onMouseMove={composeEventHandlers(props.onMouseMove, (event) => {
+                    if (event.currentTarget.contains(event.target as HTMLElement)) {
+                      if (lastPointerXRef.current !== event.clientX) {
+                        pointerXDirRef.current =
+                          event.clientX > lastPointerXRef.current ? 'right' : 'left';
+                        lastPointerXRef.current = event.clientX;
+                      }
+                    }
```
andy-hook, line 594: `onMouseMove` fires so often that it'll report the same value multiple times, this was a simple way to only update the direction if they've actually changed. The outer most condition is to just avoid it bubbling and updating needlessly up the tree.

packages/react/menu/src/Menu.tsx
```diff
@@ -692,19 +731,32 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
...
-            contentContext.onPointerGraceAreaChange([
-              { x: event.clientX, y: event.clientY },
-              { x: contentEdge, y: contentRect.top },
-              { x: contentEdge, y: contentRect.bottom },
-            ]);
-
-            pointerGraceDurationTimer.current = window.setTimeout(
+            const side = context.content?.dataset.side as Side;
+            const rightSide = side === 'right';
+            const bleed = rightSide ? -5 : +5;
```
andy-hook, line 736: This bleed value is arbitrary but ensures that the exit point is consistently inside the polygon across browsers and rtl.

packages/react/menu/src/Menu.tsx
```diff
@@ -456,18 +461,43 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+                pointerXDirRef.current
+              )
+            )
+              event.preventDefault();
           }}
           onItemLeave={React.useCallback((event) => {
-            if (!isPointerInGraceArea(event, pointerGraceAreaRef.current)) {
+            const prematureLeave = pointerXDirRef.current !== pointerGraceAreaRef.current?.side;
+            if (
+              prematureLeave ||
```
benoitgrelard, line 476: I'm struggling to see which case this covers?
If I'm leaving the item-trigger from the left side (in LTR) then I won't be in the grace area right? so the check is redundant.
I've removed the check locally and all seems to work still, but perhaps I'm missing the case this is catering for?

benoitgrelard: Actually it seems to work because the same check already exists within `isPointerMovingThroughGraceArea` so regardless this one here isn't needed I think.

benoitgrelard: I still think it's a bit odd reading this that this check would be needed at all. I'm guessing the reason is mostly because of the added bleed which will make the point actually still be inside the area?

packages/react/menu/src/Menu.tsx
benoitgrelard: This is all actually making me think that rather than tracking the pointer direction, and use the mouse leave event coordinates, if we tracked the pointer position in a ref in mousemove, we probably wouldn't need the artificial bleed and we wouldn't need the directional check, I have a feeling it would all work actually better this way and the code would be simpler.

packages/react/menu/src/Menu.tsx
```diff
@@ -680,6 +718,7 @@ const MenuSubTrigger = React.forwardRef((props, forwardedRef) => {
           contentContext.onItemEnter(event);
           if (event.defaultPrevented) return;
           if (!props.disabled && !context.open && !openTimerRef.current) {
+            contentContext.onPointerGraceAreaChange(null);
```
benoitgrelard, line 721: Why was this one needed?

benoitgrelard (commented):
It all feels really good, I've posted a few high-level comments for now.

packages/react/menu/src/Menu.tsx
andy-hook: We need to clear the previous area when moving across triggers to prevent the old area from interfering with focus behaviour.

packages/react/menu/src/Menu.tsx
andy-hook: As discussed, this was redundant 😅

packages/react/menu/src/Menu.tsx
andy-hook: Me and @benoitgrelard went into more detail with this on a call, it was something I had tried before but it had some non obvious caveats that meant we'd still need to rely on pointer movement direction. We decided to stick to the original implementation but tidy it up.

packages/react/menu/src/Menu.tsx
```diff
@@ -450,24 +454,35 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
+    const isMovingTowards = pointerDirRef.current === pointerGraceIntentRef.current?.side;
+    return isMovingTowards && isPointerInGraceArea(event, pointerGraceIntentRef.current?.area);
+  }, []);
+
   return (
     <PortalWrapper>
       <ScrollLockWrapper>
         <MenuContentProvider
           searchRef={searchRef}
           onItemEnter={(event) => {
```
benoitgrelard, line 467: Not sure how we missed this before, but this one needs `useCallback` now.

packages/react/menu/src/Menu.tsx
```diff
@@ -450,24 +454,35 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
   // the last element in the DOM (beacuse of the `Portal`)
   useFocusGuards();
 
+  const isPointerMovingToSubmenu = React.useCallback((event: React.MouseEvent) => {
+    const isMovingTowards = pointerDirRef.current === pointerGraceIntentRef.current?.side;
+    return isMovingTowards && isPointerInGraceArea(event, pointerGraceIntentRef.current?.area);
+  }, []);
```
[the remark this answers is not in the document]
benoitgrelard: Love how this reads now.

packages/react/menu/src/Menu.tsx
```diff
@@ -554,6 +569,12 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
                       searchRef.current = '';
                     }
                   })}
+                  onMouseMove={composeEventHandlers(props.onMouseMove, (event) => {
+                    const target = event.target as HTMLElement;
+                    if (event.currentTarget.contains(target) && event.movementX !== 0) {
+                      pointerDirRef.current = event.movementX >= 1 ? 'right' : 'left';
```
benoitgrelard, line 575: just a nit here, but if we're already checking for `!== 0` above, I feel like `> 0` here would express the directional intent better, rather than re-encapsulating that non-0 check.

packages/react/menu/src/Menu.tsx
andy-hook: Was wondering on this too, thought it might be a for a reason I was unaware of. Thanks for the reminder

andy-hook (approved):

PR #692, opened 2021-06-08
[DropdownMenu] Add `aria-labelledby` to follow spec

jjenzz:
**This is the first PR of a few that I will be raising with various accessibility improvements.**

- Adds `aria-labelledby` to `MenuSubContent` and the root `DropdownMenuContent`. Annoyingly the label isn't being read aloud in NVDA but I will look into this separately.
- Creates a new `DropdownMenuRoot` part because there are a bunch of attributes on the root content that are not required in the sub dropdown menus. This was also needed because the root `aria-labelledby` would override the submenu `aria-labelledby` in `MenuSubContent` if this attribute was added to both types within the `DropdownMenu` composition which resulted in incorrect ids for submenus (they would be pointing at the `DropdownMenuTrigger` id).
- Updates `aria-expanded` for sub menu triggers to follow WAI-ARIA spec
	
	image
	image

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -18,7 +18,7 @@ type Direction = 'ltr' | 'rtl';
 const DROPDOWN_MENU_NAME = 'DropdownMenu';
 
 type DropdownMenuContextValue = {
-  isSubmenu: false;
+  isRootMenu: true;
   triggerId: string;
   triggerRef: React.RefObject<HTMLButtonElement>;
   contentId: string;
@@ -28,7 +28,7 @@ type DropdownMenuContextValue = {
 };
 
 type DropdownMenuSubContextValue = {
-  isSubmenu: true;
+  isRootMenu: false;
   open: boolean;
   onOpenChange(open: boolean): void;
   onOpenToggle(): void;
@@ -58,7 +58,7 @@ const DropdownMenu: React.FC<DropdownMenuOwnProps> = (props) => {
 
   return isInsideContent ? (
     <DropdownMenuProvider
-      isSubmenu={true}
+      isRootMenu={false}
       open={open}
       onOpenChange={setOpen}
       onOpenToggle={handleOpenToggle}
@@ -90,7 +90,7 @@ const DropdownMenuRoot: React.FC<DropdownMenuRootOwnProps> = (props) => {
   const triggerRef = React.useRef<HTMLButtonElement>(null);
   return (
     <DropdownMenuProvider
-      isSubmenu={false}
+      isRootMenu={true}
       triggerId={useId()}
       triggerRef={triggerRef}
       contentId={useId()}
@@ -124,7 +124,7 @@ type DropdownMenuTriggerPrimitive = Polymorphic.ForwardRefComponent<
 const DropdownMenuTrigger = React.forwardRef((props, forwardedRef) => {
   const { as = TRIGGER_DEFAULT_TAG, ...triggerProps } = props;
   const context = useDropdownMenuContext(TRIGGER_NAME);
-  return context.isSubmenu ? null : (
+  return context.isRootMenu ? (
     <MenuPrimitive.Anchor
       type="button"
       id={context.triggerId}
@@ -149,7 +149,7 @@ const DropdownMenuTrigger = React.forwardRef((props, forwardedRef) => {
         }
       })}
     />
-  );
+  ) : null;
 }) as DropdownMenuTriggerPrimitive;
 
 DropdownMenuTrigger.displayName = TRIGGER_NAME;
@@ -174,40 +174,35 @@ type DropdownMenuContentPrimitive = Polymorphic.ForwardRefComponent<
 
 const DropdownMenuContent = React.forwardRef((props, forwardedRef) => {
   const {
-    onCloseAutoFocus,
     disableOutsidePointerEvents = true,
     disableOutsideScroll = true,
     portalled = true,
     ...contentProps
   } = props;
   const context = useDropdownMenuContext(CONTENT_NAME);
+
+  const commonProps = {
+    ...contentProps,
+    disableOutsidePointerEvents,
+    disableOutsideScroll,
+    portalled,
+    style: {
+      ...contentProps.style,
+      // re-namespace exposed content custom property
+      ['--radix-dropdown-menu-content-transform-origin' as any]: 'var(--radix-popper-transform-origin)',
+    },
+  };
+
   return (
     <ContentContext.Provider value={true}>
-      {context.isSubmenu ? (
-        <MenuPrimitive.Content
-          {...contentProps}
-          ref={forwardedRef}
-          disableOutsidePointerEvents={disableOutsidePointerEvents}
-          disableOutsideScroll={disableOutsideScroll}
-          portalled={portalled}
-          onCloseAutoFocus={onCloseAutoFocus}
-        />
-      ) : (
+      {context.isRootMenu ? (
         <MenuPrimitive.Content
           id={context.contentId}
           aria-labelledby={context.triggerId}
-          {...contentProps}
+          {...commonProps}
           ref={forwardedRef}
-          disableOutsidePointerEvents={disableOutsidePointerEvents}
-          disableOutsideScroll={disableOutsideScroll}
-          portalled={portalled}
-          style={{
-            ...props.style,
-            // re-namespace exposed content custom property
-            ['--radix-dropdown-menu-content-transform-origin' as any]: 'var(--radix-popper-transform-origin)',
-          }}
           trapFocus
-          onCloseAutoFocus={composeEventHandlers(onCloseAutoFocus, (event) => {
+          onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
             event.preventDefault();
             context.triggerRef.current?.focus();
           })}
@@ -223,6 +218,8 @@ const DropdownMenuContent = React.forwardRef((props, forwardedRef) => {
             { checkForDefaultPrevented: false }
           )}
         />
+      ) : (
+        <MenuPrimitive.Content {...commonProps} ref={forwardedRef} />
       )}
     </ContentContext.Provider>
   );
@@ -244,7 +241,7 @@ type DropdownMenuTriggerItemPrimitive = Polymorphic.ForwardRefComponent<
 
 const DropdownMenuTriggerItem = React.forwardRef((props, forwardedRef) => {
   const context = useDropdownMenuContext(TRIGGER_ITEM_NAME);
-  return context.isSubmenu ? <MenuPrimitive.SubTrigger {...props} ref={forwardedRef} /> : null;
+  return context.isRootMenu ? null : <MenuPrimitive.SubTrigger {...props} ref={forwardedRef} />;
 }) as DropdownMenuTriggerItemPrimitive;
 
 DropdownMenuTriggerItem.displayName = TRIGGER_ITEM_NAME;
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { composeEventHandlers } from '@radix-ui/primitive';
-import { useComposedRefs } from '@radix-ui/react-compose-refs';
+import { composeRefs } from '@radix-ui/react-compose-refs';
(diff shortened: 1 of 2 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -280,6 +282,7 @@ const MenuSubContent = React.forwardRef((props, forwardedRef) => {
   return context.isSubmenu ? (
     <MenuContentImpl
       id={context.contentId}
+      aria-labelledby={context.triggerId}
```
benoitgrelard, line 285: What is the `aria-labelledby` specifically for here and why is it only for sub triggers?

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -146,37 +182,49 @@ const DropdownMenuContent = React.forwardRef((props, forwardedRef) => {
...
+      {context.isSubmenu ? (
+        <MenuPrimitive.Content
+          {...contentProps}
+          ref={forwardedRef}
+          disableOutsidePointerEvents={disableOutsidePointerEvents}
+          disableOutsideScroll={disableOutsideScroll}
+          portalled={portalled}
+          onCloseAutoFocus={onCloseAutoFocus}
+        />
+      ) : (
```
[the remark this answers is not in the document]
jjenzz: Yes but the other side of the ternary adds root content specific props that the submenu content doesn't have... e.g. `id`, `aria-labelledBy`, `style` with transform origin stuff etc...

packages/react/menu/src/Menu.tsx
jjenzz: It's not only for sub triggers, I add it to the root trigger too in `DropdownMenu`. There isn't a root trigger part in `Menu` which is why this might look a bit confusing.

packages/react/dropdown-menu/src/DropdownMenu.tsx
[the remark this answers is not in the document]
benoitgrelard: Should we do it the same way we've done others with the `commonProps` approach?

packages/react/menu/src/Menu.tsx
benoitgrelard: Yeah I'm starting to struggle a bit, it feels like we have code a bit all over now. It feels like we're tracking submenu stuff is both now too, which may be ok but at first glance seems repeated.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -146,37 +182,49 @@ const DropdownMenuContent = React.forwardRef((props, forwardedRef) => {
...
-    </SubmenuContext.Provider>
+    <ContentContext.Provider value={true}>
+      {context.isSubmenu ? (
+        <MenuPrimitive.Content
+          {...contentProps}
+          ref={forwardedRef}
+          disableOutsidePointerEvents={disableOutsidePointerEvents}
+          disableOutsideScroll={disableOutsideScroll}
+          portalled={portalled}
+          onCloseAutoFocus={onCloseAutoFocus}
```
andy-hook, line 193: As we're now being explicit with the submenu content distinction I don't think we need to pass these anymore.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -146,37 +182,49 @@ const DropdownMenuContent = React.forwardRef((props, forwardedRef) => {
...
+          aria-labelledby={context.triggerId}
+          {...contentProps}
+          ref={forwardedRef}
+          disableOutsidePointerEvents={disableOutsidePointerEvents}
+          disableOutsideScroll={disableOutsideScroll}
+          portalled={portalled}
+          style={{
+            ...props.style,
+            // re-namespace exposed content custom property
+            ['--radix-dropdown-menu-content-transform-origin' as any]: 'var(--radix-popper-transform-origin)',
```
andy-hook, line 207: Are we sure we don't also need this for animation on submenu content?

benoitgrelard: Yeah we do, I think I'll this bit with the `commonProps` approach like I suggested above (https://github.com/radix-ui/primitives/pull/692#discussion_r648042127), it'll be clearer what the differences are and should be.

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: Yeah looks like it was never needed to be honest, given they would be part of `...contentProps`

packages/react/menu/src/Menu.tsx
jjenzz: We have to track `isSubmenu` stuff in both because there requirements in `DropdownMenu` **root** parts that don't apply in `DropdownMenu` **sub** parts. This logic cannot be moved to `Menu` because it applies specifically to `DropdownMenu`s only.

The `isSubmenu` boolean in `Menu` handles all the common submenu related stuff.

We initially tried to expose a `MenuTrigger` from `Menu` to DRY up a bunch of this stuff but `ContextMenu`s don't have triggers and it started to get really complicated when we tried it. We can maybe give that another go though.

benoitgrelard: Yeah I realised that after, I think where we landed makes sense.

packages/react/dropdown-menu/src/DropdownMenu.tsx
jjenzz: I had thought the same thing initially but then we would be spreading the knowledge of how submenus should behave across two different components. 

My thinking here was that we would just pass them along and leave it to `Menu` to determine how to handle them since it already decides what to do with these in the context of a submenu.

Given both of your feedback, I think I need to change the boolean from `context.isSubmenu` to `context.isRootMenu`. Because, `DropdownMenu` doesn't actually need to know if it's a submenu or not, it only need to know if its a root menu to handle `DropdowMenu` specific things slightly differently, otherwise it just passes everything through to `Menu` to handle.

packages/react/dropdown-menu/src/DropdownMenu.tsx
jjenzz: Ah, I originally removed this because the `side` is always hardcoded for submenus but didn't think about collision detection! Will move this onto both.

packages/react/dropdown-menu/src/DropdownMenu.tsx
jjenzz: @benoitgrelard @andy-hook thoughts on this one?

benoitgrelard: Oh I see I didn't open it up higher so it's more to do with the default values right? that's why they end up being passed explicitly. That makes sense to me then.

andy-hook: I much prefer this 👍

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -18,58 +18,93 @@ type Direction = 'ltr' | 'rtl';
 const DROPDOWN_MENU_NAME = 'DropdownMenu';
 
 type DropdownMenuContextValue = {
```
benoitgrelard, line 20: Should we rename this one `DropdownMenuRootContextValue`?

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -138,45 +174,54 @@ type DropdownMenuContentPrimitive = Polymorphic.ForwardRefComponent<
...
-            if (targetIsTrigger) event.preventDefault();
-          },
-          { checkForDefaultPrevented: false }
-        )}
-      />
-    </SubmenuContext.Provider>
+    <ContentContext.Provider value={true}>
+      {context.isRootMenu ? (
+        <MenuPrimitive.Content
+          id={context.contentId}
```
benoitgrelard, line 200: I would expect `id` to be a common prop no?

jjenzz: It's not a common prop though. In `DropdownMenu` it's only needed to associate the root trigger with the root menu via `aria-controls`. 

The `Menu` handles the `aria-controls` stuff for `MenuSubTrigger` already and if I left put this `id` on both (when it isn't necessary) it would override the `id` that `Menu` applies from its context and break the `aria-controls` stuff in `Menu`.

benoitgrelard: Ha right, that's done in `Menu`. This is doing my head in everytime, it's kinda hard to wrap your head around what's done where…

It's just that when I look at this code I'm like, well sumenus also need to associate their content to their triggers, so it looks like it's missing, but when you say it's done in `Menu` because it's shared, then it makes perfect sense.

packages/react/dropdown-menu/src/DropdownMenu.tsx
jjenzz: I dunno, we didn't do this in `Menu` because usually our context names are like:

- `{ComponentName}Provider` 
- `{ComponentName}ContextValue`

And in this case it's a `DropdownMenuProvider` for `DropdownMenu`. 

We did it this way in `Menu` too but happy to change them both in this PR if you think it makes more sense?

benoitgrelard: Yeah I agree in general because we only have one context to one component ratio.
Here we have one context provider too, but its type can be 2 things.

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: Perhaps we can add some comments for these cases to save our future selves from getting confused.

jjenzz: I dunno if a comment is needed here honestly... `DropdownMenu` introduces a **new** `Trigger` part so it logically needs to add `aria-controls` stuff from that **new** part. It doesn't need to do that for submenus because it doesn't introduce a new dropdown specific part/relationship there, it just re-uses all the `Menu` parts.

For example, we don't do `aria-controls` stuff in `Accordion` for the `Collapsible` parts, we just re-compose it.

benoitgrelard: Yeah that's very true, that's probably the diff that what caught my attention.

packages/react/dropdown-menu/src/DropdownMenu.tsx
jjenzz: Would you expect me to declare the type alongside the `DropdownRootMenu` component or leave it at the top like this?

benoitgrelard: No I think I'd still group them as they are mostly just related to the context itself.

jjenzz: Cool, done 👍

packages/react/menu/src/Menu.tsx
```diff
@@ -41,7 +41,7 @@ const SUB_CLOSE_KEYS: Record<Direction, string[]> = {
 
 const MENU_NAME = 'Menu';
 
-type MenuContextValue = {
+type MenuRootContextValue = {
   isSubmenu: false;
```
benoitgrelard, line 45: Would there be any value in updating these to `isRootMenu` instead for consistency?

One advantage I can think of, which I noted a few times whilst reviewing the PRs but never commented on is that in ternaries root would come before sub which to me is more readable or makes more sense to my brain 😀

But again at the time before I didn't think that was enough of a reason to comment on it or ask for a change.
Perhaps now it is?

jjenzz: > Would there be any value in updating these to `isRootMenu` instead for consistency?

I don't think so. The intent is different here. In `DropdownMenu` I need to know if it is the root menu to do root menu specific things related to the new trigger part.

In `Menu` we want to know if they are in a submenu.

benoitgrelard: Fair enough, I don't think it matters much what you test for as in the end it's the same result but I can see why we'd want to keep it closer to what we're catering for. Looking at it, it would reverse some ternaries in the way I mentioned but more of them would then be the wrong way, so you're probably right 😀

jjenzz: If you would prefer `root` parts to render before sub parts in the ternary then I could always do `!context.isSubmenu`? 😏

benoitgrelard (approved):
I'm happy with this! ♥️ 
Good to go for me 👍

PR #693, opened 2021-06-08
[ContextMenu] Bring inline with `DropdownMenu` changes

jjenzz:
- I realised when doing [the `DropdownMenu` bits](https://github.com/radix-ui/primitives/pull/692) that we didn't need two separate contexts for identifying `isInsideContent` and `isSubmenu` so updated this to follow the reduced approach in `DropdownMenu`
- I also thought there isn't much use us having `useControllableState` in sub context menus because the parent menu can't be controlled. They wouldn't be able to force open submenus without first forcing open parent menu which isn't possible. The only benefit the previous approach had was that they could automatically have submenus open after someone opened the parent menu via  `onContextMenu` event but I can't think when this use-case would be relevant. Thoughts?

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -130,7 +130,7 @@ type ContextMenuContentPrimitive = Polymorphic.ForwardRefComponent<
 >;
 
 const ContextMenuContent = React.forwardRef((props, forwardedRef) => {
-  const { disableOutsidePointerEvents = true, offset, ...contentProps } = props;
+  const { offset, ...contentProps } = props;
   const context = useContextMenuContext(CONTENT_NAME);
 
   const commonProps = {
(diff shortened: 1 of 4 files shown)


benoitgrelard (commented):
Yeah I think that makes sense for the control stuff, happy to remove all that code for as long as we can't control the top-level one. If this changes then we can bring it back.

andy-hook:
Really like that we're standardising on this, I agree on the control point.

benoitgrelard (dismissed):

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -182,51 +144,51 @@ const ContextMenuContent = React.forwardRef((props, forwardedRef) => {
   };
 
   return (
-    <InsideContentContext.Provider value={true}>
-      {isSubmenu ? (
-        <MenuPrimitive.Content {...commonProps} ref={forwardedRef} />
+    <ContentContext.Provider value={true}>
+      {context.isRootMenu ? (
+        <ContextMenuRootContent {...commonProps} ref={forwardedRef} />
```
benoitgrelard, line 149: Not sure if we've been doing this in others, but the `ref` is also a common prop I guess.

benoitgrelard (commented):
Yeah this makes it all quite a bit clearer!

packages/react/context-menu/src/ContextMenu.tsx
andy-hook: We've kept it separate in other cases afaik. Leaving it out of `commonProps` aligns better in my head because of the typical spread pattern we use with regular props e.g. `<Component {...props} ref={forwardedRef}`/>`

benoitgrelard: Yeah I was wondering that too.

jjenzz: yeah, we haven't done that elsewhere, and I find it easier to see that we definitely passed refs through correctly when they're in the render like this.

benoitgrelard: 🚀

benoitgrelard (approved):

PR #694, opened 2021-06-08
[DropdownMenu] Enable `onEntryFocus` for root menu

jjenzz:
- Fixes accessibility in NVDA
- Follows WAI-ARIA spec that says `Enter` and `Space` keypresses should focus first item.

I did this by listening for keyboard and mouse events to update an `isKeyboardUser` boolean in state. I chose this approach because the previous approach meant firing an `onKeyOpen` event from the trigger and we don't have a root trigger in `Menu`.


benoitgrelard (commented):
I'm not sure I like the approach here.

I think the requirement for focusing the first item when pressing space/enter/down and focus last from pressing up is only relevant to `DropdownMenuTrigger` so I would expect to see it implemented there.

At a glance it doesn't really make sense to me to see this done in `Menu` as that concept doesn't apply to `ContextMenu`.

jjenzz:
> when pressing space/enter/down and focus last from pressing up is only relevant to `DropdownMenuTrigger`

It's relevant to `MenuSubTrigger` also. As you can see we had this logic for `MenuSubTrigger` already but it was handled differently via an `onKeyOpen` callback in `Menu`, for submenus **only**.

I couldn't use that same approach for root triggers because there isn't a root trigger in `Menu`. Also, [the spec says that when a `menu` opens via keyboard then the first item should be focused](https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction-12).

image

Any kind of `menu`.

> At a glance it doesn't really make sense to me to see this done in `Menu` as that concept doesn't apply to `ContextMenu`.

If the `ContextMenu` is opened via keyboard then it does apply. These changes allow that to happen:

https://user-images.githubusercontent.com/175330/121256702-bf9f1e00-c8a4-11eb-9eb9-853aa2df9690.mp4

benoitgrelard:
Interesting, I hadn't considered the context menu could be opened like that.
I do feel though that we keep coming back to the fact that `Menu` doesn't have a `Trigger` part. I know we moved away from that but I'm wondering now…

jjenzz:
> I do feel though that we keep coming back to the fact that Menu doesn't have a Trigger part. I know we moved away from that but I'm wondering now…

I had given similar feedback a while ago so @andy-hook and I had tried to move the trigger into the `Menu` but it opened a can of worms. It wasn't as simple as it seems. 

It might be worth revisiting now though as we were in a bit of a rabbit hole with all the submenu stuff at the time. Either way though, I think I still prefer this solution in regard to focusing the first item on open. 

This feels more like like our `onPointerDownOutside` or `onFocusOutside` stuff. With those we're saying it should close if those events occur externally and here we're saying it should open with the first item focused if a keydown happens externally.

jjenzz:
I just realised this probably doesn't need to be state though so I'll update that part... just needs to be a value we can reference when `context.open` changes to `true`

benoitgrelard:
> > I do feel though that we keep coming back to the fact that Menu doesn't have a Trigger part. I know we moved away from that but I'm wondering now…
> 
> I had given similar feedback a while ago so @andy-hook and I had tried to move the trigger into the `Menu` but it opened a can of worms. It might be worth revisiting now though as we were in a bit of a rabbit hole with all the submenu stuff at the time but it's only `DropdownMenu` that has a trigger atm. It's probably a premature abstraction at this stage.

Yeah I think I agree at this stage.

> Either way though, I think I still prefer this solution in regard to focusing the first item on open.
> 
> This feels more like like our `onPointerDownOutside` or `onFocusOutside` stuff. With those we're saying it should close if those events occur externally and here we're saying it should open with the first item focused if a keydown happens externally.
> 
> The WAI-ARIA spec is saying that if they're using a keyboard and it opens then it should focus the first item... It's not saying "when it opens from the trigger" but "when a `menu` opens" and that's what the logic here communicates. That's how I got the `ContextMenu` stuff for free.

Yeah I have to admit it's pretty cool how the context menu just worked with this mouse keys accessibility thing!

andy-hook:
> it's only DropdownMenu that has a trigger atm. It's probably a premature abstraction at this stage.

Yeah.. that's where we landed when we last spoke about this. I can see the argument for moving it, especially as we've discovered more and more requirements but tbh I'm not convinced now is the right time to do so.

> The WAI-ARIA spec is saying that if they're using a keyboard and it opens then it should focus the first item... It's not saying "when it opens from the trigger" but "when a menu opens" and that's what the logic here communicates. That's how I got the ContextMenu stuff for free.

I like this approach personally, makes more sense to me.

packages/react/menu/src/Menu.tsx
```diff
@@ -74,13 +73,26 @@ type MenuOwnProps = {
 const Menu: React.FC<MenuOwnProps> = (props) => {
   const { open = false, children, onOpenChange } = props;
   const [content, setContent] = React.useState<MenuContentElement | null>(null);
+  const [isKeyboardUser, setIsKeyboardUser] = React.useState(false);
```
benoitgrelard, line 76: Like you said, probably doesn't need to be state as it doesn't affect rendering.

packages/react/menu/src/Menu.tsx
```diff
@@ -74,13 +73,26 @@ type MenuOwnProps = {
...
   const handleOpenChange = useCallbackRef(onOpenChange);
   const computedDirection = useDirection(content, props.dir);
 
+  React.useEffect(() => {
+    const handleKeyDown = () => setIsKeyboardUser(true);
+    const handleMouseMove = () => setIsKeyboardUser(false);
+    document.addEventListener('keydown', handleKeyDown, { capture: true });
+    document.addEventListener('mousemove', handleMouseMove, { capture: true });
```
benoitgrelard, line 84: Should we use pointer events here instead to include touch better?

packages/react/menu/src/Menu.tsx
benoitgrelard: Not a fan of the name `isKeyboardUser` as well, it's a bit misleading to me because for example I use both keyboard and mouse. To me it's more the last intent used I guess.

packages/react/menu/src/Menu.tsx
```diff
@@ -521,7 +508,10 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
                 loop={loop}
                 currentTabStopId={currentItemId}
                 onCurrentTabStopIdChange={setCurrentItemId}
-                onEntryFocus={onEntryFocus}
+                onEntryFocus={(event) => {
```
benoitgrelard, line 511: This only takes into account the "focus first item" right? are we going to do the focus last when pressing up?

jjenzz:
> I just realised this probably doesn't need to be state though so I'll update that part... just needs to be a value we can reference when `context.open` changes to `true`

Turns out i can't do this because when `context.open` state changes at submenu level that state change obviously won't trigger a re-render of the parent context so it wouldn't have the updated reference. Also, when a `menu` is open you can switch from being a keyboard user to a mouse user but the ref change wouldn't update the context consumers (obvs) so this happens:

CleanShot 2021-06-10 at 11 50 22


`onItemLeave` is firing and because it thinks i'm a keyboard user still, it's focusing the first item. So yeah, needs to be state.

packages/react/menu/src/Menu.tsx
jjenzz: > Like you said, probably doesn't need to be state as it doesn't affect rendering.

Turns out i can't do this because when `context.open` state changes at submenu level that state change obviously won't trigger a re-render of the parent context so it wouldn't have the updated reference. Also, when a `menu` is open you can switch from being a keyboard user to a mouse user but the ref change wouldn't update the context consumers (obvs) so this happens:

CleanShot 2021-06-10 at 11 50 22

`onItemLeave` is firing and because it thinks i'm a keyboard user still, it's focusing the first item. So yeah, needs to be state.

> it's a bit misleading to me because for example I use both keyboard and mouse

Would `isUsingKeyboard` make more sense?

benoitgrelard: I'm not sure I'm following. If the ref itself was being passed down, then `.current` on it would always be up to date no?

`isUsingKeyboard` sounds good to me 👍

packages/react/menu/src/Menu.tsx
jjenzz: oh bum, hadn't thought of that 🤔  we should do... hmmm

packages/react/menu/src/Menu.tsx
jjenzz: A `ref` change doesn't won't trigger provider consumers to re-render so the consumers will still have the previous ref from the previous render closure.

jjenzz: Oh wait, the ref would be fine if i passed the whole ref lol, it's cos I was passing `isKeyboardUser={isKeyboarRef.curren}` ... my bad 😅  too much COVID in my veins

packages/react/menu/src/Menu.tsx
benoitgrelard: That one would have to be handled as an `onKeyDown` on the trigger I guess, but the issue is the items tracked are inside `Menu` right? not inside `DropdownMenu`. We could potentially export the `useCollection` hook from `Menu` to get them in `DropdownMenu`… 🤔

packages/react/menu/src/Menu.tsx
benoitgrelard: Those were my thoughts :0

packages/react/menu/src/Menu.tsx
```diff
@@ -74,13 +73,26 @@ type MenuOwnProps = {
...
   const [content, setContent] = React.useState<MenuContentElement | null>(null);
+  const isUsingKeyboardRef = React.useRef(false);
   const handleOpenChange = useCallbackRef(onOpenChange);
   const computedDirection = useDirection(content, props.dir);
 
+  React.useEffect(() => {
+    const handleKeyDown = () => (isUsingKeyboardRef.current = true);
+    const handlePointerMove = () => (isUsingKeyboardRef.current = false);
+    document.addEventListener('keydown', handleKeyDown, { capture: true });
+    document.addEventListener('pointermove', handlePointerMove, { capture: true });
```
benoitgrelard, line 84: I think we should use `pointerdown` no? because on touch screen `pointermove` may not happen, but a `pointerdown` would always happen, or potentially we do both to cover all basis.

jjenzz: we would need both, which just got me thinking - I dunno if we need `pointer` events at all because how would you navigate by keyboard on a touch device?

benoitgrelard: You definitely can, mixed input right? iPads have both touch and keyboard support for example.

benoitgrelard: Life is to complicated now, can't assume anything anymore 😀 hahaha

jjenzz: haha true, I'll add both events in that case

packages/react/menu/src/Menu.tsx
```diff
@@ -79,12 +79,14 @@ const Menu: React.FC<MenuOwnProps> = (props) => {
 
   React.useEffect(() => {
     const handleKeyDown = () => (isUsingKeyboardRef.current = true);
-    const handlePointerMove = () => (isUsingKeyboardRef.current = false);
+    const handlePointer = () => (isUsingKeyboardRef.current = false);
     document.addEventListener('keydown', handleKeyDown, { capture: true });
```
benoitgrelard, line 83: by the way, random thought now, but why is capture phase important for all these?

jjenzz: Good point, not sure what my reasoning was before 🤔 probably not needed.

packages/react/menu/src/Menu.tsx
jjenzz: We agreed we're going to solve this in a separate PR right? I'll add a bullet to the `submenu` PR description if so

benoitgrelard: yep 👍

jjenzz:
I remembered why I bound capture events btw... 

**without capture**


https://user-images.githubusercontent.com/175330/121559914-ee390800-ca0e-11eb-8952-c540487590e1.mp4



**with capture**

https://user-images.githubusercontent.com/175330/121559700-ba5de280-ca0e-11eb-8866-6255f3e03b5c.mp4

Event steps:

- Key to open (`isUsingKeyboardRef.current === true`)
- Escape key to close so I am still focused on trigger
- Wiggle mouse (`isUsingKeyboardRef.current === false`)
- Press enter to open 

That last step doesn't set `isUsingKeyboardRef.current` to `true` before the content opens so the first item doesn't focus. Capture phase ensures we set it before any other key events execute.

benoitgrelard (approved):

PR #698, opened 2021-06-09
[FocusScope] Update API for modality changes

benoitgrelard:
This is the first one of a stack of PRs (#700) to bring in new modality changes.

>Note: base branch is `modality` (#700), not `main`.

### DIFF
--- a/packages/react/focus-scope/src/FocusScope.test.tsx
+++ b/packages/react/focus-scope/src/FocusScope.test.tsx
@@ -18,7 +18,7 @@ describe('FocusScope', () => {
     beforeEach(() => {
       rendered = render(
         <div>
-          <FocusScope as="form" wrapped contained>
+          <FocusScope as="form" loop trapped>
             <TestField label={INNER_NAME_INPUT_LABEL} />
             <TestField label={INNER_EMAIL_INPUT_LABEL} />
             <button>{INNER_SUBMIT_LABEL}</button>
@@ -59,7 +59,7 @@ describe('FocusScope', () => {
     beforeEach(() => {
       rendered = render(
         <div>
-          <FocusScope as="form" wrapped contained>
+          <FocusScope as="form" loop trapped>
             <TestField label={INNER_NAME_INPUT_LABEL} tabIndex={-1} />
             <TestField label={INNER_EMAIL_INPUT_LABEL} />
             <button>{INNER_SUBMIT_LABEL}</button>
@@ -92,7 +92,7 @@ describe('FocusScope', () => {
     beforeEach(() => {
       rendered = render(
         <div>
-          <FocusScope as="form" wrapped contained>
+          <FocusScope as="form" loop trapped>
             <TestField label={INNER_NAME_INPUT_LABEL} />
             <button onBlur={handleLastFocusableElementBlur}>{INNER_SUBMIT_LABEL}</button>
           </FocusScope>
--- a/packages/react/focus-scope/src/FocusScope.test.tsx
+++ b/packages/react/focus-scope/src/FocusScope.test.tsx
@@ -18,12 +18,10 @@ describe('FocusScope', () => {
     beforeEach(() => {
       rendered = render(
         <div>
-          <FocusScope trapped>
-            <form>
-              <TestField label={INNER_NAME_INPUT_LABEL} />
-              <TestField label={INNER_EMAIL_INPUT_LABEL} />
-              <button>{INNER_SUBMIT_LABEL}</button>
-            </form>
+          <FocusScope as="form" wrapped contained>
+            <TestField label={INNER_NAME_INPUT_LABEL} />
+            <TestField label={INNER_EMAIL_INPUT_LABEL} />
+            <button>{INNER_SUBMIT_LABEL}</button>
           </FocusScope>
           <TestField label="other" />
           <button>some outer button</button>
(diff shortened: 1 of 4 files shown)


packages/react/focus-scope/src/FocusScope.test.tsx
```diff
@@ -18,12 +18,10 @@ describe('FocusScope', () => {
     beforeEach(() => {
       rendered = render(
         <div>
-          <FocusScope trapped>
-            <form>
-              <TestField label={INNER_NAME_INPUT_LABEL} />
-              <TestField label={INNER_EMAIL_INPUT_LABEL} />
-              <button>{INNER_SUBMIT_LABEL}</button>
-            </form>
```
andy-hook, line 26: Just curious why we're removing `form` in these tests. Is it simply redundant?

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -21,10 +21,16 @@ type FocusScopeOwnProps = Polymorphic.Merge<
...
      * (default: false)
      */
-    trapped?: boolean;
+    wrapped?: boolean;
+
+    /**
+     * Whether focus should be contained inside the `FocusScope`.
+     * (default: false)
+     */
+    contained?: boolean;
```
andy-hook, line 33: Is it worth being more descriptive on the difference?

packages/react/focus-scope/src/FocusScope.test.tsx
benoitgrelard: We're not really removing it, since a change I made recently, `FocusScope` is actually a primitive rendering an element so here I'm simply removing the `form` but rendering `FocusScope as="form"` instead basically removing an extra wrapper.

packages/react/focus-scope/src/FocusScope.tsx
benoitgrelard: Will do.

packages/react/focus-scope/src/FocusScope.test.tsx
andy-hook: Ah I see it now, was skimming before.

andy-hook (approved):

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -21,10 +21,21 @@ type FocusScopeOwnProps = Polymorphic.Merge<
...
+    wrapped?: boolean;
+
+    /**
+     * Whether focus should be contained inside the `FocusScope`.
+     * If `true`, tabbing out of the last tabbable element inside `FocusScope`
+     * will do nothing. Also, when moving focus programatically outside of it, focus will
+     * be brought back in.
+     * (default: false)
+     */
+    contained?: boolean;
```
jjenzz, line 38: I was bored and wanted to jot down a few things quickly before this is merged:

- Thoughts on `loop` instead of `wrapped`? When something "wraps" something it "encloses" it, so it sounds like this would put the focus around the whole scope (on the content). 
- Thoughts on `trap` instead of `contained`? For some reason `trap` communicates that it cannot escape a bit better to me.
- For `contained` where it says tabbing the last element will do nothing, it makes it sound like `contained` and `wrapped` cannot be used together because they contradict each other.

I think something along these lines is a bit clearer:

```suggestion
    /**
     * When `true`, loops tab focus within the focus scope. Tabbing from last item
     * will focus first tabbable and shift+tab from first item will focus last tababble.
     * @defaultValue false
     */
    loop?: boolean;

    /**
     * When `true`, focus cannot escape the focus scope via tab keypress or a programmatic focus.
     * @defaultValue false
     */
    trap?: boolean;
```

benoitgrelard: I like the suggestions 👍 
On the naming, how do we feel about the verbs vs, past participle stuff? ie. `trap` vs `trapped`?
Also I think we need to think how these get renamed in upper layers.

andy-hook: Really like these suggestions, particularly `trapped`. Past participle is my preference for the consistency.

benoitgrelard: Just catching up with this now, what are we going for then?

- `contained` => `trapped`?
- `loop` => `?` ?

Interesting point to note also that we have a somewhat similar concept with a `loop` prop also on all the things that use roving focus (menus, tabs, radio-group, toolbar), so probably worth taking into consideration in this discussion too.

jjenzz: I guess if we're going past participle then it would be `trapped` and `looped` but i'm also fine with `loop`.

andy-hook: Same. Whichever feels better holistically to you.

andy-hook: Oh and at the risk of bike shedding here, what's been the approach with the past participle stuff in the public apis? I favour intuitive over dogmatic consistency tbh but would we be saying `trapFocus` should become `trappedFocus` (and why is it `loop` and not `loopFocus` or `loopedFocus`?)

Anyway, shut me down if that's out of scope.

jjenzz: > I favour intuitive over dogmatic consistency tbh

Same. The DOM APIs aren't consistent for DOM APIs either tbh. They have a `loop` attribute so my preference would be `trapped` and `loop`. 

> and why is it `loop` and not `loopFocus` or `loopedFocus`?

Since we're discussing props on a `FocusScope` component, it seems redundant to repeat the "focus" part. That would be like having a `cat` object with `catName`, `catAge`, `catBreed` properties otherwise.

```js
const cat = {
	catName: 'Joplin',
	catAge: 2,
	catBreed: 'Maine Coon' 
};

// vs

const cat = {
	name: 'Joplin',
	age: 2,
	breed: 'Maine Coon' 
};
```

benoitgrelard: I'm happy with that 👍

benoitgrelard:
@jjenzz @andy-hook I've incorporated your feedback now 👍

packages/react/focus-scope/src/FocusScope.tsx
andy-hook: > Since we're discussing props on a FocusScope component, it seems redundant to repeat the "focus" part. That would be like having a cat object with catName, catAge, catBreed properties otherwise.

Yeah totally, was referring to public props on primitives eg places where we use `TrapFocus` but that's irrelevant in the context of this PR 👍

jjenzz (approved):

PR #699, opened 2021-06-09
[Dialog] Add support for non-modal dialogs

benoitgrelard:
This PR builds on top of #698 and adds support for non-modal dialogs.

I have done this by splitting the internals in a few components:

- `DialogContent` which will render:
  - `DialogContentModalImpl` or `DialogContentNonModalImpl` which will both render:
    - `DialogContentPart`

Before I go much further, I'm still in 2 minds about this whether it's any clearer or not so I'd love to get your thoughts on it.

What pushed me towards this was that otherwise with a single component, I ended up with lots of `context.modal ? ` checks or wiring throughout the component.

### DIFF
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -159,7 +159,7 @@ const DialogOverlayImpl = React.forwardRef((props, forwardedRef) => {
 const CONTENT_NAME = 'DialogContent';
 
 type DialogContentOwnProps = Polymorphic.Merge<
-  Polymorphic.OwnProps<typeof DialogContentModalImpl | typeof DialogContentNonModalImpl>,
+  Polymorphic.OwnProps<typeof DialogContentModal | typeof DialogContentNonModal>,
   {
     /**
      * Used to force mounting when more control is needed. Useful when
@@ -170,7 +170,7 @@ type DialogContentOwnProps = Polymorphic.Merge<
 >;
 
 type DialogContentPrimitive = Polymorphic.ForwardRefComponent<
-  Polymorphic.IntrinsicElement<typeof DialogContentModalImpl | typeof DialogContentNonModalImpl>,
+  Polymorphic.IntrinsicElement<typeof DialogContentModal | typeof DialogContentNonModal>,
   DialogContentOwnProps
 >;
 
@@ -180,9 +180,9 @@ const DialogContent = React.forwardRef((props, forwardedRef) => {
   return (
     <Presence present={forceMount || context.open}>
       {context.modal ? (
-        <DialogContentModalImpl {...contentProps} ref={forwardedRef} />
+        <DialogContentModal {...contentProps} ref={forwardedRef} />
       ) : (
-        <DialogContentNonModalImpl {...contentProps} ref={forwardedRef} />
+        <DialogContentNonModal {...contentProps} ref={forwardedRef} />
       )}
     </Presence>
   );
@@ -190,17 +190,17 @@ const DialogContent = React.forwardRef((props, forwardedRef) => {
 
 DialogContent.displayName = CONTENT_NAME;
 
-type DialogContentImplOwnProps = Omit<
-  Polymorphic.OwnProps<typeof DialogContentPart>,
+type DialogContentTypeOwnProps = Omit<
+  Polymorphic.OwnProps<typeof DialogContentImpl>,
   'trapFocus' | 'disableOutsidePointerEvents'
 >;
 
-type DialogContentImplPrimitive = Polymorphic.ForwardRefComponent<
-  Polymorphic.IntrinsicElement<typeof DialogContentPart>,
-  DialogContentImplOwnProps
+type DialogContentTypePrimitive = Polymorphic.ForwardRefComponent<
+  Polymorphic.IntrinsicElement<typeof DialogContentImpl>,
+  DialogContentTypeOwnProps
 >;
 
-const DialogContentModalImpl = React.forwardRef((props, forwardedRef) => {
+const DialogContentModal = React.forwardRef((props, forwardedRef) => {
   const context = useDialogContext(CONTENT_NAME);
   const contentRef = React.useRef<HTMLDivElement>(null);
   const composedRefs = useComposedRefs(forwardedRef, contentRef);
@@ -214,10 +214,10 @@ const DialogContentModalImpl = React.forwardRef((props, forwardedRef) => {
   return (
     <Portal>
       <RemoveScroll>
-        <DialogContentPart
+        <DialogContentImpl
           {...props}
           ref={composedRefs}
-          // we make sure focus isn't contained once `DialogContent` has been closed
+          // we make sure focus isn't trapped once `DialogContent` has been closed
           // (closed !== unmounted when animating out)
           trapFocus={context.open}
           disableOutsidePointerEvents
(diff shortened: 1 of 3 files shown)


packages/react/dialog/src/Dialog.tsx
```diff
@@ -164,19 +175,111 @@ const DialogContent = React.forwardRef((props, forwardedRef) => {
...
+  const composedRefs = useComposedRefs(forwardedRef, contentRef);
+
+  // aria-hide everything except the content (better supported equivalent to setting aria-modal)
+  React.useEffect(() => {
+    const content = contentRef.current;
+    if (content) return hideOthers(content);
+  }, []);
+
+  return (
+    <Portal>
```
benoitgrelard, line 211: Annoyingly, we can't put the `Portal` as part of the `DialogContentPart` implementation because it needs to wrap the `RemoveScroll` component so that remove scroll works correctly…

benoitgrelard:
@andy-hook @jjenzz this is now rebased and ready to review again.

benoitgrelard:
Not sure why also, but Chromatic seems to be off here, it's showing as if we've removed `DropdownMenu` submenu stuff… Not sure how to fix that…

packages/react/dialog/src/Dialog.tsx
```diff
@@ -114,37 +124,38 @@ type DialogOverlayPrimitive = Polymorphic.ForwardRefComponent<
...
   Polymorphic.IntrinsicElement<typeof Primitive>,
   DialogOverlayImplOwnProps
 >;
 
 const DialogOverlayImpl = React.forwardRef((props, forwardedRef) => {
+  const context = useDialogContext(OVERLAY_NAME);
   return (
     <Portal>
-      <Primitive {...props} ref={forwardedRef} />
+      <Primitive data-state={getState(context.open)} {...props} ref={forwardedRef} />
```
jjenzz, line 146: :+1:

packages/react/dialog/src/Dialog.tsx
```diff
@@ -82,7 +92,7 @@ const DialogTrigger = React.forwardRef((props, forwardedRef) => {
       {...triggerProps}
       as={as}
       ref={composedTriggerRef}
-      onClick={composeEventHandlers(props.onClick, () => context.onOpenChange(true))}
+      onClick={composeEventHandlers(props.onClick, context.onOpenToggle)}
```
jjenzz, line 95: Why do dialog triggers toggle now? Is that a thing? 🤔

packages/react/dialog/src/Dialog.tsx
```diff
@@ -164,19 +175,112 @@ const DialogContent = React.forwardRef((props, forwardedRef) => {
...
+  Polymorphic.OwnProps<typeof DialogContentPart>,
+  'trapFocus' | 'disableOutsidePointerEvents'
+>;
+
+type DialogContentImplPrimitive = Polymorphic.ForwardRefComponent<
+  Polymorphic.IntrinsicElement<typeof DialogContentPart>,
+  DialogContentImplOwnProps
+>;
+
+const DialogContentModalImpl = React.forwardRef((props, forwardedRef) => {
```
jjenzz, line 199: Wondering why these aren't just named `DialogContentModal` and `DialogContentNonModal`? I thought we usually only have `Impl` parts when the former is being used and we need it to render a separate part internally.

I went looking for their respective non-impl parts and couldn't find them.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -164,19 +175,112 @@ const DialogContent = React.forwardRef((props, forwardedRef) => {
...
+  // aria-hide everything except the content (better supported equivalent to setting aria-modal)
+  React.useEffect(() => {
+    const content = contentRef.current;
+    if (content) return hideOthers(content);
+  }, []);
+
+  return (
+    <Portal>
+      <RemoveScroll>
+        <DialogContentPart
```
jjenzz, line 213: Can we rename this to `DialogContentImpl` for consistency? It looks like you might have copied `Slider` but I've just updated that one too to match everything else. It was the only component that did this "Part" thing, everywhere else we use "Impl".

packages/react/dialog/src/Dialog.tsx
```diff
@@ -164,19 +175,112 @@ const DialogContent = React.forwardRef((props, forwardedRef) => {
...
+          {...props}
+          ref={composedRefs}
+          // we make sure focus isn't contained once `DialogContent` has been closed
```
jjenzz, line 216:
```suggestion
          // we make sure focus isn't trapped once `DialogContent` has been closed
```

jjenzz (changes requested):

packages/react/dialog/src/Dialog.tsx
benoitgrelard: Now that this introduces non-modal dialogs then yes you can click the trigger again, so feels like we should have the same default as for popover for example.

packages/react/dialog/src/Dialog.tsx
benoitgrelard: I think the `Impl` naming convention we have is a bit general, not always due to `Presence` or similar. That said, I'm very happy to change this here.

packages/react/dialog/src/Dialog.tsx
benoitgrelard: You mean instead of `DialogContentPart`?

benoitgrelard: Haha yeah I think I looked at `Slider` 😀

andy-hook:
> Chromatic seems to be off here, it's showing as if we've removed DropdownMenu submenu stuff… Not sure how to fix that…

I'm super confused by this too 🤔  could it be their service having a hiccup? I can't find any changes that would cause it..

packages/react/dialog/src/Dialog.tsx
jjenzz: This still feels weird to me but i guess we can just go with it for now and see what feedback we get. Unless @colmtuite has any input?

packages/react/dialog/src/Dialog.tsx
jjenzz: I usually expect that when there is an `Impl` part that there is an associated non-impl part which is why the impl name had to be used. That's not the case here though which tripped me up. Would be great to change if you wouldn't mind 🙏

packages/react/dialog/src/Dialog.tsx
jjenzz: > You mean instead of `DialogContentPart`?
Yep :+1:

benoitgrelard:
Incorporated your feedback @jjenzz

packages/react/dialog/src/Dialog.tsx
```diff
@@ -123,37 +128,38 @@ type DialogOverlayPrimitive = Polymorphic.ForwardRefComponent<
...
-
 /* -------------------------------------------------------------------------------------------------
  * DialogContent
  * -----------------------------------------------------------------------------------------------*/
 
 const CONTENT_NAME = 'DialogContent';
 
 type DialogContentOwnProps = Polymorphic.Merge<
-  Polymorphic.OwnProps<typeof DialogContentImpl>,
+  Polymorphic.OwnProps<typeof DialogContentModal | typeof DialogContentNonModal>,
```
jjenzz, line 162: Interesting, does this work? I think I usually do the following for this sort of thing: 

```jsx
type DialogContentOwnProps = Polymorphic.Merge<
	| Polymorphic.OwnProps<typeof DialogContentModal>
	| Polymorphic.OwnProps<typeof DialogContentNonModal>,
	{ ... }
> 
```

If yours works we can probably clean up a *bunch* of boilerplate elsewhere 😅

jjenzz (approved):

packages/react/dialog/src/Dialog.tsx
benoitgrelard: 😬 as far as I can tell, but I think they also both have the same API so I wouldn't know.
I guess we should try it on APIs were it does make a difference?

benoitgrelard: My gut is now telling me that this is probably wrong 😀

PR #707, opened 2021-06-14
[ContextMenu][DropdownMenu] Fix menu focus after pointer interaction

andy-hook:
See https://github.com/radix-ui/primitives/pull/682#issuecomment-860507424


benoitgrelard (commented):
I would have thought this takes care of it already no?
Is something not working correctly here?

```tsx
onOpenAutoFocus={(event) => {
  // when opening a submenu, focus content for keyboard users only
  if (context.isUsingKeyboardRef.current) ref.current?.focus();
  event.preventDefault();
}}
```

andy-hook:
That handler only gets called on mount (in this case the original pointer open) it doesn't get run again on keyboard navigation.

benoitgrelard:
Ah yes great point, thanks. Perhaps we should add a comment explaining that as the code kinda looks redundant at first glance.

benoitgrelard (approved):

PR #709, opened 2021-06-14
[ContextMenu][DropdownMenu] Improve touch/mouse handling separation

benoitgrelard:
This PR fixes some issues with touch events (see #682 or below) by being more explicit about how we handle specifically touch vs. mouse using pointer events.

> With touch inputs there is some odd behaviour when layered click targets overlap e.g. after a collision adjustment (seen on IOS but can reproduce in emulated dev tool)

https://user-images.githubusercontent.com/11708259/121864338-b4108480-ccf4-11eb-8e7d-c914ae84de56.mp4

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -571,14 +571,15 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
                       searchRef.current = '';
                     }
                   })}
-                  onPointerMove={composeEventHandlers(props.onPointerMove, (event) => {
-                    if (event.pointerType === 'mouse') {
+                  onPointerMove={composeEventHandlers(
+                    props.onPointerMove,
+                    whenMouse((event) => {
                       const target = event.target as HTMLElement;
                       if (event.currentTarget.contains(target) && event.movementX !== 0) {
                         pointerDirRef.current = event.movementX > 0 ? 'right' : 'left';
                       }
-                    }
-                  })}
+                    })
+                  )}
                 />
               </RovingFocusGroup>
             </DismissableLayer>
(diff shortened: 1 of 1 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -572,10 +571,12 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
...
-                  onMouseMove={composeEventHandlers(props.onMouseMove, (event) => {
-                    const target = event.target as HTMLElement;
-                    if (event.currentTarget.contains(target) && event.movementX !== 0) {
-                      pointerDirRef.current = event.movementX > 0 ? 'right' : 'left';
+                  onPointerMove={composeEventHandlers(props.onPointerMove, (event) => {
+                    if (event.pointerType === 'mouse') {
+                      const target = event.target as HTMLElement;
+                      if (event.currentTarget.contains(target) && event.movementX !== 0) {
+                        pointerDirRef.current = event.movementX > 0 ? 'right' : 'left';
+                      }
```
andy-hook, line 579: Do you think it would be cleaner to wrap these conditions up in a self documenting util? [we did something like this before](https://github.com/radix-ui/primitives/blob/ceeb027884a110611467d91287baf47cf6fccfdf/packages/react/hover-card/src/HoverCard.tsx#L196-L199)

What do you think?

benoitgrelard: I like that. There are 3 pointer inputs right? `mouse`, `pen`, `touch`. How should we treat `pen`? My guess would be that `pen` is basically a less coarse version of `touch` so we should treat them the same?

andy-hook: Yeah I'd agree with that, at least in our general cases.

andy-hook (approved):

Issue #714, opened 2021-06-16
[DropdownMenu] can't hover sub menus in Safari in some cases


hadihallak:
### Current Behavior
Trying to hover a sub-menu will hide it

https://user-images.githubusercontent.com/13556017/122227142-4e242880-cebf-11eb-9348-912c227f453f.mov

### Expected behavior

should be able to hover sub-menus

### Reproducible example

[CodeSandbox Template](https://t.co/loQ9XX1uB3?amp=1)

### Additional context

Note that this is only breaking for me in the collision detection demo and that it works fine in the [other centered demo](https://jmq8h.csb.app/) 
### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |     @radix-ui/react-dropdown-menu    |     0.0.20    |
| React            | n/a     |         |
| Browser          |    Safari     |    14.1.1     |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

PR #716, opened 2021-06-16
[DropdownMenu][ContextMenu] Submenu Safari compat

andy-hook:
Fixes https://github.com/radix-ui/primitives/issues/714

The issue was caused by a wrongly reported direction in Safari because it [inconsistently always returns `0`](https://codesandbox.io/s/silent-currying-wn7ny) on `event.movementX` only when it's a pointer event.

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -566,11 +566,11 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
                     props.onPointerMove,
                     whenMouse((event) => {
                       const target = event.target as HTMLElement;
-                      const valueHasChanged = lastPointerXRef.current !== event.clientX;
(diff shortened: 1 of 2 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -565,8 +566,14 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
                     props.onPointerMove,
                     whenMouse((event) => {
                       const target = event.target as HTMLElement;
-                      if (event.currentTarget.contains(target) && event.movementX !== 0) {
-                        pointerDirRef.current = event.movementX > 0 ? 'right' : 'left';
+                      const valueHasChanged = lastPointerXRef.current !== event.clientX;
```
benoitgrelard, line 569: Can we perhaps name this `pointerXHasChanged`? It's a bit too generic otherwise.

benoitgrelard (approved):

PR #720, opened 2021-06-17
[ScrollArea] Add `data-orientation` to `Scrollbar`

### LINKED ISSUE #719: [ScrollArea] `Scrollbar` part doesn't have `data-orientation` attribute for styling convenience
### Current Behavior

Currently to properly style horizontal and vertical scrollbar, you'd have to create 2 different styled components, or use 2 classnames, etc.

### Expected behavior

Generally we expose `data-orientation` for this purpose so we can style conditionally inside one class.

benoitgrelard:
Fix #719

### DIFF
--- a/.yarn/versions/8f41e580.yml
+++ b/.yarn/versions/8f41e580.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-scroll-area": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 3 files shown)


andy-hook (approved):
lgtm. It might be worth us adding an example of using multiple bars when we update the docs.

Issue #593, opened 2021-04-13
[Accordion] Closing an item in a `type="single"` controlled accordion requires 2 clicks


benoitgrelard:
cc @peduarte

PR #722, opened 2021-06-17
[ToggleGroup] Fix issue when unpressing an item in single controlled group

### LINKED ISSUE #721: [ToggleGroup] Unpressing an item in a type="single" controlled `ToggleGroup` requires 2 clicks
Exactly the same problem we had in #593 and treated in  #594

benoitgrelard:
Fixes #721

This was the exact same issue we had with `Accordion` in #593 and treated in #594.

### DIFF
--- a/packages/react/toggle-group/src/ToggleGroup.test.tsx
+++ b/packages/react/toggle-group/src/ToggleGroup.test.tsx
@@ -47,7 +47,7 @@ describe('given a single ToggleGroup', () => {
           fireEvent.click(two);
         });
 
-        it('should change value to `""`', () => {
+        it('should change value to empty string', () => {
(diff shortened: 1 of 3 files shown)


packages/react/toggle-group/src/ToggleGroup.test.tsx
```diff
@@ -47,8 +47,8 @@ describe('given a single ToggleGroup', () => {
           fireEvent.click(two);
         });
 
-        it('should change value to `undefined`', () => {
-          expect(handleValueChange).toHaveBeenCalledWith(undefined);
+        it('should change value to `""`', () => {
```
andy-hook, line 50:
```suggestion
        it('should change value to empty string', () => {
```

Perhaps?

andy-hook (approved):

PR #724, opened 2021-06-17
[RadioGroup] Remove unnecessary dep

benoitgrelard:
Same, just a small thing I saw whilst opening files.


andy-hook (approved):
Good spot, i'm surprised these make it past our lint check. Normally not a big deal but perhaps we should update them to fail on prepush/ci to cover cases where we might have included unneeded deps like this.

Issue #615, opened 2021-04-23
[Slider] Thumb loses focus when value changes


jjenzz:
Originally reported here: https://discord.com/channels/752614004387610674/803656530259738674/834838499906551869

Strangely this cannot be reproduced in our project or in a CodeSandbox but I was able to decipher that the `Thumb` is unmounting/mounting when the value changes.

We have the following logic that renders the thumb:

```jsx
return value !== undefined ? <SliderThumbImpl ... /> : null;
```

However, some local logging in Darko's project shows that the `value` never changes to `undefined` so I'm baffled. If I remove this conditional though, everything works fine in his project. 

I'm parking for now as we have other priorities but one to investigate when we have time.

jjenzz:
This can be easily replicated by spinning up a new snowpack build from this template https://github.com/snowpackjs/snowpack/tree/main/create-snowpack-app/app-template-react-typescript

```
npx create-snowpack-app react-snowpack --template @snowpack/app-template-react-typescript
```

jjenzz:
Issue also occurring when there are multiple thumbs in a CRA project

andy-hook:
Related https://github.com/radix-ui/primitives/issues/615

I was curious so did some digging on this today. The issue originates in https://github.com/radix-ui/primitives/pull/731 but is not the cause, I think it's an underlying issue that's been obfuscated.

I'm 99% sure it's caused by `StrictMode` exposing non-idempotent code in the [collection index handling](https://github.com/radix-ui/primitives/blob/63b54f0a3b242eeef767587f9714acb9a131dd5f/packages/react/slider/src/collection.tsx#L18-L23), this is why we see it occur in CLI scaffolded projects like CRA, create-snowpack-app etc as they most likely ship with it enabled. `StrictMode` also only runs in development which maps to other reports we've [had in Discord](https://discord.com/channels/752614004387610674/803656530259738674/867299726998765608)

Repro: https://codesandbox.io/s/awesome-wozniak-tsokb?file=/App.js

I think the basis of the problem is that code is running more often than expected in `Strict` making the index [returned in `useSliderCollectionItem` ](https://github.com/radix-ui/primitives/blob/63b54f0a3b242eeef767587f9714acb9a131dd5f/packages/react/slider/src/Slider.tsx#L496-L502) off by one (thus `value === undefined`)

[...]

PR #731, opened 2021-06-21
[Slider] Move focus to correct thumb when using keyboard

### LINKED ISSUE #730: [Slider] Focus doesn't jump to next thumb when crossing in multiple slider type
### Current Behavior

Focus remains on left thumb

### Expected behavior

Should move to right thumb

### Reproducible example

https://user-images.githubusercontent.com/1539897/122745848-572b4600-d281-11eb-92f6-00b250742af2.mov

jjenzz:
Fixes #730

- Thumb focusing logic fires every time value changes (this fixes `Home` and `End` keypress too)
- Updated to use PointerEvents instead which meant I could remove some `setTimeout` logic
- Renamed `SliderPart` to `SliderImpl` for consistency

There is the following outstanding issue which I will fix in a separate PR:

https://user-images.githubusercontent.com/175330/122761147-d164c600-d293-11eb-8ad9-0c3ac7621a9c.mp4

When the thumb reaches the same value as another thumb, I have to press the arrow one extra time than I should need to for it to move.

### DIFF
--- a/packages/react/slider/package.json
+++ b/packages/react/slider/package.json
@@ -26,6 +26,7 @@
     "@radix-ui/react-primitive": "workspace:*",
     "@radix-ui/react-use-controllable-state": "workspace:*",
     "@radix-ui/react-use-direction": "workspace:*",
+    "@radix-ui/react-use-layout-effect": "workspace:*",
     "@radix-ui/react-use-previous": "workspace:*",
     "@radix-ui/react-use-size": "workspace:*"
   },
--- a/packages/react/slider/package.json
+++ b/packages/react/slider/package.json
@@ -24,7 +24,6 @@
     "@radix-ui/react-context": "workspace:*",
     "@radix-ui/react-polymorphic": "workspace:*",
     "@radix-ui/react-primitive": "workspace:*",
-    "@radix-ui/react-use-callback-ref": "workspace:*",
     "@radix-ui/react-use-controllable-state": "workspace:*",
     "@radix-ui/react-use-direction": "workspace:*",
     "@radix-ui/react-use-previous": "workspace:*",
--- a/packages/react/slider/src/Slider.tsx
+++ b/packages/react/slider/src/Slider.tsx
@@ -3,7 +3,6 @@ import { clamp } from '@radix-ui/number';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { createContext } from '@radix-ui/react-context';
-import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
 import { useControllableState } from '@radix-ui/react-use-controllable-state';
 import { useDirection } from '@radix-ui/react-use-direction';
 import { usePrevious } from '@radix-ui/react-use-previous';
--- a/packages/react/slider/src/Slider.tsx
+++ b/packages/react/slider/src/Slider.tsx
@@ -92,54 +92,36 @@ const Slider = React.forwardRef((props, forwardedRef) => {
   const [values = [], setValues] = useControllableState({
     prop: value,
     defaultProp: defaultValue,
-    onChange: onValueChange,
+    onChange: (value) => {
+      const thumbs = [...thumbRefs.current];
+      thumbs[valueIndexToChangeRef.current]?.focus();
+      onValueChange(value);
+    },
   });
 
   function handleSlideStart(value: number) {
     const closestIndex = getClosestValueIndex(values, value);
-    updateValues(value, closestIndex).then((valueIndexToChange) => {
-      /**
-       * Browsers fire event handlers before executing their event implementation
-       * so they can check if `preventDefault` was called first. Therefore,
-       * if we focus the thumb on slide start (`mousedown`), the browser will execute
-       * their `mousedown` implementation after our focus which will instantly
-       * `blur` the thumb again (because it effectively clicks off the thumb).
-       *
-       * We use a `setTimeout`  to move the focus to the next tick (after the
-       * mousedown) to ensure focus on mousedown.
-       */
-      window.setTimeout(() => focusThumb(valueIndexToChange), 0);
-    });
+    updateValues(value, closestIndex);
   }
 
   function handleSlideMove(value: number) {
-    updateValues(value, valueIndexToChangeRef.current).then(focusThumb);
+    updateValues(value, valueIndexToChangeRef.current);
   }
 
-  function updateValues(value: number, atIndex: number): Promise<number> {
+  function updateValues(value: number, atIndex: number) {
     const decimalCount = getDecimalCount(step);
     const snapToStep = roundValue(Math.round((value - min) / step) * step + min, decimalCount);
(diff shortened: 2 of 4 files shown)


packages/react/slider/src/Slider.tsx
```diff
@@ -465,19 +376,29 @@ const SliderPart = React.forwardRef((props, forwardedRef) => {
...
-       *
-       * Also ensures that slider receives all pointer events after mouse down
-       * even when mouse moves outside the document.
-       */
       onPointerDown={composeEventHandlers(props.onPointerDown, (event) => {
-        (event.target as HTMLElement).setPointerCapture(event.pointerId);
+        const target = event.target as HTMLElement;
+        target.setPointerCapture(event.pointerId);
+        // Prevent browser focus behaviour because we focus a thumb manually when values change.
+        event.preventDefault();
```
jjenzz, line 383: This change allowed me to remove the `setTimeout` stuff from [here](https://github.com/radix-ui/primitives/pull/731/files#diff-6bc9aae1ceceb7cfd3e1b1f419113c46a6d88dd4515e280bac28c9b1313eae99L101). We couldn't do this before because of [this](https://github.com/radix-ui/primitives/pull/731/files#diff-6bc9aae1ceceb7cfd3e1b1f419113c46a6d88dd4515e280bac28c9b1313eae99L440) but we're no longer using mouse events.

packages/react/slider/src/Slider.tsx
```diff
@@ -465,19 +376,29 @@ const SliderPart = React.forwardRef((props, forwardedRef) => {
...
+        // Touch devices have a delay before focusing so won't focus if touch immediately moves
+        // away from target (sliding). We want thumb to focus regardless.
+        if (context.thumbs.has(target)) {
```
jjenzz, line 386: Checking context instead of using the `isThumb` method. This allowed me to remove the `data-radix-slider-thumb` attribute that we had left over because of that `isThumb` util.

packages/react/slider/src/Slider.tsx
```diff
@@ -92,56 +92,38 @@ const Slider = React.forwardRef((props, forwardedRef) => {
...
-       * Browsers fire event handlers before executing their event implementation
-       * so they can check if `preventDefault` was called first. Therefore,
-       * if we focus the thumb on slide start (`mousedown`), the browser will execute
-       * their `mousedown` implementation after our focus which will instantly
-       * `blur` the thumb again (because it effectively clicks off the thumb).
-       *
-       * We use a `setTimeout`  to move the focus to the next tick (after the
-       * mousedown) to ensure focus on mousedown.
-       */
-      window.setTimeout(() => focusThumb(valueIndexToChange), 0);
```
benoitgrelard, line 111: So good that we can remove this! Were you able to check the regression on this one? I can't remember myself how to check for it.

packages/react/slider/src/Slider.tsx
```diff
@@ -390,70 +347,24 @@ type SliderPartPrimitive = Polymorphic.ForwardRefComponent<
...
-  const removeMouseEventListeners = useCallbackRef(() => {
-    document.removeEventListener('mousemove', handleSlideMouseMove);
-    document.removeEventListener('mouseup', removeMouseEventListeners);
-    onSlideMouseUp();
-  });
-  const removeTouchEventListeners = useCallbackRef(() => {
-    document.removeEventListener('touchmove', handleSlideTouchMove);
-    document.removeEventListener('touchend', removeTouchEventListeners);
-    onSlideTouchEnd();
-  });
```
benoitgrelard, line 419: So just so I'm clear, we don't need the document handlers because we now use pointer events with pointer capture? So even if we move quickly and our cursor escapes the thumb it will still work?

packages/react/slider/src/Slider.tsx
```diff
@@ -92,56 +92,38 @@ const Slider = React.forwardRef((props, forwardedRef) => {
   const [values = [], setValues] = useControllableState({
     prop: value,
     defaultProp: defaultValue,
-    onChange: onValueChange,
+    onChange: (value) => {
+      const thumbs = [...thumbRefs.current];
+      thumbs[valueIndexToChangeRef.current]?.focus();
```
benoitgrelard, line 97: Isn't this an issue if someone is controlling the slider from another input programatically? focus would jump to it right?

benoitgrelard:
Love all the cleanup!
Wondering if these changes also fix #615 possibly?

packages/react/slider/src/Slider.tsx
jjenzz: Yeah, to test you have to click somewhere on the track and the thumb should focus.

benoitgrelard: That was the one! Thanks! 🙏

packages/react/slider/src/Slider.tsx
jjenzz: We were using document events before so that they were only bound when a drag started. With pointer events we can check `hasPointerCapture` to determine if they're dragging instead.

packages/react/slider/src/Slider.tsx
jjenzz: Ah yeah... hmmm... good point. Let me rethink this.

jjenzz:
> Wondering if these changes also fix #615 possibly?

I wondered that too but we're unable to replicate the issue locally. When we release this I'll pull into that repo that I have locally and check.

jjenzz:
In fact, if I run a `build` and copy paste build output into that repo I can verify that way. Will give that a go.

packages/react/slider/src/Slider.tsx
benoitgrelard: nice 👍

packages/react/slider/src/Slider.tsx
jjenzz: Ah, no this isn't a problem because this only fires when `setValue` is called internally. If the value is changed externally we don't call `setValue`, you'd end up in a loop otherwise.

benoitgrelard: Ah yes because we took care of that in our `useControllableState` hook, that's right! Awesome then!

benoitgrelard (approved):

packages/react/slider/package.json
```diff
@@ -24,9 +24,9 @@
     "@radix-ui/react-context": "workspace:*",
     "@radix-ui/react-polymorphic": "workspace:*",
     "@radix-ui/react-primitive": "workspace:*",
-    "@radix-ui/react-use-callback-ref": "workspace:*",
     "@radix-ui/react-use-controllable-state": "workspace:*",
     "@radix-ui/react-use-direction": "workspace:*",
+    "@radix-ui/react-use-layout-effect": "workspace:*",
```
jjenzz, line 29: We use this package in `collection.tsx` that was moved into `Slider` a while back but we forgot to add the dep.

jjenzz:
> Wondering if these changes also fix #615 possibly?

Nope, looks like this PR actually makes it worse and I can't figure out why snowpack is reinstantiating the thumb, it's so weird 😬

PR #733, opened 2021-06-21
[Slider] Fix sticky value with keyboard

### LINKED ISSUE #732: [Slider] Extra keypress required to move thumb
https://user-images.githubusercontent.com/175330/122761147-d164c600-d293-11eb-8ad9-0c3ac7621a9c.mp4

When the thumb reaches the same value as another thumb, I have to press the arrow one extra time than I should need to for it to move.

jjenzz:
Fixes #732

### DIFF
--- a/packages/react/slider/src/Slider.tsx
+++ b/packages/react/slider/src/Slider.tsx
@@ -116,7 +116,7 @@ const Slider = React.forwardRef((props, forwardedRef) => {
       const nextValues = getNextSortedValues(prevValues, nextValue, atIndex);
       if (hasMinStepsBetweenValues(nextValues, minStepsBetweenThumbs * step)) {
         valueIndexToChangeRef.current = nextValues.indexOf(nextValue);
-        return nextValues[atIndex] !== prevValues[atIndex] ? nextValues : prevValues;
+        return String(nextValues) === String(prevValues) ? prevValues : nextValues;
       } else {
         return prevValues;
       }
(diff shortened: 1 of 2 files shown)


packages/react/slider/src/Slider.tsx
```diff
@@ -117,7 +117,7 @@ const Slider = React.forwardRef((props, forwardedRef) => {
       const nextValues = getNextSortedValues(prevValues, nextValue, atIndex);
       if (hasMinStepsBetweenValues(nextValues, minStepsBetweenThumbs * step)) {
         valueIndexToChangeRef.current = nextValues.indexOf(nextValue);
-        return nextValues[atIndex] !== prevValues[atIndex] ? nextValues : prevValues;
+        return String(nextValues) === String(prevValues) ? prevValues : nextValues;
```
benoitgrelard, line 120: I understand the change, but I can't wrap my head around why the previous way would create the keyboard bug yet this wouldn't. 🤔

benoitgrelard (approved):

packages/react/slider/src/Slider.tsx
jjenzz: For anyone else following along, without this change we ended up with something like this:

```
[15,16,20] = atIndex = 0, valueIndexToChange = 0
[16,16,20] = atIndex = 0, valueIndexToChange = 0
[16,17,20] = atIndex = 0, valueIndexToChange = 1
```

That last array would then check `nextValues[0] !== prevValues[0]` (`16 !== 16`) which would be equal and return the `prevValues`.

PR #735, opened 2021-06-21
Add initial Cypress configuration

andy-hook:
Originally tested the [component test runner](https://docs.cypress.io/guides/component-testing/introduction) but a prerequisite is to use webpack for compilation and we'd prefer to avoid that additional complexity right now.

I've added a simple test to `Dialog` to ensure things work as expected.

In the future we might want to colocate these alongside package code but for now I think starting simple is best.

The biggest issue I ran into was `detached from DOM` errors when executing `click` after initial load. Possibly related to React lifecycle / storybook, I did a lot of research but most reported solutions were either problematic or convoluted. The simplest way I've found is to force a next tick by chaining `delay(0)` after the `visit()`.

More context here:
https://github.com/cypress-io/cypress/issues/7306
https://twitter.com/kentcdodds/status/1239658244387115008


cypress/support/commands.js
```diff
@@ -0,0 +1,30 @@
...
+// Cypress.Commands.add("dismiss", { prevSubject: 'optional'}, (subject, options) => { ... })
+//
+//
+// -- This will overwrite an existing command --
+// Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+Cypress.Commands.add('visitStory', (storyName, options) => {
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  return cy.visit(`iframe.html?id=components-${storyName}`, options).wait(0);
+});
```
benoitgrelard, line 30: Love that neat custom command!

cypress/integration/Dialog.ts
```diff
@@ -0,0 +1,11 @@
+describe('Dialog', () => {
```
benoitgrelard, line 1: I agree it would be good to collocate these, but happy for us to look at this in a separate PR.

package.json
```diff
@@ -11,8 +11,11 @@
     "lint": "eslint packages --ext .js,.ts,.tsx",
     "types:check": "tsc --skipLibCheck --noEmit",
     "test": "jest",
-    "test:ci": "jest --ci",
+    "test:ci": "jest --ci && yarn cypress:ci",
     "storybook": "start-storybook -p 9009 --ci",
+    "cypress:ci": "start-server-and-test storybook http://localhost:9009 cypress:run",
```
benoitgrelard, line 16: Is the intention to run this only on CI then?

benoitgrelard (commented):
♥️

package.json
andy-hook: To test locally we’d use `cypress:dev` for the interactive runner or `cypress:run` for headless, it’s implicit that storybook also needs to be running though. Happy to make it more explicit if we think it’d be easier.

cypress/integration/Dialog.ts
andy-hook: Definitely, I'd prefer to write a few bigger tests first to get a feel for the setup and uncover any existing issues before moving to a monorepo setup.

package.json
benoitgrelard: I’m ok with it as is t be honest

benoitgrelard (approved):

jjenzz (approved):
Doesn't look like you hit the same babel config issues I was having? Strange, no idea what I'd done 😅

andy-hook:
> Doesn't look like you hit the same babel config issues I was having? Strange, no idea what I'd done 😅

haha I had them too, luckily bumping `babel/core` solved them.

PR #737, opened 2021-06-22
[Slot] Improve types compatibility

### LINKED ISSUE #736: Trying to render a Trigger component as a Slot throws type error
### Current Behavior

After updating radix packages to their latest versions, I noticed that for 3 of them (@radix-ui/react-dialog, @radix-ui/react-alert-dialog and @radix-ui/react-popover), when I try to render their `Trigger` components `as` `Slot` (@radix-ui/react-slot), I get some typing error, preventing me from doing so.

### Expected behavior

As with previous versions, the `Trigger` components in the above mentioned packages should not throw any errors when I try to render them as `Slot`s.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/gracious-kepler-6eki9?file=/src/Component.tsx)

### Additional context

If I set the packages to older versions, everything works as expected: https://codesandbox.io/s/optimistic-keldysh-b8wss
It seems that even with react-slot on the latest version, the example works as long as react-popover is @<=0.0.15.

### Your environment

benoitgrelard:
Fixes #736

### DIFF
--- a/packages/react/slot/src/Slot.tsx
+++ b/packages/react/slot/src/Slot.tsx
@@ -5,7 +5,7 @@ import { composeRefs } from '@radix-ui/react-compose-refs';
  * Slot
  * -----------------------------------------------------------------------------------------------*/
 
-type SlotProps = { children: React.ReactNode };
+type SlotProps = { children?: React.ReactNode };
 
 const Slot = React.forwardRef<HTMLElement, SlotProps>((props, forwardedRef) => {
   const { children, ...slotProps } = props;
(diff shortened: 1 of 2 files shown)


packages/react/slot/src/Slot.tsx
```diff
@@ -5,7 +5,7 @@ import { composeRefs } from '@radix-ui/react-compose-refs';
  * Slot
  * -----------------------------------------------------------------------------------------------*/
 
-type SlotProps = { children: React.ReactNode };
+type SlotProps = { children?: React.ReactNode };
```
andy-hook, line 8: What was going wrong under the hood here? I looked at the linked sandbox but the polymorphic errors are a bit cryptic and I'm unclear on how this change resolves it 😅

jjenzz (approved):

andy-hook (approved):

PR #738, opened 2021-06-22
[ScrollArea] Fix `forceMount` type on `Scrollbar`

### LINKED ISSUE #728: [ScrollArea] ScrollArea.Scrollbar loses types after being put through a Stitches styled call
### Current Behavior
When styling the Scrollbar of ScrollArea like this:
```
const StyledScrollbar = styled(ScrollArea.Scrollbar, {
...
});
```

Where styled is a stitches styled, and if `forceMount` is passed on or any of the primitive props to it, there are Typescript errors:

image

### Expected behaviour

Types are preserved even after putting the component through a JSS styling function.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/jovial-payne-j2dqg?file=/src/App.tsx)

### Additional context

The bug occurs after wrapping a Radix component in a Stitches `styled` function. None of the props exposed and documented in the docs are then recognized by typescript.

### Your environment

| Radix Package(s) |    ScrollArea     |     0.0.14    |
| Browser          |    Chrome     |    92.0.4515.51     |
| Node             | n/a     |     15    |
| npm/yarn         |    yarn     |    1.22.5     |
| Operating System |    MacOS     |    Big Sur 11.3.1     |

jjenzz:
Fixes #728 

The bug was caused because of the `ScrollAreaScrollbarVisible` part of the union which doesn't have a `forceMount` prop. It looks like this was confusing the type implementation for Stitches and styled-components.

It works fine without these libraries e.g. `React.ComponentProps<typeof ScrollArea.Scrollbar>` reports fine but I've made this change anyway to please external libs. Not sure why they break.

### DIFF
--- a/packages/react/scroll-area/src/ScrollArea.tsx
+++ b/packages/react/scroll-area/src/ScrollArea.tsx
@@ -206,6 +206,7 @@ type ScrollAreaScrollbarPrimitive = Polymorphic.ForwardRefComponent<
 >;
 
 const ScrollAreaScrollbar = React.forwardRef((props, forwardedRef) => {
+  const { forceMount, ...scrollbarProps } = props;
   const context = useScrollAreaContext(SCROLLBAR_NAME);
   const { onScrollbarXEnabledChange, onScrollbarYEnabledChange } = context;
   const isHorizontal = props.orientation === 'horizontal';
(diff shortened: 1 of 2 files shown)


packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -196,17 +196,12 @@ ScrollAreaViewport.displayName = VIEWPORT_NAME;
...
 const SCROLLBAR_NAME = 'ScrollAreaScrollbar';
 
-type ScrollAreaScrollbarOwnProps =
-  | Polymorphic.OwnProps<typeof ScrollAreaScrollbarAuto>
-  | Polymorphic.OwnProps<typeof ScrollAreaScrollbarHover>
-  | Polymorphic.OwnProps<typeof ScrollAreaScrollbarScroll>
-  | Polymorphic.OwnProps<typeof ScrollAreaScrollbarVisible>;
-
+type ScrollAreaScrollbarOwnProps = Polymorphic.Merge<
+  Polymorphic.OwnProps<typeof ScrollAreaScrollbarVisible>,
```
jjenzz, line 200: They all eventually render a `ScrollAreaScrollbarVisible`. Not ideal that we have to type this assumption here but not sure what else we can do.

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -196,17 +196,12 @@ ScrollAreaViewport.displayName = VIEWPORT_NAME;
 
 const SCROLLBAR_NAME = 'ScrollAreaScrollbar';
 
-type ScrollAreaScrollbarOwnProps =
-  | Polymorphic.OwnProps<typeof ScrollAreaScrollbarAuto>
-  | Polymorphic.OwnProps<typeof ScrollAreaScrollbarHover>
-  | Polymorphic.OwnProps<typeof ScrollAreaScrollbarScroll>
-  | Polymorphic.OwnProps<typeof ScrollAreaScrollbarVisible>;
-
+type ScrollAreaScrollbarOwnProps = Polymorphic.Merge<
```
benoitgrelard, line 199: ok so we end up with simplified types whilst slightly inaccurate in the sense that if you use the visible one, TS will let you add a `forceMount` prop without complaints even though it doesn't apply?

packages/react/scroll-area/src/ScrollArea.tsx
benoitgrelard: ok so we end up with simplified types whilst slightly inaccurate in the sense that if you use the visible one, TS will let you add a forceMount prop without complaints even though it doesn't apply?

jjenzz: If you use `type="always"` it will say it accepts a `forceMount` prop which is kinda irrelevant there because `type="always"` is always mounted by nature of the type haha. 

So yeah, it will accept the `forceMount` prop but do nothing with it.

benoitgrelard (approved):

PR #739, opened 2021-06-22
[Popover] Explicitly focus trigger on unmount

### LINKED ISSUE #518: [FocusScope] Focus not restored to trigger when popover contains input with autoFocus attribute
### Current Behavior

In a `@radix-ui/react-popover`, focus is not restored to the trigger on close (`ESC`) if the popover contains an input with the `autoFocus` attribute.

### Expected behavior

Focus is restored to the trigger on close.

### Reproducible example

[Codesandbox reproduction this way ==>](https://codesandbox.io/s/amazing-blackwell-61llb)

### Additional comment

I don't know if this is a bug per se; it might be a misuse of the `autoFocus` attribute on my part. But since the behaviour caught me off guard I'm opening this issue.

jjenzz:
Fixes #518

### DIFF
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -240,7 +240,10 @@ const DialogContentImpl = React.forwardRef((props, forwardedRef) => {
             // (closed !== unmounted when animating out)
             trapped={context.open}
             onMountAutoFocus={onOpenAutoFocus}
-            onUnmountAutoFocus={onCloseAutoFocus}
+            onUnmountAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
+              event.preventDefault();
+              context.triggerRef.current?.focus();
+            })}
           >
             <DismissableLayer
               role="dialog"
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -240,7 +240,7 @@ const DialogContentImpl = React.forwardRef((props, forwardedRef) => {
             // (closed !== unmounted when animating out)
             trapped={context.open}
             onMountAutoFocus={onOpenAutoFocus}
-            onUnmountAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
+            onUnmountAutoFocus={composeEventHandlers(onCloseAutoFocus, (event) => {
               event.preventDefault();
               context.triggerRef.current?.focus();
             })}
--- a/packages/react/focus-scope/src/FocusScope.tsx
+++ b/packages/react/focus-scope/src/FocusScope.tsx
@@ -103,18 +103,21 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
   React.useEffect(() => {
     if (container) {
       focusScopesStack.add(focusScope);
-      const PREVIOUSLY_FOCUSED_ELEMENT = document.activeElement as HTMLElement | null;
+      const previouslyFocusedElement = document.activeElement as HTMLElement | null;
+      const hasAutoFocusedCandidate = container.contains(previouslyFocusedElement);
 
-      // we need to setup the listeners before we `dispatchEvent`
-      container.addEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
-      container.addEventListener(AUTOFOCUS_ON_UNMOUNT, onUnmountAutoFocus);
+      if (!hasAutoFocusedCandidate) {
+        // we need to setup the listeners before we `dispatchEvent`
+        container.addEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
+        container.addEventListener(AUTOFOCUS_ON_UNMOUNT, onUnmountAutoFocus);
 
-      const mountEvent = new Event(AUTOFOCUS_ON_MOUNT, EVENT_OPTIONS);
-      container.dispatchEvent(mountEvent);
-      if (!mountEvent.defaultPrevented) {
-        focusFirst(getTabbableCandidates(container), { select: true });
-        if (document.activeElement === PREVIOUSLY_FOCUSED_ELEMENT) {
-          focus(container);
+        const mountEvent = new Event(AUTOFOCUS_ON_MOUNT, EVENT_OPTIONS);
+        container.dispatchEvent(mountEvent);
+        if (!mountEvent.defaultPrevented) {
+          focusFirst(getTabbableCandidates(container), { select: true });
+          if (document.activeElement === previouslyFocusedElement) {
+            focus(container);
+          }
         }
       }
 
@@ -128,7 +131,7 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
           const unmountEvent = new Event(AUTOFOCUS_ON_UNMOUNT, EVENT_OPTIONS);
           container.dispatchEvent(unmountEvent);
           if (!unmountEvent.defaultPrevented) {
-            focus(PREVIOUSLY_FOCUSED_ELEMENT ?? document.body, { select: true });
+            focus(previouslyFocusedElement ?? document.body, { select: true });
           }
           // we need to remove the listener after we `dispatchEvent`
           container.removeEventListener(AUTOFOCUS_ON_UNMOUNT, onUnmountAutoFocus);
@@ -184,10 +187,10 @@ FocusScope.displayName = FOCUS_SCOPE_NAME;
  * Stops when focus has actually moved.
  */
 function focusFirst(candidates: HTMLElement[], { select = false } = {}) {
-  const PREVIOUSLY_FOCUSED_ELEMENT = document.activeElement;
+  const previouslyFocusedElement = document.activeElement;
   for (const candidate of candidates) {
     focus(candidate, { select });
-    if (document.activeElement !== PREVIOUSLY_FOCUSED_ELEMENT) return;
+    if (document.activeElement !== previouslyFocusedElement) return;
   }
 }
 
--- a/packages/react/focus-scope/src/FocusScope.tsx
+++ b/packages/react/focus-scope/src/FocusScope.tsx
@@ -104,13 +104,11 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
     if (container) {
       focusScopesStack.add(focusScope);
       const previouslyFocusedElement = document.activeElement as HTMLElement | null;
-      const hasAutoFocusedCandidate = container.contains(previouslyFocusedElement);
+      const hasFocusedCandidate = container.contains(previouslyFocusedElement);
 
-      if (!hasAutoFocusedCandidate) {
-        // we need to setup the listeners before we `dispatchEvent`
-        container.addEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
-
+      if (!hasFocusedCandidate) {
         const mountEvent = new Event(AUTOFOCUS_ON_MOUNT, EVENT_OPTIONS);
+        container.addEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
         container.dispatchEvent(mountEvent);
         if (!mountEvent.defaultPrevented) {
           focusFirst(getTabbableCandidates(container), { select: true });
--- a/packages/react/focus-scope/src/FocusScope.tsx
+++ b/packages/react/focus-scope/src/FocusScope.tsx
@@ -109,7 +109,6 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
       if (!hasAutoFocusedCandidate) {
         // we need to setup the listeners before we `dispatchEvent`
         container.addEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
-        container.addEventListener(AUTOFOCUS_ON_UNMOUNT, onUnmountAutoFocus);
 
         const mountEvent = new Event(AUTOFOCUS_ON_MOUNT, EVENT_OPTIONS);
         container.dispatchEvent(mountEvent);
@@ -129,6 +128,7 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
         // See: https://github.com/facebook/react/issues/17894
         setTimeout(() => {
           const unmountEvent = new Event(AUTOFOCUS_ON_UNMOUNT, EVENT_OPTIONS);
+          container.addEventListener(AUTOFOCUS_ON_UNMOUNT, onUnmountAutoFocus);
           container.dispatchEvent(unmountEvent);
           if (!unmountEvent.defaultPrevented) {
             focus(previouslyFocusedElement ?? document.body, { select: true });
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -480,14 +480,12 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
               event.preventDefault();
               contentRef.current?.focus();
             })}
-            onUnmountAutoFocus={(event) => {
+            onUnmountAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
               // skip autofocus on unmount if clicking outside is permitted and it happened
               if (!disableOutsidePointerEvents && isPointerDownOutsideRef.current) {
                 event.preventDefault();
-              } else {
-                onCloseAutoFocus?.(event);
               }
-            }}
+            })}
           >
             <DismissableLayer
               as={Slot}
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -480,7 +480,7 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
               event.preventDefault();
               contentRef.current?.focus();
             })}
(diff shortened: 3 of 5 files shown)


packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -103,7 +110,8 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
   React.useEffect(() => {
     if (container) {
       focusScopesStack.add(focusScope);
-      const PREVIOUSLY_FOCUSED_ELEMENT = document.activeElement as HTMLElement | null;
```
jjenzz, line 106: Never understood why we uppercase this, we don't do this for any of our other variables.

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -22,10 +22,16 @@ type FocusScopeOwnProps = Polymorphic.Merge<
...
-     * (default: false)
+     * @defaultValue false
      */
     trapped?: boolean;
 
+    /**
+     * The element to auto focus when unmounted.
+     * @defaultValue false
+     */
+    unmountAutoFocusElement?: HTMLElement | null;
```
jjenzz, line 33: I'm not attached to this name so feel free to suggest alternatives 😛

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -103,7 +110,8 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
   React.useEffect(() => {
     if (container) {
       focusScopesStack.add(focusScope);
-      const PREVIOUSLY_FOCUSED_ELEMENT = document.activeElement as HTMLElement | null;
+      const activeElement = document.activeElement as HTMLElement | null;
+      const previouslyFocusedElement = unmountAutoFocusElement || activeElement;
```
andy-hook, line 114: This and the first conditional in the next block is slightly confusing to me, does `unmountAutoFocusElement` become previously focused? why wouldn't it be covered by `document.activeElement` per the previous code?

packages/react/focus-scope/src/FocusScope.tsx
andy-hook: Honestly can't think of a better one either 😅

packages/react/focus-scope/src/FocusScope.tsx
andy-hook: This bothered me as well haha.

packages/react/focus-scope/src/FocusScope.tsx
benoitgrelard: I think it's the naming that's not really correct anymore.
Before `previouslyFocusElement` made sense because it was `activeElement` at the time of mounting.
But not that's not really true if we say it's `unmountAutoFocusElement || activeElement`.

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -113,7 +121,7 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
       container.dispatchEvent(mountEvent);
       if (!mountEvent.defaultPrevented) {
         focusFirst(getTabbableCandidates(container), { select: true });
-        if (document.activeElement === PREVIOUSLY_FOCUSED_ELEMENT) {
+        if (document.activeElement === previouslyFocusedElement) {
```
benoitgrelard, line 124: The meaning of this conditional has now changed I think because of the newly added var, we need to be careful about this.

Before this line was there to make sure we focused the container if the focus was still where it was before mounting (even after running all the successive potential `focusFirst`).

Now it means something else because of the added `unmountAutoFocusElement` that can be arbitrarily passed in.

I think we basically still need the concept of `previouslyFocusedElement` (being `document.activeElement` at the point of mount) separate from the new var.

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -128,7 +136,7 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
           const unmountEvent = new Event(AUTOFOCUS_ON_UNMOUNT, EVENT_OPTIONS);
           container.dispatchEvent(unmountEvent);
           if (!unmountEvent.defaultPrevented) {
-            focus(PREVIOUSLY_FOCUSED_ELEMENT ?? document.body, { select: true });
+            focus(previouslyFocusedElement ?? document.body, { select: true });
```
benoitgrelard, line 139: I think this is the only place where we really need `unmountAutoFocusElement ?? document.activeElement ?? document.body`

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -137,7 +145,7 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
         }, 0);
       };
     }
-  }, [container, onMountAutoFocus, onUnmountAutoFocus, focusScope]);
+  }, [container, unmountAutoFocusElement, onMountAutoFocus, onUnmountAutoFocus, focusScope]);
```
benoitgrelard, line 148: Something else has changed here semantically because of the new var which now became a dep.
The side effect of focusing back out should really only be when unmounting, but now it's different because it would execute if someone changed the passed `unmountAutoFocusElement`.

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -184,10 +192,10 @@ FocusScope.displayName = FOCUS_SCOPE_NAME;
  * Stops when focus has actually moved.
  */
 function focusFirst(candidates: HTMLElement[], { select = false } = {}) {
-  const PREVIOUSLY_FOCUSED_ELEMENT = document.activeElement;
+  const previouslyFocusedElement = document.activeElement;
```
benoitgrelard, line 195: 👍

packages/react/popover/src/Popover.tsx
```diff
@@ -260,6 +263,7 @@ const PopoverContentImpl = React.forwardRef((props, forwardedRef) => {
           // we make sure we're not trapping once it's been closed
           // (closed !== unmounted when animating out)
           trapped={trapFocus && context.open}
+          unmountAutoFocusElement={context.trigger}
```
benoitgrelard, line 266: Are we going to be explicit in all the use-cases of `FocusScope`? If so, should it be a required prop and not bother about `document.activeElement` as a fallback?

I guess there are still places like context menu where it's the default.
Makes me think though, I think it was in `DropdownMenu` that we forced focusing the trigger, because we used `mousedown` to trigger open, which meant that activeElement wasn't set on time, so that seems to be a good place for your new prop too, rather than preventing default in the `onCloseAutoFocus` and running our own focus.

packages/react/focus-scope/src/FocusScope.tsx
benoitgrelard: haha no reason whatsoever, no idea why I did that 😀

packages/react/popover/src/Popover.tsx
jjenzz: > Are we going to be explicit in all the use-cases of `FocusScope`? If so, should it be a required prop and not bother about `document.activeElement` as a fallback?

I thought about this too but we use `FocusScope` in `Menu` which has no trigger so couldn't really make it required there. Same for `ContextMenu`, if this was required what would I pass there?

benoitgrelard:
Interestingly also, I just realised that Chromatic caught something, and focus is in a different place than before.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -223,10 +223,7 @@ const DropdownMenuRootContent = React.forwardRef((props, forwardedRef) => {
       disableOutsideScroll={disableOutsideScroll}
       portalled={portalled}
       trapFocus
-      onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
-        event.preventDefault();
-        context.triggerRef.current?.focus();
-      })}
+      closeAutoFocusRef={context.triggerRef}
```
jjenzz, line 226: @benoitgrelard I'm not sure how I feel about this. 

This is making me wonder why we wouldn't have an `openAutoFocusRef` also now so things like `AlertDialog` wouldn't need to `event.preventDefault()` in `onOpenAutoFocus`, they'd just pass a node.

That's then making me wonder why we need the `onCloseAutoFocus` and `onOpenAutoFocus` events at all....

Is this PR the right approach or is there another way we could solve the original issue? I'm still so full of brain fog 🙈

benoitgrelard: That would be going back to a direction we had originally and that's what most libs do, but I feel like we moved away for a reason. I do like the approach we have that we "do the thing" and people can prevent that and do their own thing instead.

For the original issue, we could simply do that? prevent default and specifically focus the trigger?

jjenzz: The original issue happens because when focus scope gets `document.activeElement` it is the `input` inside the focus scope because the `autoFocus` attribute focuses it before the `FocusScope` effect runs. So, when the focus scope closes, it tries to focus the input from inside content, but it has been unmounted.

> For the original issue, we could simply do that? prevent default and specifically focus the trigger?

That's true. I guess I thought we needed this change so all the other components would work if something was `autoFocus`ed inside them but I forgot we already do this `event.preventDefault()` thing in those components so it wouldn't be a problem there. DOH! Not sure why I didn't think of that, will give it a go!

jjenzz:
@benoitgrelard Okay... this is ready for you again now 🙂

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -103,18 +103,21 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
...
-      const mountEvent = new Event(AUTOFOCUS_ON_MOUNT, EVENT_OPTIONS);
-      container.dispatchEvent(mountEvent);
-      if (!mountEvent.defaultPrevented) {
-        focusFirst(getTabbableCandidates(container), { select: true });
-        if (document.activeElement === PREVIOUSLY_FOCUSED_ELEMENT) {
-          focus(container);
+      const previouslyFocusedElement = document.activeElement as HTMLElement | null;
+      const hasAutoFocusedCandidate = container.contains(previouslyFocusedElement);
+
+      if (!hasAutoFocusedCandidate) {
```
jjenzz, line 109: I added a check so that if `autoFocus` has been applied to a candidate, it won't run our `onMountAutoFocus` logic. 

Without this, the container was always focused and the `autoFocus` attribute was ignored because line 118 would always be true given that the `previouslyFocusedElement` would be the `autoFocus`ed element.

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -103,18 +103,21 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
...
-      container.addEventListener(AUTOFOCUS_ON_UNMOUNT, onUnmountAutoFocus);
-
-      const mountEvent = new Event(AUTOFOCUS_ON_MOUNT, EVENT_OPTIONS);
-      container.dispatchEvent(mountEvent);
-      if (!mountEvent.defaultPrevented) {
-        focusFirst(getTabbableCandidates(container), { select: true });
-        if (document.activeElement === PREVIOUSLY_FOCUSED_ELEMENT) {
-          focus(container);
+      const previouslyFocusedElement = document.activeElement as HTMLElement | null;
+      const hasAutoFocusedCandidate = container.contains(previouslyFocusedElement);
```
benoitgrelard, line 107: I get how this works because of the order it's run, but perhaps we should be more explicit by checking that it's inside and that it has `autofocus` attribute?

packages/react/dialog/src/Dialog.tsx
```diff
@@ -240,7 +240,10 @@ const DialogContentImpl = React.forwardRef((props, forwardedRef) => {
             // (closed !== unmounted when animating out)
             trapped={context.open}
             onMountAutoFocus={onOpenAutoFocus}
-            onUnmountAutoFocus={onCloseAutoFocus}
+            onUnmountAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
```
benoitgrelard, line 243: This one isn't strictly necessary right now is it?

packages/react/popover/src/Popover.tsx
```diff
@@ -261,13 +261,12 @@ const PopoverContentImpl = React.forwardRef((props, forwardedRef) => {
...
-              event.preventDefault();
-            } else {
-              onCloseAutoFocus?.(event);
-            }
-          }}
+          onUnmountAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
+            const isPointerOutside = isPointerDownOutsideRef.current;
+            const isAllowedPointerOutside = isPointerOutside && !disableOutsidePointerEvents;
+            if (!isAllowedPointerOutside) context.triggerRef.current?.focus();
+            event.preventDefault();
```
benoitgrelard, line 268: nit:

```suggestion
            event.preventDefault();
            if (!isAllowedPointerOutside) context.triggerRef.current?.focus();
```

I think generally when we take over an event, we prevent default first and then do our action. I know it doesn't matter technically but I feel like it reads better in terms of the intent.

packages/react/focus-scope/src/FocusScope.tsx
jjenzz: Interesting, perhaps I should rename this to `hasFocusedCandidate` instead because it doesn't necessarily need to be tied to `autoFocus` attribute. 

If something has already been focused inside by any means at the time that it opens (e.g. programatically) then we don't need to auto focus anything ourselves.

packages/react/dialog/src/Dialog.tsx
jjenzz: Without it, the same bug could occur in dialog right?

packages/react/focus-scope/src/FocusScope.tsx
benoitgrelard: Yeah I guess that would make sense.

packages/react/dialog/src/Dialog.tsx
benoitgrelard: I guess technically it may happen with Safari because it won't focus the button?

jjenzz: It will be an issue in any browser AFAIK...

**Without this change**

https://user-images.githubusercontent.com/175330/123978222-98c19c80-d9b7-11eb-8504-9ad080b77c21.mp4


**With this change**

https://user-images.githubusercontent.com/175330/123978038-6fa10c00-d9b7-11eb-8eb6-1c416c54dc97.mp4

benoitgrelard: Why though? Did we remove the fallback to previous `activeElement`?

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -126,9 +126,10 @@ const FocusScope = React.forwardRef((props, forwardedRef) => {
         // See: https://github.com/facebook/react/issues/17894
         setTimeout(() => {
           const unmountEvent = new Event(AUTOFOCUS_ON_UNMOUNT, EVENT_OPTIONS);
+          container.addEventListener(AUTOFOCUS_ON_UNMOUNT, onUnmountAutoFocus);
```
jjenzz, line 129: @benoitgrelard I've moved this here where it's relevant because I accidentally wrapped it in the `hasFocusedCandidate` conditional before 🙈

packages/react/dialog/src/Dialog.tsx
jjenzz: I have an input inside the dialog with an `autoFocus` attribute (to match the original bug report) so:
- when it opens, the previous `activeElement` will be the `input` at the time the `FocusScope` effect runs
- when it closes it will focus the `input` and not the trigger which gets unmounted and we lose focus

benoitgrelard: Oh wow, that's super unexpected that the autofocus's input steals it earlier. I think I saw Diego mention something like that on twitter, and that it's potentially a React bug.

benoitgrelard: https://twitter.com/diegohaz/status/1408180632933388289

jjenzz: I'm not sure if it is a React bug. If you think about it, React will render before it runs the effects. The render will mean the browser will `autoFocus` the input, then the effect will run and get that as the `activeElement`.

benoitgrelard: Yeah I guess the browser can do it quicker than React.

jjenzz:
@benoitgrelard @andy-hook I believe all feedback is addressed now. Let me know if you're happy for this to go in 🙂

packages/react/menu/src/Menu.tsx
```diff
@@ -480,14 +480,12 @@ const MenuContentImpl = React.forwardRef((props, forwardedRef) => {
               event.preventDefault();
               contentRef.current?.focus();
             })}
-            onUnmountAutoFocus={(event) => {
+            onUnmountAutoFocus={composeEventHandlers(onCloseAutoFocus, (event) => {
```
benoitgrelard, line 483: 👍

benoitgrelard (approved):

Issue #210, opened 2020-10-19
[Context] Improve Tooltip Error Message when Arrow is outside Position


peduarte:
I was playing around with the Tooltip and I didn't think about the fact that the `Arrow` needed to be inside the `Position` Component.

I got this error:

>  `Popper.Arrow` must be used within `Popper`

![image](https://user-images.githubusercontent.com/372831/96458153-e4b12480-1220-11eb-9bf6-781de3c33fe4.png)

Im not sure how doable this is, but it'd be great if the error message was:

>  `Tooltip.Arrow` must be used within `Tooltip.Position`

I think that would be a more helpful message since it represents what I've done. I understand that the `Tooltip` is built on top of the lower level `Popper`, but as I user I haven't imported or used that at all.

Cheers!

chaance:
This would be a nice improvement for sure but it's not on the priority list right now, so I'm going to close the issue. We'll make space for more DX improvements after getting through a long backlog of bugs.

PR #741, opened 2021-06-22
[Dialog] Add optional `Title` and `Description` parts

### LINKED ISSUE #729: [Dialog] Consider adding optional `Title` and `Description` parts like in `AlertDialog`
This stemmed from some user feedback on Discord where someone realised that labelling a dialog via `aria-label` or `aria-labelledby` was necessary (they found it in WAI-ARIA docs).

https://discord.com/channels/752614004387610674/803656530259738674/856455668754808832

I explained that some labelling is left to the user, but that we added `Title`/`Description` parts to `AlertDialog` as it was a more specific pattern. But I honestly can't see any reason against exposing those as optional parts for `Dialog` if it can simplify labelling for most people without having to faff around too much.

Let me know what you think and if we agree I can take a look.

cc  @jjenzz @andy-hook

benoitgrelard:
Closes #729

This PR:
- brings `Title` and `Description` from `AlertDialog` into `Dialog` itself
- improves upon existing dev warnings with new requirements
  - description is optional for `Dialog` but required for `AlertDialog` so that's reflected in the warnings
  - "localizes" the messages based on which primitive it is
  - links to docs

### DIFF
--- a/packages/react/alert-dialog/src/AlertDialog.tsx
+++ b/packages/react/alert-dialog/src/AlertDialog.tsx
@@ -34,11 +34,9 @@ const [
   useAlertDialogContentContext,
 ] = createContext<AlertDialogContentContextValue>(CONTENT_NAME);
 
-type AlertDialogContentOwnProps = Omit<Polymorphic.OwnProps<typeof DialogPrimitive.Content>, 'id'>;
-
 type AlertDialogContentPrimitive = Polymorphic.ForwardRefComponent<
   Polymorphic.IntrinsicElement<typeof DialogPrimitive.Content>,
-  AlertDialogContentOwnProps
+  Polymorphic.OwnProps<typeof DialogPrimitive.Content>
 >;
 
 const AlertDialogContent = React.forwardRef((props, forwardedRef) => {
--- a/packages/react/alert-dialog/src/AlertDialog.tsx
+++ b/packages/react/alert-dialog/src/AlertDialog.tsx
@@ -47,7 +47,7 @@ const AlertDialogContent = React.forwardRef((props, forwardedRef) => {
   const cancelRef = React.useRef<React.ElementRef<typeof AlertDialogCancel> | null>(null);
 
   return (
-    <DialogPrimitive.LabelWarningContext.Provider
+    <DialogPrimitive.LabelWarningProvider
       value={React.useMemo(
         () => ({ contentName: CONTENT_NAME, titleName: TITLE_NAME, docsSlug: 'alert-dialog' }),
         []
@@ -73,7 +73,7 @@ const AlertDialogContent = React.forwardRef((props, forwardedRef) => {
           {process.env.NODE_ENV === 'development' && <DescriptionWarning contentRef={contentRef} />}
         </DialogPrimitive.Content>
       </AlertDialogContentProvider>
-    </DialogPrimitive.LabelWarningContext.Provider>
+    </DialogPrimitive.LabelWarningProvider>
   );
 }) as AlertDialogContentPrimitive;
 
--- a/packages/react/alert-dialog/src/AlertDialog.tsx
+++ b/packages/react/alert-dialog/src/AlertDialog.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import { createContext } from '@radix-ui/react-context';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
+import * as DialogPrimitive from '@radix-ui/react-dialog';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { extendPrimitive } from '@radix-ui/react-primitive';
-import * as DialogPrimitive from '@radix-ui/react-dialog';
+import { Slottable } from '@radix-ui/react-slot';
 
 import type * as Polymorphic from '@radix-ui/react-polymorphic';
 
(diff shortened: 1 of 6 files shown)


packages/react/alert-dialog/src/AlertDialog.tsx
```diff
@@ -47,112 +34,44 @@ const [
   useAlertDialogContentContext,
 ] = createContext<AlertDialogContentContextValue>(CONTENT_NAME);
 
-type AlertDialogContentOwnProps = Omit<
-  Polymorphic.OwnProps<typeof DialogPrimitive.Content>,
-  'refToFocusOnOpen' | 'id'
```
benoitgrelard, line 52: some very old leftover here `refToFocusOnOpen` 😀

packages/react/alert-dialog/src/AlertDialog.tsx
```diff
@@ -47,112 +34,44 @@ const [
...
-      })}
+    <DialogPrimitive.AccessibilityDevWarningsContext.Provider
+      value={React.useMemo(
+        () => ({
+          descriptionRequired: true,
+          partNames: { content: CONTENT_NAME, title: TITLE_NAME, description: DESCRIPTION_NAME },
+          docsUrl: 'https://radix-ui.com/primitives/docs/components/alert-dialog',
+        }),
+        []
+      )}
```
benoitgrelard, line 56: I've used a React context defined in `Dialog` with a default value, which we override here in `AlertDialog` to provide the necessary details for this specific primitive requirements.

Let me know if you can find another way, I started originally going via props, but it bothered me to have "private" props exposed publicly just because they were needed to be passed from `AlertDialog`'s internals into `Dialog`.

Interestingly, this could maybe be a way to solve the longstanding #210.

benoitgrelard: One thing to note here is that this particular bit of code here won't be dead-eliminated in prod, let me know if you'd like me to find a way to remodel this conditionally to only include it in dev or if you think it's unnecessary complexity.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -220,47 +232,104 @@ const DialogContentImpl = React.forwardRef((props, forwardedRef) => {
...
+              })}
+              // When focus is trapped, a focusout event may still happen.
+              // We make sure we don't trigger our `onDismiss` in such case.
+              onFocusOutside={(event) => event.preventDefault()}
+              onDismiss={() => context.onOpenChange(false)}
+            />
+          </FocusScope>
+        </RemoveScroll>
+      </Portal>
+      {process.env.NODE_ENV === 'development' && (
```
benoitgrelard, line 278: This change now eliminates the need for `Slottable` that existed in `AlertDialog` because now we know when we're in `DialogContentImpl` it's because we're mounted.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -296,28 +365,101 @@ function getState(open: boolean) {
...
+  ariaLabel?: string;
+  ariaLabelledBy?: string;
+  ariaDescribedBy?: string;
+}> = (props) => {
+  const { ariaLabel, ariaLabelledBy, ariaDescribedBy } = props;
+  const context = useDialogContext('AccessibilityDevWarnings');
+  const devWarningsContext = React.useContext(AccessibilityDevWarningsContext);
+  const { partNames } = devWarningsContext;
+
+  const LABEL_WARNING = `\`${partNames.content}\` requires a label for the component to be accessible for screen reader users.
```
benoitgrelard, line 384: Mostly copied component from what was in `AlertDialog` but brought the warnings inline here so I could add the localised names from context.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -296,28 +365,101 @@ function getState(open: boolean) {
...
+    );
+    if (!hasLabel) {
+      console.warn(LABEL_WARNING);
+    }
+
+    const hasDescription = Boolean(
+      (ariaDescribedBy && document.getElementById(ariaDescribedBy)) ||
+        (context.descriptionId && document.getElementById(context.descriptionId))
+    );
+    if (devWarningsContext.descriptionRequired && !hasDescription) {
```
benoitgrelard, line 414: And also check for description being required or not.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -296,28 +365,101 @@ function getState(open: boolean) {
   return open ? 'open' : 'closed';
 }
 
+const AccessibilityDevWarningsContext = React.createContext({
```
benoitgrelard, line 368: New context to provide info about warnings.

packages/react/alert-dialog/src/AlertDialog.tsx
```diff
@@ -47,112 +34,44 @@ const [
   useAlertDialogContentContext,
 ] = createContext<AlertDialogContentContextValue>(CONTENT_NAME);
 
-type AlertDialogContentOwnProps = Omit<
-  Polymorphic.OwnProps<typeof DialogPrimitive.Content>,
-  'refToFocusOnOpen' | 'id'
->;
+type AlertDialogContentOwnProps = Omit<Polymorphic.OwnProps<typeof DialogPrimitive.Content>, 'id'>;
```
jjenzz, line 37: I know you haven't changed this but curious if you know why we remove the `id` prop here? I don't think we do this anywhere else, we sorta just allow people to break things.

benoitgrelard: Yeah true, I didn't give this one much thought, I'll check it and compare with others.

benoitgrelard: Yeah totally right, we usually allow all overriding, plus there was no reason to remove it here ever cause we were not doing anything different with `id` in `AlertDialogContent` specifically 🤷‍♂️

jjenzz (changes requested):
Some initial feedback to discuss. I'll finish reviewing when we've decided the outcome of this in case things change.

packages/react/alert-dialog/src/AlertDialog.tsx
```diff
@@ -47,112 +34,42 @@ const [
...
-      {...dialogContentProps}
-      ref={forwardedRef}
-      onOpenAutoFocus={composeEventHandlers(dialogContentProps.onOpenAutoFocus, (event) => {
-        event.preventDefault();
-        cancelRef.current?.focus({ preventScroll: true });
-      })}
+    <DialogPrimitive.AccessibilityDevWarningsContext.Provider
+      value={React.useMemo(
+        () => ({
+          descriptionRequired: true,
```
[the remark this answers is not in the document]
benoitgrelard: Yeah that's a great point on the `descriptionRequired`, not sure why I went that way, if it is only required in `AlertDialog` makes more sense to do it there.

I'll try and improve the whole thing based on your feedback and sandbox.

Thanks!

packages/react/dialog/src/Dialog.tsx
```diff
@@ -296,28 +358,70 @@ function getState(open: boolean) {
...
   //
   Root,
   Trigger,
   Overlay,
   Content,
+  Title,
+  Description,
   Close,
+  //
+  LabelWarningContext,
```
jjenzz, line 418: nit: Can we expose just `LabelWarningProvider` instead please? It looks like we only need the provider externally and we could also simplify consumption then:

```jsx
const LabelWarningProvider: React.FC<LabelWarningContextValue> = props => {
  return <LabelWarningContext.Provider value={props} />;
}
```

```jsx
<Dialog.LabelWarningProvider contentName="" titleName="" docsSlug="">
</Dialog.LabelWarningProvider>
``` 

vs

```jsx
<Dialog.LabelWarningContext.Provider value={React.useMemo(() => ({ contentName: '', titleName: '', docsSlug: '' ), [])}>
</Dialog.LabelWarningContext.Provider>
``` 

I'm not too bothered about doing the memoization internally though, it was more `Dialog.LabelWarningProvider` vs `Dialog.LabelWarningContext.Provider` that was bothering me so the following would suffice if you prefer 😅

```jsx
const LabelWarningProvider = LabelWarningContext.Provider;
export { LabelWarningProvider };
```

jjenzz (approved):
:+1: looking good. Just one little nit that I'll leave up to you 😛

packages/react/dialog/src/Dialog.tsx
benoitgrelard: I went with the latter, because the former was missing the fact it should render `children` which brings us back in the same situation as `createContext`.

PR #743, opened 2021-06-23
[ContextMenu] Add long press support for touch/pen

### LINKED ISSUE #701: [ContextMenu] iOS and iPadOS touch support
## Bug report

I'm not sure of the right way to classify this issue: is it a bug report, feature request, or discussion? The `contextmenu` event doesn't fire on iOS and iPadOS when using touch input (it does on iPad OS when using a mouse), but my general belief is that all components should work cross browser and cross device (latest versions only) even if some browser and device specific workarounds are required.

### Current Behavior

The Radix context menu never appears on iOS and iPadOS when using touch input.

### Expected behavior

To be able to access the Radix context menu on iOS and iPadOS when using touch input.

### Reproducible example

Use the doc page demos on an iOS or iPadOS device: https://radix-ui.com/primitives/docs/components/context-menu

### Suggested solution

Android fires the `contextmenu` event on long press, so maybe implement a long press trigger on iOS and iPadOS?

### Your environment

| Radix Package(s) |    Context Menu     |    v0.0.19     |
| React            | n/a     |    v17    |
| Operating System |    iOS and iPadOS     |    v14     |

benoitgrelard:
Closes #701

This PR adds support for long press to trigger opening `ContextMenu` with touch/pen.

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -82,6 +82,17 @@ const ContextMenuTrigger = React.forwardRef((props, forwardedRef) => {
   const virtualRef = React.useRef({
     getBoundingClientRect: () => DOMRect.fromRect({ width: 0, height: 0, ...pointRef.current }),
   });
+  const longPressTimerRef = React.useRef(0);
+  const clearLongPress = React.useCallback(
+    () => window.clearTimeout(longPressTimerRef.current),
+    []
+  );
+  const open = (event: React.MouseEvent | React.PointerEvent) => {
+    pointRef.current = { x: event.clientX, y: event.clientY };
+    context.onOpenChange(true);
+  };
+
+  React.useEffect(() => clearLongPress, [clearLongPress]);
 
   return (
     <ContentContext.Provider value={false}>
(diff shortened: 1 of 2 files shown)


rafgraph:
I think this looks good, but I didn’t clone and test. My only suggestion would be to set `-webkit-touch-callout: none` as an inline style on the element to prevent the iOS “context menu” from appearing (this acts like calling `preventDefault`).

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -82,6 +82,17 @@ const ContextMenuTrigger = React.forwardRef((props, forwardedRef) => {
   const virtualRef = React.useRef({
     getBoundingClientRect: () => DOMRect.fromRect({ width: 0, height: 0, ...pointRef.current }),
   });
+  const longPressTimerRef = React.useRef(0);
+  const clearLongPress = React.useCallback(
+    () => window.clearTimeout(longPressTimerRef.current),
+    []
+  );
+  const open = (event: React.MouseEvent | React.PointerEvent) => {
```
jjenzz, line 90: Can we rename this to `handleOpen` for consistency with our other event handlers please?

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -91,10 +102,24 @@ const ContextMenuTrigger = React.forwardRef((props, forwardedRef) => {
         as={as}
         ref={forwardedRef}
         onContextMenu={composeEventHandlers(props.onContextMenu, (event) => {
+          // clearing the long press here because some platforms already supports
```
jjenzz, line 105:
```suggestion
          // clearing the long press here because some platforms already support
```

jjenzz (approved):

jjenzz (approved):

rafgraph:
One edge case to watch out for is if the user puts two touch points down. It looks like `handleOpen` will be called twice, which is fine if it’s idempotent. Alternatively the context menu could cancel when a second touch point is added. I don’t have a strong preference, just wanted to mention the edge case.

rafgraph:
Also with 2 touch points it looks like 2 timers are set but only 1 is tracked as a `ref` and cleared in `useEffect`, which might result in a bug.

benoitgrelard:
Thanks @rafgraph, I'll check the double touch case. Instinctively I thought it'd be ok cause I'm using pointer events rather than touch events, but I'll check how it behaves.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -113,6 +113,8 @@ const ContextMenuTrigger = React.forwardRef((props, forwardedRef) => {
         onPointerDown={composeEventHandlers(
           props.onPointerDown,
           whenTouchOrPen((event) => {
+            // clear the long press here in case there's multiple touch points
+            clearLongPress();
```
rafgraph, line 117: The least complex solution 👍👍

PR #744, opened 2021-06-23
[DropdownMenu] Add submenu int tests

andy-hook:
This PR adds initial interaction tests for submenus on `DropdownMenu`. I will raise a separate improvement to cover `ContextMenu`.

I've gone with the `.spec` standard on filenames purely because this matches the Cypress docs and differentiates from our jest tests, I also like that these types of tests document what our expected behaviour should be i.e the "spec". Happy for this to be whatever though.

Still feeling out the correct level for some of the tests and these are not meticulously extensive for all scenarios, however, they should replace the bulk of manual testing we'd been doing while building the feature. I was intentional to only cover the submenu work as we can progressively add additional tests for other `DropdownMenu` functionality as / when we see the need.

Let me know if i've missed anything obvious.


cypress/integration/DropdownMenu.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      it('should close when exiting trigger away from submenu', () => {
+        cy.findByText('Open').click();
+
+        movePointerTo('Bookmarks →');
+        cy.findByText('Inbox').should('exist');
+
+        exitPointerRightToLeft('Bookmarks →');
+        cy.findByText('Inbox').should('not.exist');
+      });
```
jjenzz, line 106: We tend to write our tests like "given, when, then" (nested describes) which really helps avoid test setup duplication, and with one assertion per test to isolate failures (making them easier to debug/fix): 

```jsx
describe('given an open menu', () => {
	beforeEach(() => {
		cy.findByText('Open').click();
	});

	describe('when moving mouse pointer on submenu trigger', () => {
		beforeEach(() => {
			movePointerTo('Bookmarks →');
		});

		it('should open the submenu', () => {
			cy.findByText('Inbox').should('exist');
		});

		it('should not focus first item', () => {
        	cy.findByText('Inbox').should('not.be.focused');
      	});

		describe('and moving pointer towards submenu', () => {
			beforeEach(() => {
				exitPointerLeftToRight('Bookmarks →');
			});

			it('should keep the submenu open', () => {
				cy.findByText('Inbox').should('exist');
			});

			it('should keep the main menu open', () => {
				cy.findByText('Bookmarks →').should('exist');
			});
		});

		describe('and moving pointer away from submenu', () => {
			beforeEach(() => {
				exitPointerRightToLeft('Bookmarks →');
			});

			it('should close the submenu', () => {
				cy.findByText('Inbox').should('not.exist');
			});

			it('should close the main menu', () => {
				cy.findByText('Bookmarks →').should('exist');
			});
		});

		describe('and moving pointer to trigger sibling item', () => {
			beforeEach(() => {
				movePointerTo('New Window');
			});

			it('should close the submenu', () => {
				cy.findByText('Inbox').should('not.exist');
			});

			it('should keep the main menu open', () => {
				cy.findByText('Bookmarks →').should('exist');
			});
		});

		describe('and moving pointer to trigger sibling trigger', () => {
			beforeEach(() => {
				movePointerTo('Tools →');
			});

			it('should close the submenu', () => {
				cy.findByText('Inbox').should('not.exist');
			});

			it('should keep the main menu open', () => {
				cy.findByText('Bookmarks →').should('exist');
			});
		});

		describe('and clicking a submenu item', () => {
			beforeEach(() => {
	        	cy.findByText('Inbox').click();				
			});

			it('should close the submenu', () => {
				cy.findByText('Inbox').should('not.exist');
			});

			it('should close the main menu', () => {
				cy.findByText('Bookmarks →').should('not.exist');
			});
		});

		describe('and pressing escape', () => {
			beforeEach(() => {
				cy.get('body').type('{esc}');
			});

			it('should close the submenu', () => {
				cy.findByText('Inbox').should('not.exist');
			});

			it('should close the main menu', () => {
				cy.findByText('Bookmarks →').should('not.exist');
			});
		});
	});

	describe('when in RTL mode', () => {
		beforeEach(() => {
			cy.findByText('Right-to-left').click();
		});

		describe('and moving pointer ....', () => {});
	});
});
```

Hopefully there are enough examples here to explain what I mean but happy to have a call to run through it if it's not clear 🙂

jjenzz (changes requested):
Excited to start getting some of these in! 

I've left some feedback on how we tend to handle our test architecture below. Let me know if you wanna call.

cypress/integration/DropdownMenu.spec.ts
andy-hook: Thanks for this! your example has already highlighted some omissions that would help the tests.

On the "given, when, then" point, I really like this! but there are some caveats with cypress which can make it challenging. The [best practice doc](https://docs.cypress.io/guides/references/best-practices#Creating-tiny-tests-with-a-single-assertion ) sums some of those up.

The other issue is the way that assertions in Cypress have a concept of [retry-ability](https://docs.cypress.io/guides/core-concepts/retry-ability), it’s kinda their answer to common flake in e2e, which for the most part makes tests more reliable but using it with this suggested approach can make failures sometimes less clear.

If you’re free today would be great to jump on a call to discuss, perhaps we can come out with an improved structure while working around the constraints 🙂

jjenzz: Ah, thanks for the link. Seems they clearly recommend against this but that makes my heart sad because I find this sooooo much easier to reason about. Also, why would they even provide these APIs if it's an anti-pattern 🤔 

Their reasons against it seem more opinionated except for the last point about perf. Obviously that's an issue but how much slower are we talking?

With the current tests it's not clear which part of the test is the bit described by the test and which parts are just repeated set up 🙈 

Let's have a catch up on this today so I can maybe understand how much of an issue the issues they describe really are.

jjenzz: We caught up on this and we're going to stick with the general approach @andy-hook has here but try make the code a bit more self documented to help with some of my confusions. 

Andy also helped me realise that it's not hugely important to be able to reason about the code because with Cypress we can step through the test visually which is super helpful.

andy-hook:
@jjenzz I've restructured the tests following our discussion and added a couple of cases to cover collision changes on the pointer grace.

Stopped short on the self documenting variables for every text selection, ended up feeling more difficult to follow with the dev runner as it shows all of the real labels in context, happy to revisit though if we still find it hard to follow.

cypress/integration/DropdownMenu.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+        pointerOver('Inbox');
+        pointerOver('Bookmarks →');
+        cy.findByText('Inbox').should('be.visible');
+      });
+
+      it(
+        'should close submenu when moving pointer away but remain open when moving towards',
+        {
+          viewportWidth: 550,
+        },
```
andy-hook, line 31: Passing in a slimmer width to trigger the collision detection, I found this to be more robust than calling [cy.viewport](https://docs.cypress.io/api/commands/viewport) from within the body

cypress/integration/DropdownMenu.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+        () => {
+          // Moving away
+          pointerOver('Bookmarks →');
+          cy.findByText('Inbox').should('be.visible');
+          pointerExitRightToLeft('Bookmarks →');
+          cy.findByText('Inbox').should('not.exist');
+
+          // Moving towards
+          pointerOver('Bookmarks →');
+          cy.findByText('Inbox').should('be.visible');
```
andy-hook, line 41: This looks redundant but what's really happening is we're leaning on the retry-ability of the assertion to cover the 100ms timeout on the open, the alternative is to chain a `wait(100)` but that couples us closer to the implementation.

You'll notice this in a few places, perhaps there's a better way to present it.

jjenzz (approved):

andy-hook:
Merging to test Modality, current runtime on CI seems fine but I suspect we’ll want to use a [dedicated action](https://github.com/cypress-io/github-action) before long to run these in parallel.

PR #751, opened 2021-06-29
[Popover] Add support for modality

andy-hook:
This PR builds on top of #700 and adds support for modality options in popovers.

The approach taken here mirrors that of the [Dialog changes](https://github.com/radix-ui/primitives/pull/699).

### DIFF
--- a/packages/react/popover/src/Popover.stories.tsx
+++ b/packages/react/popover/src/Popover.stories.tsx
@@ -222,7 +222,11 @@ export const Chromatic = () => (
     <h2>Open</h2>
     <Popover defaultOpen>
       <PopoverTrigger className={triggerClass}>open</PopoverTrigger>
-      <PopoverContent className={contentClass} sideOffset={5}>
+      <PopoverContent
+        className={contentClass}
+        sideOffset={5}
+        onFocusOutside={(event) => event.preventDefault()}
+      >
         <PopoverClose className={closeClass}>close</PopoverClose>
         <PopoverArrow className={arrowClass} width={20} height={10} />
       </PopoverContent>
@@ -230,7 +234,11 @@ export const Chromatic = () => (
 
     <h2 style={{ marginTop: 100 }}>Open with reordered parts</h2>
     <Popover defaultOpen>
-      <PopoverContent className={contentClass} sideOffset={5}>
+      <PopoverContent
+        className={contentClass}
+        sideOffset={5}
+        onFocusOutside={(event) => event.preventDefault()}
+      >
         <PopoverClose className={closeClass}>close</PopoverClose>
         <PopoverArrow className={arrowClass} width={20} height={10} />
       </PopoverContent>
@@ -291,7 +299,7 @@ export const Chromatic = () => (
       <PopoverAnchor style={{ padding: 20, background: 'gainsboro' }}>
         <PopoverTrigger className={triggerClass}>open</PopoverTrigger>
       </PopoverAnchor>
-      <PopoverContent className={contentClass}>
+      <PopoverContent className={contentClass} onFocusOutside={(event) => event.preventDefault()}>
         <PopoverClose className={closeClass}>close</PopoverClose>
         <PopoverArrow className={arrowClass} width={20} height={10} />
       </PopoverContent>
--- a/packages/react/popover/src/Popover.stories.tsx
+++ b/packages/react/popover/src/Popover.stories.tsx
@@ -30,6 +30,21 @@ export const Styled = () => {
   );
 };
 
+export const NonModal = () => {
+  return (
+    <>
+      <Popover modal={false}>
+        <PopoverTrigger className={triggerClass}>open</PopoverTrigger>
+        <PopoverContent className={contentClass} sideOffset={5}>
+          <PopoverClose className={closeClass}>close</PopoverClose>
+          <PopoverArrow className={arrowClass} width={20} height={10} offset={10} />
+        </PopoverContent>
+      </Popover>
+      <input style={{ marginLeft: 10 }} />
+    </>
+  );
+};
+
 export const Controlled = () => {
   const [open, setOpen] = React.useState(false);
   return (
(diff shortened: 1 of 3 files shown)


packages/react/popover/src/Popover.tsx
```diff
@@ -40,10 +41,11 @@ type PopoverOwnProps = {
   open?: boolean;
   defaultOpen?: boolean;
   onOpenChange?: (open: boolean) => void;
+  modal?: boolean;
 };
 
 const Popover: React.FC<PopoverOwnProps> = (props) => {
-  const { children, open: openProp, defaultOpen, onOpenChange } = props;
+  const { children, open: openProp, defaultOpen, onOpenChange, modal = true } = props;
```
andy-hook, line 48: Is defaulting to `modal` the expected here? I'm in two minds as whether it should be the other way around.

packages/react/popover/src/Popover.tsx
```diff
@@ -167,15 +170,83 @@ const PopoverContent = React.forwardRef((props, forwardedRef) => {
...
+      {...props}
+      ref={composedRefs}
+      // we make sure we're not trapping once it's been closed
+      // (closed !== unmounted when animating out)
+      trapFocus={context.open}
+      disableOutsidePointerEvents
+      // When focus is trapped, a `focusout` event may still happen.
+      // We make sure we don't trigger our `onDismiss` in such case.
+      onFocusOutside={composeEventHandlers(props.onFocusOutside, (event) => event.preventDefault())}
+    />
```
andy-hook, line 216: Does modal mode need right click handling as well?

andy-hook (commented):
Important to note that the toggle issue remains for non modals:

https://user-images.githubusercontent.com/11708259/123776604-59b41e00-d8c7-11eb-92ab-7917ef4efd04.mp4



I think it's more exacerbated with the explicit modes, especially if non modal were to be the default. Not sure what we should do here.

packages/react/popover/src/Popover.tsx
benoitgrelard: Yeah I also feel like `modal={false}` is a better default for `Popover`. How do we feel about having different defaults for different primitives?

> Note: that was already the case but with the granular props instead.

packages/react/popover/src/Popover.tsx
```diff
@@ -246,69 +314,31 @@ const PopoverContentImpl = React.forwardRef((props, forwardedRef) => {
...
 
-  // Hide everything from ARIA except the content
-  React.useEffect(() => {
-    const content = contentRef.current;
-    if (content) return hideOthers(content);
-  }, []);
-
   return (
     <PortalWrapper>
       <ScrollLockWrapper>
```
benoitgrelard, line 319: The outside scrolling should be part of modality, so it shouldn't be in the `Impl` anymore, but rather only in the modal version.

packages/react/popover/src/Popover.tsx
```diff
@@ -246,69 +314,31 @@ const PopoverContentImpl = React.forwardRef((props, forwardedRef) => {
...
   useFocusGuards();
 
-  // Hide everything from ARIA except the content
-  React.useEffect(() => {
-    const content = contentRef.current;
-    if (content) return hideOthers(content);
-  }, []);
-
   return (
     <PortalWrapper>
```
benoitgrelard, line 318: This differs from `Dialog` in that we offer the ability to not portal, that said it probably will need to be duplicated in both modal/non-modal versions because of this: https://github.com/radix-ui/primitives/pull/699#discussion_r648391771

benoitgrelard (commented):
I've left some initial feedback for you @andy-hook.

packages/react/popover/src/Popover.tsx
andy-hook: Good point, was wondering about this.

packages/react/popover/src/Popover.tsx
andy-hook: I agree, I think it should default to `false`. I'm fine with differing defaults, I think we should optimise for the common case and it (kind of) was the original behaviour. This does make the closing toggle issue much more obvious though.

packages/react/popover/src/Popover.tsx
andy-hook: Discussed with the team and the answer here is no, we prevent right click dismiss on primitives with an `Overlay` part only e.g. `Dialog`. In that example we consider it contiguous with the primitive itself.

packages/react/popover/src/Popover.tsx
```diff
@@ -167,14 +170,106 @@ const PopoverContent = React.forwardRef((props, forwardedRef) => {
...
+          isPointerDownOutsideRef.current = false;
+        })}
+        onPointerDownOutside={composeEventHandlers(props.onPointerDownOutside, (event) => {
+          const { originalEvent } = event.detail;
+          const isLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === false;
+          isPointerDownOutsideRef.current = isLeftClick;
+        })}
+        onInteractOutside={composeEventHandlers(props.onInteractOutside, (event) => {
+          // prevent dismissing when clicking the trigger
+          // as it's already setup to close, otherwise it would close and immediately open.
```
benoitgrelard, line 265: nit: we didn't update the comment either in `Dialog` we should

```suggestion
          // prevent dismissing when clicking the trigger
          // (or focusing it also as some platforms will focus on pointer down)
          // as it's already setup to close, otherwise it would close and immediately open.
```

packages/react/popover/src/Popover.stories.tsx
```diff
@@ -222,15 +222,23 @@ export const Chromatic = () => (
     <h2>Open</h2>
     <Popover defaultOpen>
       <PopoverTrigger className={triggerClass}>open</PopoverTrigger>
-      <PopoverContent className={contentClass} sideOffset={5}>
+      <PopoverContent
+        className={contentClass}
+        sideOffset={5}
+        onFocusOutside={(event) => event.preventDefault()}
```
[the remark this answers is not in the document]
andy-hook: Exactly, I figured this was still a valid test of the open state but a necessary evil given the story configuration.

packages/react/popover/src/Popover.tsx
andy-hook: Good spot, thanks. I went with this in the end https://github.com/radix-ui/primitives/pull/751/commits/3d92dfb1872a4fb82e7706b406b4bf2f2e9180ba

Let me know if that works for you.

benoitgrelard: Love it.

benoitgrelard (approved):

PR #752, opened 2021-06-29
Add no-unused-var lint rule

andy-hook:
This PR adds `@typescript-eslint/no-unused-vars` to our lint rules.

I've been caught out a few times when passing destructured props down across several comps and not having a signal to let me know if i'd forgotten to thread a given prop. It should also help mitigate [small things like this](https://github.com/radix-ui/primitives/pull/724).

One drawback is that we'd trigger a warning when intentionally omitting a given prop via `...rest`, but it appears we currently don't do this very often outside of a couple of tests.

Let me know what you think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,6 +130,12 @@
       }
     },
     "rules": {
+      "@typescript-eslint/no-unused-vars": [
+        1,
+        {
+          "vars": "local"
+        }
+      ],
       "prefer-const": [
         1,
         {
--- a/packages/react/slider/src/collection.stories.tsx
+++ b/packages/react/slider/src/collection.stories.tsx
@@ -155,10 +155,10 @@ function Item({ children, disabled = false, ...props }: ItemProps) {
 const [createTabsCollection, useTabsItem, useTabs] = createCollection('Tabs');
 const [createPanelsCollection, usePanelsItem, usePanels] = createCollection('Panels');
 
-const TabsContext = React.createContext({
-  selectedIndex: 0,
-  setSelectedIndex: (index: number) => {},
-});
+const TabsContext = React.createContext<{
+  selectedIndex: number;
+  setSelectedIndex: React.Dispatch<React.SetStateAction<number>>;
+}>({ selectedIndex: 0, setSelectedIndex: () => {} });
 
 const Tabs = createTabsCollection(function Tabs({ children }) {
   const tabs = useTabs();
--- a/.yarn/versions/7023f969.yml
+++ b/.yarn/versions/7023f969.yml
@@ -1,0 +1,2 @@
+declined:
+  - primitives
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -130,6 +130,12 @@
       }
     },
     "rules": {
+      "@typescript-eslint/no-unused-vars": [
```
benoitgrelard, line 133: Curious why it's a `@typescript-eslint` rule, isn't this a normal js rule anyway?

packages/react/slider/src/collection.stories.tsx
```diff
@@ -155,10 +155,10 @@ function Item({ children, disabled = false, ...props }: ItemProps) {
 const [createTabsCollection, useTabsItem, useTabs] = createCollection('Tabs');
 const [createPanelsCollection, usePanelsItem, usePanels] = createCollection('Panels');
 
-const TabsContext = React.createContext({
-  selectedIndex: 0,
-  setSelectedIndex: (index: number) => {},
-});
+const TabsContext = React.createContext<{
```
benoitgrelard, line 158: This change doesn't seem related to having unused vars, what was the reason here?

package.json
andy-hook: It's preferred for compat with types. It's generally recommended to [disable the base rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md) but I assume this is already handled by one of our presets. Only setting `no-unused-vars` will invoke both.

packages/react/slider/src/collection.stories.tsx
andy-hook: `index` was technically an unused var. It also aligns closer to [other uses](https://github.com/radix-ui/primitives/blob/94d7f8d6a64db1a518296478a4c4186bfbbc7f62/packages/react/dismissable-layer/src/DismissableLayer.tsx#L286-L290).

package.json
benoitgrelard: Thanks for the explanation.

benoitgrelard (approved):

jjenzz:
> I've been caught out a few times when passing destructured props down across several comps and not having a signal to let me know if i'd forgotten to thread a given prop.

While I'm all for this change, I'm super curious why TS wouldn't have caught the props threading issue? I'm worried if it might be another issue with the polymorphic types 🙈

jjenzz (approved):

andy-hook:
> While I'm all for this change, I'm super curious why TS wouldn't have caught the props threading issue? I'm worried if it might be another issue with the polymorphic types 🙈

Ah I should have been clearer on this one, don’t worry it’s not related to the typings. A recent example would be when I needed to raise an existing handler, in some cases I might forget to pass the original destructured prop back in if there’re a lot of other props being passed.  

e.g. [See here where I’ve raised some handlers](https://github.com/radix-ui/primitives/pull/751/files#diff-5c4477c3aab85a2bd407a2e527442d3131d9c637552cebe61e8403503617d449R331-R333), originally I messed up by forgetting to re-pass one of the props back in after moving it. If the linter had called out the unused prop in the destructure I would have noticed immediately.

Admittedly that's kind of a unique example, but it’s a handy rule for the little things like that.

PR #758, opened 2021-07-01
[AlertDialog] Explicit modality

andy-hook:
Force `Modal` mode on `AlertDialog` and prevent dismiss on click outside.

### DIFF
--- a/packages/react/alert-dialog/src/AlertDialog.tsx
+++ b/packages/react/alert-dialog/src/AlertDialog.tsx
@@ -14,8 +14,10 @@ import type * as Polymorphic from '@radix-ui/react-polymorphic';
 
 const ROOT_NAME = 'AlertDialog';
 
-const AlertDialog: React.FC<React.ComponentProps<typeof DialogPrimitive.Root>> = (props) => (
-  <DialogPrimitive.Root {...props} />
+type AlertDialogProps = Omit<React.ComponentProps<typeof DialogPrimitive.Root>, 'modal'>;
+
+const AlertDialog: React.FC<AlertDialogProps> = (props) => (
+  <DialogPrimitive.Root {...props} modal={true} />
 );
 
 AlertDialog.displayName = ROOT_NAME;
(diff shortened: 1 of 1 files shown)


packages/react/alert-dialog/src/AlertDialog.tsx
```diff
@@ -14,8 +14,10 @@ import type * as Polymorphic from '@radix-ui/react-polymorphic';
 
 const ROOT_NAME = 'AlertDialog';
 
-const AlertDialog: React.FC<React.ComponentProps<typeof DialogPrimitive.Root>> = (props) => (
-  <DialogPrimitive.Root {...props} />
+type AlertDialogProps = Omit<React.ComponentProps<typeof DialogPrimitive.Root>, 'modal'>;
```
andy-hook, line 17: Do I have this right? we want to enforce this opinion on consumers?

benoitgrelard: Yeah I think so.

packages/react/alert-dialog/src/AlertDialog.tsx
```diff
@@ -62,6 +64,9 @@ const AlertDialogContent = React.forwardRef((props, forwardedRef) => {
             event.preventDefault();
             cancelRef.current?.focus({ preventScroll: true });
           })}
+          onPointerDownOutside={composeEventHandlers(contentProps.onPointerDownOutside, (event) =>
```
benoitgrelard, line 67: I think we also need to prevent the escape keydown one.

benoitgrelard: May be worth rejigging what's exposed from `Dialog` to facilitate the changes here also.

benoitgrelard: Ha cool, good to know. I suppose that's still a meaningful user action, as opposed to clicking outside by mistake potentially.

benoitgrelard: We probably should just hardcode it then and remove it from the types.

andy-hook: > We probably should just hardcode it then and remove it from the types.

That's fine with me, are we sure consumers won't need access to the handler for a different purpose?

PR #759, opened 2021-07-01
[AlertDialog] Don't compose onPointerDownOutside

andy-hook:
See https://github.com/radix-ui/primitives/pull/758#discussion_r662357477

### DIFF
--- a/packages/react/alert-dialog/src/AlertDialog.tsx
+++ b/packages/react/alert-dialog/src/AlertDialog.tsx
@@ -39,7 +39,7 @@ const [
 
 type AlertDialogContentPrimitive = Polymorphic.ForwardRefComponent<
   Polymorphic.IntrinsicElement<typeof DialogPrimitive.Content>,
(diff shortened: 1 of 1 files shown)


packages/react/alert-dialog/src/AlertDialog.tsx
```diff
@@ -39,7 +39,7 @@ const [
 
 type AlertDialogContentPrimitive = Polymorphic.ForwardRefComponent<
   Polymorphic.IntrinsicElement<typeof DialogPrimitive.Content>,
-  Polymorphic.OwnProps<typeof DialogPrimitive.Content>
+  Omit<Polymorphic.OwnProps<typeof DialogPrimitive.Content>, 'onPointerDownOutside'>
```
benoitgrelard, line 42: I think we normally lift this up when there are modifications, most of the time really.

benoitgrelard (approved):

PR #764, opened 2021-07-02
[Popover] Restore `checkForDefaultPrevented`

andy-hook:
We'll be discussing the use of these composition utils (and this option) separately, however for now it's best to restore these to match the existing behaviour.

### DIFF
--- a/packages/react/popover/src/Popover.tsx
+++ b/packages/react/popover/src/Popover.tsx
@@ -226,8 +226,10 @@ const PopoverContentModal = React.forwardRef((props, forwardedRef) => {
           disableOutsidePointerEvents
           // When focus is trapped, a `focusout` event may still happen.
           // We make sure we don't trigger our `onDismiss` in such case.
-          onFocusOutside={composeEventHandlers(props.onFocusOutside, (event) =>
-            event.preventDefault()
+          onFocusOutside={composeEventHandlers(
+            props.onFocusOutside,
+            (event) => event.preventDefault(),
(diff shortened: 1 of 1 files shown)


packages/react/popover/src/Popover.tsx
```diff
@@ -255,21 +257,30 @@ const PopoverContentNonModal = React.forwardRef((props, forwardedRef) => {
...
+        onPointerDownOutside={composeEventHandlers(
+          props.onPointerDownOutside,
+          (event) => {
+            const { originalEvent } = event.detail;
+            const isLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === false;
+            isPointerDownOutsideRef.current = isLeftClick;
+          },
+          { checkForDefaultPrevented: false }
+        )}
+        onInteractOutside={composeEventHandlers(
```
andy-hook, line 269: This differs in that it covers the focus where before we didn't need to, not sure if that makes much difference where this is concerned.

benoitgrelard (approved):
Do we need to bring them back into the modality branch for `Dialog` too? I have a feeling I might have taken them out.

andy-hook:
> Do we need to bring them back into the modality branch for `Dialog` too? I have a feeling I might have taken them out.

If you did then it wasn't anytime recent, nothing in the modality diff that I can see.

benoitgrelard:
I guess the idea is to not change the current behaviour so I'd say leave it as is then.

PR #767, opened 2021-07-05
[Dialog][AlertDialog][Popover][DropdownMenu] Prevent pointer-events reenabling prematurely (mobile)

### LINKED ISSUE #631: [DropdownMenu] reopens when trigger is press on mobile
### Current Behavior

When the dropdown menu is open, pressing the trigger button closes and reopens the dropdown menu
https://streamable.com/vuqvg4

### Expected behavior

Expect the dropdown menu to be closed when trigger is pressed

### Reproducible example

https://radix-ui.com/primitives/docs/components/dropdown-menu#basic-example

### LINKED ISSUE #761: [Dialog] Re-opens on trigger click when modal
After some investigation into https://github.com/radix-ui/primitives/issues/631 I realised this cannot be fixed as easily as `DropdownMenu` so I've split these issues for individual investigation.

### LINKED ISSUE #762: [AlertDialog] Re-opens on trigger click
After some investigation into https://github.com/radix-ui/primitives/issues/631 I realised this cannot be fixed as easily as `DropdownMenu` so I've split these issues for individual investigation.

### LINKED ISSUE #763: [Popover] Re-opens on trigger click
After some investigation into https://github.com/radix-ui/primitives/issues/631 I realised this cannot be fixed as easily as `DropdownMenu` so I've split these issues for individual investigation.

jjenzz:
Fixes #631
Fixes #761 
Fixes #762 
Fixes #763

I managed to fix these all in one go eventually. 

@benoitgrelard  I believe I've also managed to provide a solution that merges both of our suggestions. My main concern was to keep the fix within `useBodyPointerEvents` since the issue is a side-effect of pointer-events enabling too soon, but I've only enabled the delay on `pointerdown` (because like you say, that's **when** it's an issue) and it's a touch or pen pointer.

Happy to continue to discuss alternate solutions as well if we have any because obviously, `setTimeout`s are never ideal. 

---------

UPDATE: Delay has been removed in favour of waiting for `click` event. I gave myself the idea here https://github.com/radix-ui/primitives/pull/767#issuecomment-874220750 😅

### DIFF
--- a/packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
+++ b/packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
@@ -24,7 +24,7 @@ function useBodyPointerEvents({ disabled }: { disabled: boolean }) {
       document.removeEventListener('pointerdown', handlePointerDown);
       document.removeEventListener('pointerup', handlePointerUp);
     };
-  }, [isTouchOrPenPressedRef]);
+  }, []);
 
   useLayoutEffect(() => {
     if (disabled) {
--- a/packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
+++ b/packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
@@ -5,20 +5,26 @@ let changeCount = 0;
 let originalBodyPointerEvents: string;
 
 function useBodyPointerEvents({ disabled }: { disabled: boolean }) {
-  const isTouchOrPenPressRef = React.useRef(false);
+  const isTouchOrPenPressedRef = React.useRef(false);
+  const isMousePressedRef = React.useRef(false);
 
   React.useEffect(() => {
     const handlePointerDown = (event: PointerEvent) => {
-      isTouchOrPenPressRef.current = event.pointerType !== 'mouse';
+      const isMouse = event.pointerType === 'mouse';
+      isTouchOrPenPressedRef.current = !isMouse;
+      isMousePressedRef.current = isMouse;
     };
-    const handlePointerUp = () => (isTouchOrPenPressRef.current = false);
+    const handlePointerUp = () => {
+      isTouchOrPenPressedRef.current = false;
+      isMousePressedRef.current = false;
+    };
     document.addEventListener('pointerdown', handlePointerDown);
     document.addEventListener('pointerup', handlePointerUp);
     return () => {
       document.removeEventListener('pointerdown', handlePointerDown);
       document.removeEventListener('pointerup', handlePointerUp);
     };
-  }, [isTouchOrPenPressRef]);
+  }, [isTouchOrPenPressedRef]);
 
   useLayoutEffect(() => {
     if (disabled) {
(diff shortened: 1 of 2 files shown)


jjenzz:
Alternatively, we change `onPointerDownOutside` to `onClickOutside`. I know that will make closing feel a bit slower but I'm not sure that making it feel more instantaneous is worth these side effects/timers.

packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
```diff
@@ -1,21 +1,76 @@
...
+           * - if long pressesing it, the events will execute after the longpress delay.
+           */
+          document.addEventListener('click', resetPointerEvents, { once: true });
+        } else if (isMousePressedRef.current) {
+          /**
+           * We force pointer-events to remain disabled until `pointerup` in-case a mouse pointer
+           * down re-enables pointer-events. Otherwise, events bound to inert controls could
+           * execute after pointer-events have been re-enabled, e.g. `select` event.
+           */
+          document.addEventListener('pointerup', resetPointerEvents, { once: true });
```
jjenzz, line 71: Fixes this issue https://modulz.slack.com/archives/C01T7FAM3CP/p1625499762023000?thread_ts=1625495026.018900&cid=C01T7FAM3CP

packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
```diff
@@ -1,21 +1,76 @@
...
+           * - if long pressesing it, the events will execute after the longpress delay.
+           */
+          document.addEventListener('click', resetPointerEvents, { once: true });
+        } else if (isMousePressedRef.current) {
+          /**
+           * We force pointer-events to remain disabled until `pointerup` otherwise, events
+           * bound to inert controls could execute after pointer-events have been re-enabled,
+           * e.g. `select` event.
+           */
+          document.addEventListener('pointerup', resetPointerEvents, { once: true });
```
jjenzz, line 71: Fixes this issue https://modulz.slack.com/archives/C01T7FAM3CP/p1625499762023000?thread_ts=1625495026.018900&cid=C01T7FAM3CP

packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
```diff
@@ -1,21 +1,76 @@
...
+      isTouchOrPenPressedRef.current = false;
+      isMousePressedRef.current = false;
+    };
+    document.addEventListener('pointerdown', handlePointerDown);
+    document.addEventListener('pointerup', handlePointerUp);
+    return () => {
+      document.removeEventListener('pointerdown', handlePointerDown);
+      document.removeEventListener('pointerup', handlePointerUp);
+    };
+  }, [isTouchOrPenPressedRef]);
```
norman-ags, line 27: Hello, I may be wrong about this but I think `useEffect` doesn't run when a ref changes.

On a side note, I've been waiting for this PR. Thanks for doing this!

benoitgrelard: Yes I think you're right, the dependency isn't needed.

packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
```diff
@@ -1,21 +1,76 @@
...
+
   useLayoutEffect(() => {
     if (disabled) {
       if (changeCount === 0) {
         originalBodyPointerEvents = document.body.style.pointerEvents;
       }
+
+      function resetPointerEvents() {
+        if (changeCount === 0) {
```
benoitgrelard, line 36: Nice to move this in here now 👍

packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
benoitgrelard: This part is new right?
Can you explain a bit more what you mean about "inert" controls? I'm not sure I follow.

benoitgrelard (commented):
Love the direction of this new solution!

One thing I've been wondering: does this also negate the need for the change I have made in `DismissableLayer`'s `usePointerDownOutside`?

andy-hook:
Love this!

> does this also negate the need for the change I have made in DismissableLayer's usePointerDownOutside?

Just for my benefit, are we talking about [this](https://github.com/radix-ui/primitives/pull/654/files#diff-a256967b057472a242da252da7a135b1c3cff0cbcda31180fcbcbad33726a0b0R195-R215)?

packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
jjenzz: A control that cannot be interacted with. Same definition as WAI ARIA.

packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
jjenzz: You are correct, the dependency isn't needed but I always thought we were against empty dependency arrays on this project for some reason. Happy to remove though.

jjenzz:
> One thing I've been wondering: does this also negate the need for the change I have made in `DismissableLayer`'s `usePointerDownOutside`?

@benoitgrelard Potentially, I'll try replicate the issue you were experiencing and confirm.

packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
benoitgrelard: I think we just follow the react hook eslint rules, so basically if it moans about something missing, it should be in there. In this case it doesn't moan which means it's safe.

jjenzz:
@benoitgrelard Yep, looks like it's fixed there too 🎉

jjenzz:
Spoke to Benoit about his issue and we think the fix there is still required since the fix in this PR will only apply to modal things.

benoitgrelard (approved):

PR #768, opened 2021-07-05
[DropdownMenu][ContextMenu] Add modality support

andy-hook:
This PR adds modality support via the `modal` prop in `DropdownMenu` and `ContextMenu`.

I've made the changes in the underlying `Menu` prim. It differs slightly from `Dialog` / `Popover` because `MenuContentImpl` also provides common functionality for submenus and I didn't want to conflate the `Modal` modes straight away in case the submenu impl needs to change in the near term.

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -29,10 +29,11 @@ const [ContextMenuProvider, useContextMenuContext] = createContext<ContextMenuCo
 type ContextMenuOwnProps = {
   onOpenChange?(open: boolean): void;
   dir?: Direction;
+  modal?: boolean;
 };
 
 const ContextMenu: React.FC<ContextMenuOwnProps> = (props) => {
-  const { children, onOpenChange, dir } = props;
+  const { children, onOpenChange, dir, modal } = props;
   const [open, setOpen] = React.useState(false);
   const isInsideContent = React.useContext(ContentContext);
   const handleOpenChangeProp = useCallbackRef(onOpenChange);
@@ -53,7 +54,7 @@ const ContextMenu: React.FC<ContextMenuOwnProps> = (props) => {
     </ContextMenuProvider>
   ) : (
     <ContextMenuProvider isRootMenu={true} open={open} onOpenChange={handleOpenChange}>
-      <MenuPrimitive.Root dir={dir} open={open} onOpenChange={handleOpenChange}>
+      <MenuPrimitive.Root dir={dir} open={open} onOpenChange={handleOpenChange} modal={modal}>
         {children}
       </MenuPrimitive.Root>
     </ContextMenuProvider>
@@ -141,7 +142,7 @@ const ContentContext = React.createContext(false);
 
 type ContextMenuContentOwnProps = Omit<
   Polymorphic.OwnProps<typeof MenuPrimitive.Content>,
-  'trapFocus' | 'disableOutsideScroll' | 'portalled' | 'side' | 'align'
+  'portalled' | 'side' | 'align'
 >;
 
 type ContextMenuContentPrimitive = Polymorphic.ForwardRefComponent<
@@ -183,15 +184,10 @@ type ContextMenuRootContentPrimitive = Polymorphic.ForwardRefComponent<
 >;
 
 const ContextMenuRootContent = React.forwardRef((props, forwardedRef) => {
-  const { disableOutsidePointerEvents = true, ...contentProps } = props;
-  const context = useContextMenuContext(CONTENT_NAME);
   return (
     <MenuPrimitive.Content
-      {...contentProps}
+      {...props}
       ref={forwardedRef}
-      disableOutsidePointerEvents={context.open ? disableOutsidePointerEvents : false}
-      trapFocus
-      disableOutsideScroll
       portalled
       side="right"
       sideOffset={2}
--- a/packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
@@ -50,6 +50,61 @@ export const Styled = () => (
   </div>
 );
 
+export const NonModal = () => {
+  return (
+    <div
+      style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', height: '100vh' }}
+    >
+      <div style={{ display: 'inline-flex', alignItems: 'center', flexDirection: 'column' }}>
+        <DropdownMenu modal={false}>
+          <DropdownMenuTrigger className={triggerClass}>Open</DropdownMenuTrigger>
+          <DropdownMenuContent className={contentClass} sideOffset={5}>
+            <DropdownMenuItem className={itemClass} onSelect={() => console.log('undo')}>
+              Undo
+            </DropdownMenuItem>
+            <DropdownMenuItem className={itemClass} onSelect={() => console.log('redo')}>
+              Redo
+            </DropdownMenuItem>
+            <DropdownMenuSeparator className={separatorClass} />
+            <DropdownMenu>
+              <DropdownMenuTriggerItem className={subTriggerClass}>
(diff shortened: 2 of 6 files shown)


packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -141,7 +142,7 @@ const ContentContext = React.createContext(false);
 
 type ContextMenuContentOwnProps = Omit<
   Polymorphic.OwnProps<typeof MenuPrimitive.Content>,
-  'trapFocus' | 'disableOutsideScroll' | 'portalled' | 'side' | 'align'
```
andy-hook, line 144: These are now narrowed at the `Menu` level

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -183,15 +184,10 @@ type ContextMenuRootContentPrimitive = Polymorphic.ForwardRefComponent<
...
 
 const ContextMenuRootContent = React.forwardRef((props, forwardedRef) => {
-  const { disableOutsidePointerEvents = true, ...contentProps } = props;
-  const context = useContextMenuContext(CONTENT_NAME);
   return (
     <MenuPrimitive.Content
-      {...contentProps}
+      {...props}
       ref={forwardedRef}
-      disableOutsidePointerEvents={context.open ? disableOutsidePointerEvents : false}
```
andy-hook, line 192: It wasn't clear to me what purpose this `open` check was serving?

packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
```diff
@@ -357,13 +407,13 @@ export const Chromatic = () => {
...
       <h2>Open</h2>
-      <DropdownMenu defaultOpen>
+      <DropdownMenu defaultOpen modal={false}>
         <DropdownMenuTrigger className={triggerClass}>Open</DropdownMenuTrigger>
         <DropdownMenuContent
           className={contentClass}
           sideOffset={5}
           avoidCollisions={false}
-          disableOutsideScroll={false}
+          onFocusOutside={(event) => event.preventDefault()}
```
andy-hook, line 416: See https://github.com/radix-ui/primitives/pull/751#discussion_r662394163

packages/react/menu/src/Menu.tsx
```diff
@@ -224,16 +228,36 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
   MenuRootContentOwnProps
 >;
 
 const MenuRootContent = React.forwardRef((props, forwardedRef) => {
   const context = useMenuContext(CONTENT_NAME);
+
+  return context.modal ? (
+    <MenuRootContentModal {...props} ref={forwardedRef} />
+  ) : (
+    <MenuRootContentNonModal {...props} ref={forwardedRef} />
```
andy-hook, line 245: Same as prior implementations. I think this is still the cleanest way to provide the two modes though I am a little concerned at the size of `Menu`

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -139,6 +147,9 @@ const DropdownMenuTrigger = React.forwardRef((props, forwardedRef) => {
         // only call handler if it's the left button (mousedown gets triggered by all mouse buttons)
         // but not when the control key is pressed (avoiding MacOS right click)
         if (event.button === 0 && event.ctrlKey === false) {
+          // prevent trigger focusing when opening
+          // this allows the content to be given focus without competition
+          if (!context.open) event.preventDefault();
```
andy-hook, line 152: In non modal mode the removal of trapping seems to let the trigger steal focus from content on open

packages/react/context-menu/src/ContextMenu.tsx
jjenzz: Historically we've done that because `closed !== unmounted` when animating out. So, this will have enabled pointer events outside while the animation was playing out.

jjenzz:
The changes look great but should we be allowing non modal menus?

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: I'm confused about this change. It feels like there's a deeper issue here if something is happening.

packages/react/menu/src/Menu.tsx
```diff
@@ -244,9 +268,40 @@ const MenuRootContent = React.forwardRef((props, forwardedRef) => {
...
 
   return (
-    <MenuContentImpl {...props} ref={composedRefs} onDismiss={() => context.onOpenChange(false)} />
+    <MenuContentImpl
+      {...props}
+      ref={composedRefs}
+      // we make sure we're not trapping once it's been closed
+      // (closed !== unmounted when animating out)
+      trapFocus={context.open}
+      disableOutsidePointerEvents
```
benoitgrelard, line 277: I think this is where https://github.com/radix-ui/primitives/pull/768/files?diff=split&w=1#r663963751 should move to now?

benoitgrelard:
> The changes look great but should we be allowing non modal menus? 💭

Interesting, what are you thinking here?

jjenzz:
> Interesting, what are you thinking here?

I guess I'm struggling to think how/when/why consumer would choose between one or the other for menus. What's an example of a situation where `modal` wouldn't be correct?

I'm wondering if we could just hardcode them to either modal or non-modal and it would be fine?

packages/react/dropdown-menu/src/DropdownMenu.tsx
jjenzz: It makes sense to me I think 💭 

The `mousedown` event waits for your handler to execute so it can check if you have called `event.preventDefault()` to prevent focusing. Therefore, when you mousedown on the trigger button: 

- it opens the menu 
- `FocusScope` kicks in and opens the container
- the browser `mousedown` handler checks if `event.preventDefault` was called
- it wasn't, so focuses the trigger

I do wonder if this would still be a problem if we used `onPointerDown` though instead. Did you try that already @andy-hook since we want to move in that direction anyway?

andy-hook: I'm seeing exactly what @jjenzz is describing for either events, though it's a good point that this should probably be a pointer event, the underlying issue remains though.

benoitgrelard: I'm just confused why this issue is new and wasn't present before I guess.

jjenzz: Yeah, although I would have thought there was something covering it up before (a race condition perhaps) and that this is more accurate given we know that `onMouseDown` will focus the button.

jjenzz:
I'm gonna leave this one to @benoitgrelard for approval since he is way more experienced on this side of the codebase than me.

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: afaik these were always focus trapped before, now without a trap the focus is allowed to shift back to the trigger

packages/react/menu/src/Menu.tsx
andy-hook: I think so, but it looks like this change previously only applied to `ContextMenu`? I'm seeing some strangeness on `DropdownMenu` with this change so need to investigate.

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: https://codesandbox.io/s/eloquent-williamson-qwp9y seems it should be working, at least in the current configuration 🤔 

That said, if we switch to `pointerDown` we will have the same issue. But I am curious what's happening under the hood here.

benoitgrelard (approved):
Looks great to me, tested in storybook too and works really well 👏

PR #771, opened 2021-07-08
[DropdownMenu][Popover][Dialog] Apply explicit trigger auto focus

andy-hook:
This change applies the focus fixes first added in https://github.com/radix-ui/primitives/pull/739 but now accounts for the two modality modes.
The above changes were intentionally removed during a rebase to make their addition easier to review and cleaner in history.


packages/react/dialog/src/Dialog.tsx
```diff
@@ -347,6 +339,19 @@ const DialogContentImpl = React.forwardRef((props, forwardedRef) => {
...
+          onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => {
+            preventAutoFocusRef.current = false;
+          })}
+          onPointerDownOutside={composeEventHandlers(
+            props.onPointerDownOutside,
+            (event) => {
+              const originalEvent = event.detail.originalEvent;
+              const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
+              const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
+              preventAutoFocusRef.current = isRightClick || !props.disableOutsidePointerEvents;
```
andy-hook, line 351: Rather than focusing on a left pointer down I've instead excluded the auto focus when right click or pointer events enabled (e.g. non modal).

This was the cleanest way I could see to handle the differing behaviour between the modality modes, let me know what you think though.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -225,8 +228,21 @@ const DropdownMenuRootContent = React.forwardRef((props, forwardedRef) => {
...
+      onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => {
+        preventAutoFocusRef.current = false;
       })}
+      onPointerDownOutside={composeEventHandlers(
+        props.onPointerDownOutside,
+        (event) => {
+          const originalEvent = event.detail.originalEvent;
+          const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
+          const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
+          preventAutoFocusRef.current = isRightClick || !context.modal;
```
andy-hook, line 242: I don't like that we need to pass `modal` down in context to make this work. I've had to do this because we have no access to the trigger at the `Menu` level. An alternative is to move the trigger into `Menu` but I'm a little concerned at the sheer number of concerns that `Menu` is already handling.

If anyone can think of an alternative I'd be open to it.

packages/react/menu/src/Menu.tsx
```diff
@@ -301,23 +300,9 @@ const MenuRootContentNonModal = React.forwardRef((props, forwardedRef) => {
...
-      onPointerDownOutside={composeEventHandlers(
-        props.onPointerDownOutside,
-        (event) => {
-          const originalEvent = event.detail.originalEvent as MouseEvent;
-          const isLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === false;
-          isPointerDownOutsideRef.current = isLeftClick;
-        },
+      onCloseAutoFocus={composeEventHandlers(
+        props.onCloseAutoFocus,
+        (event) => event.preventDefault(),
```
andy-hook, line 305: `Menu` will now always expect auto focus will be managed manually (where before it was leaning into the implicit auto focus behaviour hence it working fine with dropdown in the previous config)

packages/react/dialog/src/Dialog.tsx
jjenzz: I feel like I must be missing something here because my immediate thought is to wonder why we can't have the separate implementations in the different modal type components... like:

```js
const DialogContentModal = () => {
	// ...
	return (
		<DialogContentImpl 
			disableOutsidePointerEvents
			onUnmountAutoFocus={composeEventHandlers(onCloseAutoFocus, (event) => {
          		event.preventDefault();
          		if (!preventAutoFocusRef.current) context.triggerRef.current?.focus();
        	})}
			onPointerDownOutside={event => {
				const originalEvent = event.detail.originalEvent;
				const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
				const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
				preventAutoFocusRef.current = isRightClick
			}}
		/>
	);
}

// Doesn't need the trigger focus because pointer events aren't disabled?
const DialogContentNonModal = () => {
	// ...
	return (
		<DialogContentImpl 
			disableOutsidePointerEvents
			onUnmountAutoFocus={composeEventHandlers(onCloseAutoFocus, (event) => {
				event.preventDefault();
			})}
		/>
	);
}

Comment (jjenzz) [reply packages/react/dialog/src/Dialog.tsx:351]: Although, I'm curious why trigger focus doesn't happen when `disableOutsidePointerEvents` is `false` in your logic here? I think the original ticket was to specifically explicitly focus the trigger but this looks like it wouldn't get explicitly focused now in a non-modal dialog?

**Comment by benoitgrelard** [comment]
> Focus trigger on any pointerDown outside unless it's a "right click"

Did you consider the overlay too? I think we were saying from modal right click is allowed but only if there's no overlay right? ie. all but Dialog.

**Comment by andy-hook** [comment]
> Did you consider the overlay too? I think we were saying from modal right click is allowed but only if there's no overlay right? ie. all but Dialog.

I was referring to the auto focus of the trigger which I understood as never wanting to do that on right click? or were you referring to the "focusing of another element on right click" thing? e.g. I can right click on a textarea to copy some text even when modal?

**Comment by benoitgrelard** [comment]
Yeah I was referring to the latter. Sorry about that, so you mean this PR is purely only about the former?

**Comment by benoitgrelard** [comment]
Although it's kinda related, because when right clicking outside of a dialog which has an overlay, I would expect the focus to go back to the trigger.

**Comment by andy-hook** [comment]
> Yeah I was referring to the latter. Sorry about that, so you mean this PR is purely only about the former?

It was for the former at the time as the latter relies on [this change ](https://github.com/radix-ui/primitives/pull/772) but I can probably update this now if needed.

> Although it's kinda related, because when right clicking outside of a dialog which has an overlay, I would expect the focus to go back to the trigger.

Maybe my wires are crossed but I assumed we didn't want anything to happen when right clicking an overlay? e.g. it would remain open so focusing the trigger then wouldn't make sense?

**Comment by benoitgrelard** [comment]
> Maybe my wires are crossed but I assumed we didn't want anything to happen when right clicking an overlay? e.g. it would remain open so focusing the trigger then wouldn't make sense?

My bad, I'm the one who's getting confused, you're 100% correct.
This is hurting my brain… 🤯

File: packages/react/dialog/src/Dialog.tsx (hunk 1 above)
Comment (andy-hook) [reply packages/react/dialog/src/Dialog.tsx:351]: Yeah we can totally do this though I stopped short because we can't do it for `DropdownMenu` unless we move the trigger to `Menu`. I suppose we could have the implementation be different for that case.. or even bite the bullet and move the trigger?

**Comment by andy-hook** [comment]
@benoitgrelard @jjenzz – I've updated the PR with the suggested change

> Yeah I was referring to the latter.

On this point, I updated the branch with [this change](https://github.com/radix-ui/primitives/pull/772) which seems to get us most of the way there (@benoitgrelard you win, I actually thought this wouldn't be possible at all!)

https://user-images.githubusercontent.com/11708259/125306496-2fcb1480-e327-11eb-815a-fef9e6b4e39d.mp4

**Comment by benoitgrelard** [comment]
Amazing! Yeah that's what I thought that this is what would fix it. So good 🙂
Is this ready for review now then?

**Comment by andy-hook** [comment]
> Is this ready for review now then?

👍

File: packages/react/dialog/src/Dialog.tsx [hunk 4]
```diff
@@ -221,15 +222,19 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
...
+            event.preventDefault();
+            if (!isRightClickOutsideRef.current) context.triggerRef.current?.focus();
+          })}
           onPointerDownOutside={composeEventHandlers(props.onPointerDownOutside, (event) => {
-            const originalEvent = event.detail.originalEvent as MouseEvent;
-            const isRightClick =
-              originalEvent.button === 2 ||
-              (originalEvent.button === 0 && originalEvent.ctrlKey === true);
+            const originalEvent = event.detail.originalEvent;
+            const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
```
benoitgrelard, line 231: I like this! `ctrlLeftClick` 👍

packages/react/dialog/src/Dialog.tsx
```diff
@@ -221,15 +222,19 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
           // (closed !== unmounted when animating out)
           trapFocus={context.open}
           disableOutsidePointerEvents
+          onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
+            event.preventDefault();
+            if (!isRightClickOutsideRef.current) context.triggerRef.current?.focus();
```
benoitgrelard, line 227: Do we actually need this check here?

Because I see below that we say we don't close if it's a right click, therefore I'm thinking we never get to `onCloseAutoFocus` in the case of a right click?

packages/react/dialog/src/Dialog.tsx
```diff
@@ -244,24 +249,20 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
 
 const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
   const context = useDialogContext(CONTENT_NAME);
-  const isPointerDownOutsideRef = React.useRef(false);
+  const isEscapeKeyPressRef = React.useRef(false);
```
benoitgrelard, line 252: nit: Shall we name this `isEscapeKeyDown` considering that's the name of the handler?

packages/react/popover/src/Popover.tsx
```diff
@@ -224,6 +225,21 @@ const PopoverContentModal = React.forwardRef((props, forwardedRef) => {
...
+            props.onPointerDownOutside,
+            (event) => {
+              const originalEvent = event.detail.originalEvent;
+              const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
+              const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
+
+              isRightClickOutsideRef.current = isRightClick;
+            },
+            { checkForDefaultPrevented: false }
+          )}
```
benoitgrelard, line 242: 👍

packages/react/popover/src/Popover.tsx
```diff
@@ -240,7 +256,7 @@ const PopoverContentModal = React.forwardRef((props, forwardedRef) => {
 const PopoverContentNonModal = React.forwardRef((props, forwardedRef) => {
   const { portalled = true, ...contentNonModalProps } = props;
   const context = usePopoverContext(CONTENT_NAME);
-  const isPointerDownOutsideRef = React.useRef(false);
+  const isEscapeKeyPressRef = React.useRef(false);
```
benoitgrelard, line 259: Same nit on the name here

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -215,6 +217,7 @@ type DropdownMenuRootContentPrimitive = Polymorphic.ForwardRefComponent<
 const DropdownMenuRootContent = React.forwardRef((props, forwardedRef) => {
   const { portalled = true, ...contentProps } = props;
   const context = useDropdownMenuContext(CONTENT_NAME);
+  const preventAutoFocusRef = React.useRef(false);
```
benoitgrelard, line 220: nit: can we rename this more explicitly `preventCloseAutoFocusRef`?

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: It doesn't bother me too much, I think ultimately this is easier to understand than if we had to move around the whole trigger requirement into `Menu`. We've tried that before and it was pretty confusing if I remember correctly.

benoitgrelard (commented):
It looks good to me overall, just left one question and a few nits.

packages/react/dialog/src/Dialog.tsx
jjenzz: agree, i liked this clarity too :+1:

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -225,8 +228,21 @@ const DropdownMenuRootContent = React.forwardRef((props, forwardedRef) => {
       portalled={portalled}
       onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
         event.preventDefault();
-        context.triggerRef.current?.focus();
+        if (!preventCloseAutoFocusRef.current) context.triggerRef.current?.focus();
+      })}
+      onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => {
+        preventCloseAutoFocusRef.current = false;
       })}
```
jjenzz, line 235: Do we need `{ checkForDefaultPrevented: false }` here? In fact, do we need it on all of the handlers that set the handlers in this PR that set `isPointerDownOutsideRef` or `preventCloseAutoFocusRef` ?

benoitgrelard (approved):

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: I thought we said we would re-assess all this separately?

andy-hook: Honestly I wasn't on which way to go here, we did say we'd assess it separately so I just carried over the existing stance from the pointer handlers in `Popover` and `Menu`.

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: Yeah true, I really didn't really want to go there a third time haha

jjenzz (approved):

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: Oh that's true, i'll revert it on `onEscapeKeyDown` just to keep it inline with how it was for now.

PR #775, opened 2021-07-13
[DismissableLayer] Prevent issue when mounting via `pointerdown`

benoitgrelard:


jjenzz (approved):

Luncher:
Hi @benoitgrelard @jjenzz Can you explain what it means by "mounting via pointerdown", thanks!

PR #777, opened 2021-07-13
[Slider] Prevent thumb focus when disabled

jjenzz:


benoitgrelard (approved):
Just missing versioning, all good otherwise 👍

PR #778, opened 2021-07-13
[Collection] Widen usage of items

benoitgrelard:
Previously `CollectionSlot` included the provider part, so it typically meant that whatever we decided was the root of the collection (in terms of DOM scoping) was also the subtree in which the items would be available.

This was fine for all current usage, but `Select` needs to be able to access the items, when typing whilst the trigger is focused, and we wouldn't necessarily want to add an extra element surrounding the entire select just in order to provide a root for DOM scoping.

So I've basically split up `CollectionSlot` in 2 separate components (`CollectionProvider` and `CollectionRootSlot`).
It doesn't really change the usage much for current components, and it means I can now bring the `CollectionProvider` much higher up the tree as necessary.

### DIFF
--- a/packages/react/collection/src/Collection.tsx
+++ b/packages/react/collection/src/Collection.tsx
@@ -2,11 +2,9 @@ import React from 'react';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { Slot } from '@radix-ui/react-slot';
 
-type SlotProps = React.ComponentProps<typeof Slot>;
-
 function createCollection<ItemElement extends HTMLElement, ItemData>() {
   /* -----------------------------------------------------------------------------------------------
-   * Collection
+   * CollectionProvider
    * ---------------------------------------------------------------------------------------------*/
 
   type CollectionElement = HTMLElement;
@@ -17,21 +15,39 @@ function createCollection<ItemElement extends HTMLElement, ItemData>() {
   };
   const Context = React.createContext<ContextValue>({} as any);
 
-  const COLLECTION_SLOT_NAME = 'CollectionSlot';
+  const PROVIDER_NAME = 'CollectionProvider';
 
-  const CollectionSlot = React.forwardRef<CollectionElement, SlotProps>((props, forwardedRef) => {
+  const CollectionProvider: React.FC = (props) => {
     const { children } = props;
     const ref = React.useRef<CollectionElement>(null);
-    const composedRefs = useComposedRefs(forwardedRef, ref);
     const itemMap = React.useRef<ContextValue['itemMap']>(new Map()).current;
     return (
       <Context.Provider value={React.useMemo(() => ({ itemMap, collectionRef: ref }), [itemMap])}>
-        <Slot ref={composedRefs}>{children}</Slot>
+        {children}
       </Context.Provider>
     );
-  });
+  };
 
-  CollectionSlot.displayName = COLLECTION_SLOT_NAME;
+  CollectionProvider.displayName = PROVIDER_NAME;
+
+  /* -----------------------------------------------------------------------------------------------
+   * CollectionRootSlot
+   * ---------------------------------------------------------------------------------------------*/
+
+  const ROOT_SLOT_NAME = 'CollectionRootSlot';
+
+  type SlotProps = React.ComponentProps<typeof Slot>;
+
+  const CollectionRootSlot = React.forwardRef<CollectionElement, SlotProps>(
+    (props, forwardedRef) => {
+      const { children } = props;
+      const context = React.useContext(Context);
+      const composedRefs = useComposedRefs(forwardedRef, context.collectionRef);
+      return <Slot ref={composedRefs}>{children}</Slot>;
+    }
+  );
+
+  CollectionRootSlot.displayName = ROOT_SLOT_NAME;
 
   /* -----------------------------------------------------------------------------------------------
    * CollectionItem
@@ -84,7 +100,7 @@ function createCollection<ItemElement extends HTMLElement, ItemData>() {
     };
   }
 
-  return [CollectionSlot, CollectionItemSlot, useCollection] as const;
+  return [CollectionProvider, CollectionRootSlot, CollectionItemSlot, useCollection] as const;
 }
 
 export { createCollection };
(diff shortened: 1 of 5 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -209,15 +211,17 @@ const MenuContent = React.forwardRef((props, forwardedRef) => {
...
-    <Presence present={forceMount || context.open}>
-      <CollectionSlot>
-        {context.isSubmenu ? (
-          <MenuSubContent {...contentProps} ref={forwardedRef} />
-        ) : (
-          <MenuRootContent {...contentProps} ref={forwardedRef} />
-        )}
-      </CollectionSlot>
-    </Presence>
+    <CollectionProvider>
```
benoitgrelard, line 214: I've put this one above `Presence` because `Presence` clones its child and adds a ref to it.

jjenzz:
I'm guessing it will it be something structurally something along these lines for `Select`?

```jsx
<CollectionProvider>
	<Select>
		<SelectTrigger />
		<CollectionRootSlot>
			<SelectContent>
				<CollectionItemSlot><SelectItem /></CollectionItemSlot>
				<CollectionItemSlot><SelectItem /></CollectionItemSlot>
				<CollectionItemSlot><SelectItem /></CollectionItemSlot>
			</SelectContent>
		</CollectionRootSlot>
	</Select>
</CollectionProvider>
```

packages/react/collection/src/Collection.tsx
```diff
@@ -17,21 +15,39 @@ function createCollection<ItemElement extends HTMLElement, ItemData>() {
...
+
+  /* -----------------------------------------------------------------------------------------------
+   * CollectionRootSlot
+   * ---------------------------------------------------------------------------------------------*/
+
+  const ROOT_SLOT_NAME = 'CollectionRootSlot';
+
+  type SlotProps = React.ComponentProps<typeof Slot>;
+
+  const CollectionRootSlot = React.forwardRef<CollectionElement, SlotProps>(
```
jjenzz, line 41: Can we name this `CollectionSlot` still? Curious why we needed the name change.

jjenzz:
I know this is a separate discussion but I'm super confused why we put the `Slot` internally on the collection components instead of doing `<Collection as={Slot}>` / `<CollectionItem as={Slot} />` like we do elsewhere.

Wondering if it's worth creating a ticket to update it for consistency or if there was a reason we couldn't do that?

benoitgrelard:
> I'm guessing it will it be something **structurally** along these lines for `Select`?
> 
> ```js
> <CollectionProvider>
> 	<Select>
> 		<SelectTrigger />
> 		<CollectionRootSlot>
> 			<SelectContent>
> 				<CollectionItemSlot><SelectItem /></CollectionItemSlot>
> 				<CollectionItemSlot><SelectItem /></CollectionItemSlot>
> 				<CollectionItemSlot><SelectItem /></CollectionItemSlot>
> 			</SelectContent>
> 		</CollectionRootSlot>
> 	</Select>
> </CollectionProvider>
> ```

Exactly 👌

packages/react/collection/src/Collection.tsx
benoitgrelard: I renamed it because I thought it was now clearer that it was about the DOM slots, one for the `root` and one for each `item`.

Happy to revert that bit though if you think it doesn't help anything.

benoitgrelard:
> I know this is a separate discussion but I'm super confused why we put the `Slot` internally on the collection components instead of doing `<Collection as={Slot}>` / `<CollectionItem as={Slot} />` like we do elsewhere.
> 
> Wondering if it's worth creating a ticket to update it for consistency or if there was a reason we couldn't do that which I've no doubt forgotten?

We did this together, because we were having problems with types because there's a generic, and it's not statically analysable because it's returned from a dynamic function.

jjenzz:
> We did this together back then remember?

Yes I remember we did it for a reason but I couldn't remember the reason, hence asking.

> It's because we were having problems with types because there's a generic

Eek. TypeScript ruining all the fun! 

If you fancy being super generous would you mind slipping a comment in the code about that while we're here? haha. Happy for us to do it later in a separate PR but I'm really keen to get these implicit things documented better in the code for mine (because my memory is awful 😅 ) and @andy-hook's sake.

packages/react/collection/src/Collection.tsx
jjenzz: Yeah I don't think it is helping because usually we name our components like `Menu` and `MenuItem` not `MenuRoot` and `MenuItem`. 

I ignore the `Slot` part of the name as I read it mentally like [this](https://github.com/radix-ui/primitives/pull/778#issuecomment-879169603).

benoitgrelard:
> If you fancy being super generous would you mind slipping a comment in the code about that while we're here? haha. Happy for us to do it later in a separate PR but I'm really keen to get these implicit things documented better in the code for mine (because my memory is awful 😅 ) and @andy-hook's sake.

Absolutely, will save us more question later.

benoitgrelard:
That's all done now.

packages/react/collection/src/Collection.tsx
```diff
@@ -2,11 +2,14 @@ import React from 'react';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { Slot } from '@radix-ui/react-slot';
 
-type SlotProps = React.ComponentProps<typeof Slot>;
+// We have resorted to returning slots directly rather than exposing primitives that can then
+// be slotted like `<CollectionItem as={Slot}>…</CollectionItem>`. This is because we were having
+// issues with types because there are generics at play and it's not statically analyzable because
+// the components are returned by a dynamic function.
```
andy-hook, line 8:
```suggestion
// We have resorted to returning slots directly rather than exposing primitives that can then
// be slotted like `<CollectionItem as={Slot}>…</CollectionItem>`.
// This is because we encountered issues with generic types, static analysis 
// and the dynamic function used to return the components.
```

jjenzz (approved):

andy-hook (approved):

packages/react/collection/src/Collection.tsx
jjenzz: > This is because we encountered issues with generic types that cannot be statically analyised
> due to being created dynamically via `createCollection`.

jjenzz:
Small improvement to the comment because it sounds like they're all different issues atm. When they're actually the precursors to the issue right? https://github.com/radix-ui/primitives/pull/778#discussion_r669059369

benoitgrelard:
@jjenzz done

andy-hook:
>  it sounds like they're all separate issues atm when they're actually the precursors to the one issue right?

ah true! didn't mean to block here, just found the original a bit of a mouthful to read back 😅

benoitgrelard:
Haha yeah that's my bad english, we need Pedro's succinct touch to write our comments 😀

PR #779, opened 2021-07-13
[Polymorphic] Allow `ComponentType` props to override

jjenzz:
I saw [this discussion](https://github.com/radix-ui/primitives/discussions/774) raised today and it got me wondering why we explicitly always made the `OwnProps` take precedence. I really can't seem to remember but all tests pass if the `as` component props take precedence. Do you recall why we did it the other way around @benoitgrelard ?

While I was there I had a brainwave on some perf improvements which I have also introduced here but want to test this all more thoroughly (including the prop override reorder) when I have time so will leave this in draft for now for discussion.

Quick test to show polymorphism still working:

https://user-images.githubusercontent.com/175330/125493802-7efd7119-0378-4927-9a97-0442b90be87c.mp4

### DIFF
--- a/packages/react/polymorphic/src/polymorphic.ts
+++ b/packages/react/polymorphic/src/polymorphic.ts
@@ -36,20 +36,18 @@ interface ForwardRefComponent<
   /**
    * When `as` prop is passed, use this overload.
    * Merges original own props (without DOM props) and the inferred props
-   * from `as` element with the own props taking precendence.
+   * from `as` element with the inferred taking precendence.
    *
    * We explicitly avoid `React.ElementType` and manually narrow the prop types
    * so that events are typed when using JSX.IntrinsicElements.
    */
-  <
-    As extends
-      | keyof JSX.IntrinsicElements
-      | React.ComponentType<any> = NarrowIntrinsic<IntrinsicElementString>
-  >(
+  <As = NarrowIntrinsic<IntrinsicElementString>>(
     props: As extends keyof JSX.IntrinsicElements
-      ? Merge<JSX.IntrinsicElements[As], OwnProps & { as: As }>
+      ? Merge<OwnProps, Merge<JSX.IntrinsicElements[As], { as: As }>>
       : As extends React.ComponentType<infer P>
-      ? Merge<P, OwnProps & { as: As }>
+      ? Merge<OwnProps, Merge<P, { as: As }>>
+      : As extends ''
+      ? { as: keyof JSX.IntrinsicElements }
       : never
   ): React.ReactElement | null;
 }
--- a/.yarn/versions/4a422b8c.yml
+++ b/.yarn/versions/4a422b8c.yml
@@ -1,0 +1,39 @@
+releases:
+  "@radix-ui/react-accessible-icon": patch
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-announce": patch
+  "@radix-ui/react-arrow": patch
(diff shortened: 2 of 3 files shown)


benoitgrelard:
I cannot immediately remember no, but my hunch is there must be a good reason for it, because I remember us doing this specifically and there's even a comment for it… 🤔

jjenzz:
I vaguely remember it might have been something to do with a scenario where the `OwnProps` contain a prop with the same name as a DOM prop. 

In that case we want the `as` props `OwnProps` to override but don't want its DOM props to override the conflicting `OwnProps`.

E.g.

```jsx
const ComponentOwnProps = {
	onChange(thing: boolean): void;
};

const Component = () => { /* ... */ };

// thing would be `event` type from `input`
return <Component as="input" onChange={thing => console.log(thing)} />;
```

I'm not sure how common this conflict is likely to be though 🤔

benoitgrelard:
Ha yes, you might be onto something. Isn't it when we want to remove the DOM prop by doing `never`?

packages/react/polymorphic/src/polymorphic.ts
```diff
@@ -36,20 +36,18 @@ interface ForwardRefComponent<
...
-  >(
+  <As = NarrowIntrinsic<IntrinsicElementString>>(
     props: As extends keyof JSX.IntrinsicElements
-      ? Merge<JSX.IntrinsicElements[As], OwnProps & { as: As }>
+      ? Merge<OwnProps, Merge<JSX.IntrinsicElements[As], { as: As }>>
       : As extends React.ComponentType<infer P>
-      ? Merge<P, OwnProps & { as: As }>
+      ? Merge<OwnProps, Merge<P, { as: As }>>
+      : As extends ''
+      ? { as: keyof JSX.IntrinsicElements }
```
jjenzz, line 50: This means the component types will only have to worry about the huge `JSX.IntrinsicElements` object when someone is in the process of adding the `as` prop.

jjenzz:
> Ha yes, you might be onto something. Isn't it when we want to remove the DOM prop by doing `never`?

That might be it!!! In which case, this change can't go in (yet). 

I feel like that `never` thing needs more thought because conceptually it seems to make sense to me that whatever is passed to `as` would take precedence. When I have some time I'll have a think about an alternate way to solve the `never` thing.

benoitgrelard:
> I feel like that `never` thing needs more thought because conceptually it seems to make sense to me that whatever is passed to `as` would take precedence.

I agree with this sentiment too, when I saw the issue I thought, yeah that seems like what you'd want.

jjenzz:
Thought some more about this over the weekend and realised there is good reason why we did that. Because, that's how polymorphism works 🙈 😅   

Although it seems to make sense that the types for the component passed to `as` would take precedence, functionality-wise, that's not the case: https://codesandbox.io/s/kind-worker-83vfn 

I've split the performance improvements into a separate PR and I'll need to have a think on whether the expectation is possible here in some way (overriding `OwnProps` with **custom props** from passed component).

jjenzz:
Closing this for now because of the above ☝️

marceloadsj:
Make sense 😆.

> ...I'm wondering if primitive types should union instead of override but tbh, it seems the expectation isn't correct... 🤔
Yep, I think the way it is currently working is the correct way, because of this: https://github.com/radix-ui/primitives/discussions/774#discussioncomment-1022799

So, it seems something more related to the right way of thinking about polymorphism on components, than anything else.
I still have to align better my mind with that paradigm 😅. Thanks a lot!

Issue #617, opened 2021-04-23
[Tooltip] On a Button that opens a modal, Tooltip reappears when modal is closed


lin-jessica:
## Question

I've got a button with a Tooltip that opens a Modal when clicked. When the modal is closed, focus state returns to that button causing the Tooltip to reappear. What are my options for making this less weird for seeing-eye users?

Some more details reported on [Reach/ui](https://github.com/reach/reach-ui/issues/790)

benoitgrelard:
Hi there, yes this is something we have also faced in the Modulz app.
We have some ideas here but we haven't gotten around to it yet.
We will update you once we start working on it.

jjenzz:
For reference, here was the sandbox I had created a while back demoing things we had discussed https://codesandbox.io/embed/usefocusenter-forked-x6i0n

rafgraph:
I'm having a related issue when switching tabs or applications, the tooltip can reappear when switching back to the tab with the tooltip. This happens when the tooltip trigger still has focus but the tooltip is no longer shown, then when switching tabs/apps, the browser calls focus on the tooltip trigger when switching back, causing the tooltip to show again. Note that this happens on Chrome v91 MacOS, but not Safari MacOS as Safari doesn't blur-and-re-focus on tab/app switching.

One idea would be to show the tooltip when `focus` is caused by keyboard input, but not when `focus` is caused by mouse or touch input (just use hover and long press to trigger the tooltip from mouse and touch input). Something like [Event From](https://github.com/rafgraph/event-from) might help with this.

benoitgrelard:
I may not be understanding correctly, but I don't think that use-case is an issue.
ie. If the trigger was focused and tooltip was showing before you command+tab out of the browser, and you come back, the trigger is still focused so the tooltip should still show.

rafgraph:
The issue is that it also happens when the tooltip is not showing before switching tabs/apps (when the trigger still has focus).

https://user-images.githubusercontent.com/11911299/121581809-49153400-c9fc-11eb-9119-fa5b48cf4def.mov

benoitgrelard:
Ah yes I see, thanks for the extra video. That's kind of a different issue from the original one though, I would say less important/annoying too. Would you mind creating a separate issue for it so we can track the different things separately?

rafgraph:
[...]

PR #783, opened 2021-07-15
[Tooltip] Prevent tooltip from opening on tab change

### LINKED ISSUE #705: [Tooltip] Reappears when switching tabs/apps
## Bug report

Branched from related issue: https://github.com/radix-ui/primitives/issues/617

### Current Behavior

When switching tabs or applications, the tooltip can reappear. This happens when the tooltip trigger still has focus (but the tooltip is not shown). When switching tabs/apps, the browser calls focus on the tooltip trigger when switching back, causing the tooltip to show again. Note that this happens on Chrome v91 MacOS, but not Safari MacOS as Safari doesn't blur-and-re-focus on tab/app switching.

### Expected behavior

The tooltip should not show when switching tabs/apps (unless it is shown when leaving the tab/app).

### Reproducible example

https://user-images.githubusercontent.com/11911299/121581809-49153400-c9fc-11eb-9119-fa5b48cf4def.mov

### Your environment

| Radix Package(s) |    Tooltip     |    0.0.18     |
| React            | n/a     |    17     |
| Browser          |    Chrome     |     91    |

jjenzz:
Fixes #705 

This seemed like a simple enough fix so jumped on it quick for a break.

It's different to #617 which happens because the content that the trigger opens is considered to be outside the tooltip tree, so returning to the tooltip trigger behaves like a brand new focus.

Happy to close this if we think the two issues can be resolved by the same solution, I'm just not sure if they can.

### DIFF
--- a/.yarn/versions/b521ae54.yml
+++ b/.yarn/versions/b521ae54.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -115,10 +115,16 @@ const Tooltip: React.FC<TooltipOwnProps> = (props) => {
   }, [contentId]);
 
   React.useEffect(() => {
-    const handleFocus = () => (hasWindowFocusedRef.current = true);
+    let previouslyFocusedElement: Element | null;
+    const handleFocus = () => (hasWindowFocusedRef.current = previouslyFocusedElement === trigger);
+    const handleBlur = () => (previouslyFocusedElement = document.activeElement);
     window.addEventListener('focus', handleFocus);
-    return () => window.removeEventListener('focus', handleFocus);
-  }, []);
+    window.addEventListener('blur', handleBlur);
+    return () => {
+      window.removeEventListener('focus', handleFocus);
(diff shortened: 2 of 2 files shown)


benoitgrelard:
You are right it's a different issue than #617 as you noted. However I'm not sure this is the right solution for this one because it only handles tab change, not document focus change. As in, if my page is still visible, but I command+tab to another app, or click in another app, and then back to the browser with my page.

jjenzz:
I wasn't aware those other scenarios were an issue, apologies. I've updated the event listener to handle all cases.

benoitgrelard (commented):
I'm seeing some inconsistent behaviour with this now. The tooltip is not always showing when it should, hard to explain but I can show you on a call if you want, ping me.

jjenzz:
Yeah I see issues too which aren't easy to solve because a `blur` event is triggered if you go from the button to the browser chrome, and a `focus` if you go from chrome to button. 

I've pushed a solution I thought would might solve it but doesn't because of the above issue. 

Okay, like everything, it seems this is much trickier to solve than i initially anticipated so I've pushed something I tried (which doesn't work because of the above issue) so we have a history of what was tried and I'm going to close the PR.

The remaining issue with this code (as far as I can tell) is the following:

- Click button
- Click the address bar
- Tab to button

Tooltip should appear but doesn't.

https://user-images.githubusercontent.com/175330/125800593-f74a267d-44c8-4c95-be3c-1684140718f5.mp4

PR #784, opened 2021-07-16
[Polymorphic] Improve performance

jjenzz:
It's really hard to know for sure if tweaks to polymorphic types are improving performance but I picked a huge component (`ScrollArea`) and recorded a video of me tweaking its types (with and without these changes) to show how TS is able to more quickly report issues now.

The main change here is to remove the `extends keyof JSX.IntrinsicElements` from the generic which would have been slowing down **all components** regardless of whether the `as` prop was being used.

Instead, I check to see if `as` extends an empty string and if so, I type it as `keyof JSX.IntrinsicElements` to enable intellisense when used.

**INTELLISENSE**

https://user-images.githubusercontent.com/175330/125993174-b89e107b-4caa-4485-af6a-55c18b1e9347.mp4

**BEFORE** (~8s)

https://user-images.githubusercontent.com/175330/125992071-b1aa176b-6e08-4717-9b84-d6bb2ec77b22.mp4

**AFTER** (~5s)

https://user-images.githubusercontent.com/175330/125992101-02282990-5ddf-470c-b721-6421b83c7a77.mp4

I'll continue to think of ways to improve it further but this definitely feels better.

### DIFF
--- a/packages/react/polymorphic/src/polymorphic.ts
+++ b/packages/react/polymorphic/src/polymorphic.ts
@@ -15,8 +15,6 @@ type OwnProps<E> = E extends ForwardRefComponent<any, infer P> ? P : {};
  */
 type IntrinsicElement<E> = E extends ForwardRefComponent<infer I, any> ? I : never;
 
-type NarrowIntrinsic<E> = E extends keyof JSX.IntrinsicElements ? E : never;
-
 type ForwardRefExoticComponent<E, OwnProps> = React.ForwardRefExoticComponent<
   Merge<E extends React.ElementType ? React.ComponentPropsWithRef<E> : never, OwnProps & { as?: E }>
 >;
@@ -41,7 +39,7 @@ interface ForwardRefComponent<
    * We explicitly avoid `React.ElementType` and manually narrow the prop types
    * so that events are typed when using JSX.IntrinsicElements.
    */
-  <As = NarrowIntrinsic<IntrinsicElementString>>(
+  <As = IntrinsicElementString>(
     props: As extends ''
       ? { as: keyof JSX.IntrinsicElements }
       : As extends React.ComponentType<infer P>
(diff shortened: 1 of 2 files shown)


andy-hook:
Ran a quick test with `--diagnostics` and looks like this change is more than halving the check time 😮 

image

packages/react/polymorphic/src/polymorphic.ts
```diff
@@ -41,15 +41,13 @@ interface ForwardRefComponent<
...
-  <
-    As extends
-      | keyof JSX.IntrinsicElements
-      | React.ComponentType<any> = NarrowIntrinsic<IntrinsicElementString>
-  >(
-    props: As extends keyof JSX.IntrinsicElements
-      ? Merge<JSX.IntrinsicElements[As], OwnProps & { as: As }>
+  <As = IntrinsicElementString>(
+    props: As extends ''
+      ? { as: keyof JSX.IntrinsicElements }
```
andy-hook, line 46: Not super familiar with these typings. Is the gist that the conditionals are progressively narrowing from simple to more complex only as necessary?

packages/react/polymorphic/src/polymorphic.ts
```diff
@@ -41,15 +41,13 @@ interface ForwardRefComponent<
    * We explicitly avoid `React.ElementType` and manually narrow the prop types
    * so that events are typed when using JSX.IntrinsicElements.
    */
-  <
-    As extends
-      | keyof JSX.IntrinsicElements
-      | React.ComponentType<any> = NarrowIntrinsic<IntrinsicElementString>
```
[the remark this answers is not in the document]
jjenzz: Oh yes, good catch!

packages/react/polymorphic/src/polymorphic.ts
jjenzz: Yes, and we do this instead of checking `As extends React.ElementType` because that breaks event typing for intrinsics.

jjenzz:
> Ran a quick test with `--diagnostics` and looks like this change is more than halving the check time 

Never used that before, TIL 😅

andy-hook (approved):
Looks good from my side, works well and the perf improvement is mega 👍

PR #794, opened 2021-07-22
[Slider] React.StrictMode fixes

### LINKED ISSUE #615: [Slider] Thumb loses focus when value changes

### LINKED ISSUE #773: [Slider] Not working in Next.js when reactStrictMode is enabled
## Bug report

Hello,
I copy [basic example](https://www.radix-ui.com/docs/primitives/components/slider#basic-example) with stitches from official docs.
I use it in Next.js app with enabled [react strict mode](https://nextjs.org/docs/api-reference/next.config.js/react-strict-mode) in next.config.js

### Current Behavior

Server warning `Did not expect server HTML to contain a <span> in <span>`...
Can't change value with thumb.

### Expected behavior

When `reactStrictMode: false` it work as expected.

### Reproducible example

[StackBlitz](https://stackblitz.com/edit/nextjs-fcxaig)

jjenzz:
----------

**NOTE:** PR will be easier to review with hidden whitespace changes

_______

Fixes #615 
Fixes #773 

Replaces our old `collection` logic with our new `Collection`. This fixes the `React.StrictMode` issues that have been reported lately, but means the thumbs will no longer be in the correct position on initial render for SSR as our new `Collection` component doesn't handle this case. 

More information [here](https://github.com/radix-ui/primitives/issues/615#issuecomment-884850078). Thanks to @andy-hook for narrowing down collection code as the issue.

It seems this is a less jarring experience for now so I'm thinking we can look into improving SSR experience across our components separately.

**Tested in:**

- [X] NextJS React.StrictMode
- [X] Create React App React.StrictMode
- [X] Snowpack React.StrictMode

https://user-images.githubusercontent.com/175330/126637939-40385aaa-2d8d-4793-9097-a8755844e444.mp4


packages/react/slider/src/Slider.stories.tsx
```diff
@@ -176,6 +176,26 @@ export const WithinForm = () => {
   );
 };
 
+export const Strict = () => (
```
andy-hook, line 179: Should we go ahead and add the `strictMode` flag in storybook to cover this instead?

packages/react/slider/src/Slider.tsx
```diff
@@ -493,12 +497,18 @@ type SliderThumbPrimitive = Polymorphic.ForwardRefComponent<
...
-  const { ref: collectionRef, index } = useSliderCollectionItem();
-  const ref = useComposedRefs(forwardedRef, collectionRef);
   const context = useSliderContext(THUMB_NAME);
+  const { getItems } = useCollectionItem();
+  const [thumb, setThumb] = React.useState<React.ElementRef<typeof SliderThumbImpl> | null>(null);
+  const composedRefs = useComposedRefs(forwardedRef, (node) => setThumb(node));
+  const index = React.useMemo(
+    () => (thumb ? getItems().findIndex((item) => item.ref.current === thumb) : 0),
+    [getItems, thumb]
+  );
```
jjenzz, line 507: I thought this part would mean all thumbs would render at `0` on the range since we require a re-render from state to `findIndex` but seems okay. There's something about React that I still need to learn here apparently.

packages/react/slider/src/Slider.tsx
```diff
@@ -493,12 +497,18 @@ type SliderThumbPrimitive = Polymorphic.ForwardRefComponent<
...
 
 const SliderThumb = React.forwardRef((props, forwardedRef) => {
-  const { ref: collectionRef, index } = useSliderCollectionItem();
-  const ref = useComposedRefs(forwardedRef, collectionRef);
   const context = useSliderContext(THUMB_NAME);
+  const { getItems } = useCollectionItem();
+  const [thumb, setThumb] = React.useState<React.ElementRef<typeof SliderThumbImpl> | null>(null);
+  const composedRefs = useComposedRefs(forwardedRef, (node) => setThumb(node));
+  const index = React.useMemo(
+    () => (thumb ? getItems().findIndex((item) => item.ref.current === thumb) : 0),
```
andy-hook, line 505: Curious, could `value` technically be inaccurate on the first render if we're falling back to `0` here?

packages/react/slider/src/Slider.stories.tsx
jjenzz: We can maybe do that separately because all stories will need testing thoroughly and any issues logged when we do that work. Perhaps I can update the ticket you created for it to remind whoever does it to remove it from this story?

packages/react/slider/src/Slider.tsx
jjenzz: Yep, that's the whole SSR index issue I was explaining. The old collection stuff tried to solve this.

andy-hook: This is challenging my understanding as well haha

packages/react/slider/src/Slider.tsx
andy-hook: I think the server is rendering all thumb index's as `0` right now but it appears to work fine here because we only have one thumb. `context.values[0]` will be accurate for the first but with more I think we'll see the mismatch still?

jjenzz: Ah man you're right, my brain isn't working. I was expecting it to be positioned at `0` on the range but it is `0` value index haha. So yeah, issue still remains.

andy-hook: I wonder then, if we want to tackle the SSR separately perhaps we choose not to render the thumbs while we're waiting on the node? it would at least appear more visually consistent until we can come up with an alternative counting method that works with `strict`?

jjenzz: I can't get the `node` if I don't render the thumb... haha

jjenzz: Ooo, I can `hidden` it tho... :+1:

andy-hook: > I can't get the `node` if I don't render the thumb, the node is the thumb... 🙈 haha

haha, of course 😂

jjenzz: @andy-hook I've hidden them for a render instead 🙂

packages/react/slider/src/Slider.tsx
```diff
@@ -486,28 +488,44 @@ SliderRange.displayName = RANGE_NAME;
...
+   * Until we figure out SSR indexes, we hide the thumbs on initial render while we work out
+   * the index. Otherwise, SSR will place the thumbs in the wrong position and snap into the
+   * right position during hydration which would be jarring.
+   */
+  return isInitialRenderRef.current || index !== -1 ? (
+    <SliderThumbImpl
+      {...props}
+      ref={composedRefs}
+      index={index}
+      style={isInitialRenderRef.current ? { display: 'none' } : undefined}
```
andy-hook, line 521: Think we need to spread in consumer styles? is there any consequence to `display: none` vs `visibility: hidden`?

packages/react/slider/src/Slider.tsx
```diff
@@ -486,28 +488,44 @@ SliderRange.displayName = RANGE_NAME;
...
+
+  React.useEffect(() => {
+    isInitialRenderRef.current = false;
+  }, []);
+
+  /**
+   * Until we figure out SSR indexes, we hide the thumbs on initial render while we work out
+   * the index. Otherwise, SSR will place the thumbs in the wrong position and snap into the
+   * right position during hydration which would be jarring.
+   */
```
andy-hook, line 515:
```suggestion
 /**
   * We hide thumbs on the first render otherwise SSR will initially place them
   * in the wrong position and later snap them into the correct position during hydration
   * which would be visually jarring.
   * 
   * Previously we accounted for this by managing indexes inside the collection ahead of time,
   * but this caused problems with `React.StrictMode` so we're falling back to this for now.
   */
```

Nit: How about something like this?

andy-hook, line 515:
```suggestion
  /**
   * We hide thumbs on the first render while we work out the index, otherwise SSR will
   * initially place them in the wrong position and later snap them into the correct position
   * during hydration which would be visually jarring.
   * 
   * Previously we accounted for this by managing the indexes inside of a collection ahead of time
   * but this caused problems with `React.StrictMode`, so we're falling back to this for now.
   */
```

Is that any better for our future selves?

packages/react/slider/src/Slider.tsx
jjenzz: The thumbs are generally absolutely positioned on top of the slider so they can overlap. Hiding them should be fine but good point on consumer styles, will add those instead of `undefined`!

packages/react/slider/src/Slider.tsx
jjenzz: I've tweaked the top bit but not added the bottom bit as it doesn't seem worthwhile to have comments in the code that say "previously we did this". We have Github history in commits and issues for that. Thoughts? 🙏

andy-hook (approved):

packages/react/slider/src/collection.tsx
```diff
@@ -1,130 +0,0 @@
-import * as React from 'react';
```
benoitgrelard, line 1: Nice to see all this go, maybe the package will even slim down a bit 😀 
Not sure…

packages/react/slider/src/Slider.tsx
```diff
@@ -486,28 +488,44 @@ SliderRange.displayName = RANGE_NAME;
...
+  React.useEffect(() => {
+    isInitialRenderRef.current = false;
+  }, []);
+
+  /**
+   * We hide thumbs on the first render while we work out the index, otherwise SSR will
+   * render them in the wrong position before they snap into the correct position during
+   * hydration which would be visually jarring.
+   */
+  return isInitialRenderRef.current || index !== -1 ? (
```
benoitgrelard, line 516: If we're checking for the initial render case here, is the `index !== -1` just a safety to not render the whole thing if for some reason that thumb wasn't tracked? I can't think of any reason why this check would be necessary right now.

packages/react/slider/src/Slider.tsx
```diff
@@ -486,28 +488,44 @@ SliderRange.displayName = RANGE_NAME;
...
+  );
+
+  React.useEffect(() => {
+    isInitialRenderRef.current = false;
+  }, []);
+
+  /**
+   * We hide thumbs on the first render while we work out the index, otherwise SSR will
+   * render them in the wrong position before they snap into the correct position during
+   * hydration which would be visually jarring.
```
benoitgrelard, line 514: Might have been ok to leave it as is rather than doing more work specifically to hide it?
Are you thinking that we'll deal with this anyway differently at some point?

benoitgrelard (commented):
Nice work on this! 👍

packages/react/slider/src/collection.tsx
jjenzz: I wondered the same haha 🤞

packages/react/slider/src/Slider.tsx
jjenzz: It's necessary for this case:

```jsx
<Slider.Root value={[20]}>
	...
	<Slider.Thumb />
	<Slider.Thumb />
</Slider.Root>
```

Notice how there are two thumbs but only one value? The second thumb won't render. We handled this case before with the `value !== undefined` check but perhaps it's okay to let them break it and remove redundant JSX instead?

packages/react/slider/src/Slider.tsx
jjenzz: You think it's okay for them to render in the wrong position and snap into a different position during hydration?

The plan is to fix the SSR issue eventually so that it renders in the correct position server-side.

packages/react/slider/src/Slider.tsx
benoitgrelard: Yeah that’s probably fine too seeing as it’s just a user error that they would figure out pretty easily.

packages/react/slider/src/Slider.tsx
benoitgrelard: It’s not great but I wonder how visible it is in a real use case. But yeah regardless SSR is something we need to look at some point. We might end up with some controlled overrides (like providing the index yourself) when you want to make sure it’s ok on first server render.

packages/react/slider/src/Slider.tsx
jjenzz: I can create a PR for that then :+1:

packages/react/slider/src/Slider.tsx
benoitgrelard: I think these kind of solutions would definitely be the easiest. Simple trade offs.

packages/react/slider/src/Slider.tsx
jjenzz: Oh actually, I just thought about this some more... the `SliderImpl` requires an `index`. So, it would be wrong for me to pass an index that doesn't exist. 

Perhaps it would be less confusing if I made the index fallback to `undefined` and did a `index !== undefined` check here instead?

PR #798, opened 2021-07-26
[Slider] Improve SSR fallback logic

jjenzz:
These changes stemmed from some original feedback [here](https://github.com/radix-ui/primitives/pull/794#discussion_r676404750).

I removed the `isInitialRenderRef` check because it seems a clearer approach was to check that we have a `value` like before.


benoitgrelard (approved):
👌

PR #799, opened 2021-07-26
[Dialog][Popover][DropdownMenu] Adjust non-modal auto focus behaviour

andy-hook:
Previously non-modal auto focus was not covering use of the close button. This change applies auto focus for all interactions except when pointer down outside.

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -20,6 +20,7 @@ type ContextMenuContextValue = {
   isRootMenu: boolean;
   open: boolean;
   onOpenChange(open: boolean): void;
+  modal: boolean;
 };
 
 const [ContextMenuProvider, useContextMenuContext] = createContext<ContextMenuContextValue>(
@@ -33,7 +34,7 @@ type ContextMenuOwnProps = {
 };
 
 const ContextMenu: React.FC<ContextMenuOwnProps> = (props) => {
-  const { children, onOpenChange, dir, modal } = props;
+  const { children, onOpenChange, dir, modal = true } = props;
   const [open, setOpen] = React.useState(false);
   const isInsideContent = React.useContext(ContentContext);
   const handleOpenChangeProp = useCallbackRef(onOpenChange);
@@ -47,13 +48,23 @@ const ContextMenu: React.FC<ContextMenuOwnProps> = (props) => {
   );
 
   return isInsideContent ? (
-    <ContextMenuProvider isRootMenu={false} open={open} onOpenChange={handleOpenChange}>
+    <ContextMenuProvider
+      isRootMenu={false}
+      open={open}
+      onOpenChange={handleOpenChange}
+      modal={modal}
+    >
       <MenuPrimitive.Sub open={open} onOpenChange={handleOpenChange}>
         {children}
       </MenuPrimitive.Sub>
     </ContextMenuProvider>
   ) : (
-    <ContextMenuProvider isRootMenu={true} open={open} onOpenChange={handleOpenChange}>
+    <ContextMenuProvider
+      isRootMenu={true}
+      open={open}
+      onOpenChange={handleOpenChange}
+      modal={modal}
+    >
       <MenuPrimitive.Root dir={dir} open={open} onOpenChange={handleOpenChange} modal={modal}>
         {children}
       </MenuPrimitive.Root>
@@ -184,6 +195,9 @@ type ContextMenuRootContentPrimitive = Polymorphic.ForwardRefComponent<
 >;
 
 const ContextMenuRootContent = React.forwardRef((props, forwardedRef) => {
+  const context = useContextMenuContext(CONTENT_NAME);
+  const shouldPreventFocusRef = React.useRef(false);
+
   return (
     <MenuPrimitive.Content
       {...props}
@@ -192,6 +206,22 @@ const ContextMenuRootContent = React.forwardRef((props, forwardedRef) => {
       side="right"
       sideOffset={2}
       align="start"
+      onCloseAutoFocus={(event) => {
+        props.onCloseAutoFocus?.(event);
+
+        if (!event.defaultPrevented && shouldPreventFocusRef.current) {
+          event.preventDefault();
+        }
+
+        shouldPreventFocusRef.current = false;
+      }}
+      onInteractOutside={(event) => {
+        props.onInteractOutside?.(event);
+
+        if (!event.defaultPrevented && !context.modal) {
+          shouldPreventFocusRef.current = true;
+        }
+      }}
     />
   );
 }) as ContextMenuRootContentPrimitive;
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -256,13 +256,30 @@ const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
         ref={forwardedRef}
         trapFocus={false}
         disableOutsidePointerEvents={false}
-        onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
-          event.preventDefault();
-          if (!isPointerDownOutsideRef.current) context.triggerRef.current?.focus();
-        })}
         onPointerDownOutside={composeEventHandlers(
           props.onPointerDownOutside,
           () => (isPointerDownOutsideRef.current = true),
+          { checkForDefaultPrevented: false }
+        )}
+        onCloseAutoFocus={composeEventHandlers(
+          props.onCloseAutoFocus,
+          (event) => {
+            if (!event.defaultPrevented && !isPointerDownOutsideRef.current) {
+              context.triggerRef.current?.focus();
+            }
+            event.preventDefault();
+            isPointerDownOutsideRef.current = false;
+          },
+          { checkForDefaultPrevented: false }
+        )}
+        onEscapeKeyDown={composeEventHandlers(
+          props.onEscapeKeyDown,
+          () => (isPointerDownOutsideRef.current = false),
+          { checkForDefaultPrevented: false }
+        )}
+        onFocus={composeEventHandlers(
+          props.onFocus,
+          () => (isPointerDownOutsideRef.current = false),
           { checkForDefaultPrevented: false }
         )}
         onInteractOutside={composeEventHandlers(props.onInteractOutside, (event) => {
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -247,7 +247,8 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
 
 const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
   const context = useDialogContext(CONTENT_NAME);
-  const isEscapeKeyDownRef = React.useRef(false);
+  const isPointerDownOutsideRef = React.useRef(false);
+
   return (
     <Portal>
       <DialogContentImpl
@@ -257,11 +258,15 @@ const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
         disableOutsidePointerEvents={false}
         onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
           event.preventDefault();
-          if (isEscapeKeyDownRef.current) context.triggerRef.current?.focus();
+          if (!isPointerDownOutsideRef.current) context.triggerRef.current?.focus();
         })}
-        onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => {
-          isEscapeKeyDownRef.current = true;
-        })}
+        onPointerDownOutside={composeEventHandlers(
+          props.onPointerDownOutside,
+          () => {
+            isPointerDownOutsideRef.current = true;
+          },
+          { checkForDefaultPrevented: false }
+        )}
         onInteractOutside={composeEventHandlers(props.onInteractOutside, (event) => {
           // Prevent dismissing when clicking the trigger.
           // As the trigger is already setup to close, without doing so would
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -258,18 +258,16 @@ const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
         disableOutsidePointerEvents={false}
         onCloseAutoFocus={(event) => {
           props.onCloseAutoFocus?.(event);
-          const userPrevented = event.defaultPrevented;
 
-          if (!userPrevented) {
+          if (!event.defaultPrevented && !isInteractOutsideRef.current) {
             event.preventDefault();
-            if (!isInteractOutsideRef.current) context.triggerRef.current?.focus();
+            context.triggerRef.current?.focus();
           }
 
           isInteractOutsideRef.current = false;
         }}
         onInteractOutside={(event) => {
           props.onInteractOutside?.(event);
-          const userPrevented = event.defaultPrevented;
 
           // Prevent dismissing when clicking the trigger.
           // As the trigger is already setup to close, without doing so would
@@ -280,7 +278,7 @@ const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
           const target = event.target as HTMLElement;
           const targetIsTrigger = context.triggerRef.current?.contains(target);
           if (targetIsTrigger) event.preventDefault();
-          else if (!userPrevented) isInteractOutsideRef.current = true;
+          else if (!event.defaultPrevented) isInteractOutsideRef.current = true;
         }}
       />
     </Portal>
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -247,7 +247,7 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
 
 const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
   const context = useDialogContext(CONTENT_NAME);
-  const isInteractOutsideRef = React.useRef(false);
+  const shouldFocusTriggerRef = React.useRef(true);
 
   return (
     <Portal>
@@ -260,14 +260,17 @@ const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
           props.onCloseAutoFocus?.(event);
 
           if (!event.defaultPrevented) {
+            // Always prevent auto focus because we either focus manually or want user agent focus
             event.preventDefault();
-            if (!isInteractOutsideRef.current) context.triggerRef.current?.focus();
+            if (shouldFocusTriggerRef.current) context.triggerRef.current?.focus();
           }
 
-          isInteractOutsideRef.current = false;
+          shouldFocusTriggerRef.current = true;
         }}
         onInteractOutside={(event) => {
           props.onInteractOutside?.(event);
+
+          if (!event.defaultPrevented) shouldFocusTriggerRef.current = false;
 
           // Prevent dismissing when clicking the trigger.
           // As the trigger is already setup to close, without doing so would
@@ -278,7 +281,6 @@ const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
           const target = event.target as HTMLElement;
           const targetIsTrigger = context.triggerRef.current?.contains(target);
           if (targetIsTrigger) event.preventDefault();
-          else if (!event.defaultPrevented) isInteractOutsideRef.current = true;
         }}
       />
     </Portal>
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -247,7 +247,7 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
 
 const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
   const context = useDialogContext(CONTENT_NAME);
-  const isPointerDownOutsideRef = React.useRef(false);
+  const isInteractOutsideRef = React.useRef(false);
 
   return (
     <Portal>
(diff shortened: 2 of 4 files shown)


packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -229,8 +229,6 @@ const DropdownMenuRootContent = React.forwardRef((props, forwardedRef) => {
       onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
         event.preventDefault();
         if (!preventCloseAutoFocusRef.current) context.triggerRef.current?.focus();
-      })}
-      onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => {
```
andy-hook, line 233: We need to reset the ref here between closes, this change previously assumed it would only ever be closed via `escape` key. Now it will reset after every close regardless.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -257,11 +258,15 @@ const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
...
-          isEscapeKeyDownRef.current = true;
+          if (!isPointerDownOutsideRef.current) context.triggerRef.current?.focus();
         })}
+        onPointerDownOutside={composeEventHandlers(
+          props.onPointerDownOutside,
+          () => {
+            isPointerDownOutsideRef.current = true;
+          },
+          { checkForDefaultPrevented: false }
+        )}
```
jjenzz, line 269: if you fancy being generous haha, nit 🙈 :

```suggestion
        onPointerDownOutside={composeEventHandlers(
          props.onPointerDownOutside,
          () => (isPointerDownOutsideRef.current = true),
          { checkForDefaultPrevented: false }
        )}
```

packages/react/popover/src/Popover.tsx
```diff
@@ -269,11 +269,15 @@ const PopoverContentNonModal = React.forwardRef((props, forwardedRef) => {
...
-          isEscapeKeyDownRef.current = true;
+          if (!isPointerDownOutsideRef.current) context.triggerRef.current?.focus();
         })}
+        onPointerDownOutside={composeEventHandlers(
+          props.onPointerDownOutside,
+          () => {
+            isPointerDownOutsideRef.current = true;
+          },
+          { checkForDefaultPrevented: false }
+        )}
```
jjenzz, line 280: same nit 🙈 :

```suggestion
        onPointerDownOutside={composeEventHandlers(
          props.onPointerDownOutside,
          () => (isPointerDownOutsideRef.current = true),
          { checkForDefaultPrevented: false }
        )}
```

jjenzz (dismissed):

jjenzz:
❤️

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: This is one of those cases where we need to make sure they can't cancel that bit of the behaviour then right? Because right now, if they prevent default to run their own focus, the ref won't be reset.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -247,7 +247,8 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
 
 const DialogContentNonModal = React.forwardRef((props, forwardedRef) => {
   const context = useDialogContext(CONTENT_NAME);
-  const isEscapeKeyDownRef = React.useRef(false);
+  const isPointerDownOutsideRef = React.useRef(false);
```
benoitgrelard, line 250: This one is never getting reset in `DialogContentNonModal` or `PopoverContentNonModal`?

andy-hook: They are reset during teardown / remount. `DropdownMenuRootContent` is unusual because this is handled one level lower by `Menu`, which means the top level state persists unless we reset manually.

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: I think that depends, the current assumption makes this moot because if we `preventDefault` on `onCloseAutoFocus` then we're opting out of the built in focus handling all together right? or is the expectation that the user is cancelling the default behaviour of **the handler**, and that the conditions with which it is run should persist? e.g. the `onCloseAutoFocus` handler itself is subject to pointer conditions, not just the focus behaviour within it.

packages/react/dialog/src/Dialog.tsx
benoitgrelard: Teardown/remount isn't enough though, because you can have a pointer down but no close (so no teardown) for example when preventing default in `onPointerDownOutside` or in `onInteractOutside`.

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: Well I think it's more about ensuring the "integrity" of the system. That's kinda when we need to think about which bit they can or cannot prevent. ie. here it's important that it gets reset properly when it's actually closed.

packages/react/dialog/src/Dialog.tsx
andy-hook: Good point and made me question another assumption, given the the case you describe we should probably be resetting when interacting _inside_ as well.

benoitgrelard: It's really tricky, because basically what you want is reset it if there was a pointer down that didn't result in a close. But you don't really know this yet.

andy-hook: Yeah.. the simplest way I could see was to explicitly handle ref clearing https://github.com/radix-ui/primitives/pull/799/commits/b74b6cfed252e009e5e9f952bd50cd2bb8519027 it's a bit verbose though.

benoitgrelard: Yeah I don't love it haha, gonna have to think some more about this.
Any ideas @jjenzz?

jjenzz: I'm not sure I'm fully understanding the issue. How do I replicate from within the `modality` branch?

- Open dialog
- Press close button
- Trigger doesn't auto focus?

I've just tried these steps in the `modality` branch and the trigger focuses fine so I'm missing something here for sure.

andy-hook: It should focus if you close via the button. Try opening non-modal with keyboard and closing with return / space, it should focus the trigger but it doesn't

jjenzz: I'm probably missing something but would this suffice?

image


- Preventing default in `onPointerDownOutside` or `onFocusOutside` would prevent dismissing and `onInteractOutside` so would never set the ref to `true`.
- Dismissing via outside interaction only would set it to `true` which would close it and bring it back into it's default `false` state upon remount.

benoitgrelard: > … and `onInteractOutside` so would never set the ref to true.

That's not the case right? `onInteractOutside` will run even if you call `preventDefault` in `onPointerDownOutside` or `onFocusOutside`.

benoitgrelard: But now it's making me think, perhaps it's as simple as resetting it in `onInteractOutside` when `!event.defaultPrevented`?

jjenzz: > `onInteractOutside` will run even if you call `preventDefault` in `onPointerDownOutside` or `onFocusOutside`.

Nope, which is really baffling to me but it doesn't. Because yah, I originally did what you suggested and console.logged `event.defaultPrevented` because I wanted to check that, but it never logged ...

benoitgrelard: hmmm

image

The only thing behind the `defaultPrevented` check is `onDismiss`…

jjenzz: If we want to be sure though, seems this should be all we need:

```diff
- else  isInteractOutsideRef.current = true;
+ else if (!event.defaultPrevented) isInteractOutsideRef.current = true;
```

jjenzz: Yep, I have no idea what's going on there. It's made me worry that we might have a load of logic elsewhere that isn't running 😅

jjenzz: If I change it to this then it calls it fine, but with the `event` param it doesn't

image

jjenzz: Looks like there is a bug somewhere which we can fix separately, and we should definitely do the `!event.defaultPrevented` check in the meantime because `onInteractOutside` should be called.

andy-hook: Did I open another can of worms here? haha. I've updated from the pro tips, is this what you had in mind? https://github.com/radix-ui/primitives/pull/799/commits/dfe047989d4ac612b3d1836fb9853d60415620b8

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -254,6 +248,11 @@ const DropdownMenuRootContent = React.forwardRef((props, forwardedRef) => {
           // focus on pointer down, creating the same issue.
           const targetIsTrigger = context.triggerRef.current?.contains(event.target as HTMLElement);
           if (targetIsTrigger) event.preventDefault();
+          else if (!event.defaultPrevented) {
+            const originalEvent = event.detail.originalEvent as PointerEvent;
+            const isModalLeftClick = context.modal && originalEvent.button === 0;
```
andy-hook, line 253: Do we foresee any issue with this? was easier to move it from `onPointerDownOutside` as we have to account for both modes in this comp.

packages/react/dialog/src/Dialog.tsx
benoitgrelard: haha I "think" that's correct, but man this is a real slippery area 😀 
I'm gonna have to take a deeper look at it and actually play with it myself a bit to get my head around it properly.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -247,21 +247,26 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
...
-        onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, () => {
-          isEscapeKeyDownRef.current = true;
-        })}
+        onCloseAutoFocus={composeEventHandlers(
+          props.onCloseAutoFocus,
+          (event) => {
+            if (!event.defaultPrevented && !isInteractOutsideRef.current) {
+              context.triggerRef.current?.focus();
+            }
+            event.preventDefault();
```
benoitgrelard, line 265: Man, this is all getting so confusing because of our `checkForDefaultPrevented` thing haha. Next step will really be to do something about this because it's melting my brain 😀

Here for example, it feels weird to see a check for `!event.defaultPrevented` and also have `checkForDefaultPrevented: false` 😀

Also, the placement of our own `event.preventDefault()` bothers me…
- I understand we need to do it in **ALL** cases because either:
  - we don't want to run any focus at all (ie. you've clicked somewhere outside, so natural focus should occur)
  - we want explicit focus to trigger so we prevent the default close auto focus
- So, instinctively I want to put it as the first line at the beginning of the handler, but we can't because we immediately check for `!event.defaultPrevented` (because we need to know if it's been prevented by the user, not us)

As a test I tried to see how it would look like if we didn't use `composeEventHandlers` and ended with this which definitely makes me think we should move away from it soon because I find it way clearer:

```tsx
  props.onCloseAutoFocus?.(event);
  const userPrevented = event.defaultPrevented;

  if (!userPrevented) {
    event.preventDefault();
    if (!isInteractOutsideRef.current) context.triggerRef.current?.focus();
  }

  isInteractOutsideRef.current = false;
}}
```

With the current way we end up with a double negative sort of situation, I find it super hard to reason about, especially because for example our own `event.preventDefault()` is run even if the user has already prevented. I know it doesn't make any difference, but it makes it quite confusing when I read it.

What do we all think about this? Do we wait? or do we bite the bullet and start here with this change before updating everywhere later?

andy-hook: I'm glad you bring this up because this whole thing often makes my head spin and I find myself wondering whether I'm just dense or if this approach needs adjustment.

In your example, will the consumer setting

```jsx
onCloseAutoFocus={(event) => event.preventDefault()}
```

Revert the auto focus behaviour to the implicit, last focused element thing?

benoitgrelard: No if the user prevent defaults, no focus behaviour will happen because the check is done in `FocusScope` and we pass it down: `onUnmountAutoFocus={onCloseAutoFocus}`.

andy-hook: My 2c is that your suggestion makes A LOT more sense and I would prefer to go with it here. I also support incrementally moving away from the utils if we agree that's the right thing to do.

benoitgrelard: Yeah at this point I think I'd also say we should do it incrementally as it seems pointless to have those things use it even though we all feel it's making things worse and harder to understand.

jjenzz: Aw man, I totally agree with all of this and **MUCH** prefer your suggestion Benoit. I don't think we need to make huge sweeping codebase changes if we decide to change direction, I think it's fine to start here and avoid `composeEventHandlers` going forward and have a separate task to update legacy stuff.

benoitgrelard: Yay for all agreeing 🎉 
Let's goooo!

andy-hook: I've tweaked this for non-modal for now https://github.com/radix-ui/primitives/pull/799/commits/cdc24481073ed22d40c98bc816316ec58a0f719d will update the others in those files separately if we're good.

Question on moving away from the utils, in the common case does this mean we'll just be adopting an inline condition in place?

e.g.

```jsx
onClick={composeEventHandlers(onClick, () => {
  ...
})}
      
// Will become

onClick={(event) => {
  props.onClick?.(event)
  const userPrevented = event.defaultPrevented

  if (!userPrevented) {
   ...
  }
}}
```

jjenzz: Yeah although I'm find with just doing:

```jsx
onClick={(event) => {
  props.onClick?.(event)
  if (!event.defaultPrevented) {
   ...
  }
}}
```

It's not necessarily "userPrevented". We could prevent things internally via a capture event or something for example.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -247,31 +247,41 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
...
+          props.onCloseAutoFocus?.(event);
+          const userPrevented = event.defaultPrevented;
+
+          if (!userPrevented) {
+            event.preventDefault();
+            if (!isInteractOutsideRef.current) context.triggerRef.current?.focus();
+          }
+
+          isInteractOutsideRef.current = false;
+        }}
```
jjenzz, line 269: Was there a reason we didn't go with this?

```suggestion
        onCloseAutoFocus={(event) => {
          props.onCloseAutoFocus?.(event);
          if (!event.defaultPrevented && !isInteractOutsideRef.current) {
			context.triggerRef.current?.focus();
          }

          isInteractOutsideRef.current = false;
          event.preventDefault();
        }}
```

The current one confused me because `preventDefault` is always happening either way. Either they are or we are so we may as well `event.preventDefault()` all the time. It didn't make sense why that had to be part of the conditional.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -247,31 +247,41 @@ const DialogContentModal = React.forwardRef((props, forwardedRef) => {
...
           // cause it to close and immediately open.
           //
           // We use `onInteractOutside` as some browsers also
           // focus on pointer down, creating the same issue.
-          const targetIsTrigger = context.triggerRef.current?.contains(event.target as HTMLElement);
+          const target = event.target as HTMLElement;
+          const targetIsTrigger = context.triggerRef.current?.contains(target);
           if (targetIsTrigger) event.preventDefault();
-        })}
+          else if (!userPrevented) isInteractOutsideRef.current = true;
```
jjenzz, line 283: again, I'm fine with just `else if (!event.defaultPrevented)` here because it's not necessarily the user preventing it.

jjenzz (approved):
some nits, but not gonna hold up the PR for them. Rest looks good :+1:

packages/react/dialog/src/Dialog.tsx
jjenzz: In fact, shouldn't we only be preventing default if we're focusing ourselves? so it would be in **both** conditionals:

```jsx
          props.onCloseAutoFocus?.(event);
          if (!event.defaultPrevented && !isInteractOutsideRef.current) {
            event.preventDefault();
            context.triggerRef.current?.focus();
          }
          isInteractOutsideRef.current = false;
        }}
```

?

andy-hook: Mainly for the ordering reason in Benoit's comment https://github.com/radix-ui/primitives/pull/799#discussion_r678099342 if we take it out of the condition then we need to ensure it comes after.

packages/react/dialog/src/Dialog.tsx
andy-hook: Yeah I'm on board with this 👍

packages/react/dialog/src/Dialog.tsx
andy-hook: > In fact, shouldn't we only be preventing default if we're focusing ourselves?

Ah I didn't see this comment before I added mine, yeah I think you're right!

jjenzz (approved):
Code-wise looks good to me but I'll leave it to Benoit to make sure it's all good behaviour-wise since he has a much deeper understanding than me here 😅

packages/react/dialog/src/Dialog.tsx
benoitgrelard: No, that's not right, we need to prevent **all the time** not just when we're focusing something manually. Like I explained here: https://github.com/radix-ui/primitives/pull/799#discussion_r678099342

> I understand we need to do it in ALL cases because either:
> - **we don't want to run any focus at all (ie. you've clicked somewhere outside, so natural focus should occur)**
> - we want explicit focus to trigger so we prevent the default close auto focus

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -226,37 +226,34 @@ const DropdownMenuRootContent = React.forwardRef((props, forwardedRef) => {
...
-          const targetIsTrigger = context.triggerRef.current?.contains(event.target as HTMLElement);
-          if (targetIsTrigger) event.preventDefault();
-        },
-        { checkForDefaultPrevented: false }
-      )}
+      onCloseAutoFocus={(event) => {
+        props.onCloseAutoFocus?.(event);
+
+        if (!event.defaultPrevented && !isInteractOutsideRef.current) {
+          event.preventDefault();
```
benoitgrelard, line 233: `DropdownMenu.Root` can be modal or non-modal right?
So in this case this is correct here t only `preventDefault` if we focus ourselves.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -226,37 +226,34 @@ const DropdownMenuRootContent = React.forwardRef((props, forwardedRef) => {
...
+          if (!isInteractOutsideRef.current) context.triggerRef.current?.focus();
```
benoitgrelard, line 234: This check is already in the above conditional.

packages/react/popover/src/Popover.tsx
```diff
@@ -267,28 +267,30 @@ const PopoverContentNonModal = React.forwardRef((props, forwardedRef) => {
...
-            const targetIsTrigger = context.triggerRef.current?.contains(target);
-            if (targetIsTrigger) event.preventDefault();
-          },
-          { checkForDefaultPrevented: false }
-        )}
+        onCloseAutoFocus={(event) => {
+          props.onCloseAutoFocus?.(event);
+
+          if (!event.defaultPrevented && !isInteractOutsideRef.current) {
+            event.preventDefault();
```
benoitgrelard, line 274: That one is wrong too I think, same as non modal dialog.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -226,37 +226,34 @@ const DropdownMenuRootContent = React.forwardRef((props, forwardedRef) => {
...
+        //
+        // We use `onInteractOutside` as some browsers also
+        // focus on pointer down, creating the same issue.
+        const target = event.target as HTMLElement;
+        const targetIsTrigger = context.triggerRef.current?.contains(target);
+        if (targetIsTrigger) event.preventDefault();
+        else if (!event.defaultPrevented) {
+          const originalEvent = event.detail.originalEvent as PointerEvent;
+          const isModalLeftClick = context.modal && originalEvent.button === 0;
+          isInteractOutsideRef.current = !isModalLeftClick;
```
benoitgrelard, line 254: I'm still struggling to follow the double negatives a bit, like we set a `isInteractOutside` which is about what "happened" and set it to the negative of something.

Then later we do `if (!isInteractOutside…)`

I feel like it would be easier for the ref we track to be about the "what" rather than the "how", ie. something like `shouldReturnFocus`.

Can we go on a call and pair quickly to see if it would help?

benoitgrelard (approved):

packages/react/dialog/src/Dialog.tsx
andy-hook: @benoitgrelard is on the money here, it always needs to happen in non-modal to avoid focus fighting on outside clicks etc. https://github.com/radix-ui/primitives/pull/799/commits/fb6e32cca44c9a271a9ada0db4d75713e230e308 this whole thing got a little confusing after a while 😅

andy-hook:
@jjenzz – Me and @benoitgrelard had a call on this and made some readability changes plus reverted some wrong assumptions I had with the last round of changes. Going to merge now and we can discuss more in the main modality branch if needed.

packages/react/dialog/src/Dialog.tsx
jjenzz: Ah of course... wow, yeah, my head is spinning 😅

jjenzz: On the whole "coming after" thing. What's wrong with this? In the olden days this was actually the recommended place to put it to ensure progressive enhancement didn't break if there was a JS error in the handling logic.

jjenzz: See this example where when JS fails, the app still works. **if the event is prevented after** https://codepen.io/pen/?editors=1010

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -192,6 +206,22 @@ const ContextMenuRootContent = React.forwardRef((props, forwardedRef) => {
...
       align="start"
+      onCloseAutoFocus={(event) => {
+        props.onCloseAutoFocus?.(event);
+
+        if (!event.defaultPrevented && shouldPreventFocusRef.current) {
+          event.preventDefault();
+        }
+
+        shouldPreventFocusRef.current = false;
+      }}
```
jjenzz, line 217: @andy-hook I saw your comment about further discussing in the `modality` branch but i find it's easier to follow if we keep all discussion related to this change in one place.

So, on this one, curious why we no longer focus the trigger here when we prevent default? We had to do it explicitly in `main` to fix `autoFocus` issues.

andy-hook: This change was to match the non-modal behaviour we have in the other components. That is that when clicking outside in non-modal, focus shouldn't be returned.

Not familiar with the issue you mentioned, do you have a link to it? if this will cause issues would be good to consider it.

@benoitgrelard probably has a better explanation as he originally brought this to my attention.

jjenzz: Oh wait, this is `ContextMenu` and they can't `autoFocus` a menu item so it's okay to fallback onto `FocusScope`s auto focus behaviour in this case :+1:

This was the original `autoFocus` issue if you're curious https://github.com/radix-ui/primitives/issues/518. I went through all our components and explicitly focused the trigger on unmount but looks like I also decided against applying it to `ContextMenu` at the time.

PR #802, opened 2021-07-28
[ContextMenu][DropdownMenu] Fix test flake

andy-hook:
It took me a while to figure this one out but the flake we were seeing in our Cypress tests is caused by the variability in measurement speed inside popper when it first renders.

See [here where we apply unmeasured styles](https://github.com/radix-ui/primitives/blob/207014676bce54a97bf61d0e703439765ea62a45/packages/core/popper/src/popper.ts#L355-L363) to position popper elements off screen, the tests don't have a problem with this in 90% of cases as it happens so fast not to cause a problem. The issue starts when these measurements don't happen fast enough and cypress starts it's interaction on the offscreen elements (as they technically exist in dom) this will make the menu visibility logic execute with inaccurate coords and remove the content, failing the test.

Important to note that this doesn't happen with synthetic events, but `realEvents` uses [CDP](https://chromedevtools.github.io/devtools-protocol/) to produce real native events and produces a technically valid failure.

The solution then is to simply assert on visibility before performing the hover interaction. This prompts Cypress to wait for the measurements to complete thus the element styles position content into visible space and all works as it should.

Here're my reduced test results:

image

### DIFF
--- a/cypress/integration/ContextMenu.spec.ts
+++ b/cypress/integration/ContextMenu.spec.ts
@@ -69,18 +69,21 @@ describe('ContextMenu', () => {
         // Disabled item
         pointerOver('Bookmarks →');
         cy.findByText('Inbox').should('be.visible');
+        pointerExitRightToLeft('Bookmarks →');
         pointerOver('Print…');
         cy.findByText('Inbox').should('not.exist');
 
         // Trigger item
         pointerOver('Bookmarks →');
         cy.findByText('Inbox').should('be.visible');
+        pointerExitRightToLeft('Bookmarks →');
         pointerOver('Tools →');
         cy.findByText('Inbox').should('not.exist');
 
         // Disabled trigger item
         pointerOver('Bookmarks →');
         cy.findByText('Inbox').should('be.visible');
+        pointerExitRightToLeft('Inbox');
         pointerOver('History →');
         cy.findByText('Inbox').should('not.exist');
       });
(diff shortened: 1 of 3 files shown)


cypress/integration/ContextMenu.spec.ts
```diff
@@ -69,18 +69,21 @@ describe('ContextMenu', () => {
         // Disabled item
         pointerOver('Bookmarks →');
         cy.findByText('Inbox').should('be.visible');
+        pointerExitRightToLeft('Bookmarks →');
```
benoitgrelard, line 72: Curious why we're adding these new `pointerExitRightToLeft` steps everywhere?

benoitgrelard (commented):
Amazing work on this 👏

jjenzz (approved):
Wow, the write up on this is great :+1:

Same question as Benoit, but all good otherwise!

cypress/integration/ContextMenu.spec.ts
andy-hook: I removed the `focus` call in pointer utils because it was technically not an accurate representation of the interaction and I wanted to make sure it hadn't masked any inaccurate tests.   It ended up failing these two because the exit grace works so well in the story that the sibling items are still within the zone (using default coords) so the menu never closes.

Anyway,   I added the directional exits to place this outside the zone before moving to the item, but your comment made me look at it again and I realised this test isn’t actually that helpful as users cant “hop” across items like this in the real world. I’ve instead consolidated them to cover a more helpful scenario while avoiding the need to handle this https://github.com/radix-ui/primitives/pull/802/commits/5d492779644b6f309f6b78aa728d6bb21e0611f7

benoitgrelard (approved):

Issue #634, opened 2021-05-05
Cannot compose `DropdownMenu.Item` and `Dialog.Trigger` in latest


jjenzz:
In previous versions it was possible to do `<Dialog.Trigger as={DropdownMenu.Item} />` to trigger a dialog from a dropdown menu item but this is no longer working.

### Before

https://codesandbox.io/s/unruffled-cache-g7cnm?file=/src/App.js

### After

https://codesandbox.io/s/dark-wave-3w8kk?file=/src/App.tsx

kaisermann:
Curious fact: if you add a `disabled` prop to the `Dialog.trigger` it works (by click only)

kaisermann:
It seems that handling the `onSelect` on [`mouseUp`](https://github.com/radix-ui/primitives/blob/main/packages/react/menu/src/Menu.tsx#L624) AND closing the dropdown menu is what's causing it to stop working.

andrevenancio:
Whats the status on this bug? What's the official way of having a Dialog being triggered from within a Dropdown menu?

steveruizok:
Yall are going to hate this but here's a hacky solution for a button that both triggers a dialog and closes a dropdown. https://codesandbox.io/s/jovial-cloud-7d4mn

andrevenancio:
> Yall are going to hate this but here's a hacky solution for a button that both triggers a dialog and closes a dropdown. https://codesandbox.io/s/jovial-cloud-7d4mn

It works for simple architecture, but on my specific case I want to abstract the Dialog away from the Dropdown because its re-used in another part of the app. So it makes sense to make a component out of. 

So if that's the case, my Dialog would have a timeout and changes the state that needs to be read by the dropdown.. Would love to see how the team at Modulz is using this. I think a context menu firing a Dialog might be a common pattern

jgoux:
> https://codesandbox.io/s/jovial-cloud-7d4mn

Indeed it works, but in practice don't you want to group each `MenuItem` along with its Dialog like : 
```tsx
<MenuContent>
  <MyDialog>
    <MenuItem as={DialogTrigger}>Open my dialog</MenuItem>
  </MyDialog>
  <MyOtherDialog>
    <MenuItem as={DialogTrigger}>Open my other dialog</MenuItem>
  </MyOtherDialog>
</MenuContent>
```

It would be something like this : https://codesandbox.io/s/naughty-mirzakhani-hkqxn?file=/src/App.js

[...]

PR #812, opened 2021-08-03
[DropdownMenu][ContextMenu] Prevent text selection on menu close in Firefox

### LINKED ISSUE #742: [DropdownMenu] Using mouse to select drop down item traps mouse in active text selection
## Bug report

This bug seems to be happening in **Firefox only**. I haven't noticed this bug in Chrome or Safari.

First affected version: `@radix-ui/react-dropdown-menu` @ **0.0.20**.

### Current Behavior

Bug:
- Click once to open a drop down menu
- Click once more to select an item from the menu
- The mouse now gets trapped in an active text selection on the underlying content

## ✅ Behavior before 0.0.20
https://user-images.githubusercontent.com/83844/123067049-eca91000-d410-11eb-895c-9815b31129bd.mp4

## ❌ Behavior since 0.0.20
https://user-images.githubusercontent.com/83844/123067151-05b1c100-d411-11eb-954e-4eecfd8156f8.mp4

### Expected behavior

When clicking to select an item from the drop down menu, I'd expect no start of a text selection to happen.

### Reproducible example

This is just an arbitrary example that already existed in the gallery, not my example. It just features a drop down menu. It shows that all normal usage of this component will illustrate the bug (be sure to use Firefox, though).

[CodeSandbox Template](https://codesandbox.io/s/jmq8h)

### Suggested solution

?

### Your environment

| Radix Package(s) |    @radix-ui/react-dropdown-menu     |   0.0.20      |
| React            | n/a     |    16.14.0     |
| Browser          |     Firefox    |    89.0.1 (64-bit)     |
| Operating System |    macOS     |   10.15.7  (19H114)      |

jjenzz:
Hopefully the code comments explain clearly what is going on here 🤞 

Fixes #742.

--------------

UPDATE: The original change I made here was no longer needed as the changes I made to [fix `Dialog.Trigger` / `DropdownMenu.Item` composition](https://github.com/radix-ui/primitives/pull/818) also fixed this.

I've updated the code comments to explain that it covers this case too and the explanation as to why it fixes this is as follows:

>  Browsers wait for our handler logic to execute before executing their own implementation in case we have prevented their default behaviour. Therefore, the menu will unmount before the browser's `pointerup` logic has executed causing `pointerdown` to be stuck in some browsers. In FF this resulted in text selection occurring on the page when moving the mouse after close. We move the unmount logic till after the browser receives the pointerup to avoid this.

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -672,7 +672,8 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
       }}
       onPointerUp={composeEventHandlers(props.onPointerUp, (event) => {
         // Pointer down can move to a different menu item which should activate it on pointer up.
-        // We dispatch a click for selection to allow composition with click based triggers.
+        // We dispatch a click for selection to allow composition with click based triggers and to
+        // prevent Firefox from getting stuck in text selection mode when the menu closes.
         if (!isPointerDownRef.current) event.currentTarget?.click();
       })}
       onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
--- a/.yarn/versions/48277b4f.yml
+++ b/.yarn/versions/48277b4f.yml
@@ -1,0 +1,3 @@
+declined:
+  - primitives
+  - "@radix-ui/react-menu"


packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
```diff
@@ -155,6 +155,33 @@ export const Modality = () => {
   );
 };
 
+export const TextSelection = () => (
```
jjenzz, line 158: Left this in so you can test PR but I'll probably remove it once we're happy it's all working. Not sure this story is needed?

andy-hook:
I remember exploring a similar approach for fixing https://github.com/radix-ui/primitives/issues/634 but abandoned because we discussed doing something with event composition instead.

I've just tested that issue and it does indeed fix it as well 👍  this raises the question of whether we're ok introducing this timer, I'm personally fine with it for now given the absence of an alternative approach.

Do we see any unexpected issues resulting from the timer though?

packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
andy-hook: It's a bit of an edge really, appreciated it being here to test but I don't think we need to keep it around imo.

packages/react/menu/src/Menu.tsx
```diff
@@ -652,10 +653,24 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
...
+       * We move the close logic till after the browser receives the pointerup to avoid this.
+       */
+      rootCloseTimerRef.current = window.setTimeout(context.onRootClose, 0);
     }
   };
 
+  React.useEffect(() => {
+    const rootCloseTimer = rootCloseTimerRef.current;
+    return () => window.clearTimeout(rootCloseTimer);
+  });
```
andy-hook, line 672: Just wondering why we need the assignment here? is it wrap style thing? also should we provide an empty deps array?

packages/react/menu/src/Menu.tsx
```diff
@@ -652,10 +653,24 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
...
-      context.onRootClose();
+      /**
+       * Browsers wait for our handler logic to execute before executing their own implementation
+       * in case we have prevented the default behaviour. Therefore, the root will unmount before
+       * the browser's `pointerup` logic has executed causing `pointerdown` to be stuck in some
+       * browsers. In FF this results in text selection occuring on the page when moving the mouse
+       * after close.
+       *
+       * We move the close logic till after the browser receives the pointerup to avoid this.
+       */
```
andy-hook, line 664: Wondering whether we need to adjust this comment given that this fixes other issues https://github.com/radix-ui/primitives/pull/812#issuecomment-891889618

packages/react/menu/src/Menu.tsx
jjenzz: Ah oops, should definitely add the dep array. I split it into a variable because of habit 🙈  will inline that too.

packages/react/menu/src/Menu.tsx
jjenzz: Interesting... It feels a bit odd to comment here that we delay to allow composition with `onClick` because that means we're baking logic into components for external dependencies when components shouldn't really have knowledge of, or handling for, external things.

I can't actually get the `DropdownMenu.Item` + `Dialog.Trigger` to work locally with this change though (I made some tweaks to the code so that the timeout only applies where needed).

jjenzz: Ah, I can get it to work... ignore me.

jjenzz: It seems that the previous way I wrote this still didn't fix the composition issue for keyboard users but now I'm a bit hesitant to push this change given that it masks that other issue a bit.

andy-hook (approved):

nvie:
Amazing, thanks for the fix!

Issue #745, opened 2021-06-24
[DropdownMenu] Item as anchor does not navigate to the expected destination


AlvinLing:
### Current Behavior

When a DropdownMenu.Item is rendered as an anchor, clicking the item closes the menu.

`<DropdownMenu.Item as="a" href="/foo">Click Me</DropdownMenu.Item>`

### Expected behavior

When a DropdownMenu.Item is rendered as an anchor, the user should be navigated to the provided href value.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/loving-star-29eqk?file=/pages/index.js)

### Suggested solution

Adding `onSelect={ (e) => { e.preventDefault() } }` fixes the issue but it would be nice if that was not required.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |    @radix-ui/react-dropdown-menu     |   0.0.22      |
| React            | n/a     |    17.0.2     |
| Browser          |   Chrome      |     91.0.4472.106 (Official Build) (x86_64)    |
| Node             | n/a     |    14.15.4     |
| npm/yarn         |   npm      |   7.6.1      |
| Operating System |   MacOS      |    11.4     |

jjenzz:
This was fixed by #818 and will be in the next release.

PR #818, opened 2021-08-10
[DropdownMenu][Menu] Allow dialog composition

jjenzz:
While doing this, I noticed that the DX for composing multiple dialogs from menu items is not ideal from a DX perspective. However, this fix will at least make single/multiple possible while we work on improving the DX.

Issue (not sure whether to mark fixed given DX issues): #634


packages/react/menu/src/Menu.tsx
```diff
@@ -661,15 +665,21 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
...
         const isTypingAhead = contentContext.searchRef.current !== '';
         if (disabled || (isTypingAhead && event.key === ' ')) return;
         if (SELECTION_KEYS.includes(event.key)) {
-          // prevent page scroll if using the space key to select an item
-          if (event.key === ' ') event.preventDefault();
-          handleSelect();
+          event.currentTarget.click();
+          // prevent default browser behaviour for selection keys, they should trigger a
+          // selection only. for example, stops page from scrolling when hitting space key.
+          event.preventDefault();
```
jjenzz, line 682: This is needed for `Enter` keypress also otherwise the `Dialog` will not open. I am really stumped why this is needed though. There isn't a `keydown` event in the `Dialog.Trigger` so we're not preventing any of our implementation, just browser stuff but I dno what that "stuff" is 🙈

Any ideas?

jjenzz:
@andy-hook I think this is ready now 😬

packages/react/menu/src/Menu.tsx
andy-hook: Not knowing the "stuff" is bugging me too. I realised that the `Dialog` does open, it's just that the close button is triggered as well, causing it to close immediately.

Could this be related to any of the implicit form submission behaviour associated with `Enter`? I can see why it would given we're not in a form context, I'm just struggling to see what else the browser could be doing.

jjenzz: > it's just that the close button is triggered as well, causing it to close immediately.

This should only be happening if you hold the `Enter` key down which is how the browser works unfortunately https://codepen.io/jjenzz/pen/rNmPwqj

Is it happening when you just press enter and don't hold it?

jjenzz: Let's have a quick call because I'm not seeing that here 🤔

jjenzz: Oh gosh, I see what you mean now hahaha... when I *don't* put the `event.preventDefault()` in.

> without preventing default

Missed that part. Interesting... yeah okay, let me try something quick.

packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
```diff
@@ -312,6 +313,215 @@ export const WithLabels = () => (
...
+            Open
+          </DropdownMenuTrigger>
+
+          <DropdownMenuContent
+            className={contentClass}
+            sideOffset={5}
+            onCloseAutoFocus={(event) => {
+              // prevent focusing dropdown trigger when it closes from a dialog trigger
+              if (isDialogOpenRef.current) event.preventDefault();
+            }}
```
jjenzz, line 379: @benoitgrelard We shouldn't really have to do this because the `DropdownMenu` already has logic to prevent its trigger from focusing if there was a focus outside of it. However, [the `setTimeout` in `FocusScope` for `onCloseAutoFocus`](
https://github.com/radix-ui/primitives/blob/main/packages/react/focus-scope/src/FocusScope.tsx#L133) is re-ordering events and breaking the expected behaviour here.

We have discussed removing the `setTimeout` from `FocusScope` given that v17 has been out for a while and we are a new modern library but we're reluctant to make that change as part of this PR (not sure of the knock on effects in other primz).

We're going to leave this in the story for now to demonstrate the workaround but it would be good to discuss this in more detail with you when you return.

packages/react/menu/src/Menu.tsx
```diff
@@ -661,15 +665,25 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
       {...itemProps}
       ref={composedRefs}
       disabled={disabled}
-      // we handle selection on `pointerUp` rather than `click` to match native menus implementation
-      onPointerUp={composeEventHandlers(props.onPointerUp, handleSelect)}
+      onClick={composeEventHandlers(props.onClick, handleSelect)}
+      onPointerDown={() => (isPointerDownRef.current = true)}
```
andy-hook, line 669: Think we forgot to compose the handler

packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
```diff
@@ -312,6 +313,215 @@ export const WithLabels = () => (
   </div>
 );
 
+export const WithAsDialogTrigger = () => {
```
andy-hook, line 316: nit: would something like `withItemAsDialogTrigger` or `withComposedDialogTrigger` read better?

andy-hook (approved):

packages/react/menu/src/Menu.tsx
garand: Looks like the same thing I was running into here. 🤔 

https://discord.com/channels/752614004387610674/872115819490996244/872416709607309353

andy-hook: > Looks like the same thing I was running into here.

Yep! we were actually using one of those examples to help debug. I was quite surprised to see the events working this way after calling `focus` but it explains what we were seeing in that thread.

packages/react/menu/src/Menu.tsx
```diff
@@ -644,15 +644,19 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
...
   const handleSelect = () => {
     const menuItem = ref.current;
     if (!disabled && menuItem) {
       const itemSelectEvent = new Event(ITEM_SELECT, { bubbles: true, cancelable: true });
       menuItem.addEventListener(ITEM_SELECT, (event) => onSelect?.(event), { once: true });
       menuItem.dispatchEvent(itemSelectEvent);
-      if (itemSelectEvent.defaultPrevented) return;
-      context.onRootClose();
+      if (itemSelectEvent.defaultPrevented) {
+        isPointerDownRef.current = false;
```
benoitgrelard, line 656: Just to make sure I understand fully, we're resetting here because the menu won't close, the item won't unmount and so the ref won't be reset automatically on remount?

packages/react/menu/src/Menu.tsx
```diff
@@ -661,15 +665,28 @@ const MenuItem = React.forwardRef((props, forwardedRef) => {
...
-      onPointerUp={composeEventHandlers(props.onPointerUp, handleSelect)}
+      onClick={composeEventHandlers(props.onClick, handleSelect)}
+      onPointerDown={(event) => {
+        props.onPointerDown?.(event);
+        isPointerDownRef.current = true;
+      }}
+      onPointerUp={composeEventHandlers(props.onPointerUp, (event) => {
+        // Pointer down can move to a different menu item which should activate it on pointer up.
+        // We dispatch a click for selection to allow composition with click based triggers.
+        if (!isPointerDownRef.current) event.currentTarget?.click();
```
benoitgrelard, line 676: I find it a bit odd that if users add an `onClick` listener to the item, it will trigger even when it's not a click (down and up), I'm guessing that was the only way we managed to get it to work with `Dialog` trigger?

packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
jjenzz: Because the `modal` scenario traps focus in the dialog so when it tries to focus the dropdown trigger on `onCloseAutoFocus` it can't.

packages/react/menu/src/Menu.tsx
jjenzz: the ref will always be reset automatically on remount, this is specifically for a case where someone prevents it from closing. Without this, this would break:

1. on the same item `pointer down -> pointer up`
2. `event.preventDefault()` in `onSelect`
3. on a different item `pointer down -> pointer move -> pointer up on the item from step 1`

That pointer up on the item from step one would think they pointered down on it too because it wouldn't have been reset when it was kept open.

packages/react/menu/src/Menu.tsx
jjenzz: > I'm guessing that was the only way we managed to get it to work with Dialog trigger?

Yes and I've also realised this conveniently fixes #745 too.

I'm okay with it firing a click tbh because that's what we're logically determining to be a click on the item - the circumstances in which it would activate. I guess we can see what complaints we get here from consumers, if any.

PR #819, opened 2021-08-11
[AlertDialog][ContextMenu][Dialog][Popover][DropdownMenu] Fix pointer-events reset

jjenzz:
Pointer events were not resetting correctly when pop ups were opening from within other pop ups. This was initially reported here regarding `Dropdown`/`Dialog` composition: https://discord.com/channels/752614004387610674/803656530259738674/874553651417477162

This change ensures we decrease the count at the same time as the reset (when the unmount events have executed).

### DIFF
--- a/packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
+++ b/packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
@@ -33,6 +33,7 @@ function useBodyPointerEvents({ disabled }: { disabled: boolean }) {
       }
 
       function resetPointerEvents() {
+        changeCount--;
         if (changeCount === 0) {
           document.body.style.pointerEvents = originalBodyPointerEvents;
         }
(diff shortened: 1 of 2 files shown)


garand:
👀 👀 👀

Seems like this would fix this? https://codesandbox.io/s/friendly-noyce-rin40

andy-hook:
> Seems like this would fix this?

Yep, this change resolves it 👍

packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx
```diff
@@ -33,6 +33,7 @@ function useBodyPointerEvents({ disabled }: { disabled: boolean }) {
       }
 
       function resetPointerEvents() {
+        changeCount--;
```
benoitgrelard, line 36: I think that's my bad on this one, when we rejigged the code before I went on holidays I didn't catch that this bit wasn't moved with it.

jjenzz: Nah, my fault for not moving it 😅  complete oversight.

PR #828, opened 2021-08-18
[Presence] Allow close animation only

### LINKED ISSUE #635: [Presence] Animation fails to run in certain conditions
### Current Behavior

When providing a single exit animation, `Presence` will only animate into the unmounted state once.

### Expected behavior

It should always perform the animation when transitioning to an unmounted state.

### Reproducible example

https://codesandbox.io/s/inspiring-dirac-zheqh

Hide content -> animation works -> show content -> hide content -> no animation

jjenzz:
I'm still struggling with some `Toast` animation bits which I'll continue to try solve separately but this at least fixes #635 for the timebeing.

### DIFF
--- a/packages/react/presence/src/Presence.tsx
+++ b/packages/react/presence/src/Presence.tsx
@@ -31,7 +31,7 @@ function usePresence(present: boolean) {
   const [node, setNode] = React.useState<HTMLElement>();
   const stylesRef = React.useRef<CSSStyleDeclaration>({} as any);
   const prevPresentRef = React.useRef(present);
-  const prevAnimationNameRef = React.useRef<string>();
+  const prevAnimationNameRef = React.useRef<string>('none');
   const initialState = present ? 'mounted' : 'unmounted';
   const [state, send] = useStateMachine(initialState, {
     mounted: {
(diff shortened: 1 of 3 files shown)


packages/react/presence/src/Presence.tsx
```diff
@@ -47,6 +47,11 @@ function usePresence(present: boolean) {
     },
   });
 
+  React.useEffect(() => {
+    const currentAnimationName = getAnimationName(stylesRef.current);
+    prevAnimationNameRef.current = state === 'mounted' ? currentAnimationName : 'none';
```
jjenzz, line 52: The cause of the original issue was that when animating out the first time the `prevAnimationNameRef.current` was becoming `hGfvx` or something (the name of the close animation) and then when remounting there was no animation so `prevAnimationNameRef.current` was still set to the close animation.

That meant that the second time it tried to close, the previous and current animation values would be the same and it wouldn't suspend unmount as it though there wasn't an animation initiating.

This change ensures that when the component unmounts we always set the previous animation to `none`.

packages/react/presence/src/Presence.tsx
```diff
@@ -96,33 +101,20 @@ function usePresence(present: boolean) {
...
       };
-
-      const handleAnimationStart = (event: AnimationEvent) => {
-        if (event.target === node) {
-          // if animation occurred, store its name as the previous animation.
-          prevAnimationNameRef.current = getAnimationName(stylesRef.current);
-        }
-      };
-
-      node.addEventListener('animationstart', handleAnimationStart);
```
jjenzz, line 107: I remembered that we added all this stuff before for a reason so I went through GitHub history and validated the issues that this originally solved are still working here.

benoitgrelard (approved):

andy-hook (approved):

PR #832, opened 2021-08-25
[All] Replace polymorphism with `asChild` prop

jjenzz:


package.json
```diff
@@ -131,12 +131,20 @@
...
     },
     "rules": {
-      "@typescript-eslint/no-unused-vars": [
+      "import/no-anonymous-default-export": [
         1,
         {
-          "vars": "local"
+          "allowObject": true
         }
       ],
```
jjenzz, line 139: When i updated eslint, we started getting errors in our stories where it does

```jsx
export default { Title: 'Components/AccessibleIcon' };
```

This rule change allows default object exports.

package.json
```diff
@@ -131,12 +131,20 @@
...
+      "react/jsx-pascal-case": [
+        2,
+        {
+          "allowNamespace": true
+        }
+      ],
```
jjenzz, line 145: I had to upgrade eslint and apply this rule to allow `Primitive.button` (so lowercase `button`).

package.json
```diff
@@ -131,12 +131,20 @@
...
+      "no-unused-vars": "off",
+      "@typescript-eslint/no-unused-vars": "off",
```
jjenzz, line 147: In the eslint upgrade, the configuration for this has been moved to `tsconfig.json`.

packages/react/toolbar/src/Toolbar.tsx
```diff
@@ -146,23 +119,23 @@ ToolbarLink.displayName = LINK_NAME;
...
-}) as ToolbarToggleGroupPrimitive;
+type ToolbarToggleGroupElement = React.ElementRef<typeof ToggleGroupPrimitive.Root>;
+type ToolbarToggleGroupProps = React.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Root>;
+
+const ToolbarToggleGroup = React.forwardRef<ToolbarToggleGroupElement, ToolbarToggleGroupProps>(
+  (props, forwardedRef) => {
+    const context = useToolbarContext(TOGGLE_GROUP_NAME);
+    return (
+      // @ts-ignore
+      <ToggleGroupPrimitive.Root
```
jjenzz, line 130: In the next PR I'll be exporting all of our component types so that we no longer need to do `React.ComponentPropsWithoutRef<typeof ...>` because that utility flattens union which breaks [discriminated unions](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions).

In the meantime, I've ignored this so we can start reviewing this work separately from that piece.

PR #838, opened 2021-09-02
[All] Avoid inlining built type declarations

jjenzz:
------------------

**Note**: PR will be easier to review with hidden whitespace changes.

------------------

- Reduces our type def file sizes by favouring interfaces for our prop types. Further reading: https://github.com/microsoft/TypeScript/issues/37151#issuecomment-681927567
- Uses `React.ElementRef<>` for external components only
- Uses `Radix.ComponentPropsWithRef<>` for external components only

_Those last two bullets weren't required to fix the inlining but I find it made the types easier to digest. Happy to change this though if others don't agree._

An additional benefit of using interfaces is that we no longer need the `MergeProps` utility because interfaces will moan when there are conflicts in an extended interface. Therefore, we can `Omit` specific conflicts manually inline only when needed. This [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIggSXNeTkDeAUMqcjAPYUBcyAzmFKAOYDcRAvu0QPQ-IAVABbA6yaFApQxdCsgDuKANYgK8hSgwQAJsjByABqgC2wMAB50WHLEQQANMgDklCk4B8BoqEi38A+QpsXzwUCAAPSBBtMStg3DtCEjJXWhAAV2MAI2h2DiIiMABPAAcUKwFSlABeJLJyKloGJhA2Tm4+QRExbQoIOhAnMHEoSSh6OUVkYGMSgBtgBDM5ovFo5HSShTMhZAMMADdoAz0qhREEXdFz1amtbR09IRQDAFloZggABUkSuhP0mBgHMAHSFM4BCiVMrIWoVM4AMjqKUayAy2VynAKCAoIAYIzGABF0i1mABhXF0TIlIG42iQ6E1ZGkVLIACM7SAA
) shows the advantage over using `type`s.

### DIFF
--- a/.yarn/versions/4c05b7ba.yml
+++ b/.yarn/versions/4c05b7ba.yml
@@ -1,0 +1,41 @@
+releases:
+  "@radix-ui/react-accessible-icon": patch
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-announce": patch
+  "@radix-ui/react-arrow": patch
+  "@radix-ui/react-aspect-ratio": patch
+  "@radix-ui/react-avatar": patch
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-collapsible": patch
+  "@radix-ui/react-collection": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
+  "@radix-ui/react-dismissable-layer": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-focus-scope": patch
+  "@radix-ui/react-hover-card": patch
+  "@radix-ui/react-label": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-popper": patch
+  "@radix-ui/react-portal": patch
+  "@radix-ui/react-presence": patch
+  "@radix-ui/react-primitive": patch
+  "@radix-ui/react-progress": patch
+  "@radix-ui/react-radio-group": patch
+  "@radix-ui/react-roving-focus": patch
+  "@radix-ui/react-scroll-area": patch
+  "@radix-ui/react-separator": patch
+  "@radix-ui/react-slider": patch
+  "@radix-ui/react-slot": patch
+  "@radix-ui/react-switch": patch
+  "@radix-ui/react-tabs": patch
+  "@radix-ui/react-toggle": patch
+  "@radix-ui/react-toggle-group": patch
+  "@radix-ui/react-toolbar": patch
+  "@radix-ui/react-tooltip": patch
+  "@radix-ui/react-visually-hidden": patch
+
+declined:
+  - primitives
--- a/packages/react/checkbox/src/Checkbox.tsx
+++ b/packages/react/checkbox/src/Checkbox.tsx
@@ -27,15 +27,13 @@ type CheckboxContextValue = {
 const [CheckboxProvider, useCheckboxContext] = createContext<CheckboxContextValue>(CHECKBOX_NAME);
 
 type CheckboxElement = React.ElementRef<typeof Primitive.button>;
-type CheckboxProps = Radix.MergeProps<
-  Radix.ComponentPropsWithoutRef<typeof Primitive.button>,
-  {
-    checked?: CheckedState;
-    defaultChecked?: CheckedState;
-    required?: boolean;
-    onCheckedChange?(checked: CheckedState): void;
-  }
->;
+type PrimitiveButtonProps = Radix.ComponentPropsWithoutRef<typeof Primitive.button>;
+interface CheckboxProps extends Omit<PrimitiveButtonProps, 'checked' | 'defaultChecked'> {
+  checked?: CheckedState;
+  defaultChecked?: CheckedState;
+  required?: boolean;
+  onCheckedChange?(checked: CheckedState): void;
+}
 
 const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>((props, forwardedRef) => {
   const {
@@ -115,16 +113,14 @@ Checkbox.displayName = CHECKBOX_NAME;
 const INDICATOR_NAME = 'CheckboxIndicator';
 
 type CheckboxIndicatorElement = React.ElementRef<typeof Primitive.span>;
-type CheckboxIndicatorProps = Radix.MergeProps<
-  Radix.ComponentPropsWithoutRef<typeof Primitive.span>,
-  {
-    /**
-     * Used to force mounting when more control is needed. Useful when
-     * controlling animation with React animation libraries.
-     */
-    forceMount?: true;
-  }
->;
+type PrimitiveSpanProps = Radix.ComponentPropsWithoutRef<typeof Primitive.span>;
+interface CheckboxIndicatorProps extends PrimitiveSpanProps {
+  /**
+   * Used to force mounting when more control is needed. Useful when
+   * controlling animation with React animation libraries.
+   */
+  forceMount?: true;
+}
 
 const CheckboxIndicator = React.forwardRef<CheckboxIndicatorElement, CheckboxIndicatorProps>(
   (props, forwardedRef) => {
@@ -148,11 +144,12 @@ CheckboxIndicator.displayName = INDICATOR_NAME;
 
 /* ---------------------------------------------------------------------------------------------- */
 
-type BubbleInputProps = Omit<Radix.ComponentPropsWithoutRef<'input'>, 'checked'> & {
+type InputProps = Radix.ComponentPropsWithoutRef<'input'>;
+interface BubbleInputProps extends Omit<InputProps, 'checked'> {
   checked: CheckedState;
   control: HTMLElement | null;
   bubbles: boolean;
-};
+}
 
 const BubbleInput = (props: BubbleInputProps) => {
   const { control, checked, bubbles = true, ...inputProps } = props;
@@ -212,3 +209,4 @@ export {
   Root,
   Indicator,
 };
+export type { CheckboxProps, CheckboxIndicatorProps };
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -26,11 +26,11 @@ type ContextMenuContextValue = {
 const [ContextMenuProvider, useContextMenuContext] =
   createContext<ContextMenuContextValue>(CONTEXT_MENU_NAME);
 
-type ContextMenuProps = {
+interface ContextMenuProps {
   onOpenChange?(open: boolean): void;
   dir?: Direction;
   modal?: boolean;
-};
+}
 
 const ContextMenu: React.FC<ContextMenuProps> = (props) => {
   const { children, onOpenChange, dir, modal = true } = props;
@@ -80,7 +80,8 @@ ContextMenu.displayName = CONTEXT_MENU_NAME;
 const TRIGGER_NAME = 'ContextMenuTrigger';
 
 type ContextMenuTriggerElement = React.ElementRef<typeof Primitive.span>;
-type ContextMenuTriggerProps = Radix.ComponentPropsWithoutRef<typeof Primitive.span>;
+type PrimitiveSpanProps = Radix.ComponentPropsWithoutRef<typeof Primitive.span>;
+interface ContextMenuTriggerProps extends PrimitiveSpanProps {}
 
 const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMenuTriggerProps>(
   (props, forwardedRef) => {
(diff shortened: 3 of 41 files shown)


benoitgrelard:
> An additional benefit of using interfaces is that we no longer need the MergeProps utility because interfaces will moan when there are conflicts in an extended interface. Therefore, we can Omit specific conflicts manually inline only when needed. This TS playground shows the advantage over using types.

Thanks for providing that playground that's really great!

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -27,15 +27,13 @@ type CheckboxContextValue = {
...
-  Radix.ComponentPropsWithoutRef<typeof Primitive.button>,
-  {
-    checked?: CheckedState;
-    defaultChecked?: CheckedState;
-    required?: boolean;
-    onCheckedChange?(checked: CheckedState): void;
-  }
->;
+type PrimitiveButtonProps = Radix.ComponentPropsWithoutRef<typeof Primitive.button>;
+interface CheckboxProps extends Omit<PrimitiveButtonProps, 'checked' | 'defaultChecked'> {
```
benoitgrelard, line 31: I like this to be honest, it's clearer than the `MergeProps` util as well

packages/react/popover/src/Popover.tsx
```diff
@@ -295,32 +286,31 @@ const PopoverContentNonModal = React.forwardRef<PopoverContentTypeElement, Popov
 
 /* -----------------------------------------------------------------------------------------------*/
 
+type PopoverContentImplElement = React.ElementRef<typeof PopperPrimitive.Content>;
```
benoitgrelard, line 289: Any reason why this one is higher up and not close to the props one?
Not sure if I missed others like this too earlier…

packages/react/primitive/src/Primitive.tsx
```diff
@@ -3,32 +3,32 @@ import { Slot } from '@radix-ui/react-slot';
...
-
-type PrimitiveForwardRefComponent<E extends React.ElementType> = React.ForwardRefExoticComponent<
-  MergeProps<React.ComponentPropsWithRef<E>, PrimitiveProps>
->;
-
 // Temporary while we await merge of this fix:
 // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55396
-type PropsWithoutRef<P> = P extends { ref?: any } ? Pick<P, Exclude<keyof P, 'ref'>> : P;
+// prettier-ignore
+type PropsWithoutRef<P> = P extends any ? ('ref' extends keyof P ? Pick<P, Exclude<keyof P, 'ref'>> : P) : P;
```
benoitgrelard, line 9: This has changed a little, would you mind explaining the changes roughly?

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -86,10 +84,11 @@ RadioGroup.displayName = RADIO_GROUP_NAME;
...
 type RadioGroupItemElement = React.ElementRef<typeof Radio>;
-type RadioGroupItemProps = Radix.MergeProps<
-  Omit<Radix.ComponentPropsWithoutRef<typeof Radio>, 'onCheck'>,
-  { value: string; name?: never }
->;
+type RadioProps = Radix.ComponentPropsWithoutRef<typeof Radio>;
+interface RadioGroupItemProps extends Omit<RadioProps, 'onCheck'> {
+  value: string;
+  name?: never;
+}
```
benoitgrelard, line 91: do we still need to do this now, or could we get away with doing:

```suggestion
interface RadioGroupItemProps extends Omit<RadioProps, 'onCheck' | 'name'> {
  value: string;
```

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -72,15 +72,15 @@ RovingFocusGroup.displayName = GROUP_NAME;
...
-  RovingFocusGroupOptions & {
-    currentTabStopId?: string | null;
-    defaultCurrentTabStopId?: string;
-    onCurrentTabStopIdChange?: (tabStopId: string | null) => void;
-    onEntryFocus?: (event: Event) => void;
-  }
->;
+type PrimitiveDivProps = Radix.ComponentPropsWithoutRef<typeof Primitive.div>;
+interface RovingFocusGroupImplProps
+  extends Omit<PrimitiveDivProps, 'dir'>,
```
benoitgrelard, line 77: That's interesting, because that `dir` one would kinda go unnoticed previously.

packages/react/popover/src/Popover.tsx
jjenzz: I did it so all the `Props` types were together and the `Element` types were first. Otherwise the element type kinda gets lost:

```jsx
type FocusScopeProps = Radix.ComponentPropsWithoutRef<typeof FocusScope>;
type DismissableLayerProps = Radix.ComponentPropsWithoutRef<typeof DismissableLayer>;
type PopperContentProps = Radix.ComponentPropsWithoutRef<typeof PopperPrimitive.Content>;
type PopoverContentImplElement = React.ElementRef<typeof PopperPrimitive.Content>;
interface PopoverContentImplProps
  extends PopperContentProps,
    Omit<DismissableLayerProps, 'onDismiss'> {
	// ...
}
```

What would we prefer?

.yarn/versions/4c05b7ba.yml
```diff
@@ -0,0 +1,41 @@
+releases:
+  "@radix-ui/react-accessible-icon": patch
```
benoitgrelard, line 2: I wonder about all these if we should decline them all here, because we don't want 0.1.1 or whether this is ok because there's another commit asking for a minor, therefore it takes the highest one?

packages/react/primitive/src/Primitive.tsx
jjenzz: Ah yeah, so... the original typedef that `DefinitelyTyped` had ensured that there was no unnecessary `Pick`ing over things by checking if `'ref' extends keyof P` but my type would still `Pick` if `P` was an empty object, so you'd end up with `Pick<{}, Exclude<keyof {}, 'ref'>>`.

I've changed it now so that `P extends any` activates the distributive conditional over the union and then applies the same original logic that the `DefinitelyTyped` types had. Hopefully this is more likely to be approved sooner by them too as it is less of a change to their original defs.

packages/react/toolbar/src/Toolbar.tsx
```diff
@@ -119,7 +116,8 @@ ToolbarLink.displayName = LINK_NAME;
 const TOGGLE_GROUP_NAME = 'ToolbarToggleGroup';
 
 type ToolbarToggleGroupElement = React.ElementRef<typeof ToggleGroupPrimitive.Root>;
-type ToolbarToggleGroupProps = Radix.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Root>;
+type ToggleGroupProps = Radix.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Root>;
+type ToolbarToggleGroupProps = ToggleGroupProps;
```
andy-hook, line 120: Should we be using an interface here?

packages/react/radio-group/src/RadioGroup.tsx
jjenzz: Good catch! 👀!!

packages/react/roving-focus/src/RovingFocusGroup.tsx
jjenzz: Yep, I've found that the interfaces really make you think more!

.yarn/versions/4c05b7ba.yml
jjenzz: AFAIK it is okay because the release will pick the highest version and roll with that.

packages/react/popover/src/Popover.tsx
benoitgrelard: I wouldn't mind this even:

```tsx
type FocusScopeProps = Radix.ComponentPropsWithoutRef<typeof FocusScope>;
type DismissableLayerProps = Radix.ComponentPropsWithoutRef<typeof DismissableLayer>;

type PopperContentProps = Radix.ComponentPropsWithoutRef<typeof PopperPrimitive.Content>;
type PopoverContentImplElement = React.ElementRef<typeof PopperPrimitive.Content>;
interface PopoverContentImplProps
  extends PopperContentProps,
    Omit<DismissableLayerProps, 'onDismiss'> {
	// ...
}
```

jjenzz: I've done this everywhere else:

```jsx
type PopoverContentImplElement = React.ElementRef<typeof PopperPrimitive.Content>;
type PopperContentProps = Radix.ComponentPropsWithoutRef<typeof PopperPrimitive.Content>;
interface PopoverContentImplProps extends PopperContentProps {
	// ...
}
```

As in, the props that it extends are after the element type... and since `FocusScopeProps` and `DismissableLayerProps` are also props it extends, I put those in the middle too.

I'm happy to move all props that it extends above the element type but would need to change that in every file too for consistency. e.g:

```jsx
type FocusScopeProps = Radix.ComponentPropsWithoutRef<typeof FocusScope>;
type DismissableLayerProps = Radix.ComponentPropsWithoutRef<typeof DismissableLayer>;
type PopperContentProps = Radix.ComponentPropsWithoutRef<typeof PopperPrimitive.Content>;

type PopoverContentImplElement = React.ElementRef<typeof PopperPrimitive.Content>;
interface PopoverContentImplProps
  extends PopperContentProps,
    Omit<DismissableLayerProps, 'onDismiss'> {
	// ...
}
```

packages/react/dialog/src/Dialog.tsx
```diff
@@ -367,7 +357,8 @@ const DialogContentImpl = React.forwardRef<DialogContentImplElement, DialogConte
 const TITLE_NAME = 'DialogTitle';
 
 type DialogTitleElement = React.ElementRef<typeof Primitive.h2>;
-type DialogTitleProps = Radix.ComponentPropsWithoutRef<typeof Primitive.h2>;
+type PrimitiveHeading2Props = Radix.ComponentPropsWithoutRef<typeof Primitive.h2>;
```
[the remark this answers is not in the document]
jjenzz: Ah, I'll change the other one to `PrimitiveHeading3Props` in case we have a component that extends more than one heading type.

packages/react/toolbar/src/Toolbar.tsx
jjenzz: I meant to leave a note on this one. We cannot make this an `interface` because the `ToggleGroupProps` is a union of two interfaces and an interface cannot extend a union.

It will moan at you if you try to so we will know it is extending a union and is safe to be a `type`:

image

I realise it is inconsistent though so is confusing. I'm not sure what else to do here though. Ideally we would figure out generic prop types and then we wouldn't need any discriminated unions for our prop types.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -147,10 +148,8 @@ const CONTENT_NAME = 'ContextMenuContent';
 const ContentContext = React.createContext(false);
 
 type ContextMenuContentElement = React.ElementRef<typeof MenuPrimitive.Content>;
-type ContextMenuContentProps = Omit<
-  Radix.ComponentPropsWithoutRef<typeof MenuPrimitive.Content>,
-  'portalled' | 'side' | 'align'
->;
+type MenuProps = Radix.ComponentPropsWithoutRef<typeof MenuPrimitive.Content>;
+interface ContextMenuContentProps extends Omit<MenuProps, 'portalled' | 'side' | 'align'> {}
```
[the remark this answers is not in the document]
jjenzz: I've removed this type altogether because `MenuContentProps` is already declared in this file so I've just extended that.

packages/react/popover/src/Popover.tsx
benoitgrelard: I'm not fussed, let's leave it as is then as it's consistent already.

jjenzz: One thing that I found somewhat annoying in some of the other files is that multiple components might re-use the same type like:

```jsx
type MenuLabelElement = React.ElementRef<typeof Primitive.div>;
type PrimitiveDivProps = Radix.ComponentPropsWithoutRef<typeof Primitive.div>;
type MenuLabelProps extends PrimitiveDivProps {};

// ... later i have to extend `PrimitiveDivProps` but it's declared randomly elsewhere in the file

type MenuSeparatorElement = React.ElementRef<typeof Primitive.div>;
type MenuSeparatorProps extends PrimitiveDivProps {};
```

This made me wonder if we should just declare all external types upfront at the top of the file like:

```jsx
type PrimitiveDivProps = Radix.ComponentPropsWithoutRef<typeof Primitive.div>;

/* -------------------------------------------------------------------------------------------
 * MenuLabel
 * ------------------------------------------------------------------------------------------- */
type MenuLabelElement = React.ElementRef<typeof Primitive.div>;
type MenuLabelProps extends PrimitiveDivProps {};

/* -------------------------------------------------------------------------------------------
 * MenuSeparator
 * ------------------------------------------------------------------------------------------- */
type MenuSeparatorElement = React.ElementRef<typeof Primitive.div>;
type MenuSeparatorProps extends PrimitiveDivProps {};
```

That way, this example you're referring to would be like:

```jsx
type FocusScopeProps = Radix.ComponentPropsWithoutRef<typeof FocusScope>;
type DismissableLayerProps = Radix.ComponentPropsWithoutRef<typeof DismissableLayer>;
type PopperContentProps = Radix.ComponentPropsWithoutRef<typeof PopperPrimitive.Content>;

// lots of other components...

/* -------------------------------------------------------------------------------------------
 * PopoverContentImpl
 * ------------------------------------------------------------------------------------------- */
type PopoverContentImplElement = React.ElementRef<typeof PopperPrimitive.Content>;
interface PopoverContentImplProps extends PopperContentProps, Omit<DismissableLayerProps, 'onDismiss'> {
	// ...
}
```

Thoughts?

packages/react/toolbar/src/Toolbar.tsx
andy-hook: Ah I see what you mean. Not much we can do at the moment then 👍

andy-hook (approved):

benoitgrelard (approved):

PR #842, opened 2021-09-06
Expose Accordion/ToggleGroup sub types

benoitgrelard:


benoitgrelard (commented):
What about `ToolbarToggleGroup`?

jjenzz:
Hm, not sure what we do for that one 🤔

jjenzz:
In fact, we could just do the `Extract` thing and then remove my comment! That removes the inconsistent `type` alias issue here too:

image

benoitgrelard:
> Hm, not sure what we do for that one 🤔

Yeah that was my thoughts exactly 😀

benoitgrelard:
> In fact, we could just do the `Extract` thing and then remove my comment! That removes the inconsistent `type` alias issue here too:
> 
> image

That's not a bad shout, then we can also expose both props the same way the other 2?

jjenzz:
> That's not a bad shout, then we can also expose both props the same way the other 2?

As in, expose **all** the props including the `type`?

That actually makes more sense to me cos our other prop types are **all** the props. Consumers can `Omit` to remove the `type` if they want.

benoitgrelard:
I meant once we've done this extract thing for internal purpose, then we can export `ToolbarToggleGroupSingleProps` and `ToolbarToggleGroupMultipleProps` like I did expose for `Accordion` and `ToggleGroup`.

Is that what you meant?

jjenzz:
> Is that what you meant?

Yeah I think we should export those types but they are actually `ToolbarToggleGroupWithSingleProps` and `ToolbarToggleGroupWithMultipleProps` (I named them wrong in my screenshot) so we should export the `With` props from the other components too instead of the ones with the `type` prop. Consumers would just need to do this:


```jsx
import type { AccordionWithSingleProps, AccordionWithMultiplProps } from '@radix-ui/react-accordion';

const Accordion = (props: Omit<AccordionWithSingleProps, 'type'>) => (
  <AccordionPrimitive type="single" collapsible {...props} />
) 
```

benoitgrelard:
Why can't we do the omit for them and expose that though?

jjenzz:
> Why can't we do the omit for them and expose that though?

We can but that's not consistent with any of the other prop types that we expose. The props we export are always the props for the component they consume but here we would be removing some of the components props.

benoitgrelard:
But it's consistent with the other 2, that's what I'm after. It's never gonna be consistent with the other components because we're exposing sub-types really so they aren't comparable.

jjenzz:
If we're not fussed about the consistency of our prop types matching our exposed API the sure.

packages/react/toggle-group/src/ToggleGroup.tsx
```diff
@@ -276,4 +276,10 @@ export {
   Root,
   Item,
 };
-export type { ToggleGroupWithSingleProps, ToggleGroupWithMultipleProps, ToggleGroupItemProps };
+export type {
+  ToggleGroupSingleProps,
+  ToggleGroupMultipleProps,
+  ToggleGroupWithSingleProps,
+  ToggleGroupWithMultipleProps,
```
jjenzz, line 283: Should we get rid of these two then if we're always going to expose the props without the `type`?

benoitgrelard: I removed them everywhere but then the build failed in `ToolbarToggleGroup` because it was looking for those.

PR #843, opened 2021-09-06
Add GH action to publish RC on merge

jjenzz:
~I'm not sure how to test this 😅~

~I've added links in the comments to examples I've followed so hopefully it's all good but what do we think we should do here?~

~I found [something called `act` for testing GitHub actions](https://github.com/nektos/act)  and installed Docker like it suggests but then it dawned on me that it would probably still do a publish...? so dno. I also couldn't get it to work off the bat so would require a dedicated effort.~

------------------

All tested and working smoothly when I tested with tokens that I had created. Tokens have since been switched to Stephen's which should be fine too as I've walked through correct configuration with him.

### DIFF
--- a/.github/workflows/publish-rc.yml
+++ b/.github/workflows/publish-rc.yml
@@ -1,0 +1,68 @@
+name: 'Publish Release Candidate'
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish-rc:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+      - uses: actions/checkout@v2
+        run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
+
+      - name: Install dependencies and publish
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '${{ steps.nvm.outputs.NVMRC }}'
+        run: yarn && yarn publish:next
+
+      # https://github.community/t/how-to-push-to-protected-branches-in-a-github-action/16101/30
+      - name: Branch protection OFF
+        uses: octokit/request-action@v2.x
+        with:
+          route: PUT /repos/:repository/branches/main/protection
+          repository: ${{ github.repository }}
+          required_status_checks: |
+            null
+          enforce_admins: |
+            null
+          required_pull_request_reviews: |
+            null
+          restrictions: |
+            null
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Commit release
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: 'Publish release candidate'
+
+      # https://github.community/t/how-to-push-to-protected-branches-in-a-github-action/16101/30
+      - name: Branch protection ON
+        uses: octokit/request-action@v2.x
+        with:
+          route: PUT /repos/:repository/branches/main/protection
+          repository: ${{ github.repository }}
+          # https://docs.github.com/en/rest/reference/repos#update-branch-protection-preview-notices
+          mediaType: |
+            previews:
+              - luke-cage
+          required_status_checks: |
+            strict: false
+            contexts:
+              - UI Tests
+              - check
+              - test
+          enforce_admins: |
+            true
+          required_pull_request_reviews: |
+            require_code_owner_reviews: true
+            required_approving_review_count: 1
+          restrictions: |
+            null
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish-rc.yml
+++ b/.github/workflows/publish-rc.yml
@@ -6,6 +6,7 @@ on:
       - main
     paths-ignore:
       - '**/package.json'
+      - '.yarn/versions/**'
 
 jobs:
   publish-rc:
--- a/.github/workflows/publish-rc.yml
+++ b/.github/workflows/publish-rc.yml
@@ -14,14 +14,17 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
+
+      - name: Set node version
+        id: node_version
+        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-rc.yml
```diff
@@ -0,0 +1,68 @@
...
+          # https://docs.github.com/en/rest/reference/repos#update-branch-protection-preview-notices
+          mediaType: |
+            previews:
+              - luke-cage
+          required_status_checks: |
+            strict: false
+            contexts:
+              - UI Tests
+              - check
+              - test
```
jjenzz, line 59: image

.github/workflows/publish-rc.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: octokit/request-action@v2.x
+        with:
+          route: PUT /repos/:repository/branches/main/protection
+          repository: ${{ github.repository }}
+          # https://docs.github.com/en/rest/reference/repos#update-branch-protection-preview-notices
+          mediaType: |
+            previews:
+              - luke-cage
+          required_status_checks: |
+            strict: false
```
jjenzz, line 55: `strict` determines this option
image

.github/workflows/publish-rc.yml
```diff
@@ -0,0 +1,71 @@
+name: 'Publish Release Candidate'
+
+on:
+  push:
+    branches:
+      - main
+    # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths
+    paths-ignore:
+      - '**/package.json'
```
jjenzz, line 9: > When all the path names match patterns in `paths-ignore`, the workflow will not run

Trying to prevent the RC commit to `main` from re-triggering this action.

jjenzz:
Blood sweat and tears have gone into testing this 😅 but it is now ready and working! So pleased with this 🎉

.github/workflows/publish-rc.yml
```diff
@@ -0,0 +1,46 @@
...
+          commit_message: 'Publish release candidate'
+
+      # https://docs.github.com/en/rest/reference/repos#update-branch-protection-preview-notices
+      - name: Branch protection ON
+        run: |
+          curl --request PUT \
+          --url 'https://api.github.com/repos/radix-ui/primitives/branches/main/protection' \
+          --header 'Authorization: token ${{ secrets.PAT_CHANGE_BRANCH_PROTECTION_RULES_FROM_STEPHEN }}' \
+          --header 'Accept: application/vnd.github.luke-cage-preview+json' \
+          --data '{"required_status_checks":{"strict":false,"contexts":["UI Tests","check","test"]},"enforce_admins":true,"required_pull_request_reviews":{"require_code_owner_reviews":true,"required_approving_review_count":1},"restrictions":null}'
```
jjenzz, line 46: If we ever change branch protection rules, we need to make sure we update this so that the action resets the protection back to the new configuration.

.github/workflows/publish-rc.yml
```diff
@@ -0,0 +1,46 @@
...
+      - run: yarn && yarn publish:next
+        env:
+          YARN_NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTOMATION_FROM_STEPHEN }}
+
+      - name: Branch protection OFF
+        run: |
+          curl --request PUT \
+          --url 'https://api.github.com/repos/${{ github.repository }}/branches/main/protection' \
+          --header 'Authorization: token ${{ secrets.PAT_CHANGE_BRANCH_PROTECTION_RULES_FROM_STEPHEN }}' \
+          --data '{"required_status_checks":null,"enforce_admins":false,"required_pull_request_reviews":null,"restrictions":null}'
```
andy-hook, line 32: Curious about how this endpoint works. If I was using the GUI manually I can disable `enforce_admins` to ignore the rules as an admin, I guess that wouldn't work here which is why we need to explicitly disable all of them?

Just wondering if there's a way to auth the push so that we can change fewer rules, or is it a requirement to pass the entire object?

.github/workflows/publish-rc.yml
andy-hook: How is this working exactly? does it prevent a loop from occurring after every push?

andy-hook (commented):
Awesome stuff! 🚀  🚀  just a couple of clarifying comments from me

benoitgrelard (approved):
So excited about this.
Mostly the same questions as @andy-hook.

I am also wondering what impact this has on to when we do a manual stable release, would that trigger anything when we commit the deleted version files, updated package.json or is it only on merge?

.github/workflows/publish-rc.yml
jjenzz: Yeah, because every push is going to add the updated `package.json`s that the publish creates

.github/workflows/publish-rc.yml
jjenzz: Hmmm oh yeah, you're right. I can just `enforce_admins: false` and `enforce_admins: true` instead

jjenzz:
> I am also wondering what impact this has on to when we do a manual stable release, would that trigger anything when we commit the deleted version files, updated package.json or is it only on merge?

Hmm, interesting. I guess I need to add `.yarn/versions/**` to the `paths-ignore`.

.github/workflows/publish-rc.yml
jjenzz: That's a shame... looks like the endpoint makes the other fields required: 
image

benoitgrelard: boooo 👎

andy-hook: Aw that's a shame. If it turns into a problem in the future perhaps there's a way to do `get` and restore them from that or something.

andy-hook (approved):

.github/workflows/publish-rc.yml
jjenzz: Ooo, let me try that. Might not be too tricky 🤔

andy-hook: Yeah.. if it looks remotely tricky then probably not worth spending the brain cycles on. Just a thought though.

jjenzz: It's tricky... haha. We've decided to leave as is because we rarely change these rules, if ever.

PR #844, opened 2021-09-06
[Toolbar] Replace inconsistent type alias prop type

jjenzz:
Realised we could avoid this when I was reviewing [Benoit's recent PR](https://github.com/radix-ui/primitives/pull/842#issuecomment-913781362).

It also bugged me that we had the `With` naming so I rejigged some of the naming conventions for the implementation components.

### DIFF
--- a/packages/react/accordion/src/Accordion.stories.tsx
+++ b/packages/react/accordion/src/Accordion.stories.tsx
@@ -7,7 +7,6 @@ import {
   AccordionHeader,
   AccordionTrigger,
   AccordionContent,
-  AccordionSingleProps,
 } from './Accordion';
 
 export default { title: 'Components/Accordion' };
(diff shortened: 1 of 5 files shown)


packages/react/accordion/src/Accordion.stories.tsx
```diff
@@ -677,7 +676,3 @@ const contentAttrClass = css({
   display: 'block',
   ...styles,
 });
-
-export function Accordion2(props: AccordionSingleProps) {
-  return <Accordion {...props} type="single" collapsible />;
-}
```
jjenzz, line 683: @benoitgrelard Looks like you may have added this accidentally when you were debugging the issues Pedro was having?

benoitgrelard: ah yeah my bad!

benoitgrelard (approved):

andy-hook (approved):

Issue #849, opened 2021-09-09
Dropdown menu with animation won't close when clicking outside.


norman-ags:
### Current Behavior

Dropdown menu with animation won't close when clicking outside. 

### Expected behavior

Dropdown menu should close when clicking outside. 

### Reproducible example

https://codesandbox.io/s/laughing-violet-oq7d1?file=/src/App.js

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |    DropdownMenu     |  0.1.0       |
| React            | n/a     |    17.0.2     |
| Browser          |     Microsoft Edge   |    93.0.961.38     |

andy-hook:
Hey @norman-ags, I can't see the issue from the sandbox you posted. I tested an adjusted example and it works as expected https://codesandbox.io/s/gracious-mendeleev-ulmez?file=/src/App.js

Let me know if i'm missing something!

norman-ags:
> Hey @norman-ags, I can't see the issue from the sandbox you posted. I tested an adjusted example and it works as expected https://codesandbox.io/s/gracious-mendeleev-ulmez?file=/src/App.js
> 
> Let me know if i'm missing something!

Sorry, I forgot to save it while editing. I updated the sandbox now and added this:
```
// Added scaleIn 
animation: `${scaleIn} 0.5s ease-out`
```

andy-hook:
> I updated the sandbox now

Great, thanks. For animation we recommend using the `date-state` attributes per my example so that the presence logic can track and correctly mount/unmount. If you only want to animate in you can simply use `data-state="open”`.  

https://www.radix-ui.com/docs/primitives/overview/animation

Is that helpful? or was there a reason why these wouldn't work for you?

norman-ags:
> Is that helpful? or was there a reason why these wouldn't work for you?

Actually I was just following the example in the [docs](https://www.radix-ui.com/docs/primitives/components/dropdown-menu#origin-aware-animations)

andy-hook:
> Actually, I was just following the example in the [docs](https://www.radix-ui.com/docs/primitives/components/dropdown-menu#origin-aware-animations)

Doh I see the issue now, yes this should be working so looks like a regression. Thanks for the report 👍

MildTomato:
@andy-hook also seeing the same issue. also following [docs](https://www.radix-ui.com/docs/primitives/components/dropdown-menu#origin-aware-animations)

[...]

PR #851, opened 2021-09-09
[Presence] Get `animationName` consistently

andy-hook:
Fixes https://github.com/radix-ui/primitives/issues/849

The problem was caused by some state being stuck in `isAnimating` and suspending the unmount. It seems that the animation name getting retrieved from style in [the mount effect](https://github.com/radix-ui/primitives/blob/887228c4eab10aace9132b907e8693a027ae7027/packages/react/presence/src/Presence.tsx#L53) can sometimes return as `none` under certain conditions, that's why this doesn't show in the story. I was wondering if [this comment](https://github.com/radix-ui/primitives/pull/828#discussion_r691191786) was referring to this issue in the past?

Getting the name in the start event makes this behave consistently.

### DIFF
--- a/packages/react/presence/src/Presence.stories.tsx
+++ b/packages/react/presence/src/Presence.stories.tsx
@@ -28,6 +28,35 @@ export const WithMultipleOpenAndCloseAnimations = () => (
   <Animation className={multipleOpenAndCloseAnimationsClass} />
 );
 
+export const WithDeferredMountAnimation = () => {
+  const ref = React.useRef<HTMLDivElement>(null);
+  const timerRef = React.useRef(0);
+  const [open, setOpen] = React.useState(false);
+  const [animate, setAnimate] = React.useState(false);
+
+  React.useEffect(() => {
+    if (open) {
+      timerRef.current = window.setTimeout(() => {
+        setAnimate(true);
+      }, 150);
+    } else {
+      setAnimate(false);
+      window.clearTimeout(timerRef.current);
+    }
+  }, [open]);
+
+  return (
+    <>
+      <Toggles nodeRef={ref} open={open} onOpenChange={setOpen} />
+      <Presence present={open}>
+        <div className={animate ? mountAnimationClass : noAnimationClass} ref={ref}>
+          Content
+        </div>
+      </Presence>
+    </>
+  );
+};
+
 function Animation(props: React.ComponentProps<'div'>) {
   const ref = React.useRef<HTMLDivElement>(null);
   const [open, setOpen] = React.useState(false);
@@ -92,6 +121,10 @@ const slideUp = css.keyframes({
 const slideDown = css.keyframes({
   from: { transform: 'translateY(0)' },
   to: { transform: 'translateY(30px)' },
+});
+
+const noAnimationClass = css({
+  animation: 'none',
 });
 
 const mountAnimationClass = css({
--- a/packages/react/presence/src/Presence.stories.tsx
+++ b/packages/react/presence/src/Presence.stories.tsx
@@ -36,9 +36,7 @@ export const WithDeferredMountAnimation = () => {
 
   React.useEffect(() => {
     if (open) {
-      timerRef.current = window.setTimeout(() => {
-        setAnimate(true);
-      }, 150);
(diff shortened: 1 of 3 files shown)


benoitgrelard (commented):
I seem to remember we used to have that but then it was taken out.
@jjenzz do you remember why?

jjenzz:
I removed this [here](https://github.com/radix-ui/primitives/pull/828) ... I wonder if close animation is now broken again?

andy-hook:
> Is close animation broken again now that we've re-introduced this? As in, a close animation only.

Looked ok in my testing – https://601857eb614bae0021c9b9dd-iwpsaxggjf.chromatic.com/?path=/story/components-presence--with-unmount-animation

jjenzz:
Sweet. I thought this was redundant given my other changes as the stories all still worked. Can we add a new story that highlights this issue so that any future changes don't re-introduce it please?

andy-hook:
> What makes this example different?

At the moment the implementation assumes that the animation won't change after mount, if the initial animation happens to be `none`, (in our case when a [component uses popper](https://github.com/radix-ui/primitives/blob/887228c4eab10aace9132b907e8693a027ae7027/packages/react/popper/src/Popper.tsx#L153)) then that value is what is retrieved initially in the [`getComputedStyle(node)`](https://github.com/radix-ui/primitives/blob/887228c4eab10aace9132b907e8693a027ae7027/packages/react/presence/src/Presence.tsx#L118), it doesn't have mechanism to update that value if an animation is then added after.

Here's a sandbox showing the problem

https://codesandbox.io/s/angry-wu-vis9w?file=/src/App.js

By listening on the `start` event we can update the previous animation value if it changes. This isn't reflected in the existing story because the animation is never updated like this.

jjenzz:
Ahhh, thanks for the example. I seeeee 🙂 In that case, yeah just a story to help us avoid this in future would be super helpful 🙏

packages/react/presence/src/Presence.stories.tsx
```diff
@@ -28,6 +28,35 @@ export const WithMultipleOpenAndCloseAnimations = () => (
...
+      setAnimate(false);
+      window.clearTimeout(timerRef.current);
+    }
+  }, [open]);
+
+  return (
+    <>
+      <Toggles nodeRef={ref} open={open} onOpenChange={setOpen} />
+      <Presence present={open}>
+        <div className={animate ? mountAnimationClass : noAnimationClass} ref={ref}>
```
jjenzz, line 52: I found the flicker where it appears and then animates in jarring. I thought it was a bug in the story. Thoughts on changing to this so it looks more like a real use-case?

```suggestion
        <div hidden={!animate} className={animate ? mountAnimationClass : undefined} ref={ref}>
```

packages/react/presence/src/Presence.stories.tsx
```diff
@@ -28,6 +28,35 @@ export const WithMultipleOpenAndCloseAnimations = () => (
...
+  const ref = React.useRef<HTMLDivElement>(null);
+  const timerRef = React.useRef(0);
+  const [open, setOpen] = React.useState(false);
+  const [animate, setAnimate] = React.useState(false);
+
+  React.useEffect(() => {
+    if (open) {
+      timerRef.current = window.setTimeout(() => {
+        setAnimate(true);
+      }, 150);
```
jjenzz, line 41: nit 😅

```suggestion
      timerRef.current = window.setTimeout(() => setAnimate(true), 150);
```

jjenzz (approved):

packages/react/presence/src/Presence.stories.tsx
andy-hook: Oh yeah this is better, I did wonder if this looked a bit broken without prior knowledge

andy-hook: @jjenzz I realised using `hidden` would mask the issue we want check for. I've added a description instead https://github.com/radix-ui/primitives/pull/851/commits/9b0fb0e131da7b5913bd688e363ceb43c32bbacf let me know if you think this is enough.

jjenzz: That works :+1: thanks

Issue #848, opened 2021-09-08
[DropdownMenu][Tooltip] combination fails to open menu


benoitgrelard:
### Current Behavior

Menu won't open

### Expected behavior

both tooltip and menu should work

### Reproducible example

https://codesandbox.io/s/wandering-architecture-d8uw7?file=/src/App.tsx

### Additional context

Seems to be related to refs for me because I can see the menu isn't being positioned, so probably can't measure the trigger somehow.

benoitgrelard:
After some digging, we think this is because of nested contexts (in this case `Popper`).
Similar to this issue raised already for `ContextMenu`: #681

jjenzz:
Closing as duplicate of https://github.com/radix-ui/primitives/issues/681

Issue #681, opened 2021-06-03
Components with shared underlying context break


jjenzz:
Original report https://modulz.slack.com/archives/C024220NA2V/p1622709843010000

This composition works as expected https://codesandbox.io/s/keen-greider-h10gy?file=/src/App.tsx 

This one doesn't https://codesandbox.io/s/old-bush-hnbdw?file=/src/App.tsx

jjenzz:
Benoit and I looked into this previously. It's because the two components use `Menu` under the hood so the parts will be using the closest `MenuContext` they find.

We need to make sure the separate components use separate contexts somehow.

PR #852, opened 2021-09-09
[All] Replace `React.createContext` with our `createContext` util

jjenzz:
As discussed, replaces `React.createContext` instances with our utility in preparation for the bug fixes coming for:

- #681
- #848
- #210

### DIFF
--- a/packages/react/alert-dialog/src/AlertDialog.tsx
+++ b/packages/react/alert-dialog/src/AlertDialog.tsx
@@ -80,10 +80,9 @@ const AlertDialogContent = React.forwardRef<AlertDialogContentElement, AlertDial
 
     return (
       <DialogPrimitive.LabelWarningProvider
-        value={React.useMemo(
-          () => ({ contentName: CONTENT_NAME, titleName: TITLE_NAME, docsSlug: 'alert-dialog' }),
-          []
-        )}
+        contentName={CONTENT_NAME}
+        titleName={TITLE_NAME}
+        docsSlug="alert-dialog"
       >
         <AlertDialogContentProvider cancelRef={cancelRef}>
           <DialogPrimitive.Content
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -35,7 +35,7 @@ interface ContextMenuProps {
 const ContextMenu: React.FC<ContextMenuProps> = (props) => {
   const { children, onOpenChange, dir, modal = true } = props;
   const [open, setOpen] = React.useState(false);
-  const { isInsideContent } = useContentContext(CONTEXT_MENU_NAME);
+  const contentContext = useContentContext(CONTEXT_MENU_NAME);
   const handleOpenChangeProp = useCallbackRef(onOpenChange);
 
   const handleOpenChange = React.useCallback(
@@ -46,7 +46,7 @@ const ContextMenu: React.FC<ContextMenuProps> = (props) => {
     [handleOpenChangeProp]
   );
 
-  return isInsideContent ? (
+  return contentContext.isInsideContent ? (
     <ContextMenuProvider
       isRootMenu={false}
       open={open}
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -35,7 +35,7 @@ interface ContextMenuProps {
 const ContextMenu: React.FC<ContextMenuProps> = (props) => {
   const { children, onOpenChange, dir, modal = true } = props;
   const [open, setOpen] = React.useState(false);
-  const isInsideContent = React.useContext(ContentContext);
+  const { isInsideContent } = useContentContext(CONTEXT_MENU_NAME);
   const handleOpenChangeProp = useCallbackRef(onOpenChange);
 
   const handleOpenChange = React.useCallback(
@@ -103,7 +103,7 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
     React.useEffect(() => clearLongPress, [clearLongPress]);
 
     return (
-      <ContentContext.Provider value={false}>
+      <ContentProvider isInsideContent={false}>
         <MenuPrimitive.Anchor virtualRef={virtualRef} />
         <Primitive.span
           {...props}
@@ -132,7 +132,7 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
           )}
           onPointerUp={composeEventHandlers(props.onPointerUp, whenTouchOrPen(clearLongPress))}
         />
-      </ContentContext.Provider>
+      </ContentProvider>
     );
   }
 );
@@ -145,7 +145,9 @@ ContextMenuTrigger.displayName = TRIGGER_NAME;
 
 const CONTENT_NAME = 'ContextMenuContent';
 
-const ContentContext = React.createContext(false);
+const [ContentProvider, useContentContext] = createContext(CONTENT_NAME, {
+  isInsideContent: false,
+});
 
 type ContextMenuContentElement = React.ElementRef<typeof MenuPrimitive.Content>;
 type MenuContentProps = Radix.ComponentPropsWithoutRef<typeof MenuPrimitive.Content>;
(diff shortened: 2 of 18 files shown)


packages/react/alert-dialog/src/AlertDialog.tsx
```diff
@@ -80,10 +80,9 @@ const AlertDialogContent = React.forwardRef<AlertDialogContentElement, AlertDial
 
     return (
       <DialogPrimitive.LabelWarningProvider
-        value={React.useMemo(
-          () => ({ contentName: CONTENT_NAME, titleName: TITLE_NAME, docsSlug: 'alert-dialog' }),
-          []
-        )}
+        contentName={CONTENT_NAME}
+        titleName={TITLE_NAME}
+        docsSlug="alert-dialog"
```
benoitgrelard, line 85: I'm confused how this one gets a change when I don't see any associated change to `LabelWarningProvider`.
What am I missing here?

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -35,7 +35,7 @@ interface ContextMenuProps {
 const ContextMenu: React.FC<ContextMenuProps> = (props) => {
   const { children, onOpenChange, dir, modal = true } = props;
   const [open, setOpen] = React.useState(false);
-  const isInsideContent = React.useContext(ContentContext);
+  const { isInsideContent } = useContentContext(CONTEXT_MENU_NAME);
```
benoitgrelard, line 38: I think whenever possible we do this no? `contentContext = ` and then later `contentContext.isInsideContent`

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -103,7 +103,7 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
     React.useEffect(() => clearLongPress, [clearLongPress]);
 
     return (
-      <ContentContext.Provider value={false}>
+      <ContentProvider isInsideContent={false}>
```
benoitgrelard, line 106: This is arguable more readable now too 👍

packages/react/dialog/src/Dialog.tsx
```diff
@@ -419,20 +419,18 @@ function getState(open: boolean) {
   return open ? 'open' : 'closed';
 }
 
-const LabelWarningContext = React.createContext({
+const [LabelWarningProvider, useLabelWarningContext] = createContext('DialogLabelWarning', {
```
benoitgrelard, line 422: Should we put `DialogLabelWarning` in name `const` like others?

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -301,31 +302,38 @@ function useFocusOutside(onFocusOutside?: (event: FocusOutsideEvent) => void) {
...
+  );
 
   const TotalLayerCountProvider: React.FC = ({ children }) => {
     const [total, setTotal] = React.useState(0);
-    const context = React.useMemo(() => ({ total, setTotal }), [total, setTotal]);
     return (
-      <TotalLayerCountContext.Provider value={context}>{children}</TotalLayerCountContext.Provider>
+      <TotalLayerCountProviderImpl
+        total={total}
+        onTotalIncrease={React.useCallback(() => setTotal((n) => n + 1), [])}
```
benoitgrelard, line 315: nice refactoring to align with other places 👍

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -334,22 +342,24 @@ function createTotalLayerCount(displayName?: string) {
...
-        {children}
-      </RunningLayerCountContext.Provider>
+      <RunningLayerCountProviderImp count={runningCount}>{children}</RunningLayerCountProviderImp>
     );
   };
   if (displayName) {
     RunningLayerCountProvider.displayName = displayName;
   }
 
+  const CONSUMER_NAME = 'RunningLayerCountConsumer';
```
benoitgrelard, line 359: I'm confused what rules we've used in this file (and others) to decide when to put the name in a const or not, seems like sometimes here it's inline because used only once, but here that one is also used only once.
What was the thought behind that in general?

packages/react/label/src/Label.tsx
```diff
@@ -12,7 +13,8 @@ import type * as Radix from '@radix-ui/react-primitive';
 const NAME = 'Label';
 
 type LabelContextValue = { id: string; ref: React.RefObject<HTMLSpanElement> };
-const LabelContext = React.createContext<LabelContextValue | undefined>(undefined);
+
+const [LabelProvider, useLabelContextImpl] = createContext<LabelContextValue | null>(NAME, null);
```
benoitgrelard, line 17: maybe this should be `useLabelContext` and the public one `useLabel` because it does more than just returning context?

benoitgrelard (commented):
There seems to be still something wrong with the label one, as caught by the Chromatic story.

packages/react/context-menu/src/ContextMenu.tsx
benoitgrelard: Same in other places obviously if we decide to do it.

packages/react/dismissable-layer/src/DismissableLayer.tsx
jjenzz: Mostly did it because it made the `defaultContext` easier to type 😅 nice win in the process.

packages/react/context-menu/src/ContextMenu.tsx
jjenzz: I thought that as well ... 🎉

packages/react/alert-dialog/src/AlertDialog.tsx
jjenzz: The `LabelWarningProvider` change is here https://github.com/radix-ui/primitives/pull/852/files/3e240a58c85a0cd28a02a1379388a83a48b95038#diff-18be67c22def126bf675c06beb358c92a4a8c6d1a071adf162fcd5ac9cac18bdR422

It's using  `DialogPrimitive.LabelWarningProvider`.

packages/react/context-menu/src/ContextMenu.tsx
jjenzz: Oh yeah, true, I tried to go with minimal changes and went a bit too minimal here 😆 will change for sure!

packages/react/dismissable-layer/src/DismissableLayer.tsx
jjenzz: I'm not fussed which route we take tbh, probably easier to do these inline right?

packages/react/label/src/Label.tsx
jjenzz: I thought that too but then that's a breaking change.... how do we handle semver with breaking changes now that we're in beta 😕🤔

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -301,31 +302,37 @@ function useFocusOutside(onFocusOutside?: (event: FocusOutsideEvent) => void) {
  * -----------------------------------------------------------------------------------------------*/
 
 function createTotalLayerCount(displayName?: string) {
-  const TotalLayerCountContext = React.createContext<{
-    total: number;
-    setTotal: React.Dispatch<React.SetStateAction<number>>;
-  }>({ total: 0, setTotal: () => {} });
+  const [TotalLayerCountProviderImpl, useTotalLayerCountContext] = createContext(
```
benoitgrelard, line 305: Doesn't matter for that change I guess but my brain still hurts thinking about whether this would need the same context treatment fix as the rest or not…

benoitgrelard (commented):
So with these changes, does that mean that basically we should never have `React.createContext` anywhere apart from once in `createContext`?

Not sure if you saw also but there's a conflict in `Label`.

jjenzz:
> Not sure if you saw also but there's a conflict in `Label`.

Yep I saw 🙂  resolved now!

jjenzz:
> does that mean that basically we should never have `React.createContext` anywhere apart from once in `createContext`

Yeah, gives us a central place to manipulate component context.

benoitgrelard (approved):

andy-hook (approved):

jjenzz:
I'm changing the base branch for this one to `context-scope` so it will all get merged into `main` with the new scope stuff I demoed this morning. I can test everything together thoroughly then.

PR #853, opened 2021-09-09
[Primitive] Add radix identifier

jjenzz:


packages/react/primitive/src/Primitive.tsx
```diff
@@ -29,6 +29,11 @@ const Primitive = NODES.reduce(
     [node]: React.forwardRef((props: PrimitivePropsWithRef<typeof node>, forwardedRef: any) => {
       const { asChild, ...primitiveProps } = props;
       const Comp: any = asChild ? Slot : node;
+
+      React.useEffect(() => {
+        (window as any)[Symbol.for('radix-ui')] = true;
```
benoitgrelard, line 34: I was gonna ask if we should check if it's already there but probably doesn't matter really.

benoitgrelard (approved):

andy-hook (approved):

packages/react/primitive/src/Primitive.tsx
jjenzz: I'm not sure if that matters, although, I did wonder if I should remove this at some stage. As in, keep a count somewhere like we do in `useBodyPointerEvents` and remove the attribute if that reaches `0`?

Depends what we want this to communicate I guess. Should it communicate that there is a currently a radix component on the page or that they used radix somewhere at some stage?

benoitgrelard: hmm yeah good point, we should ask @peduarte but I would say the former?

benoitgrelard: I believe with stitches you wouldn't see that script tag if no style was ever used on the page would you?

peduarte: I think we only need to communicate it once.

With stitches, it will still add a `style` tag if you create a theme, as it needs to convert them to CSS variables - even if you have no styles

peduarte:
🎉

PR #858, opened 2021-09-13
[Collapsible][Slot] Fix `React.Children.only` error

jjenzz:
Originally reported here https://discord.com/channels/752614004387610674/803656530259738674/886570819134816317

Sandbox showing issue: https://codesandbox.io/s/kind-violet-v0b1b?file=/src/App.tsx

Sometimes the `Slot` children might be a `false` boolean if someone does `isSomething && <Whatever />` and that would cause `React.Children.only` to throw its error because it expects a React Element.

In the case of `CollapsibleContent` it was doing `isOpen && children`.

### DIFF
--- a/packages/react/collapsible/src/Collapsible.tsx
+++ b/packages/react/collapsible/src/Collapsible.tsx
@@ -32,7 +32,7 @@ interface CollapsibleProps extends PrimitiveDivProps {
   defaultOpen?: boolean;
   open?: boolean;
   disabled?: boolean;
-  onOpenChange?(open?: boolean): void;
+  onOpenChange?(open: boolean): void;
 }
 
 const Collapsible = React.forwardRef<CollapsibleElement, CollapsibleProps>(
--- a/packages/react/slot/src/Slot.stories.tsx
+++ b/packages/react/slot/src/Slot.stories.tsx
@@ -59,6 +59,38 @@ export const Chromatic = () => (
       <SlotWithoutSlottable>{false}</SlotWithoutSlottable>
     </ErrorBoundary>
 
+    <h2>
+      False internal child - <span aria-hidden>✅</span>
+    </h2>
+    <ErrorBoundary>
+      <SlotWithFalseInternalChild>
+        <b data-slot-element>hello</b>
+      </SlotWithFalseInternalChild>
+    </ErrorBoundary>
+
+    <h2>
+      Null internal child - <span aria-hidden>✅</span>
+    </h2>
+    <ErrorBoundary>
+      <SlotWithNullInternalChild>
+        <b data-slot-element>hello</b>
+      </SlotWithNullInternalChild>
+    </ErrorBoundary>
+
+    <h2>
+      String consumer child - <span aria-hidden>✅</span>
+    </h2>
+    <ErrorBoundary>
+      <SlotWithoutSlottable>test</SlotWithoutSlottable>
+    </ErrorBoundary>
+
+    <h2>
+      Number consumer child - <span aria-hidden>✅</span>
+    </h2>
+    <ErrorBoundary>
+      <SlotWithoutSlottable>{1}</SlotWithoutSlottable>
+    </ErrorBoundary>
+
     <h1>With Slottable</h1>
 
     <h2>
@@ -88,6 +120,20 @@ export const Chromatic = () => (
     </ErrorBoundary>
 
     <h2>
+      String consumer child - <span aria-hidden>✅</span>
+    </h2>
+    <ErrorBoundary>
+      <SlotWithSlottable>test</SlotWithSlottable>
+    </ErrorBoundary>
+
+    <h2>
+      Number consumer child - <span aria-hidden>✅</span>
+    </h2>
+    <ErrorBoundary>
+      <SlotWithSlottable>{1}</SlotWithSlottable>
+    </ErrorBoundary>
+
+    <h2>
       Empty consumer child - <span aria-hidden>✅</span>
     </h2>
     <ErrorBoundary>
@@ -99,24 +145,6 @@ export const Chromatic = () => (
     </h2>
     <ErrorBoundary>
       <SlotWithSlottable>{false}</SlotWithSlottable>
-    </ErrorBoundary>
-
-    <h2>
-      False internal child - <span aria-hidden>✅</span>
-    </h2>
-    <ErrorBoundary>
-      <SlotWithFalseInternalChild>
-        <b data-slot-element>hello</b>
-      </SlotWithFalseInternalChild>
-    </ErrorBoundary>
-
-    <h2>
-      Null internal child - <span aria-hidden>✅</span>
-    </h2>
-    <ErrorBoundary>
-      <SlotWithNullInternalChild>
-        <b data-slot-element>hello</b>
-      </SlotWithNullInternalChild>
     </ErrorBoundary>
   </>
 );
@@ -151,13 +179,9 @@ const SlotWithSlottable = ({ children, ...props }: any) => (
 );
 
 const SlotWithFalseInternalChild = ({ children, ...props }: any) => (
-  <Slot {...props}>
-    <Slottable>{false && children}</Slottable>
-  </Slot>
+  <Slot {...props}>{false && children}</Slot>
 );
 
 const SlotWithNullInternalChild = ({ children, ...props }: any) => (
-  <Slot {...props}>
-    <Slottable>{false ? children : null}</Slottable>
-  </Slot>
+  <Slot {...props}>{false ? children : null}</Slot>
 );
--- a/packages/react/slot/src/Slot.stories.tsx
+++ b/packages/react/slot/src/Slot.stories.tsx
@@ -17,56 +17,106 @@ export const WithSlottable = () => (
 
 export const Chromatic = () => (
   <>
-    <h1>
-      One consumer child without internal slottable - <span aria-hidden>✅</span>
-    </h1>
+    <h1>Without Slottable</h1>
+
+    <h2>
+      One consumer child - <span aria-hidden>✅</span>
+    </h2>
     <ErrorBoundary>
       <SlotWithoutSlottable>
         <b data-slot-element>hello</b>
       </SlotWithoutSlottable>
     </ErrorBoundary>
 
-    <h1>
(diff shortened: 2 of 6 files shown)


packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -32,7 +32,7 @@ interface CollapsibleProps extends PrimitiveDivProps {
   defaultOpen?: boolean;
   open?: boolean;
   disabled?: boolean;
-  onOpenChange?(open?: boolean): void;
+  onOpenChange?(open: boolean): void;
```
jjenzz, line 35: While fixing this I realised this would error:

```jsx

<Collapsible ... onOpenChange={setOpen} />
```

It errors because the `onOpenChange` expects `setOpen` to accept `undefined` but as far as I know, this would/should always give them a boolean.

packages/react/slot/src/Slot.tsx
```diff
@@ -11,63 +11,24 @@ interface SlotProps {
...
-  const { children, ...slotProps } = props;
-  const child = React.Children.only(children);
-
-  return React.isValidElement(child)
-    ? React.cloneElement(child, {
-        ...mergeProps(slotProps, child.props),
-        ref: composeRefs(forwardedRef, (child as any).ref),
-      })
+  return children
+    ? React.Children.only(children) && React.isValidElement(children)
```
jjenzz, line 15: `React.Children.only` already does `React.isValidElement` under the hood but I have had to do it again here to hold TypeScript's hand and prevent the `children.props` code from erroring.

packages/react/slot/src/Slot.tsx
```diff
@@ -11,63 +11,24 @@ interface SlotProps {
...
-SlotClone.displayName = 'SlotClone';
+Slot.displayName = 'Slot';
 
 /* -------------------------------------------------------------------------------------------------
  * Slottable
  * -----------------------------------------------------------------------------------------------*/
 
 const Slottable = ({ children }: { children: React.ReactNode }) => {
-  return children as React.ReactElement;
+  return <>{children}</>;
```
jjenzz, line 31: I also noticed that components using `Slottable` (e.g. `Tooltip.Content`) would throw an error if there were no children, e.g. 

```jsx
<Tooltip.Content></Tooltip.Content>
```

Slipped this fix in while I was here. Hope that's okay 🙏

packages/react/slot/src/Slot.tsx
```diff
@@ -50,9 +50,8 @@ interface SlotCloneProps {
 
 const SlotClone = React.forwardRef<any, SlotCloneProps>((props, forwardedRef) => {
   const { children, ...slotProps } = props;
-  const child = React.Children.only(children);
-
-  return React.isValidElement(child)
+  const child = React.isValidElement(children) && React.Children.only(children);
```
benoitgrelard, line 53: I understand how this fixes the issue when I think about the fact that `React.Children.only` would error, but somehow I find it really confusing to read the diff.

Probably one of these cases where the diff is confusing, but reading the code by itself is ok though.

benoitgrelard (dismissed):

packages/react/slot/src/Slot.tsx
jjenzz: It's behaving as if `React` checks the length of the children internally first before verifying if the child is a react element. Whereas we want to check if it's an element first.

benoitgrelard: Exactly, the confusing part for me when looking at the diff is that we were checking the validity of the element regardless after. So at first glance if you don't really know how `React.Children.Only` behaves, the code kinda looks equivalent, just moved around a bit.

jjenzz: I dunno... `false && throw` isn't equivalent to `throw && false`. We needed the former to prevent it falling into the throw.

jjenzz: Okay, I've moved this all around a bit. Hopefully it's clearer 🙈

packages/react/slot/src/Slot.tsx
```diff
@@ -13,28 +13,22 @@ const Slot = React.forwardRef<HTMLElement, SlotProps>((props, forwardedRef) => {
...
+      return isSlottable(child) ? (
+        <SlotClone {...slotProps} ref={forwardedRef}>
+          {child.props.children}
+        </SlotClone>
+      ) : (
+        child
+      );
+    })
+  ) : (
+    React.Children.only(children)
```
jjenzz, line 31: This is now saying that if the length is greater than `1` and none of the children contain a `Slottable` then it is the consumer code providing more than one child so we throw the `React.Children.only` error.

packages/react/slot/src/Slot.tsx
```diff
@@ -13,28 +13,24 @@ const Slot = React.forwardRef<HTMLElement, SlotProps>((props, forwardedRef) => {
...
+          <SlotClone {...slotProps} ref={forwardedRef}>
+            {child.props.children}
+          </SlotClone>
+        ) : (
+          child
+        );
       })}
     </>
+  ) : (
+    <>{React.Children.only(children)}</>
```
jjenzz, line 33: This is now saying that if the length is greater than `1` and none of the children contain a `Slottable` then it is the consumer code providing more than one child so we throw the `React.Children.only` error.

packages/react/slot/src/Slot.tsx
```diff
@@ -13,28 +13,24 @@ const Slot = React.forwardRef<HTMLElement, SlotProps>((props, forwardedRef) => {
...
-      </SlotClone>
-    );
-  }
-
-  return (
+  return childLength === 1 ? (
+    <SlotClone {...slotProps} ref={forwardedRef}>
+      {children}
+    </SlotClone>
+  ) : React.Children.toArray(children).some(isSlottable) ? (
```
andy-hook, line 20: nit: I personally find nested ternaries difficult to read compared to the early return.

packages/react/slot/src/Slot.tsx
andy-hook: Is it expected that `< 1` should throw? seems that the `WithoutSlottable` story now throws while `WithSlottable` does not, this was the opposite before.

andy-hook (commented):
I'm not super familiar here but left a couple of small comments. Wondering whether we should we add some tests to validate the correct behaviour?

packages/react/slot/src/Slot.tsx
jjenzz: If there isn't a `Slottable` it should throw. I've updated the Chromatic stories to reflect the expectation. Let me know if they're clearer now :+1:

andy-hook: Thanks for that, it does make more sense now. What I'm still finding confusing though are the `boolean` values as children. Right now `false` won't throw yet `null` will, I kind of expected those to behave consistently?

jjenzz: Oh, let me take a look... that might be another bug haha

andy-hook: > I think that if the consumer does false && ... it should error because asChild expects a child.

I think this is where I'm getting confused because now, with this change, in collapsible `isOpen && children` wont error (that's the fix) but if I wrote it as `isOpen ? children : null` it would. I know technically why that is, but from a consumer perspective that seems inconsistent, my heads spins a bit with some of the slot stuff though so perhaps I'm not following? 🤔

jjenzz: Haha, there is a difference between what the consumer passes and what we do internally. Let's call about this quick so I explain... it is the reason why we have the `Slottable`.

However, I've actually had to revert to allowing the consumer to pass `null` or `false` because I can't seem to determine if a falsey child was provided by the consumer or our code so tests look like this now:

image

I think this is fine for now... it would have been a nice to have to make `asChild` error if there was no child but it's not the end of the world if it doesn't.

packages/react/collapsible/package.json
```diff
@@ -23,6 +23,7 @@
     "@radix-ui/react-id": "workspace:*",
     "@radix-ui/react-presence": "workspace:*",
     "@radix-ui/react-primitive": "workspace:*",
+    "@radix-ui/react-slot": "workspace:*",
```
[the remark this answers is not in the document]
jjenzz: omg, i need to stop doing too many things at once 😅

jjenzz:
@benoitgrelard would you mind taking another look at this one to make sure you're happy? I've had to make it so that if the consumer passes `{null}`, `{false}`, `'string'`, or `{1}` as a child then it will render `null` and won't error. 

Unfortunately, I couldn't figure out a way to identify if those child types were theirs or rendered internally by us so had to go this route to fix the original issue.

andy-hook (approved):
I'm happy here. Would be good to get @benoitgrelard thoughts though as you are both more familiar with the implications of this approach than me.

packages/react/slot/src/Slot.stories.tsx
```diff
@@ -4,38 +4,184 @@ import { Slot, Slottable } from './Slot';
 export default { title: 'Components/Slot' };
 
 export const WithoutSlottable = () => (
-  <AsCompWithoutSlottable as={Slot}>
```
benoitgrelard, line 7: oh wow these hadn't been updated??!

packages/react/slot/src/Slot.tsx
```diff
@@ -11,30 +11,28 @@ interface SlotProps {
...
               {child.props.children}
             </SlotClone>
+          ) : (
+            child
           );
-        }
+        })}
+      </>
+    );
+  }
```
benoitgrelard, line 30: So this case if for things like `Tooltip` where it will find a `<Slottable>` internally?

benoitgrelard (approved):
This all looks actually a lot clearer to me overall, differentiating the different scenarios in the code.

I've just left a few questions to sense check my own understanding.

packages/react/slot/src/Slot.stories.tsx
jjenzz: Yeah but they didn't matter really, they're just stories showing that `Slot` works so I left them before but changed them now because it definitely made me double take 😅

packages/react/slot/src/Slot.tsx
jjenzz: Yep 👍  it will slot **their** children and leave our children as is.

PR #859, opened 2021-09-13
[Tabs] Unmount content when inactive

### LINKED ISSUE #813: Mount / Unmount of blurred tabs
Feature request 
--------

I've been using tabs to load 2 different components. Both the components have a keyboard event listener. The tab that is blurred is still detecting the keyboard events even though the tab is not focused. I wonder if the tabs should unmount or should have a `focus` property. There are pro's and con's for both but I personal prefer the unmount version.

jjenzz:
Following up on this discussion https://github.com/radix-ui/primitives/discussions/855#discussioncomment-1316094

Closes #813.

Adding this change for discussion.

### DIFF
--- a/.yarn/versions/03a3e321.yml
+++ b/.yarn/versions/03a3e321.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-tabs": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


benoitgrelard (approved):
Change looks simple enough, did we test if it broke anything accessibility wise?

jjenzz:
I tested in VoiceOver but just tested NVDA as well to make sure 😅  It works fine :+1:

PR #861, opened 2021-09-13
[Label] Use React events to allow `stopPropagation`

jjenzz:
Been feeling unproductive with this context issue, needed to fix something before I go insane 😅

Fixes https://github.com/radix-ui/primitives/issues/609

https://user-images.githubusercontent.com/175330/133163355-bed38c31-28c1-4b44-88aa-8ecda8291900.mp4


packages/react/label/src/Label.stories.tsx
```diff
@@ -28,6 +29,56 @@ export const WithHtmlFor = () => (
...
+      <Label htmlFor="control">Label</Label>
+
+      <h1>Wrapping control - stopping propagation</h1>
+      <Label>
+        <Checkbox.Root className={controlClass}>
+          <Checkbox.Indicator className={indicatorClass} />
+        </Checkbox.Root>{' '}
+        Label.{' '}
+        <span
+          aria-hidden
```
jjenzz, line 56: Our a11y eslint rules moan about non-interactive elements having `onClick` handlers without a keyboard handler. Adding this `aria-hidden` silenced it.

packages/react/label/src/Label.stories.tsx
```diff
@@ -76,17 +127,25 @@ const rootClass = css({
...
 const controlClass = css({
-  display: 'inline-block',
+  display: 'inline-flex',
   border: '1px solid gainsboro',
   padding: 10,
   verticalAlign: 'middle',
   margin: '0 10px',
 
-  '&::after': {
+  '&::before': {
```
benoitgrelard, line 136: This changed the output of chromatic obviously, but was it intentional? The output doesn't read as well now.

jjenzz: Ah, good eyes! No need for this change, I was fiddling with styling. Will revert!

jjenzz: Story looks like this now:

https://user-images.githubusercontent.com/175330/133243671-a8d3336d-6ea2-4e75-8311-ef75052f18a4.mp4

packages/react/label/src/Label.tsx
```diff
@@ -69,8 +53,35 @@ const Label = React.forwardRef<LabelElement, LabelProps>((props, forwardedRef) =
...
-      <Primitive.span role="label" id={id} {...labelProps} ref={ref} />
+    <LabelContext.Provider
+      value={React.useMemo(() => ({ id, ref: labelRef, onControlChange: setControl }), [id])}
+    >
+      <Primitive.span
+        role="label"
+        id={id}
+        {...labelProps}
+        ref={ref}
+        onMouseDown={(event) => {
```
andy-hook, line 64: I think we need to compose `labelProps.onMouseDown`? same with `onClick`

packages/react/label/src/Label.tsx
```diff
@@ -69,8 +53,37 @@ const Label = React.forwardRef<LabelElement, LabelProps>((props, forwardedRef) =
   }, [id, htmlFor]);
 
   return (
-    <LabelContext.Provider value={React.useMemo(() => ({ id, ref: labelRef }), [id])}>
-      <Primitive.span role="label" id={id} {...labelProps} ref={ref} />
+    <LabelContext.Provider
+      value={React.useMemo(() => ({ id, ref: labelRef, onControlChange: setControl }), [id])}
```
andy-hook, line 57:
```suggestion
      value={React.useMemo(() => ({ id, onControlChange: setControl }), [id])}
```

Looks like we don't need to pass this anymore, what i'm wondering though is why TS doesn't complain about any of the properties passed in here, am I missing something?

packages/react/label/src/Label.tsx
```diff
@@ -22,37 +22,21 @@ interface LabelProps extends PrimitiveSpanProps {
 
 const Label = React.forwardRef<LabelElement, LabelProps>((props, forwardedRef) => {
   const { htmlFor, id: idProp, ...labelProps } = props;
+  const [control, setControl] = React.useState<HTMLElement | null>(null);
   const labelRef = React.useRef<HTMLSpanElement>(null);
   const ref = useComposedRefs(forwardedRef, labelRef);
```
andy-hook, line 27: Out of scope for the PR but I'm used to seeing this as `ref` and `composedRefs`

packages/react/label/src/Label.tsx
```diff
@@ -81,39 +94,15 @@ Label.displayName = NAME;
 
 const useLabelContext = (element?: HTMLElement | null) => {
   const context = React.useContext(LabelContext);
+  const { onControlChange } = context || {};
```
andy-hook, line 97: Should we use an empty obj as the initial context value instead?

packages/react/label/src/Label.tsx
jjenzz: Oh yeah, why didn't TS catch that 🤔

jjenzz: Seems abstracting the context swallows the TS errors (including in latest TS versions) https://www.typescriptlang.org/play?ts=3.9.7#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQdMAnmFnAGIQRwC8cA3nThwIAOywAuOKhhRgogOb1hMAO4QpogK4gARlij0AvvTq4xMuAGExMLAA94fbHhgA6XJWR2bou44AeTggAPgAKfiMUdGRRJgBKUwItUXxgMTgAeQBrZCYw+IEhOEoYLShROADff3cABWIAN2AAEwM4RuQAGy0sHn5+EXEpMjEsMgAaODUNOAAmKZgAC0pJcmXVsjgjKIB6ELojBmTUmHTKgCFibKxRAEYCouFzUUsX2t4BIbXR8Unp9RSBbTFZYH4bMFbEzFUrlSrVWwOepNVrtTo9Pr8d5IvYHI50E5pDJXCA3URzR6CYSwipVGpItwNCDNNpQDrdXr9Fz4NxaVBYACyWBAEDCjx4ITgEW+IzG-xmQMWoPByqh8SmAG0ALrxXGHY4pImXa63ADMlOKLzeiMcn257j5guFovFkulY1lf0WgPmStWIwh4226rg2sSMKwZVpCL8DKZLLRHMx2MceqMQA

packages/react/label/src/Label.tsx
jjenzz: Provided a default context value instead

packages/react/label/src/Label.tsx
andy-hook: 🙌

packages/react/label/src/Label.tsx
andy-hook: How annoying, as discussed earlier we could explicitly type the return / variable def to make it work but it hasn't been an issue anywhere else so probably fine tbh.

jjenzz: Yeah, there are only a couple places we do this. Shall I make a separate PR to type these to avoid this in future?

cc @benoitgrelard

andy-hook: >  Thoughts?

Agree, I'd prefer the type safety if it's not too much hassle for us to do so.

jjenzz: I've done a quick test locally and that doesn't seem to fix it either 🙈 

image

the `one` and `two` properties should be erroring.

packages/react/label/src/Label.tsx
```diff
@@ -22,37 +25,21 @@ interface LabelProps extends PrimitiveSpanProps {
...
-      const label = labelRef.current;
-
+      const label = ref.current;
       if (label && element) {
-        const removeLabelClickEventListener = addLabelClickEventListener(label, element);
         const getAriaLabel = () => element.getAttribute('aria-labelledby');
         const ariaLabelledBy = [getAriaLabel(), id].filter(Boolean).join(' ');
         element.setAttribute('aria-labelledby', ariaLabelledBy);
-
+        setControl(element);
```
andy-hook, line 41: Does this need to be state? or would a `ref` work?

jjenzz: Oooh good shout, changed!

packages/react/label/src/Label.tsx
andy-hook: Typing the return works for me

image

But as we're just talking about redundant properties I don't think it's a huge deal.

andy-hook (approved):

packages/react/label/src/Label.tsx
jjenzz: ooooo nice!

jjenzz: Proposed changes here https://github.com/radix-ui/primitives/pull/864

benoitgrelard: Yeah it's not completely broken, it seems it's just not complaining about extra properties, but it will complain in all the cases about wrong values, or missing values.

PR #866, opened 2021-09-15
[Slider] Fix zero values issue

### LINKED ISSUE #865: [Slider] zero values do not display correctly when min is less than zero
### Current Behavior

The Slider component does not display zero values correctly when the min value is set to less than zero. A component with the props `value={[0]} min={-100} max={100}` should display the knob at the center, however it instead displays it at the left side. 

image

Other values, including negative values, display correctly. A slider with the props `value={[-50]} min={-100} max={100}` would correctly display at 25% from left. 

### Expected behavior

when  zero value should display at the center of the slider.

### Reproducible example

https://codesandbox.io/s/determined-ellis-yc2zk?file=/App.js

### Your environment

"@radix-ui/react-slider": "^0.1.0",

benoitgrelard:
Fixes #865

I've added a Chromatic regression test for it too.

### DIFF
--- a/.yarn/versions/0a2f9065.yml
+++ b/.yarn/versions/0a2f9065.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-slider": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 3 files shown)


jjenzz (approved):
Sweet as a 🥜

PR #877, opened 2021-09-24
[Collection] Memoize `getItems`

benoitgrelard:


packages/react/collection/src/Collection.tsx
```diff
@@ -96,18 +96,19 @@ function createCollection<ItemElement extends HTMLElement, ItemData>() {
...
+        context.collectionRef.current!.querySelectorAll(`[${ITEM_DATA_ATTR}]`)
+      );
+      const items = Array.from(context.itemMap.values());
+      const orderedItems = items.sort(
+        (a, b) => orderedNodes.indexOf(a.ref.current!) - orderedNodes.indexOf(b.ref.current!)
+      );
+      return orderedItems;
+    }, [context.collectionRef, context.itemMap]);
+
+    return { getItems };
```
benoitgrelard, line 111: I'm wondering if it's worth simplifying the API as well now that we're here.
The hook only returns one thing, making the function the return itself would make more sense then? What do you reckon?

andy-hook: Yeah, was thinking the same

packages/react/collection/src/Collection.tsx
```diff
@@ -96,18 +96,19 @@ function createCollection<ItemElement extends HTMLElement, ItemData>() {
...
-        const orderedItems = items.sort(
-          (a, b) => orderedNodes.indexOf(a.ref.current!) - orderedNodes.indexOf(b.ref.current!)
-        );
-        return orderedItems;
-      },
-    };
+
+    const getItems = React.useCallback(() => {
+      const orderedNodes = Array.from(
+        context.collectionRef.current!.querySelectorAll(`[${ITEM_DATA_ATTR}]`)
```
andy-hook, line 102: Something else that I was wondering was the non-null assertions. Right now the expectation is that you would know to only call this when the ref is available. If you try to get the collection inside a `useMemo` call or a `useLayoutEffect` it'll throw.

How do you feel about returning null so these cases can be handled at the consumer level?

benoitgrelard: For this it might make sense to simply do this:

```tsx
context.collectionRef.current?.querySelectorAll(`[${ITEM_DATA_ATTR}]`) ?? []
```

no?

andy-hook: Yup 👍

benoitgrelard:
@andy-hook I've rejigged it a little bit differently with an early return, I think it's even clearer and doesn't need to run through the rest of the code.

andy-hook (approved):

packages/react/collection/src/Collection.tsx
```diff
@@ -96,18 +96,19 @@ function createCollection<ItemElement extends HTMLElement, ItemData>() {
...
-          (a, b) => orderedNodes.indexOf(a.ref.current!) - orderedNodes.indexOf(b.ref.current!)
-        );
-        return orderedItems;
-      },
-    };
+
+    const getItems = React.useCallback(() => {
+      const collectionNode = context.collectionRef.current;
+      if (!collectionNode) return [];
+      const orderedNodes = Array.from(collectionNode.querySelectorAll(`[${ITEM_DATA_ATTR}]`));
```
jjenzz, line 103: Ah nice, I made the same change [here](https://github.com/radix-ui/primitives/pull/852/files#diff-0ad5c595db850ab96b0d987a0bd230537711c920e37ce92ac64e6d2876dd429cR106) but it hasn't been merged into `main` yet. Great minds.

benoitgrelard: Brilliant 🙂

PR #879, opened 2021-09-27
[Collapsible][Accordion] Add content-width CSS custom property

imtiazmangerah:
### Description

This PR adds a `--radix-collapsible-content-width` CSS custom property to the Collapsible component.

### DIFF
--- a/packages/react/collapsible/src/Collapsible.tsx
+++ b/packages/react/collapsible/src/Collapsible.tsx
@@ -159,7 +159,7 @@ const CollapsibleContentImpl = React.forwardRef<
       node.style.transition = 'none';
       node.style.animation = 'none';
 
-      // set height to height from full dimensions
+      // get width and height from full dimensions
       const rect = node.getBoundingClientRect();
       heightRef.current = rect.height;
       widthRef.current = rect.width;
--- a/packages/react/collapsible/src/Collapsible.tsx
+++ b/packages/react/collapsible/src/Collapsible.tsx
@@ -144,8 +144,10 @@ const CollapsibleContentImpl = React.forwardRef<
   const composedRefs = useComposedRefs(forwardedRef, ref);
   const heightRef = React.useRef<number | undefined>(0);
   const height = heightRef.current;
+  const widthRef = React.useRef<number | undefined>(0);
+  const width = widthRef.current;
   // when opening we want it to immediately open to retrieve dimensions
-  // when closing we delay `present` to retreive dimensions before closing
+  // when closing we delay `present` to retrieve dimensions before closing
   const isOpen = context.open || isPresent;
 
   useLayoutEffect(() => {
@@ -160,6 +162,7 @@ const CollapsibleContentImpl = React.forwardRef<
       // set height to height from full dimensions
       const rect = node.getBoundingClientRect();
       heightRef.current = rect.height;
+      widthRef.current = rect.width;
 
       // kick off any animations/transitions that were originally set up
       node.style.transition = originalTransition;
@@ -184,6 +187,7 @@ const CollapsibleContentImpl = React.forwardRef<
       ref={composedRefs}
       style={{
         [`--radix-collapsible-content-height` as any]: height ? `${height}px` : undefined,
+        [`--radix-collapsible-content-width` as any]: width ? `${width}px` : undefined,
         ...props.style,
       }}
     >
--- a/packages/react/accordion/src/Accordion.stories.tsx
+++ b/packages/react/accordion/src/Accordion.stories.tsx
@@ -296,6 +296,34 @@ export const Animated = () => {
   );
 };
 
+export const Animated2D = () => {
+  const values = ['One', 'Two', 'Three', 'Four'];
+
+  return (
+    <>
+      <Accordion type="single" className={rootClass}>
+        {values.map((value) => (
+          <AccordionItem key={value} value={value} className={itemClass}>
+            <AccordionHeader className={headerClass}>
+              <AccordionTrigger className={triggerClass}>{value}</AccordionTrigger>
+            </AccordionHeader>
+            <AccordionContent className={animated2DContentClass}>
+                <div style={{ padding: 10, background: 'whitesmoke', overflow: 'hidden'}}>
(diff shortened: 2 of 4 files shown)


benoitgrelard (commented):
Hi @imtiazmangerah,

Thanks for the contribution! This is a pretty pragmatic solution to the issue you have raised.
The only thing I find a little odd is that you only really need one or the other variable depending on the orientation you are planning to lay it out.

So as I suggested on Discord, I wonder if we should consider adding the concept of `orientation` in `Collapsible` (and subsequently `Accordion`) potentially like we have in other primitives such as `Tabs` or `Slider`.

We have to think of what there would be to gain I suppose, potentially some functional styles to get you going?

cc @andy-hook, @jjenzz

imtiazmangerah:
Hi @benoitgrelard,

I do agree that this fits in with the concept of `orientation`. I did give this some thought and came up with an example use case where both variables would be required at once (2D Animation).

```js
const open2D = css.keyframes({
  from: {
    width: 0,
    height: 0,
  },
  to: {
    width: 'var(--radix-collapsible-content-width)',
    height: 'var(--radix-collapsible-content-height)',
  },
});

const close2D = css.keyframes({
  from: {
    width: 'var(--radix-collapsible-content-width)',
    height: 'var(--radix-collapsible-content-height)'
  },
  to: {
    width: 0,
    height: 0
  },
});

const animatedContentClass = css({
  overflow: 'hidden',
  '&[data-state="open"]': {
    animation: `${slideDown} 300ms ease-out`,
  },
  '&[data-state="closed"]': {
    animation: `${slideUp} 300ms ease-in`,
  },
});

export const Animated2D = () => {
  return (
    <Collapsible className={rootClass}>
      <CollapsibleTrigger className={triggerClass}>Trigger</CollapsibleTrigger>
      <CollapsibleContent className={animatedWidthContentClass}>
        <div style={{width: '100%', height: 100, background: 'red' }} />
      </CollapsibleContent>
    </Collapsible>
  );
};
```

Just throwing this example into the discussion.

andy-hook:
I’ve been thinking about this recently as well. I’m finding it difficult to see the benefit of tying this to orientation, they are both potentially useful values to have available.

Even if we were to decide that a single value is the most common use case, would we then rename the custom prop to something more generic? (less intuitive to use?) or only expose the relevant prop per orientation? (arbitrarily limiting / no upside?)

> potentially some functional styles to get you going

Did anything jump to mind here?

benoitgrelard:
>>potentially some functional styles to get you going

>Did anything jump to mind here?

I thought potentially we could offer some baked in styles for orientation but then I looked at `Tabs` and we don't. We just have our recommended styles in our stories (and in our demo in the docs), so probably not then.

packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -144,8 +144,10 @@ const CollapsibleContentImpl = React.forwardRef<
   const composedRefs = useComposedRefs(forwardedRef, ref);
   const heightRef = React.useRef<number | undefined>(0);
   const height = heightRef.current;
+  const widthRef = React.useRef<number | undefined>(0);
+  const width = widthRef.current;
   // when opening we want it to immediately open to retrieve dimensions
-  // when closing we delay `present` to retreive dimensions before closing
+  // when closing we delay `present` to retrieve dimensions before closing
```
jjenzz, line 150: 👀 👍

jjenzz (commented):
I have one small request to update a now outdated comment in the code but otherwise this looks like a good solution to me 👍 

I agree that tying it to orientation is probably unnecessarily limiting especially when this isn't causing an extra reflow or anything. We're just exposing data we were already getting.

benoitgrelard:
> I agree that tying it to orientation is probably unnecessarily limiting especially when this isn't causing an extra reflow or anything. We're just exposing data we were already getting.

I was more thinking in the context of `Accordion` as it's built on top of `Collapsible` so we need to think of it as a whole.

I agree that `Collapsible` probably doesn't need the orientation concept because it doesn't have any accessibility or functionality based on it. `Accordion` however does with regards to arrow keys, so it might make sense there.

But yeah I think for this change itself I'm happy with it, we can then take it from there for `Accordion` changes.

benoitgrelard:
> > `Accordion` however does with regards to arrow keys, so it might make sense there.
> 
> Seems `Accordion` just needs an update as part of this PR to add the following?
> 
> ```js
> ['--radix-accordion-content-width' as any]: 'var(--radix-collapsible-content-width)',
> ```
> 
> I'm not following why the animated properties need to be tied to orientation since people might want it to scale out in both directions like @imtiazmangerah's example, regardless of orientation.

Not saying the animated property is tied to orientation but rather that it just brought to my attention that we have just assumed that Accordion is always vertical (with regards to arrow keys at least). The animation part is solved like you suggested by re-aliasing the new css property in Accordion too.

jjenzz:
> but rather that it just brought to my attention that we have just assumed that Accordion is always vertical (with regards to arrow keys at least)

Oooh, yes, we should create a ticket for that!

packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -184,6 +187,7 @@ const CollapsibleContentImpl = React.forwardRef<
       ref={composedRefs}
       style={{
         [`--radix-collapsible-content-height` as any]: height ? `${height}px` : undefined,
+        [`--radix-collapsible-content-width` as any]: width ? `${width}px` : undefined,
```
jjenzz, line 190: Since this component is re-composed in `Accordion`, would you mind re-aliasing it there as part of this PR too please?

```jsx
['--radix-accordion-content-width' as any]: 'var(--radix-collapsible-content-width)',
```

imtiazmangerah: Sure - makes sense, added a usage of this in stories and the animation does look decent :smile:

jjenzz (approved):

jjenzz:
@imtiazmangerah would you mind running `yarn bump:check -i` in your branch, select `patch` for the packages it lists with changes and then commit the file it generates please?

That should pass the failing versions check.

PR #883, opened 2021-09-30
[Presence] Prevent premature unmount

andy-hook:
Raising for discussion.

`Presence` has trouble getting the latest animation value if they are updated via state inside an effect, it will retrieve a stale `animationName` when `present` changes and fall through to the `UNMOUNT`.

Here is an example, you can see the unmount animation doesn't occur:
https://codesandbox.io/s/exciting-rgb-j04y0

This change pushes the animation retrieval logic to next tick so that the most recent animation value can be checked.

### DIFF
--- a/.yarn/versions/215c2811.yml
+++ b/.yarn/versions/215c2811.yml
@@ -1,0 +1,18 @@
+releases:
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-collapsible": patch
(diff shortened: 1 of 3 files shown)


benoitgrelard:
Random sense check question, I'm sure the example is contrived, but generally our `data-state` should be computed from state? So I'm curious why we'd need a separate state that gets synced in an effect.

andy-hook:
> I'm curious why we'd need a separate state that gets synced in an effect

I had a use case where selectively setting the state was easier to reason about compared to deriving from state directly in the render.

I thought about it some more though and I don't think this change is a good idea, even my case is better served by an alternative.

PR #884, opened 2021-10-01
[Popover][Dialog] Expose `RemoveScroll` props

jjenzz:
PR for discussion since we had a request on Discord to allow pinch zoom functionality for `Dialog` when modal. They referenced an `allowPinchZoom` prop that ChakraUI expose from `RemoveScroll` so I've done the same here.

### DIFF
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -178,12 +178,17 @@ DialogContent.displayName = CONTENT_NAME;
 
 /* -----------------------------------------------------------------------------------------------*/
 
+type RemoveScrollProps = React.ComponentProps<typeof RemoveScroll>;
 type DialogContentTypeElement = DialogContentImplElement;
 interface DialogContentTypeProps
-  extends Omit<DialogContentImplProps, 'trapFocus' | 'disableOutsidePointerEvents'> {}
+  extends Omit<DialogContentImplProps, 'trapFocus' | 'disableOutsidePointerEvents'> {
+  allowPinchZoom?: RemoveScrollProps['allowPinchZoom'];
+  removeScrollbar?: RemoveScrollProps['removeScrollBar'];
+}
 
 const DialogContentModal = React.forwardRef<DialogContentTypeElement, DialogContentTypeProps>(
   (props, forwardedRef) => {
+    const { allowPinchZoom, removeScrollbar = true, ...contentModalProps } = props;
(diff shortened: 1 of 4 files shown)


packages/react/dialog/src/Dialog.tsx
```diff
@@ -178,12 +178,17 @@ DialogContent.displayName = CONTENT_NAME;
 
 /* -----------------------------------------------------------------------------------------------*/
 
+type RemoveScrollProps = React.ComponentProps<typeof RemoveScroll>;
 type DialogContentTypeElement = DialogContentImplElement;
 interface DialogContentTypeProps
-  extends Omit<DialogContentImplProps, 'trapFocus' | 'disableOutsidePointerEvents'> {}
+  extends Omit<DialogContentImplProps, 'trapFocus' | 'disableOutsidePointerEvents'> {
+  allowPinchZoom?: RemoveScrollProps['allowPinchZoom'];
+  removeScrollbar?: RemoveScrollProps['removeScrollBar'];
```
jjenzz, line 186: Thoughts on remapping these to different names? I wasn't sure if we were cool with these or not.

benoitgrelard (commented):
I guess I wonder:
- if this pinch to zoom thing is turned off, there must be a drawback to turning it on?
- `removeScrollbar`, what is this one? is it just visually the scrollbar or it's equivalent to not rendering `RemoveScroll` like we do when we don't want to prevent scroll (non-modal for example)

jjenzz:
@benoitgrelard docs for these props is here https://github.com/theKashey/react-remove-scroll#usage

jjenzz:
> I'm curious to what extent this is a problem as well, I would have thought we would want this to be the default?

I would have thought so too but I avoided going too deep into this since my thinking with this PR is to release it as a stopgap. We already have plans to replace the `RemoveScroll` library. 

Without this, things are completely inaccessible to people with visual impairments that rely on zooming. With this they may have some other trade offs but we can document that they reference the `RemoveScroll` docs (do their own research) and raise any related issues there.

jjenzz:
> Yeah I think it would make sense to add to all the comps that use `RemoveScroll`.

Sweet :+1: will do.

> Why do we also need to expose `removeScrollBar`? My vote would be to only expose the prop that fixes an issue people have asked about.

I presume this will fix the jumping issues we had raised by Colm/Pedro (can't remember who) in the docs when there is a fixed header so the header gets wider when the scrollbar is removed.

I can test this theory to be sure but seems we have indeed had an issue for this.

PR #888, opened 2021-10-01
[All] Add scoped contexts

jjenzz:


packages/core/rect/src/observeElementRect.ts
```diff
@@ -25,6 +25,7 @@ function observeElementRect(
   } else {
     // only add a callback for this element as it's already observed
     observedData.callbacks.push(callback);
+    callback(elementToObserve.getBoundingClientRect());
```
jjenzz, line 28: If a new observer was added to an element that was already observed, it wouldn't get the initial bounding rect the same way the first observer does. This fixes that and was needed to get `DropdownMenu` + `Tooltip` composition working.

jjenzz:
I'll pick up the Chromatic issues on Monday

PR #890, opened 2021-10-05
[Slot] Improvements

jjenzz:
Pedro noticed that `<Slot className="foo" />` was [giving TS errors](https://codesandbox.io/s/nostalgic-kapitsa-fspsx?file=/src/App.tsx). 

This fixes that but also removes the `event.defaultPrevented` checks we do in `Slot` when composing events. I noticed this when debugging something from Discord #help yday and thought I'd tackle it at the same time since I don't think the `Slot` should determine whether events can be prevented or not. That should be up to the components that are being composed together. I've added a story to show that events can still be prevented correctly.

### DIFF
--- a/packages/react/slot/src/Slot.stories.tsx
+++ b/packages/react/slot/src/Slot.stories.tsx
@@ -15,6 +15,44 @@ export const WithSlottable = () => (
   </SlotWithSlottable>
 );
 
+export const WithComposedEvents = () => (
+  <>
+    <h1>Should log both</h1>
+    <SlotWithPreventableEvent>
+      <button onClick={() => console.log('button click')}>Slot event not prevented</button>
+    </SlotWithPreventableEvent>
+
+    <h1>Should log "button click"</h1>
+    <SlotWithPreventableEvent>
+      <button
+        onClick={(event) => {
+          console.log('button click');
+          event.preventDefault();
+        }}
+      >
+        Slot event prevented
+      </button>
+    </SlotWithPreventableEvent>
+
+    <h1>Should log both</h1>
+    <SlotWithoutPreventableEvent>
+      <button onClick={() => console.log('button click')}>Slot event not prevented</button>
+    </SlotWithoutPreventableEvent>
+
+    <h1>Should log both</h1>
+    <SlotWithoutPreventableEvent>
+      <button
+        onClick={(event) => {
+          console.log('button click');
+          event.preventDefault();
+        }}
+      >
+        Slot event prevented
+      </button>
+    </SlotWithoutPreventableEvent>
+  </>
+);
+
 export const Chromatic = () => (
   <>
     <h1>Without Slottable</h1>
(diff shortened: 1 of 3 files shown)


packages/react/slot/src/Slot.stories.tsx
```diff
@@ -15,6 +15,44 @@ export const WithSlottable = () => (
...
+    <SlotWithoutPreventableEvent>
+      <button
+        onClick={(event) => {
+          console.log('button click');
+          event.preventDefault();
+        }}
+      >
+        Slot event prevented
+      </button>
+    </SlotWithoutPreventableEvent>
```
jjenzz, line 52: This would have broken previously because the `Slot` would have prevented the composed event.

packages/react/slot/src/Slot.tsx
```diff
@@ -101,18 +104,6 @@ function mergeProps(slotProps: AnyProps, childProps: AnyProps) {
   return { ...slotProps, ...overrideProps };
 }
 
-type EventHandler = (...args: unknown[]) => unknown;
-
-function composeHandlers(childHandler?: EventHandler, slotHandler?: EventHandler) {
-  return function handleEvent(...args) {
-    childHandler?.(...args);
-    const isDefaultPreventedEvent = args[0] instanceof Event && args[0].defaultPrevented;
```
jjenzz, line 109: This was always returning `false` anyway because it's a SyntheticEvents so `args[0] instance of Event` was always `false`.

packages/react/slot/src/Slot.stories.tsx
```diff
@@ -15,6 +15,44 @@ export const WithSlottable = () => (
...
+  </>
```
jjenzz, line 53: If we fixed [this issue](https://github.com/radix-ui/primitives/pull/890/files#r722090555) then this story wouldn't have worked which is why I think the `Slot` shouldn't handle it.

packages/react/slot/src/Slot.tsx
```diff
@@ -5,7 +5,7 @@ import { composeRefs } from '@radix-ui/react-compose-refs';
  * Slot
  * -----------------------------------------------------------------------------------------------*/
 
-interface SlotProps {
+interface SlotProps extends React.HTMLAttributes<HTMLElement> {
```
benoitgrelard, line 8: I'm confused about this, wasn't that the whole point that DOM props should go on the DOM element (child of `Slot`) rather than `Slot`?

packages/react/slot/src/Slot.tsx
```diff
@@ -88,7 +88,10 @@ function mergeProps(slotProps: AnyProps, childProps: AnyProps) {
     const isHandler = /^on[A-Z]/.test(propName);
     // if it's a handler, modify the override by composing the base handler
     if (isHandler) {
-      overrideProps[propName] = composeHandlers(childPropValue, slotPropValue);
+      overrideProps[propName] = (...args: unknown[]) => {
```
benoitgrelard, line 91: > That should be up to the components that are being composed together. I've added a story to show that events can still be prevented correctly.

Yeah I think you're actually right about that, and with that new signature, it does take care of the original issue someone had with these event compositions not receiving all params when it's not a native event handler.

:+1:

packages/react/slot/src/Slot.tsx
benoitgrelard: I see, that's interesting. Do we foresee a lot of these use cases? Generally `Slot` wouldn't be used publicly but I can see in this case what he was trying to do. I'm just curious why we didn't want to do it before with `as` cause technically it was the same thing, you might have want your internal `Button` to have some attributes on it.

jjenzz: Either way, we need this same change for our internal use. The only reason we didn't notice this was because I did `const Comp: any = asChild ? Slot : node` in `Primitive` 🙈 . If I made that all a bit stricter it probably would have caught this same issue for us.

We didn't do it before with `as` because we were trying to encourage them to put the props on the thing that actually renders the DOM element (which would be the child). They could have put them on the `Slot` and it would have worked fine... it was just more of a DX decision I believe.

benoitgrelard (approved):

Issue #869, opened 2021-09-17
Concerned about Dropdown Menu's modal accessibility


stowball:
## Bug report

By default, Dropdown Menu uses a modal mode (`modal=true`), which prevents keyboard and/or screen-reader users from ever accessing content outside `<DropdownMenuContent>` without first selecting a `<DropdownMenuItem>`.

You state that you adhere to the [Menu Button WAI-ARIA design pattern](https://www.w3.org/TR/wai-aria-practices-1.2/#menubutton), however, this pattern never mentions that a user should be trapped within the dropdown, and indeed, their examples freely let users not interact with an open dialog should they wish.

FWIW, setting `modal={false}` on your Codesandbox demo also seems to have no effect either, but I could be doing something wrong…

### Current Behavior

1. A keyboard user tabs on to the button
2. They press Enter or Space to open the menu
3. They decide they don't want to interact with it
4. They can't (shift) tab out of it to other content on the page, nor to the `<DropdownMenuTrigger>` to close it again with Enter or Space (unless they press Escape)

The same goes for screen reader users who navigate with their previous and next commands. They are also trapped within the dropdown.

### Expected behavior

Dropdowns are non-modal and are allowed to be ignored when opened. Users should be free to access content before or after it without closing it, or to be able to reach the `<DropdownMenuTrigger>` to close it, just as mouse users can.

### Reproducible example

[Your CodeSandbox Demo](https://codesandbox.io/s/0tr78?module=App.js)

### Suggested solution

Remove the modal implementation and remove the `<DropdownMenuContent>` from being within a portal so a keyboard user's (shift) tab key and a screen reader user's previous and next commands actually take them to the relevant previous and next elements in the page.

I understand that not using a portal may present other issues (such as being clipped by `overflow` or `z-index`), however, I don't know of any other way to maintain the document order for screen-reader users using previous and next (tab can be hacked around).

### Additional context

The [Menu WAI-ARIA pattern](https://www.w3.org/TR/wai-aria-practices-1.2/#menu) explicitly states:

> Tab: Moves focus to the next element in the tab sequence, and if the item that had focus is not in a menubar, closes its menu and all open parent menu containers.
> Shift + Tab: Moves focus to the previous element in the tab sequence, and if the item that had focus is not in a menubar, closes its menu and all open parent menu containers.

### Your environment

| Software         | Name(s)                       | Version      |
| ---------------- | ----------------------------- | ------------ |
[...]

PR #900, opened 2021-10-08
[Popover] Disable looping if not `modal` or `portalled`

jjenzz:
One step closer to solving https://github.com/radix-ui/primitives/issues/869

Easier to review with hidden whitespace changes.

I've just moved `FocusScope` from the `Impl` component up into the relevant modal/non-modal components and configured them independently there instead.

https://user-images.githubusercontent.com/175330/136603852-1aaca404-fb12-4aca-a154-56dd4f6bcfa8.mp4

### DIFF
--- a/.yarn/versions/e76b55f5.yml
+++ b/.yarn/versions/e76b55f5.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-popover": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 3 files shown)


andy-hook:
@jjenzz Given [recent experiments](https://github.com/radix-ui/primitives/issues/869#issuecomment-948561442) with `aria-owns` and focus handling, are we still happy with this as a stop gap solution? your PoC looks like a good contender to solve this issue across all comps that use a `modal` prop.

jjenzz:
@andy-hook Oh yeah! Interesting... let me tweak this and see how it behaves 👍

jjenzz:
Ah, yeah I remember why I just raised that PoC and didn't implement it... it's a bigger job to get that PoC into a solid change across components.

I'll close this PR for now and we can revisit based off that PoC when we have time 👍

Issue #898, opened 2021-10-08
[HoverCard] keyboard support


joshwcomeau:
### Overview

As I understand it, HoverCard is intended to provide a visual preview for sighted users, and for that reason, it's mouse-only, with no keyboard controls. I'd love to see it extended to cover focus as well, either by default or through a boolean `triggerOnFocus` prop.

### Who does this impact? Who is this for?

This would help sighted users who don't use a pointer device. I was in this category myself for a while in 2020, using voice dictation to navigate through the web. It'd help anyone who uses non-pointer input devices, like keyboards or switches.

### Additional context

It may be that this component really isn't meant for keyboard users / there's another component that works better for them (maybe `Popover`?). In this case, I think maybe the docs could help clarify this, and make clear exactly when each component should/shouldn't be used.

The way it's currently framed, I worry developers will lump two very-different groups of people together: unsighted folks who use a screen-reader, and sighted folks who use a non-pointer input device. And I think it's worth making sure that we optimize the experience for both of them!

Overall I'm a big fan of this library 💯 Thanks for all the great work!

benoitgrelard:
Hey Josh, thanks for creating this!

`HoverCard` is by design not for screen reader users, because its interaction model doesn't allow for proper focus management and enunciation of the content inside. Also it should only be used to show redundant information that is already available behind the link (trigger).

That said I think you are making a really good point about the fact that right now it is quite possibly too binary (sighted users vs. screen reader users). There's definitely an argument to be made for a sighted person that cannot use the pointer so cannot hover.

However, given that the information behind it should be redundant, and not primary info that isn't available any other way, there's also an argument for not having something pop-up everywhere that you move your focus to. I imagine for example if I browsed twitter or github with my keyboard tabbing through, and the "user" card popped up everytime I landed on a name, etc that might also be annoying.

In that sense, an approach where the devs could decide if they want to enable that or not might be good.

> In this case, I think maybe the docs could help clarify this, and make clear exactly when each component should/shouldn't be used.

This is something we have been talking about a lot, and we need to get better at in our docs for sure.

jjenzz:
> But since these elements aren't meant to contain focusable elements

As far as I understand it, `HoverCard` can contain focusable links, however, if we kept the card out of the tab sequence (portalled) and just rendered it on link focus, would that cover your expectation?

[...]

PR #902, opened 2021-10-08
[HoverCard] Open on focus

jjenzz:
Continuing this discussion / closes https://github.com/radix-ui/primitives/issues/898

It was a quick change so thought I'd do it while I had a spare minute. Happy to close PR if we don't think this is wise.

My thoughts:

- I agree with @joshwcomeau "trigger on focus" is useful
- It can be prevented if others feel differently `onFocus={event => event.preventDefault()}`
- I have added escape keypress handling to pass the [dismissable success criterion](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#dismissable) 
- It seems we should do this to pass the [content on hover or focus success criterion](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#additional-notes) which says:
	> Content which can be triggered via pointer hover should also be able to be triggered by keyboard focus

https://user-images.githubusercontent.com/175330/136620051-da222c47-1c4a-424c-a154-42a715804602.mp4

### DIFF
--- a/packages/react/hover-card/src/HoverCard.tsx
+++ b/packages/react/hover-card/src/HoverCard.tsx
@@ -165,7 +165,7 @@ const HoverCardContentImpl = React.forwardRef<
 >((props, forwardedRef) => {
   const { portalled = true, ...contentProps } = props;
   const context = useHoverCardContext(CONTENT_NAME);
-  const ref = React.useRef<React.ElementRef<typeof PopperPrimitive.Content>>(null);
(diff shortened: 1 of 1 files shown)


andy-hook:
> So do we all agree to add the timer on focus too as discussed [here](https://github.com/radix-ui/primitives/issues/898#issuecomment-939352788)?

I agree with @jjenzz, simply keeping the same delays as for pointers makes the most sense atm. I think that's how this change is currently working anyway.

PR #925, opened 2021-10-20
[HoverCard] Add DismissableLayer

### LINKED ISSUE #920: Hover Card Content can't be focused inside Dialog
## Bug report
The Hover Card content closes inside the Dialog even though is hovered.

### Current Behavior

https://user-images.githubusercontent.com/2750170/137478803-315976ce-ba27-4329-80ba-153631263160.mov

### Expected behavior

The hover card content should remain open when hovered

### Reproducible example

https://codesandbox.io/s/confident-albattani-ckgqr?file=/App.js

### Suggested solution

Screenshot 2021-10-15 at 14 20 09

The dialog and the hover card are in 2 different portals, so maybe there needs to be a way to specify that the hover card portal should be on top.

### Your environment

| Radix Package(s) |  react-dialog, react-hover-card       |    0.1.1     |
| React            | n/a     | 17        |
| Browser          | Chrome        | 94       |
| Operating System |  mac  OS     |  Big Sur       |

jjenzz:
Fixes #920.

I've also added a change to `FocusScope` here to discuss. The change prevents auto focusing anchors when content opens. It feels like this isn't really the intended behaviour in most cases but need you guys to put me in line here if there is good reason to include anchors.

https://user-images.githubusercontent.com/175330/138120932-7b8a7c5a-3618-473a-89f7-74e072025245.mp4

### DIFF
--- a/packages/react/focus-scope/src/FocusScope.tsx
+++ b/packages/react/focus-scope/src/FocusScope.tsx
@@ -111,7 +111,7 @@ const FocusScope = React.forwardRef<FocusScopeElement, FocusScopeProps>((props, 
         container.addEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
         container.dispatchEvent(mountEvent);
         if (!mountEvent.defaultPrevented) {
-          focusFirst(getTabbableCandidates(container), { select: true });
+          focusFirst(removeLinks(getTabbableCandidates(container)), { select: true });
           if (document.activeElement === previouslyFocusedElement) {
             focus(container);
           }
@@ -219,8 +219,7 @@ function getTabbableCandidates(container: HTMLElement) {
   const walker = document.createTreeWalker(container, NodeFilter.SHOW_ELEMENT, {
     acceptNode: (node: any) => {
       const isHiddenInput = node.tagName === 'INPUT' && node.type === 'hidden';
-      const isAnchor = node.tagName === 'A';
-      if (node.disabled || node.hidden || isHiddenInput || isAnchor) return NodeFilter.FILTER_SKIP;
+      if (node.disabled || node.hidden || isHiddenInput) return NodeFilter.FILTER_SKIP;
       // `.tabIndex` is not the same as the `tabindex` attribute. It works on the
       // runtime's understanding of tabbability, so this automatically accounts
       // for any kind of element that could be tabbed to.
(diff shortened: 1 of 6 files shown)


packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -201,17 +215,27 @@ const HoverCardContentImpl = React.forwardRef<
...
-        style={{
-          ...contentProps.style,
-          // re-namespace exposed content custom property
-          ['--radix-hover-card-content-transform-origin' as any]:
-            'var(--radix-popper-transform-origin)',
-        }}
-      />
+      <DismissableLayer
+        asChild
+        disableOutsidePointerEvents={disableOutsidePointerEvents}
```
benoitgrelard, line 220: I don't think all props should be exposed. It should be equivalent to those exposed by non-modal primitives I believe.

jjenzz: Looks like its only `disableOutsidePointerEvents` that I needed to remove. Let me know if I've misunderstood the non-modal code though.

benoitgrelard: I think `onDismiss` also is never exposed.

jjenzz: I had already omitted that one. I've switched it now though to manually expose the relevant ones instead of omitting the unavailable ones. It doesn't make sense to `extend` a type if the props aren't spread on that component.

packages/react/focus-scope/src/FocusScope.tsx
```diff
@@ -219,7 +219,8 @@ function getTabbableCandidates(container: HTMLElement) {
   const walker = document.createTreeWalker(container, NodeFilter.SHOW_ELEMENT, {
     acceptNode: (node: any) => {
       const isHiddenInput = node.tagName === 'INPUT' && node.type === 'hidden';
-      if (node.disabled || node.hidden || isHiddenInput) return NodeFilter.FILTER_SKIP;
+      const isAnchor = node.tagName === 'A';
```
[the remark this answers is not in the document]
benoitgrelard: I've had a chat with @jjenzz about this yesterday and we agreed to go ahead with the idea of not auto focusing links on mount as it probably is a better default for the 90% to skip links.

That said, I believe this is the wrong place for the checks as the `getTabbableCandidates` gets used for different purposes like finding out edges to loop focus around (in which case links should be included).

 It looks like it would rather need to be done precisely where we handle the autofocus on mount event, something like this (obviously the actual solution can be anything):

```diff
if (!mountEvent.defaultPrevented) {
-  focusFirst(getTabbableCandidates(container), { select: true });
+  focusFirst(removeLinks(getTabbableCandidates(container)), { select: true });
  if (document.activeElement === previouslyFocusedElement) {
    focus(container);
  }
}
```

benoitgrelard (commented):
All good from me apart from the tweak required around removing links from auto focus on mount.

andy-hook (approved):

Issue #760, opened 2021-07-02
[Portal] Explore giving more control for layering


jjenzz:
We currently apply the maximum `z-index` on portals and have had a few requests to allow more control here. This ticket is to explore ways of allowing that for a few reasons:

- People have requested it
- Can only override **all portal instances** via the `data-radix-portal` attribute atm
- Other Radix components provide **an API** to control things (consistency)
- The `data-radix-portal` attribute is an inconsistent escape hatch that we removed from other comps

An initial discussion on this topic provided a couple ideas:

- A `layer`/`zIndex` prop for portalled parts, e.g. `<DialogContent layer={3}>`
- ~A `Portal` **provider** part exposed per component that provides its props to the internal portal, e.g. `<DialogPortal style={{ zIndex: 3 }}><DialogContent /></DialogPortal>`. The provider suggestion is because we need control internally of how/where/when the portal is rendered.~
- Expose a `Portal` part for people to wrap when they want portals, e.g. `Dialog.Portal`. [Examples](https://github.com/radix-ui/primitives/issues/760#issuecomment-872929264).
- Use `Slot` internally on the `Portal`. This would reduce elements rendered as an additional win but it wouldn't work with:
	- `Popper` based parts because popper wraps its own additional `div` for positioning
	- Parts that use https://github.com/theKashey/react-remove-scroll because portal needs to wrap the `RemoveScroll` part for it to work, and `RemoveScroll` isn't slot-aware.
- Any more for any more?

benoitgrelard:
This is great! Thanks!

One other reason why this is tricky is because most (if not all) of these components have `Presence` and the concept of a `XxxImpl` and the `Portal` needs to be inside the `XxxImpl` rather than inside the `XxxContent` to avoid it being rendered at all times.

jjenzz:
> and the `Portal` needs to be inside the `XxxImpl` rather than inside the `XxxContent` to avoid it being rendered at all times.

On this, I was just about to ask why we need the provider thing also. Because, like all parts we expose, we can control when/how they render still and if they're wrapping it around `DialogContent` then it will always be around `RemoveScroll` which solves that ordering issue.

Pseudo-code:

```jsx
const DialogPortal = (props) => {
	return <Presence present={forceMount || context.open}><Portal {...props} /></Presence>
}

const DialogContent = (props) => {
	const forceMount = context.isInsidePortal || props.forceMount;
[...]

PR #936, opened 2021-10-26
[Dialog][AlertDialog] Add portal part

jjenzz:
------------------

------------------

First instalment for #760

Vlad is struggling with the inability to portal dialogs into a custom container on the new primitives website. This should unblock him while we find time to roll this out to other primitives.

### Notes for docs

- New `Portal` parts for `Dialog` and `AlertDialog`
- Component parts will not portal if not wrapped in portal part
- Portal part doesn't render a DOM node, need to apply position and `z-index` to `Overlay` and `Content`
- Part has `container` prop which must be `HTMLElement` so document example of this
    ```jsx
    const [container, setContainer] = React.useState<HTMLDivElement | null>(null);
    return (
        <>
          <Dialog.Portal container={container} />
          <div ref={setContainer} />
        </>
    );
    ```

### DIFF
--- a/packages/react/portal/src/Portal.stories.tsx
+++ b/packages/react/portal/src/Portal.stories.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Portal } from './Portal';
+import { Portal, UnstablePortal } from './Portal';
 
 export default { title: 'Components/Portal' };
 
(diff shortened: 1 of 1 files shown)


packages/react/portal/src/Portal.tsx
```diff
@@ -66,13 +66,31 @@ const Portal = React.forwardRef<PortalElement, PortalProps>((props, forwardedRef
...
+interface UnstablePortalProps {
+  container?: HTMLElement | null;
+}
+
+const UnstablePortal: React.FC<UnstablePortalProps> = (props) => {
+  const { container = globalThis?.document?.body, children } = props;
+  return container ? ReactDOM.createPortal(<>{children}</>, container) : null;
+};
+
+Portal.displayName = UNSTABLE_PORTAL_NAME;
```
csantos-nydig, line 84: `UnstablePortal.displayName = UNSTABLE_PORTAL_NAME;`

😄

jjenzz: 😅  good spot! Fancy creating a new PR?

PR #939, opened 2021-10-28
[RadioGroup] Prevent radio activation on manual focus

### LINKED ISSUE #938: [RadioGroup] Focusing a radio item causes the radio group value to change as if that item was clicked
First things first: Radix is **wonderful**. Thank you for creating it!

---

### Current Behavior

Programmatically focusing an `Item` in a Radix radio group causes that `Item` to become checked.

### Expected behavior

Programmatically focusing an `Item` should not cause it to become checked.

### Reproducible example

https://codesandbox.io/s/festive-goldstine-sjt3o?file=/App.js

To repro: Click the button and see that the checked `Item` changes (unexpectedly... I'm pretty sure).

### Suggested solution

I'm not sure

### Your environment

| Radix Package(s) |    @radix-ui/react-radio-group     |     0.1.1    |
| Browser          |    Chrome     |     Version 96.0.4664.18 (Official Build) beta (arm64)    |
| npm/yarn         |    whatever codesandbox is using     |         |
| Operating System |    macOS     |     12.0.1    |

jjenzz:
Fixes #938 

-------------

-------------

### DIFF
--- a/packages/react/radio-group/src/RadioGroup.tsx
+++ b/packages/react/radio-group/src/RadioGroup.tsx
@@ -12,6 +12,8 @@ import { Radio, RadioIndicator, createRadioScope } from './Radio';
 import type * as Radix from '@radix-ui/react-primitive';
 import type { Scope } from '@radix-ui/react-context';
 
+const ARROW_KEYS = ['ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight'];
+
 /* -------------------------------------------------------------------------------------------------
  * RadioGroup
  * -----------------------------------------------------------------------------------------------*/
@@ -28,7 +30,6 @@ const useRadioScope = createRadioScope();
 type RadioGroupContextValue = {
   name?: string;
   required: boolean;
-  groupRef: React.RefObject<RadioGroupElement>;
   value?: string;
   onValueChange(value: string): void;
 };
@@ -68,8 +69,6 @@ const RadioGroup = React.forwardRef<RadioGroupElement, RadioGroupProps>(
     const labelId = useLabelContext();
     const labelledBy = ariaLabelledby || labelId;
     const rovingFocusGroupScope = useRovingFocusGroupScope(__scopeRadioGroup);
-    const ref = React.useRef<RadioGroupElement>(null);
-    const composedRefs = useComposedRefs(forwardedRef, ref);
     const [value, setValue] = useControllableState({
       prop: valueProp,
       defaultProp: defaultValue,
@@ -81,7 +80,6 @@ const RadioGroup = React.forwardRef<RadioGroupElement, RadioGroupProps>(
         scope={__scopeRadioGroup}
         name={name}
         required={required}
-        groupRef={ref}
         value={value}
         onValueChange={setValue}
       >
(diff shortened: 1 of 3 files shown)


jjenzz:
@benoitgrelard this is ready for a second look whenever you're free 🙏

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -146,7 +165,7 @@ const RadioGroupItem = React.forwardRef<RadioGroupItemElement, RadioGroupItemPro
              * We do this imperatively instead of updating `context.value` because changing via
              * state would not trigger change events (e.g. when in a form).
              */
-            if (context.value !== undefined) ref.current?.click();
+            if (isArrowKeyPressedRef.current) ref.current?.click();
```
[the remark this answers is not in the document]
benoitgrelard: I don't think it's worth abstracting anything yet really. We can always bring that logic in `RovingFocusGroup` later if we need to.

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -146,7 +165,7 @@ const RadioGroupItem = React.forwardRef<RadioGroupItemElement, RadioGroupItemPro
              * We do this imperatively instead of updating `context.value` because changing via
              * state would not trigger change events (e.g. when in a form).
              */
```
benoitgrelard, line 167: I wonder if we need to fledge this comment to explain why we do the arrow check?

benoitgrelard (approved):
Pragmatic solution, I like it! 👍

packages/react/radio-group/src/RadioGroup.tsx
jjenzz: Does this line cover it? https://github.com/radix-ui/primitives/pull/939/files#diff-39138edb5652b34ca612c34d47986e2e8eb2c97315f178b7fe237b28e0a2aeeaR164

I thought it implied arrow keypress since that's what roving focus does but let me know if you'd prefer something more explicit here.

benoitgrelard: Sorry I can't see where this is pointing to for some reason…

jjenzz: Ah it's because GitHub collapses that line ...
image

benoitgrelard: To be honest the second half of that comment confuses me now 😀
The first half is correct, although I feel like it could be clearer now that we've added the arrow keys check.

Perhaps something like:

```
Our `RovingFocusGroup` will focus item as we navigate with arrow keys.
We need to make that "check" that item when that's the case.
However, programatic focus should not "check" the focused item.
```

jjenzz: The second part of the comment is explaining why we imperatively call `.click()` to check it instead of setting `context.value` to the current items value (declarative). I think we still need that information.

Does this work?

```jsx
Our `RovingFocusGroup` will focus the radio when navigating with arrow keys and 
we need to "check" it in that case. We click it to "check" it (instead of updating 
`context.value`) so that the radio change event fires.
```

Issue #823, opened 2021-08-16
New component: HoverMenu


andy-hook:
# HoverMenu

An easy to use website navigation pattern with out of the box accessibility and flexible, open components to support a variety of layouts. Useful for large website navigation styles and list or feature heavy marketing.

## Examples

https://user-images.githubusercontent.com/11708259/129596534-5bf460cf-0817-418c-b1f5-495c78924970.mp4

https://user-images.githubusercontent.com/11708259/132385616-37bc6618-d489-4480-b317-4fde35767a76.mp4

https://user-images.githubusercontent.com/11708259/132385755-56ed5c2d-a8fb-44cd-8b3c-8095e0a0c5a8.mp4

https://user-images.githubusercontent.com/11708259/129596571-3027a2af-8d4a-464f-81b2-747fe7acdc2c.mp4

## Requirements

- Support single links alongside hoverable "menus" in top level nav list
- Support pointer grace when moving from item to menu
- Full keyboard control, close on escape, tab to move between menus, arrow up / down to navigate items within
- Full touch support
- Support "tiered" menus, either via nesting or additional part
- Provides `data-state` helper hooks for animating content from left to right / right to left, depending on cursor origin, visibility etc.
- Provide an active item "highlight" part that can be easily animated to move between items
- Provide custom properties with useful values to assist with animation

## Differences to `MenuBar`
`MenuBar` is a [defined pattern](https://www.w3.org/TR/wai-aria-practices-1.1/#menu) and its usage is primarily for application functionality. `HoverMenu` is intended for use as website navigation which among other things, brings with it the need for additional presentation flexibility.

While accessibility needs are similar, most websites have slightly different styles of layout depending on branding, marketing goals and target audience, because of this we can’t be too prescriptive on how this primitive lays out the content and we need a way to provide solid accessibility without compromising the types of layouts consumers may want to achieve.

## Multi tier menu support
Modern UX best practice for site structure is to [favour flatter hierarchies where possible](https://www.nngroup.com/articles/flat-vs-deep-hierarchy/) however, there are cases where it's useful to provide another tier of Hoverable menus.

https://user-images.githubusercontent.com/11708259/132385956-e2bb2818-d9c6-44c6-b413-638007d80af3.mp4

https://user-images.githubusercontent.com/11708259/132385867-e6fca6de-202e-4a65-834f-80bcef5349e8.mp4

For these cases it will be useful to be able to either nest these menus within eachother, or provide an additional part.

## Complex animation
[...]

PR #946, opened 2021-10-31
Popper core: Expose translateX and translateY as variables

buzinas:
### Description

Hi, Modulz team! First, thanks for this excellent library.

I'm currently working on a project that I need to control the popover a bit more than this package needs. I made this proposed change on `index.module.js` and I'm using `patch-package` to keep them over builds etc, but I thought that I could submit it to upstream and maybe get it merged.

This change just exposes internal `translateX` and `translateY` that can then be used down the tree.

### DIFF
--- a/packages/core/popper/src/popper.ts
+++ b/packages/core/popper/src/popper.ts
@@ -311,6 +311,8 @@ function getPlacementStylesForPoint(point: Point): CSS.Properties {
     minWidth: 'max-content',
     willChange: 'transform',
     transform: `translate3d(${x}px, ${y}px, 0)`,
+    '--radix-popper-translate-x': `${x}px`,
+    '--radix-popper-translate-y': `${y}px`,
   };
 }
 


jjenzz:
Hi @buzinas  👋

Our popper package is not meant for public consumption so it only contains boilerplate/logic that is required to support our primitives. I'm curious what your use-case is here and why you've needed to resort to using our private utilities?

benoitgrelard:
> Thanks for the quick response, @jjenzz.
> 
> I'm using `@radix-ui/react-dropdown-menu` and `@radix-ui/react-popover`, both use `@radix-ui/react-popper` under the hood. They have an `align` prop that you can pass `start`, `center`, `end`, but I need a bit more control than that.
> 
> My specific use case is that, sometimes, I need to open a popover that starts at the bottom of the page (slide pop-up), or at the right of the page (sidebar) etc. I could create more `align` variants, but it seemed that exposing the position as CSS properties is just easier and lets the user smoothly control the position when they need.
> 
> I'm using `patch-package` for that, but thought that wouldn't hurt to open a PR upstream.

Hey @buzinas, I'm not sure I fully understand what you're doing. Would you be able to create a quick demo of what you are doing in codesandbox?

buzinas:
@benoitgrelard 

Normal popover, with `align="start"`: ✅ 

image

"Full" popover that I'm assuming it's impossible to reach with `radix-ui/react-popover` today: 🚫 

image

With the proposed change, I can just do something like:

```js
{
  marginLeft: 'calc(var(--radix-popper-translate-x) * -1)',
}
```

This is just an example, but there are many things I can control when I have the popover position available.

benoitgrelard:
Not sure if you saw but there are some offset props that you can use to offset a particular axis.
`sideOffset` and `alignOffset`. In your case I think you'd be after the `alignOffset`.

From the docs:
image

That said, it's interesting looking at your UI, it looks like perhaps `Popover` would not necessarily be the right component for what you're building (a navigation).

@andy-hook is currently building a new navigation primitive which looks like it would serve your needs much better.

buzinas:
Yeah, I'm using `alignOffset` in some other parts of the application, but since I don't know what the offset is without knowing where the component is located (and I don't want to use a `ref` + effect for that), it was why I decided to expose the position as a CSS property that I can then use.

I understand that my use case isn't very usual though.

benoitgrelard:
Hi @buzinas, as you say I think the use case is a bit unusual, and as I pointed out probably not the right primitive for this even in terms of accessibility.

I will close the issue for now but feel free to keep your patch and we will ping you when our navigation/hover menu primitive is out which I think is more what you are after.

Thanks!

andy-hook:
@buzinas – Just to add more context regarding the navigation primitive, we have a rough outline [here](https://github.com/radix-ui/primitives/issues/823). As @benoitgrelard mentioned this looks to be a good candidate for your use case. We'll keep you in mind after we've released.

PR #950, opened 2021-11-02
[Toolbar] Remove `toolbaritem` role

### LINKED ISSUE #949: [Toolbar] @axe-core/react flags role="toolbaritem" as invalid
### Current Behavior

image

### Expected behavior

should not flag any warnings

### Reproducible example

https://codesandbox.io/s/bold-blackwell-6lmwx?file=/src/App.tsx

### Suggested solution

should prob be radio for ToggleGroup = single and checkbox for ToggleGroup = multi?

### Additional context

not sure if all the toolbar items are valid, might be worth testing them all

### Your environment
see CSB

jjenzz:
Fixes #949

There doesn't seem to be such a role in the WAI ARIA docs and their toolbar examples don't include this either. It looks suspiciously like a bad find+replace after copy/pasting the menu/menuitem pattern.

Does anyone know of a reason why this should be kept?

### DIFF
--- a/.yarn/versions/ca283b8d.yml
+++ b/.yarn/versions/ca283b8d.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-toolbar": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


andy-hook (approved):
I've never seen this before either, seems like a pasta snafu as far as I can tell.

Issue #958, opened 2021-11-06
[Accordion] TypeError when defaultValue isn't an array


hoop71:
### Current Behavior

Passing a string to `defaultValue` when `type='multiple'` throws a TypeError when that items is closed. 

Corresponding code: https://github.com/radix-ui/primitives/blob/main/packages/react/accordion/src/Accordion.tsx#L166

https://www.loom.com/share/fba9c67b7b2e4008b06420c9509150f9

![image](https://user-images.githubusercontent.com/25015764/140594712-df0b0e0b-edf2-48da-9e04-88dba769d103.png)

### Expected behavior

Code is resilient enough to handle this type of user error, or prevents the mistake altogether.

### Reproducible example

[CodeSandbox Repro](https://codesandbox.io/s/radix-accordion-bug-c505z?file=/App.js:3002-3010)

### Suggested solution

**Quick solution**: Check for the type of the prop and make sure it doesn't get passed.  
```
diff --git a/packages/react/accordion/src/Accordion.tsx b/packages/react/accordion/src/Accordion.tsx
index 89101cdf..8c56d9b8 100644
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -42,7 +42,12 @@ const Accordion = React.forwardRef<AccordionElement, AccordionSingleProps | Acco
     }
 
     if (type === 'multiple') {
-      const multipleProps = accordionProps as AccordionImplMultipleProps;
+      const multipleProps = (
+        Array.isArray(accordionProps?.defaultValue)
+          ? accordionProps
+          : { ...accordionProps, defaultValue: [] }
+      ) as AccordionMultipleProps;
+
       return <AccordionImplMultiple {...multipleProps} ref={forwardedRef} />;
     }
```

Would also be nice to give the user feedback on what the problem is. It's not immediately obvious what the problem is.

### Additional context

This library is awesome! Keep up the great work. 🚀 

- Can the `single` and `multiple` share an API where `defaultValue: string[]`. This would allow some simplification of logic and doesn't require the user to make a decision and can remove duplicative-ish props.

<img width="759" alt="Screen Shot 2021-11-05 at 8 43 22 PM" src="https://user-images.githubusercontent.com/25015764/140595285-1765fa3c-fd00-4b4f-8aa4-f28365d7272f.png">

[...]

PR #959, opened 2021-11-06
[Accordion] Patch for TypeError on mismatched prop types when `type='multiple'` and `defaultValue='string'`

hoop71:
### Description

Patch for #958 . Details in issue. Checks `defaultValue` for type to prevent TypeError

### DIFF
--- a/.yarn/versions/5ad189b9.yml
+++ b/.yarn/versions/5ad189b9.yml
@@ -1,0 +1,3 @@
+releases:
+  "@radix-ui/react-accordion": patch
+  primitives: patch
--- a/packages/react/accordion/src/Accordion.test.tsx
+++ b/packages/react/accordion/src/Accordion.test.tsx
@@ -216,6 +216,32 @@ describe('given a multiple Accordion', () => {
   });
 });
 
+describe('given a multiple Accordion with mismatchProps', () => {
+  let handleValueChange: jest.Mock;
+  let rendered: RenderResult;
+
+  beforeEach(() => {
+    handleValueChange = jest.fn();
+    rendered = render(
+      // @ts-ignore-next-line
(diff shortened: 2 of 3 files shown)


jjenzz:
Agreed, I don't think we should silently "fix" things for people who provide an incorrect API. We should guide them to fix their incorrect implementation.

Since this is an issue for non-TS users only, my vote would be something like:

```jsx
Accordion.propTypes = {
  type(props, propName, componentName) {
    const value = props.value || props.defaultValue; 
    if (props.type === 'multiple' && value && !Array.isArray(value)) {
      return new Error('Invalid prop \`type\` of value \`${props.type}\` supplied to \`${componentName}\`. Expects \`single\` when `defaultValue` or `value` prop is type \`string\`.
    }
  },
};
```

jjenzz:
> my brain sees it kinda backwards from that, that instead of validating the type, it would be validating the `defaultValue or `value`

I thought that too initially but then you have two separate prop type validators, this seems less boilerplate.

```jsx
Accordion.propTypes = {
  value(props) {
    if(props.type === 'multiple' && typeof props.value === 'string') {
      // error
    }
    if(props.type === 'single' && Array.isArray(props.value)) {
      // error
    }
  },
  defaultValue(props) {
    if(props.type === 'multiple' && typeof props.value === 'string') {
      // error
    }
    if(props.type === 'single' && Array.isArray(props.value)) {
      // error
    }
  },
};
```

Obviously we can abstract the validator into a validator function but the alternative I provided avoids abstraction and achieves the same thing.

hoop71:
All works for me. Thanks for quick response and sharing the thinking of the library on this minor issue. Changes pushed!

hoop71:
> Sorry for the delay, thanks for this @hoop71 !

Glad to contribute!

PR #963, opened 2021-11-08
[Dialog][AlertDialog] Enable outer scrolling dialog pattern

### LINKED ISSUE #887: Scrollable Dialog
## Feature request
With the current Dialog API the only achievable scrolling behaviour when you have a dialog with longer content is to scroll inside the actual dialog, scrolling on the whole container is not possible.

### Overview

Make the Dialog API configurable in some way to allow outer scroll.

- [Chakra UI](https://chakra-ui.com/docs/overlay/modal#modal-overflow-behavior) has a config for the Dialog if you want to scroll to be inside or outside
- [Bootstrap](https://getbootstrap.com/docs/5.1/components/modal/#scrolling-long-content)

I think scrolling in dialogs is quite common so I think it impacts a lot of users of this library.

### Additional context

I've tried to implement this outer scroll behaviour with the current API, but the issue is that clicking on the overlay doesn't work because of how the elements are layered, even with `pointer-events: none` on the content wrapper https://codesandbox.io/s/late-forest-oefp7?file=/App.js

benoitgrelard:
Fixes #887

I'll explain the thoughts I had and what I did here:

## Problem
The main issue left with being able to create an outer scrolling dialog was that we have `RemoveScroll` in `DialogContent`, therefore making the overlay non-scrollable too.

## Thoughts and solution
- The overlay is only used when it's a modal dialog, which incidentally is when we want to remove scroll (sure enough, we were doing this in `DialogContentModal`)
- So I moved `RemoveScroll` inside `DialogOverlayImpl`.
- The only remaining issue was that we also disable outside pointer events from `DialogContentModal`, so that was still preventing scrolling the overlay, for that I specifically re-enable them on the overlay.
- It also required moving the `allowPinchZoom` prop as it was previously on `DialogContent`. I remember almost raising this in the previous PR review that I thought perhaps it should live at the root as it's kinda an implementation detail that it was on the "content". I guess this solidified that feeling, we could have moved the prop to the overlay but it would definitely feel odd there I think.

---

Notes for docs:
- Make sure the `allowPinchZoom` prop is documented on the right part (`Root`).

### DIFF
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -9,7 +9,6 @@ import { FocusScope } from '@radix-ui/react-focus-scope';
 import { UnstablePortal } from '@radix-ui/react-portal';
 import { Presence } from '@radix-ui/react-presence';
 import { Primitive } from '@radix-ui/react-primitive';
-import { Slot } from '@radix-ui/react-slot';
 import { useFocusGuards } from '@radix-ui/react-focus-guards';
 import { RemoveScroll } from 'react-remove-scroll';
 import { hideOthers } from 'aria-hidden';
@@ -177,7 +176,7 @@ const DialogOverlayImpl = React.forwardRef<DialogOverlayImplElement, DialogOverl
     const { __scopeDialog, ...overlayProps } = props;
     const context = useDialogContext(OVERLAY_NAME, __scopeDialog);
     return (
-      <RemoveScroll as={Slot} allowPinchZoom={context.allowPinchZoom}>
+      <RemoveScroll forwardProps allowPinchZoom={context.allowPinchZoom}>
         <Primitive.div
           data-state={getState(context.open)}
           {...overlayProps}
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -35,15 +35,21 @@ type DialogContextValue = {
   onOpenChange(open: boolean): void;
   onOpenToggle(): void;
   modal: boolean;
+  allowPinchZoom: DialogProps['allowPinchZoom'];
 };
 
 const [DialogProvider, useDialogContext] = createDialogContext<DialogContextValue>(DIALOG_NAME);
 
+type RemoveScrollProps = React.ComponentProps<typeof RemoveScroll>;
 interface DialogProps {
   open?: boolean;
   defaultOpen?: boolean;
   onOpenChange?(open: boolean): void;
   modal?: boolean;
+  /**
+   * @see https://github.com/theKashey/react-remove-scroll#usage
+   */
+  allowPinchZoom?: RemoveScrollProps['allowPinchZoom'];
   children?: React.ReactNode;
 }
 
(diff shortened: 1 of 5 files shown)


packages/react/dialog/src/Dialog.tsx
```diff
@@ -169,7 +177,15 @@ const DialogOverlayImpl = React.forwardRef<DialogOverlayImplElement, DialogOverl
     const { __scopeDialog, ...overlayProps } = props;
     const context = useDialogContext(OVERLAY_NAME, __scopeDialog);
     return (
-      <Primitive.div data-state={getState(context.open)} {...overlayProps} ref={forwardedRef} />
+      <RemoveScroll as={Slot} allowPinchZoom={context.allowPinchZoom}>
```
jjenzz, line 180: While we are making changes here... I noticed recently that `RemoveScroll` already has a `forwardProps` prop that does what `Slot` does. Soooo, we could remove the `Slot` dependency in this package if we used that instead.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -35,15 +35,21 @@ type DialogContextValue = {
...
+type RemoveScrollProps = React.ComponentProps<typeof RemoveScroll>;
 interface DialogProps {
   open?: boolean;
   defaultOpen?: boolean;
   onOpenChange?(open: boolean): void;
   modal?: boolean;
+  /**
+   * @see https://github.com/theKashey/react-remove-scroll#usage
+   */
+  allowPinchZoom?: RemoveScrollProps['allowPinchZoom'];
```
jjenzz, line 52: I wonder if we should move this prop in the other components that support it too then?

packages/react/dialog/src/Dialog.tsx
```diff
@@ -169,7 +177,15 @@ const DialogOverlayImpl = React.forwardRef<DialogOverlayImplElement, DialogOverl
     const { __scopeDialog, ...overlayProps } = props;
     const context = useDialogContext(OVERLAY_NAME, __scopeDialog);
     return (
-      <Primitive.div data-state={getState(context.open)} {...overlayProps} ref={forwardedRef} />
+      <RemoveScroll as={Slot} allowPinchZoom={context.allowPinchZoom}>
+        <Primitive.div
+          data-state={getState(context.open)}
+          {...overlayProps}
+          ref={forwardedRef}
+          // We re-enable pointerEvents on the Overlay to enable scrolling
```
jjenzz, line 185: This comment tripped me up because I thought:

> But `RemoveScroll` should only disable pointer events outside of overlay so why do we need to re-enable them?

.... then I remembered your comment in the PR description. Would you mind if we made that a little clearer somehow? Something like:

```suggestion
          // We re-enable pointer-events prevented by `Dialog.Content` to allow scrolling the overlay.
```

packages/react/dialog/src/Dialog.tsx
benoitgrelard: Ha cool, I'll try that then.

packages/react/dialog/src/Dialog.tsx
benoitgrelard: Yeah, probably in a different PR.

jjenzz (approved):

PR #965, opened 2021-11-09
[ToggleGroup] Fix `orientation` a11y

### LINKED ISSUE #964: [RovingFocusGroup] Incorrectly applies `aria-orientation` to some roles
Originally [reported in discord](https://discord.com/channels/752614004387610674/803656530259738674/907358269339738173) when using `ToggleGroup`'s `orientation` prop.

The only roles that `aria-orientation` should be supplied to are as follows:
https://www.w3.org/TR/wai-aria/#aria-orientation

jjenzz:
Fixes #964

I've removed the `aria-orientation` from `RovingFocusGroup` and applied it in the relevant components instead.

### DIFF
--- a/.yarn/versions/e0af1fb6.yml
+++ b/.yarn/versions/e0af1fb6.yml
@@ -1,0 +1,12 @@
+releases:
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-radio-group": patch
+  "@radix-ui/react-roving-focus": patch
+  "@radix-ui/react-tabs": patch
+  "@radix-ui/react-toggle-group": patch
+  "@radix-ui/react-toolbar": patch
+
+declined:
+  - primitives
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -596,8 +596,9 @@ const MenuContentImpl = React.forwardRef<MenuContentImplElement, MenuContentImpl
                 >
                   <PopperPrimitive.Content
                     role="menu"
-                    dir={context.dir}
+                    aria-orientation="vertical"
                     data-state={getOpenState(context.open)}
+                    dir={context.dir}
                     {...popperScope}
                     {...contentProps}
                     ref={composedRefs}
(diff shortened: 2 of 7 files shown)


.yarn/versions/e0af1fb6.yml
```diff
@@ -0,0 +1,10 @@
+releases:
+  "@radix-ui/react-roving-focus": patch
+  "@radix-ui/react-toggle-group": patch
+
+declined:
+  - primitives
+  - "@radix-ui/react-menu"
+  - "@radix-ui/react-radio-group"
+  - "@radix-ui/react-tabs"
+  - "@radix-ui/react-toolbar"
```
jjenzz, line 10: Declined these because functionality wise they're identical to before.

packages/react/toggle-group/src/ToggleGroup.tsx
```diff
@@ -198,7 +198,7 @@ const ToggleGroupImpl = React.forwardRef<ToggleGroupImplElement, ToggleGroupImpl
       disabled = false,
       rovingFocus = true,
       orientation,
-      dir = 'ltr',
+      dir,
```
jjenzz, line 201: I realised in the story it wouldn't inherit direction because of this.

.yarn/versions/e0af1fb6.yml
benoitgrelard: I'm always curious about this, I know the result is the same, but don't they need to be published still so you don't end up with duplicate dependencies potentially?

benoitgrelard (approved):
That seems like a sensible fix for the issue.

.yarn/versions/e0af1fb6.yml
jjenzz: > but don't they need to be published still so you don't end up with duplicate dependencies potentially?

What's your thinking here? Sounds like something I've not considered.

benoitgrelard: I'm just thinking that if some packages aren't updated, then they will be pointing to the previous version of the roving focus package, as opposed to the newly published one, which will be pointing to the newer one. So it means someone would end up with multiple versions of it their dependency tree, so less code sharing, etc.

jjenzz: ah interesting... no harm in `patch`ing them I guess. I'll do that 👍

PR #974, opened 2021-11-11
[DropdownMenu] Make sure a disabled trigger doesn't open

### LINKED ISSUE #972: [DropdownMenu] Disabled trigger doesn't prevent menu from opening
This is because we are using `onPointerDown` instead of `onClick` and not checking `disabled`.

benoitgrelard:
Closes #972

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -164,6 +164,7 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           aria-expanded={context.open ? true : undefined}
           aria-controls={context.open ? context.contentId : undefined}
           data-state={context.open ? 'open' : 'closed'}
+          data-disabled={disabled ? '' : undefined}
           disabled={disabled}
           {...triggerProps}
           ref={composeRefs(forwardedRef, context.triggerRef)}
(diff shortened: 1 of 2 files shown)


packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -164,20 +164,21 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           aria-expanded={context.open ? true : undefined}
           aria-controls={context.open ? context.contentId : undefined}
           data-state={context.open ? 'open' : 'closed'}
+          disabled={disabled}
```
andy-hook, line 167: Should we add the `data-disabled` attr as well?

andy-hook (commented):
Do you remember the reason we went for `pointerDown` here while the popover trigger is `onClick`?

benoitgrelard:
> Do you remember the reason we went for `pointerDown` here while the popover trigger is `onClick`?

Yes, because it enables quick actions in one pointer cycle, rather than 2. I can pointer down on the trigger, move, and pointer up on an item in one cycle. That's what MacOS does too.

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: Good shout, let's add it for consistency.

andy-hook (approved):

PR #975, opened 2021-11-17
[Tabs] Prop comment consistency

Luccasoli:
### Description

updating comment to keep the standard of the other comments :D


jjenzz:
Thanks for this @Luccasoli 💛

PR #977, opened 2021-11-17
[Collapsible][Accordion] Disable animation when open on mount

jjenzz:
Addresses these discussions 

- https://github.com/radix-ui/primitives/discussions/934
- https://github.com/radix-ui/primitives/discussions/976

-------------------

-------------------

### DIFF
--- a/packages/react/accordion/src/Accordion.stories.tsx
+++ b/packages/react/accordion/src/Accordion.stories.tsx
@@ -295,7 +295,7 @@ export const Animated = () => {
       </Accordion>
 
       <h1>Open by default</h1>
-      <Accordion type="single" className={rootClass} value="One">
+      <Accordion type="single" className={rootClass} defaultValue="One">
         {values.map((value) => (
           <AccordionItem key={value} value={value} className={itemClass}>
             <AccordionHeader className={headerClass}>
--- a/packages/react/accordion/src/Accordion.stories.tsx
+++ b/packages/react/accordion/src/Accordion.stories.tsx
@@ -274,7 +274,28 @@ export const Animated = () => {
       </label>
       <br />
       <br />
+      <h1>Closed by default</h1>
       <Accordion type="single" className={rootClass}>
+        {values.map((value) => (
+          <AccordionItem key={value} value={value} className={itemClass}>
+            <AccordionHeader className={headerClass}>
+              <AccordionTrigger className={triggerClass}>{value}</AccordionTrigger>
+            </AccordionHeader>
+            <AccordionContent className={animatedContentClass}>
+              {[...Array(count)].map((_, index) => (
+                <div style={{ padding: 10 }} key={index}>
+                  Per erat orci nostra luctus sociosqu mus risus penatibus, duis elit vulputate
+                  viverra integer ullamcorper congue curabitur sociis, nisi malesuada scelerisque
+                  quam suscipit habitant sed.
+                </div>
+              ))}
+            </AccordionContent>
+          </AccordionItem>
+        ))}
+      </Accordion>
+
+      <h1>Open by default</h1>
+      <Accordion type="single" className={rootClass} value="One">
         {values.map((value) => (
           <AccordionItem key={value} value={value} className={itemClass}>
             <AccordionHeader className={headerClass}>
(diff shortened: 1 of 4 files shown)


packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -201,6 +208,7 @@ const CollapsibleContentImpl = React.forwardRef<
       style={{
         [`--radix-collapsible-content-height` as any]: height ? `${height}px` : undefined,
         [`--radix-collapsible-content-width` as any]: width ? `${width}px` : undefined,
+        animation: context.isAnimationDisabled ? 'none' : undefined,
```
andy-hook, line 211: Bit of an edge but should this be placed after the spread to account for mount animations supplied via `style` prop?

Also wondering if it's worth popping a quick note to explain why we're overriding this.

jjenzz: We always allow consumer props to override our stuff so if they're supplying styles the assumption is that they are purposefully trying to override it I believe.

> Also wondering if it's worth popping a quick note to explain why we're overriding this.

Good shout

packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -201,6 +208,8 @@ const CollapsibleContentImpl = React.forwardRef<
       style={{
         [`--radix-collapsible-content-height` as any]: height ? `${height}px` : undefined,
         [`--radix-collapsible-content-width` as any]: width ? `${width}px` : undefined,
+        // We prevent consumer animation if open when mounted to prevent layout from moving around
```
andy-hook, line 211:
```suggestion
        // Prevent animations running when initially mounted in the open state
        // this avoids unexpected layout shifts
```

Is that any better? 😅

packages/react/collapsible/src/Collapsible.tsx
```diff
@@ -56,13 +57,19 @@ const Collapsible = React.forwardRef<CollapsibleElement, CollapsibleProps>(
       onChange: onOpenChange,
     });
 
+    const isAnimationDisabledRef = React.useRef(open);
```
andy-hook, line 60: nit: Would a naming adjustment better communicate the temporary nature? something like `isAnimationPreventedRef` ? for some reason I associate `disabled` with props / semi permanent state.

packages/react/collapsible/src/Collapsible.tsx
andy-hook: > the assumption is that they are purposefully trying to override it I believe?

Yeah that's probably more likely tbh

packages/react/collapsible/src/Collapsible.tsx
jjenzz: perfect 👌

jjenzz: Tweaked it a little (shorter) but it's the same gist 😛

packages/react/accordion/src/Accordion.stories.tsx
```diff
@@ -292,6 +293,26 @@ export const Animated = () => {
           </AccordionItem>
         ))}
       </Accordion>
+
+      <h1>Open by default</h1>
+      <Accordion type="single" className={rootClass} value="One">
```
benoitgrelard, line 298: nit: 
```suggestion
      <Accordion type="single" className={rootClass} defaultValue="One">
```

to bring it inline with the collapsible story so we can test subsequent open animations.

benoitgrelard (approved):

packages/react/collapsible/src/Collapsible.tsx
andy-hook: Best of both 😄

andy-hook (approved):

packages/react/accordion/src/Accordion.stories.tsx
jjenzz: Ah yes, good shout 👀

khairulhaaziq:
hi @jjenzz does this still work? im finding out that the dialog still animates with defaultOpen, is there an extensive document how I can prevent open animation on defaultOpen? thanks!

Issue #980, opened 2021-11-17
Trigger actions with spacebar (in addition to return)


DaniGuardiola:
### Overview

Some buttons like dialog or menu triggers work on `click` and on `return` key. I'd like `spacebar` to work too, as it does in other design systems.

- The vanilla `<button>` element, along with a few other input elements, does this: http://fyvr.net/a11y/spacebar-action.html
- Additional info on vanilla elements: https://www.last-child.com/keyboard-accessibility-with-the-space-bar/

### Examples in other libraries

- `useButton` (and `usePress`) from `react-aria` does this. See this blog post for more info: https://react-spectrum.adobe.com/blog/building-a-button-part-1.html 

### Who does this impact? Who is this for?

- End users, who can use `spacebar` to trigger buttons.
- Developers who have a `Button` component that work with `spacebar`. Since your users will be used to having `spacebar` as an option to trigger the action, they will be disappointing when it doesn't work to open a dialog, menu, etc.
- Additionally, some teams (like mine) use something like `usePress` to get consistent pressed states (A.K.A. a JS replacement for `:active`), which shows the pressed state when the button is focused and the spacebar is pressed. Those styles are misleading if it doesn't actually work.

### Additional context

n/a

jjenzz:
Would you mind listing which components you think this applies to? A quick test of the components in our [radix-ui.com](https://radix-ui.com) docs shows that the aforementioned components do already support `space` keypress.

I can see we bind to `onKeyDown` instead of `onClick` for `DropdownMenu.Trigger`, `Tooltip.Trigger` and `Tabs.Trigger` so perhaps you are having some interoperability issues there? I'm not sure why we chose to use that event (@benoitgrelard might know more) but space key is supported as far as I can tell.

Regardless, I'd personally prefer that we simplified some of our logic by ensuring all triggers are `button`s (e.g. `Tabs.Trigger`) and use `onClick` if possible. It's been on my mind for some time.

DaniGuardiola:
I'll take another look later when I'm on the computer, but looking at the code I suspect it might be related to the use of `" "` in the key down handler. I've frequently seen code declaring both that and something like `"spacebar"`, so it might be a cross-browser issue?

Anyway, seems like with the onClick it should be okay.

[...]

PR #981, opened 2021-11-17
[DropdownMenu][Tooltip][Tabs] Use button with `onClick` for trigger

jjenzz:
[This discussion](https://github.com/radix-ui/primitives/issues/980) reminded me of something that has been bothering me for some time so thought I'd propose this PR before I head off for the day to discuss whenever.

### Notes for docs

- Breaking change for tabs. Need to apply `all: unset` or similar to `Tabs.Trigger`.

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -168,6 +168,7 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           disabled={disabled}
           {...triggerProps}
           ref={composeRefs(forwardedRef, context.triggerRef)}
+          onClick={composeEventHandlers(props.onClick, () => context.onOpenChange(true))}
           onPointerDown={composeEventHandlers(props.onPointerDown, (event) => {
             // only call handler if it's the left button (mousedown gets triggered by all mouse buttons)
             // but not when the control key is pressed (avoiding MacOS right click)
@@ -178,8 +179,8 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
               context.onOpenToggle();
             }
           })}
-          onKeyDown={composeEventHandlers(props.onKeyDown, (event: React.KeyboardEvent) => {
-            if (!disabled && [' ', 'Enter', 'ArrowDown'].includes(event.key)) {
+          onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
+            if (!disabled && event.key === 'ArrowDown') {
               event.preventDefault();
               context.onOpenChange(true);
             }
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -168,6 +168,8 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           disabled={disabled}
           {...triggerProps}
           ref={composeRefs(forwardedRef, context.triggerRef)}
+          // Handle anything that the browser considers a click for the element type if
+          // not using pointer e.g. Space keyup and Enter keydown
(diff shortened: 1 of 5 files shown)


packages/react/tabs/src/Tabs.stories.tsx
```diff
@@ -205,6 +205,7 @@ const listClass = css({
 });
 
 const RECOMMENDED_CSS__TABS__TRIGGER = {
+  all: 'unset' as any,
```
jjenzz, line 208: The `triggerClass` threw a huuuuuge TS error if I didn't cast this 🤔

image

image

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -168,6 +168,7 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           disabled={disabled}
           {...triggerProps}
           ref={composeRefs(forwardedRef, context.triggerRef)}
+          onClick={composeEventHandlers(props.onClick, () => context.onOpenChange(true))}
```
benoitgrelard, line 171: So the idea is to let the `onClick` handle the spacebar because it's native, even though the pointer side of things won't ever execute given we do it in pointer down?

I don't mind that much, but I wonder if it's perhaps more confusing this way (less explicit) as you come back later down and see, oh we're already doing the `onPointerDown`, I can remove the `onClick`.

Thoughts?

packages/react/tabs/src/Tabs.stories.tsx
benoitgrelard: weird…

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -199,11 +199,7 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
           onMouseDown={composeEventHandlers(props.onMouseDown, context.onClose)}
           onFocus={composeEventHandlers(props.onFocus, context.onFocus)}
           onBlur={composeEventHandlers(props.onBlur, context.onClose)}
-          onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
-            if (event.key === 'Enter' || event.key === ' ') {
-              context.onClose();
-            }
-          })}
+          onClick={composeEventHandlers(props.onClick, context.onClose)}
```
benoitgrelard, line 202: Same thoughts on this one, given that it really only deals with the keyboard stuff.

benoitgrelard (approved):
All quite sensible to me, just left a comment/question for some thoughts for us to decide how we feel.

packages/react/dropdown-menu/src/DropdownMenu.tsx
jjenzz: `onClick` handles "anything that is considered a click" so for buttons that includes space *keyup* and enter *keydown* which we had wrong before (we did keydown for both). I think it's better to just let the browser handle it to ensure we're not losing any of these little details.

I agree that it is a bit confusing now though that we also have the `onPointerDown` logic, so maybe I should add a comment?

benoitgrelard: Yeah I think adding a comment in a few places would be great.

jjenzz: Done 🙂  Let me know if you want me to reword anything 🙏

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -168,6 +168,8 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           disabled={disabled}
           {...triggerProps}
           ref={composeRefs(forwardedRef, context.triggerRef)}
+          // Handle anything that the browser considers a click for the element type if
```
benoitgrelard, line 171: ♥️

benoitgrelard (approved):

PR #984, opened 2021-11-18
[Accordion] Use `Collection` util

jjenzz:
Adds support for asynchronously rendered items.

### DIFF
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { createContextScope } from '@radix-ui/react-context';
+import { createCollection } from '@radix-ui/react-collection';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { useControllableState } from '@radix-ui/react-use-controllable-state';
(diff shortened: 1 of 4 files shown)


packages/react/accordion/src/Accordion.tsx
```diff
@@ -222,21 +230,22 @@ interface AccordionImplProps extends PrimitiveDivProps {
...
     const accordionRef = React.useRef<AccordionImplElement>(null);
     const composedRefs = useComposedRefs(accordionRef, forwardedRef);
+    const getItems = useCollection(__scopeAccordion);
 
     const handleKeyDown = composeEventHandlers(props.onKeyDown, (event) => {
       const target = event.target as HTMLElement;
       const isAccordionKey = ACCORDION_KEYS.includes(event.key);
+      const triggerCollection = getItems();
 
       if (!isAccordionKey || !isButton(target)) {
```
jjenzz, line 242: I realised we have a pretty nasty assumption here (that the trigger is always a `HTMLButtonElement`). It's probably fine in most cases but if someone used `asChild` to change the node, our keyboard accessibility logic would break. Shall I create a ticket to look into this?

benoitgrelard: What was the `isButton` check for in the first place?
Was it to make sure the event was in fact bubbling from an accordion trigger?
If so, regarding of what tag it is, isn't it better to just check that it's part of the triggers?
It seems it was in fact already handled before by the `if (triggerIndex === -1) return;` check?

I might be completely out of it though 😀

benoitgrelard:
Not sure why but Chromatic seems to be stuck on the previous diff (tabs)

packages/react/accordion/src/Accordion.tsx
jjenzz: Oh yeah... thank god you're here 😅 

I'll remove that `isButton` check as part of this PR while I'm here then.

benoitgrelard (approved):
All good from me apart the weird Chromatic stuff mentioned.

jjenzz:
> All good from me apart the weird Chromatic stuff mentioned.

Oh yeah, weird... I'll just approve it again and hopefully things will straighten out?

PR #995, opened 2021-11-22
[Dialog][AlertDialog] Fix `Dialog` unmount animation

jjenzz:
When I added the new `Dialog.Portal` part, I forgot to portal from inside `Presence` which meant its `context.open` check would rip the dialog parts out before `Presence` could suspend.

~The `Dialog.Portal` just renders a provider instead now so we can read that inside the relevant parts and wrap them in a portal when necessary.~ I map over the `children` and wrap them in `Presence` and `UnstablePortal` instead.

### Notes for docs

- ✨ &nbsp; New `forceMount` prop on `Dialog.Portal`.

### DIFF
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -12,6 +12,7 @@ import { Primitive } from '@radix-ui/react-primitive';
 import { useFocusGuards } from '@radix-ui/react-focus-guards';
 import { RemoveScroll } from 'react-remove-scroll';
 import { hideOthers } from 'aria-hidden';
+import { Slot } from '@radix-ui/react-slot';
 
 import type * as Radix from '@radix-ui/react-primitive';
 import type { Scope } from '@radix-ui/react-context';
@@ -128,13 +129,33 @@ DialogTrigger.displayName = TRIGGER_NAME;
 const PORTAL_NAME = 'DialogPortal';
 
 type PortalProps = React.ComponentPropsWithoutRef<typeof UnstablePortal>;
-interface DialogPortalProps extends PortalProps {}
+type DialogPortalElement = React.ElementRef<typeof UnstablePortal>;
+interface DialogPortalProps extends PortalProps {
+  children?: React.ReactNode;
+  /**
+   * Used to force mounting when more control is needed. Useful when
+   * controlling animation with React animation libraries.
+   */
+  forceMount?: true;
+}
 
-const DialogPortal: React.FC<DialogPortalProps> = (props: ScopedProps<DialogPortalProps>) => {
-  const { __scopeDialog, ...portalProps } = props;
-  const context = useDialogContext(PORTAL_NAME, __scopeDialog);
-  return context.open ? <UnstablePortal {...portalProps} /> : null;
-};
+const DialogPortal = React.forwardRef<DialogPortalElement, DialogPortalProps>(
+  (props: ScopedProps<DialogPortalProps>, forwardedRef) => {
+    const { __scopeDialog, forceMount, children, ...portalProps } = props;
+    const context = useDialogContext(PORTAL_NAME, __scopeDialog);
+    return (
+      <>
+        {React.Children.map(children, (child) => (
+          <Presence present={forceMount || context.open}>
+            <UnstablePortal {...portalProps} asChild ref={forwardedRef}>
+              {child}
+            </UnstablePortal>
+          </Presence>
+        ))}
+      </>
+    );
+  }
+);
 
 DialogPortal.displayName = PORTAL_NAME;
 
@@ -176,7 +197,7 @@ const DialogOverlayImpl = React.forwardRef<DialogOverlayImplElement, DialogOverl
     const { __scopeDialog, ...overlayProps } = props;
     const context = useDialogContext(OVERLAY_NAME, __scopeDialog);
     return (
-      <RemoveScroll forwardProps allowPinchZoom={context.allowPinchZoom}>
+      <RemoveScroll as={Slot} allowPinchZoom={context.allowPinchZoom}>
         <Primitive.div
           data-state={getState(context.open)}
           {...overlayProps}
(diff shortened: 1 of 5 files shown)


packages/react/dialog/src/Dialog.tsx
```diff
@@ -12,6 +12,7 @@ import { Primitive } from '@radix-ui/react-primitive';
 import { useFocusGuards } from '@radix-ui/react-focus-guards';
 import { RemoveScroll } from 'react-remove-scroll';
 import { hideOthers } from 'aria-hidden';
+import { Slot } from '@radix-ui/react-slot';
```
jjenzz, line 15: I realised that the `forwardProps` solution on `RemoveScroll` was breaking unmount animation... no idea why. Our `Slot` seems to combine the two better and things all work.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -127,13 +128,20 @@ DialogTrigger.displayName = TRIGGER_NAME;
...
-  const { __scopeDialog, ...portalProps } = props;
-  const context = useDialogContext(PORTAL_NAME, __scopeDialog);
-  return context.open ? <UnstablePortal {...portalProps} /> : null;
+  const { __scopeDialog, children } = props;
+  return (
+    <DialogPortalProvider scope={__scopeDialog} isInsidePortal>
+      {children}
+    </DialogPortalProvider>
+  );
 };
```
jjenzz, line 145: This has actually got me wondering if we need this part now?

It seems we could just have a `portalled` prop on the `Dialog.Root` part that we pass through context instead because the portal doesn't render any DOM anymore.

benoitgrelard: But then how would we allow custom containers?

jjenzz: Oh yeah, I've broken that part here 😅 forgot about that. I've just updated this to make sure props get forwarded.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -354,39 +372,46 @@ const DialogContentImpl = React.forwardRef<DialogContentImplElement, DialogConte
       ...contentProps
     } = props;
     const context = useDialogContext(CONTENT_NAME, __scopeDialog);
+    const portalContext = useDialogPortalContext(OVERLAY_NAME, __scopeDialog);
     const contentRef = React.useRef<HTMLDivElement>(null);
     const composedRefs = useComposedRefs(forwardedRef, contentRef);
+    const PortalWrapper = portalContext.isInsidePortal
+      ? { Comp: UnstablePortal, props: portalContext.props }
+      : { Comp: React.Fragment, props: undefined };
```
jjenzz, line 380: Hmmm, not sure how we want to handle this?

packages/react/dialog/src/Dialog.tsx
benoitgrelard: Cool, and that's what we used to do before right? `as={Slot}` because they also have an `as`.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -127,13 +128,20 @@ DialogTrigger.displayName = TRIGGER_NAME;
...
 type PortalProps = React.ComponentPropsWithoutRef<typeof UnstablePortal>;
 interface DialogPortalProps extends PortalProps {}
 
 const DialogPortal: React.FC<DialogPortalProps> = (props: ScopedProps<DialogPortalProps>) => {
-  const { __scopeDialog, ...portalProps } = props;
-  const context = useDialogContext(PORTAL_NAME, __scopeDialog);
-  return context.open ? <UnstablePortal {...portalProps} /> : null;
+  const { __scopeDialog, children, ...portalProps } = props;
+  return (
+    <DialogPortalProvider scope={__scopeDialog} props={portalProps} isInsidePortal>
```
benoitgrelard, line 141: Isn't `portalProps` going to be a new reference on every render?

packages/react/dialog/src/Dialog.tsx
```diff
@@ -354,39 +370,44 @@ const DialogContentImpl = React.forwardRef<DialogContentImplElement, DialogConte
...
-            data-state={getState(context.open)}
-            {...contentProps}
-            ref={composedRefs}
-            onDismiss={() => context.onOpenChange(false)}
-          />
-        </FocusScope>
-        {process.env.NODE_ENV === 'development' && <LabelWarning contentRef={contentRef} />}
-      </>
+      <PortalWrapper {...(portalContext.isInsidePortal ? portalContext.props : undefined)}>
+        {/* Prevent nested `Dialog`s from portalling unless they use their own `Dialog.Portal` */}
```
benoitgrelard, line 384: Can you elaborate on this a bit? I'm not sure I follow.

benoitgrelard (commented):
The changes seem to be fine, but I have to admit I don't completely understand what the original issue was and why a fix was needed.

Would be great if you could run me through the issue on a call at some point.

packages/react/dialog/src/Dialog.tsx
jjenzz: Yep 👍

packages/react/dialog/src/Dialog.tsx
jjenzz:
```jsx
<Dialog.Portal>
  <Dialog.Content>

    <Dialog.Root>
      // this one shouldn't portal and it won't be cause I've provided 
      // `isInsidePortal=false` from `Dialog.Content` otherwise it would 
      // have inherited the portal wrapped around `Dialog.Content`.
      <Dialog.Content /> 
    </Dialog.Root>

    <Dialog.Root>
      <Dialog.Portal><Dialog.Content /></Dialog.Portal> // this one should
    </Dialog.Root>

  </Dialog.Content>
</Dialog.Portal>
```

benoitgrelard (approved):

Issue #811, opened 2021-08-02
[Id] SSR – Mismatched server / client ids when running in `StrictMode`


andy-hook:
Related https://github.com/radix-ui/primitives/issues/793

Original report via Discord:

> I'm trying to create a Dialog with IdProvider and I can't understand why SSR doesn't work, I got an error: 
> Prop `aria-controls` did not match. Server: "radix-id-0-1" Client: "radix-id-0-4"

https://codesandbox.io/s/radixui-idprovider-ssr-mismatch-i3o1x

This is reproducible in our current SSR testing app since `StrictMode` was added in  https://github.com/radix-ui/primitives/pull/795

Wrapping `React.StrictMode` or setting `reactStrictMode: true` in `next.config` is equivalent afaik.

andy-hook:
Thought some more about this and then it struck me that of course this would be a problem, the implementation is proven non-deterministic in concurrent mode which is why `StrictMode` is having a cry about it and incrementing twice. 

So then I thought about how might we solve for this and I kept ending up back in the same place – wouldn't it be easier to just not do this on the server in the first place? I did some digging and turns out Reach has been down this rabbit hole already and written a nice piece of documentation on it.

https://github.com/reach/reach-ui/blob/develop/packages/auto-id/src/index.tsx

While this is a warning and not a "today" problem, it will become a bigger issue in the future so the question is whether we'd want to take a similar approach here / now.

Worth noting that I did check in on the react-aria implementation and it continues to suffer from the same issue.

csantos1113:
Any updates on this? We just ran into the same issue

andy-hook:
> Any updates on this?

Not since my follow up, one workaround is disabling `StrictMode` for now. The problem is only visible in development and production builds should behave as expected.

This issue isn't exclusive to Radix, it exists in many other `id` implementations in popular libraries. I need to speak with the team and see how we feel given the options available to us.

csantos1113:
> This issue isn't exclusive to Radix, it exists in many other id implementations in popular libraries. I need to speak with the team and see how we feel given the options available to us.

Yeah, I created an issue in react-spectrum (https://github.com/adobe/react-spectrum/issues/2231) for react-aria to see how they react and what ideas they might have as well.

Thanks for the quick response @andy-hook - and yes, I did exactly that 😅 strictMode disabled for now

Clee681:
Seeing the same issue was a `NextJS` user

julianbenegas:
[...]

PR #1006, opened 2021-11-30
[All] Remove `IdProvider` and introduce `React.useId`

jjenzz:
For older versions of React we are falling back to client-only ids. This removes the overhead of an `IdProvider` given that projects will likely migrate to React 18 eventually.

Fixes https://github.com/radix-ui/primitives/issues/811


packages/react/id/src/id.tsx
```diff
@@ -1,45 +1,16 @@
...
-    () => deterministicId || `radix-id-${context.prefix}-${++context.current}`,
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [deterministicId]
-  );
+  const [id, setId] = React.useState<string | undefined>(useReactId());
+  // React versions older than 18 will have client-side ids only.
+  useLayoutEffect(() => {
+    if (!deterministicId) setId((reactId) => reactId || String(count++));
+  }, [deterministicId]);
+  return deterministicId || `radix-${id}`.toLowerCase();
```
jjenzz, line 13: Although the `toLowerCase` looks nicer, I'm not sure that we should do this as they may be generating different cases on purpose for uniqueness given that `document.getElementById('fOo') !== document.getElementById('foo')`.

Thoughts?

packages/react/id/src/id.tsx
```diff
@@ -1,45 +1,16 @@
...
-
-  return React.useMemo(
-    () => deterministicId || `radix-id-${context.prefix}-${++context.current}`,
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [deterministicId]
-  );
+  const [id, setId] = React.useState<string | undefined>(useReactId());
+  // React versions older than 18 will have client-side ids only.
+  useLayoutEffect(() => {
+    if (!deterministicId) setId((reactId) => reactId || String(count++));
```
benoitgrelard, line 11: nit, I think that would be more accurate technically, if React were ever to return `0` as an id for example, safer not to assume anything.

```suggestion
    if (!deterministicId) setId((reactId) => reactId ?? String(count++));
```

packages/react/id/src/id.tsx
benoitgrelard, line 13: Curious why the `.toLowerCase` is needed?

jjenzz, line 13: Although the `toLowerCase` looks nicer, I'm not sure that we should do this. React might be generating different cases on purpose for uniqueness given that `document.getElementById('fOo') !== document.getElementById('foo')`.

Thoughts?

benoitgrelard: Ha, you answered [my question](https://github.com/radix-ui/primitives/pull/1006#discussion_r759442769) before I asked 😀 

Yeah I think we should leave it alone.

packages/react/id/src/id.tsx
jjenzz: That wouldn't matter since `React.useId` return type is string I believe, so `'0'` would still be fine but I'll change regardless for clarity though.

packages/react/id/src/id.tsx
jjenzz: https://github.com/radix-ui/primitives/pull/1006#discussion_r759442899

Happy to remove. It just looked nicer cos React generates uppercase letters sometimes.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -133,10 +134,9 @@ const Tooltip: React.FC<TooltipProps> = (props: ScopedProps<TooltipProps>) => {
...
-  const handleClose = React.useCallback(
-    () => stateMachine.send({ type: 'CLOSE', id: contentId, skipDelayDuration }),
-    [skipDelayDuration, contentId]
-  );
+  const handleClose = useCallbackRef(() => {
+    stateMachine.send({ type: 'CLOSE', id: contentId, skipDelayDuration });
+  });
 
   // send transition if the component unmounts
   React.useEffect(() => () => handleClose(), [handleClose]);
```
jjenzz, line 142: The `handleClose` had a `contentId` dependency which would start as `undefined` so this hook would call its cleanup when the `contentId` was set on client and instantly close the `defaultOpen` tooltips.

This fix solves it for now but I need to create a new `TooltipProvider` as part of this React 18 piece to ensure our state is all managed through React state / Context, so I'll see if I can make any improvements here then as well.

benoitgrelard: Nice catch!

benoitgrelard (approved):

packages/react/tooltip/src/Tooltip.tsx
jjenzz: Took me a while 😅

byF:
@jjenzz I guess this PR invalidates the docs at https://www.radix-ui.com/docs/primitives/overview/server-side-rendering?

benoitgrelard:
> @jjenzz I guess this PR invalidates the docs at https://www.radix-ui.com/docs/primitives/overview/server-side-rendering?

It will yes, it's not released publicly as stable yet, although it is part of the release candidates.

jjenzz:
@byF Thanks for the reminder, I've updated the original PR description to include notes on required docs updates.

andrevenancio:
I've noticed that now there's no need to have this module wrapping my app, but leaving that documentation page online is confusing..

benoitgrelard:
We are in the process of releasing and the docs are almost out.
Sorry about the inconvenience.

packages/react/id/src/id.tsx
```diff
@@ -1,45 +1,16 @@
...
-  const isRootIdProvider = currentContext === defaultIdContext;
-  return (
-    <IdProviderImpl
-      prefix={isRootIdProvider ? 0 : ++currentContext.prefix}
-      current={0}
-      {...props}
-    />
-  );
-};
+const useReactId = (React as any).useId || (() => undefined);
```
polRk, line 4: Do not use `(React as any).useId`, it's break project.

polRk (commented):
Why did you let this happen?

packages/react/id/src/id.tsx
polRk: Popular bundlers try to import useId from React

polRk:
#1023

packages/react/id/src/id.tsx
polRk: #1023

jjenzz: This has been fixed here https://github.com/radix-ui/primitives/pull/1028

PR #1007, opened 2021-12-01
[Tooltip] Avoid managing state outside of React

### LINKED ISSUE #899: [Id][Tooltip] Multiple sibling `IdProviders` cause mirrored interactions
This is quite specific to storybook configurations because you wouldn't typically render more than one `IdProvider` inside an app.

Because of the crossover with storybook and snapshot testing, it makes sense to wrap a provider per story via decorators to stabilise the `id`, however that causes an interesting visual side effect when rendered inline with the [docs addon](https://storybook.js.org/docs/react/writing-docs/docs-page).

The problem is that multiple instances of `IdProvider` increment in sync so there are duplicate identifiers on the page and the global machine just rolls with it.

https://user-images.githubusercontent.com/11708259/136595395-0705ed61-6fba-4804-a2a4-5dcd8c586780.mp4

jjenzz:
Replaces the state machine with a `TooltipProvider` to ensure React 18 support. The `TooltipProvider` isn't really required, it will just ensure instant open when moving between tooltips.

There are a bunch of Github Issues around [tearing](https://github.com/reactwg/react-18/discussions/69) and things in React 18 and tbh, it's not all that clear to me whether tooltip would break in its current form (it seems to work) but, the React team are pretty clear that if you keep state inside React then you will be fine.

Given that the state machine has [proven to be a bit problematic anyway](https://github.com/radix-ui/primitives/issues/899) (reliance on ids), I've removed it in favour of a provider to guarantee support and remove all the state machine utility code that existed purely for this component. 

The reason I believe this will have fixed #899 is because I'm no longer relying on ids to control open/closing of other tooltips on the page. I dispatch a `document` event instead and all tooltips close if that happens. Given that [the `IdProvider` isn't a thing anymore](https://github.com/radix-ui/primitives/pull/1006) though, that's probably also gotten rid of that particular issue.

-----------------------

⚠️  &nbsp; Easier to review with hidden whitespace changes

-----------------------

### Notes for docs

- ✨  &nbsp;New `TooltipProvider` part with `skipDelayDuration` and `delayDuration` props.
- 🔥  &nbsp;Remove `skipDelayDuration` prop from `Tooltip.Root`


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -14,45 +13,114 @@ import { createPopperScope } from '@radix-ui/react-popper';
...
-const stateMachine = createStateMachine(tooltipStateChart);
+const PROVIDER_NAME = 'TooltipProvider';
+const DEFAULT_DELAY_DURATION = 700;
+const TOOLTIP_OPEN = 'tooltip.open';
+
+type ScopedProps<P = {}> = P & { __scopeTooltipProvider?: Scope };
+const [createTooltipContext, createTooltipScope] = createContextScope(PROVIDER_NAME, [
+  createPopperScope,
+]);
+const usePopperScope = createPopperScope();
```
jjenzz, line 33: Moved from here https://github.com/radix-ui/primitives/pull/1007/files#diff-da7e63370e718104438e4866d2cd43b3390689aa2c030321312c0be78602f544L38

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -14,45 +13,114 @@ import { createPopperScope } from '@radix-ui/react-popper';
...
+  onTooltipClose(): void;
+};
+
+const [TooltipProviderProvider, useTooltipProviderContext] =
+  createTooltipContext<TooltipProviderContextValue>(PROVIDER_NAME, {
+    isOpenDelayed: true,
+    delayDuration: DEFAULT_DELAY_DURATION,
+    onTooltipOpen: () => {},
+    onTooltipClose: () => {},
+  });
```
jjenzz, line 48: Supplying defaults so that the `TooltipProvider` isn't required. Without it, tooltips will all have a delay before opening.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -14,45 +13,114 @@ import { createPopperScope } from '@radix-ui/react-popper';
...
+  /**
+   * The duration from when the mouse enters the trigger until the tooltip gets opened.
+   * @defaultValue 700
+   */
+  delayDuration?: number;
+  /**
+   * How much time a user has to enter another trigger without incurring a delay again.
+   * @defaultValue 300
+   */
+  skipDelayDuration?: number;
```
jjenzz, line 60: I copied these props to the provider and removed `skipDelayDuration` from each `Tooltip` because it only makes sense in the context of *multiple* tooltips.

The provider `delayDuration` will be the default but each tooltip can override it with their own.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -61,109 +128,77 @@ interface TooltipProps {
...
         open={open}
         stateAttribute={stateAttribute}
         trigger={trigger}
         onTriggerChange={setTrigger}
-        onFocus={handleFocus}
-        onOpen={handleOpen}
-        onClose={handleClose}
+        onFocus={React.useCallback(() => setOpen(true), [setOpen])}
+        onOpen={React.useCallback(() => {
+          document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN));
```
jjenzz, line 191: I dispatch here instead of within the `TooltipProvider` so that tooltips will still close if another tooltip is opened when there isn't a `TooltipProvider`.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -61,109 +128,81 @@ interface TooltipProps {
...
 
-  // if we're controlling the component
-  // put the state machine in the appropriate state
-  useLayoutEffect(() => {
-    if (openProp === true) {
-      stateMachine.send({ type: 'OPEN', id: contentId });
-    }
-  }, [contentId, openProp]);
+  const handleOpen = React.useCallback(() => {
+    document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN));
```
jjenzz, line 181: I dispatch here instead of within the `TooltipProvider` so that tooltips will still close if another tooltip is opened when there isn't a `TooltipProvider`.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -61,109 +128,80 @@ interface TooltipProps {
...
-    if (openProp === true) {
-      stateMachine.send({ type: 'OPEN', id: contentId });
+  const prevOpen = usePrevious(open);
+  const { onTooltipOpen, onTooltipClose } = context;
+  React.useEffect(() => {
+    if (prevOpen === open) return;
+    // We do this in an effect to ensure that we don't broadcast an
+    // open/close unless open state change was allowed (when controlled).
+    if (open) {
+      document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN));
```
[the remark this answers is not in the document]
jjenzz: This has actually made me wonder if we need a provider at all now 🤔

jjenzz:
Marking as draft because I've spotted a bug 

https://user-images.githubusercontent.com/175330/144286278-8a53847e-f989-461a-8547-52ba6178ace9.mp4

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -61,109 +128,74 @@ interface TooltipProps {
...
-    stateMachine.send({ type: 'CLOSE', id: contentId, skipDelayDuration });
-  });
+  const stateAttribute = React.useMemo(() => {
+    return open ? (wasOpenDelayed ? 'delayed-open' : 'instant-open') : 'closed';
+  }, [wasOpenDelayed, open]);
 
-  // send transition if the component unmounts
-  React.useEffect(() => () => handleClose(), [handleClose]);
+  const handleOpen = React.useCallback(() => {
+    document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN));
```
jjenzz, line 168: I dispatch here instead of within the `TooltipProvider` `onTooltipOpen` handler so that tooltips will still close if another tooltip is opened when there isn't a `TooltipProvider`.

This  actually made me wonder if we need a provider at all 🤔  I thought maybe I could use document events to control open delay but a first pass at it here isn't working. I'll take another look tomorrow but I need **one** `skipDelayTimerRef` across tooltips that they all can clear I think.

jjenzz: I've had another go at this this morning, and I might be able to get it working but the code is a bit tricky to reason about. It also conceptually makes more sense in the provider I think, given that it's a shared delay between multiple tooltips.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -61,109 +128,76 @@ interface TooltipProps {
...
-          setStateAttribute(context.delayed ? 'delayed-open' : 'instant-open');
-        } else {
-          setStateAttribute('closed');
-        }
-      } else {
-        setStateAttribute('closed');
-      }
+  const handleOpen = React.useCallback(() => {
+    setOpen((prevOpen) => {
+      if (!prevOpen) document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN));
```
jjenzz, line 169: I added a comment [here](https://github.com/radix-ui/primitives/pull/1007/files/6d7e52d40713c010280052e6a3ca9163c253dc1e#diff-da7e63370e718104438e4866d2cd43b3390689aa2c030321312c0be78602f544) about this event but since had to move it to fix a bug. I don't want to dispatch open unless it is opening.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -61,109 +128,76 @@ interface TooltipProps {
...
+      return true;
     });
+    onTooltipOpen();
```
jjenzz, line 172: I'm calling `onTooltipOpen` here and `onTooltipClose` further down, regardless of whether the consumer prevents open/close when controlled. I tested how it behaves in that case and obviously the `isOpenDelayed` will change even if it wasn't opened so moving to the next tooltip will instantly open. 

I tried a fix for this with a `React.useEffect` that runs when `open` changes and then fires `onTooltipOpen` / `onTooltipClose`. However, that was too late to cancel the timers in some cases. I didn't invest too much time trying to figure this all out as it feels like an extreme edge-case. My thinking is that nothing breaks and if anyone raises it we can look into it? Thoughts?

jjenzz: Oh, I just tested this exact same use-case in the previous version and it has the same issue so I guess it's clear by now that no one has noticed it except me 😅

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -182,9 +216,9 @@ interface TooltipTriggerProps extends PrimitiveButtonProps {}
 
 const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerProps>(
   (props: ScopedProps<TooltipTriggerProps>, forwardedRef) => {
-    const { __scopeTooltip, ...triggerProps } = props;
-    const context = useTooltipContext(TRIGGER_NAME, __scopeTooltip);
-    const popperScope = usePopperScope(__scopeTooltip);
+    const { __scopeTooltipProvider, ...triggerProps } = props;
```
benoitgrelard, line 219: I'm rather confused by all the renaming here, I'm sure there's a reason but I can't figure it out right now…

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -61,109 +128,76 @@ interface TooltipProps {
...
-  }, [contentId, openProp]);
+  React.useEffect(() => {
+    return () => window.clearTimeout(openTimerRef.current);
+  }, []);
 
   return (
     <PopperPrimitive.Root {...popperScope}>
-      <TooltipProvider
-        scope={__scopeTooltip}
+      <TooltipImplProvider
```
benoitgrelard, line 181: Why is this called `TooltipImplProvider` when there are no `TooltipImpl`?

packages/react/tooltip/src/Tooltip.tsx
colmtuite: Allowing different delays/durations/thresholds was intended. We've seen a bunch of products that have a delay on some tooltips but not delay on others.

benoitgrelard: Ah ok nice, I didn't know. So at least for the opening delay we can still control that per tooltip. Do you think the other is ok as a global? (ie. the time allowed between 2 tooltips)

colmtuite: I think so. I'd even bet it's ok hardcoded.

packages/react/tooltip/src/Tooltip.tsx
jjenzz: I can't call this `TooltipProvider` because there is a `TooltipProvider` part now. Not sure what to name it? The `TooltipProvider` is the provider for all tooltips and this provider is the provider for **each** tooltip.

packages/react/tooltip/src/Tooltip.tsx
jjenzz: Heh, I thought this might get raised. The uppermost component is the `TooltipProvider` now so I renamed it to be scoped by the uppermost. I did similar in `Toast` (it's `__scopeToastProvider`). 

I'm happy to change this back to `__scopeTooltip` and do `__scopeToast` for toast instead but it means that I can't define the scope at the top of the file anymore I end up with "Cannot use before defined/assigned" errors no matter where I put it.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -14,45 +13,113 @@ import { createPopperScope } from '@radix-ui/react-popper';
...
-import { tooltipStateChart } from './tooltipStateChart';
 
 import type * as Radix from '@radix-ui/react-primitive';
 import type { Scope } from '@radix-ui/react-context';
 
+type ScopedProps<P = {}> = P & { __scopeTooltip?: Scope };
+const [createTooltipContext, createTooltipScope] = createContextScope('Tooltip', [
+  createPopperScope,
+]);
+const usePopperScope = createPopperScope();
```
benoitgrelard, line 25: I wonder if these ones should stay closer to where they're used, it's less to do with the definition of the scope, but more with using it in a part that uses a scope. What do you think?

benoitgrelard (approved):
All good from me, just left a small note.

packages/react/tooltip/src/Tooltip.tsx
jjenzz: There are multiple parts that will use this scope though... so do I just define it alongside the first one that uses it?

benoitgrelard: Ah ok, makes sense then. 👍

PR #1011, opened 2021-12-02
[Tooltip] Remove `type="button"`

### LINKED ISSUE #993: [Tooltip] wrapping `input` element defaults its type to `button`
## Bug report

Wrapping a native `input` component without explicit type with a `Tooltip` defaults its type attribute to `'button'`.

This behaviour is not documented, nor should it be expected.

### Current Behavior

Screenshot from demo code reused to wrap native inputs:
image

First input element becomes a button

### Expected behavior

I would expect input wrapped with a `Tooltip` to behave just as if it wasn't wrapped, hence having no type forced if none is set.

### Reproducible example

https://codesandbox.io/s/keen-heisenberg-c43y8

### LINKED ISSUE #1010: [Tooltip] Trigger assumes type="button"
### Current Behavior

https://github.com/radix-ui/primitives/blob/main/packages/react/tooltip/src/Tooltip.tsx#L192

The `TooltipTrigger` assumes a button, but I think it's also common to use tooltips with links. If `asChild={true}` is used with an `<a>` child, then the anchor ends up as `<a type="button" ...>`.

### Expected behavior

I'd expect to be able to use an `<a>` child without it receiving `type="button"`.

### Reproducible example

https://codesandbox.io/s/tooltip-trigger-type-kepl8

### Suggested solution

It may not make sense to pass the `type` attribute through when using `asChild={true}`.

### Your environment

| Radix Package(s) |  tooltip       |  0.1.1       |
| React            | n/a     |  17.0.2       |
| Browser          |  n/a       |  n/a       |
| Assistive tech   |  n/a       |   n/a      |
| Node             | n/a     | n/a        |
| npm/yarn         |  n/a       |  n/a       |
| Operating System |  n/a       | n/a        |

jjenzz:
Closes #1010 
Closes #993

I haven't done this for our other components that add `type="button"` since  I don't believe we have had reports regarding those yet. Not sure what thoughts are here though?

### DIFF
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -189,7 +189,6 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
     return (
       <PopperPrimitive.Anchor asChild {...popperScope}>
         <Primitive.button
-          type="button"
           aria-describedby={context.open ? context.contentId : undefined}
           data-state={context.stateAttribute}
           {...triggerProps}
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -189,6 +189,8 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
     return (
       <PopperPrimitive.Anchor asChild {...popperScope}>
         <Primitive.button
+          // We purposefully avoid adding `type=button` here because tooltip triggers are also
+          // commonly anchors and the anchor `type` attribute signifies MIME type.
           aria-describedby={context.open ? context.contentId : undefined}
           data-state={context.stateAttribute}
           {...triggerProps}
--- a/.yarn/versions/83127137.yml
+++ b/.yarn/versions/83127137.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -189,7 +189,6 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
     return (
       <PopperPrimitive.Anchor asChild {...popperScope}>
         <Primitive.button
-          type="button"
```
benoitgrelard, line 192: With the now singled-out inconsistency, I wonder if we should leave a comment to replace this to explain why we don't do it here? I can imagine one of us in a few months coming back and adding it back by mistake.

What do you think?

benoitgrelard (approved):
>I haven't done this for our other components that add type="button" since I don't believe we have had reports regarding those yet. Not sure what thoughts are here though?

I have been wondering exactly the same thing. Trying to rationalise it, I feel like `Tooltip.Trigger` is the only trigger whose functionality isn't entirely relying on a `click`/`pointerdown` action. That makes me think that all the other triggers where we do `type="button"` have no business being replaced with other HTML elements as it wouldn't really be semantic anymore.

All that to say I think we're ok to do it just here for now and see what comes out on others if anything.

benoitgrelard (approved):

benoitgrelard:
This PR removed `type=button` because a tooltip trigger can be more than a button, it can sometimes be on a link, or even an input of many types.

bel0v:
@benoitgrelard I get that, but that seems to be only the `asChild` use case

PR #1015, opened 2021-12-03
[Slot] Prevent `key` warning

### LINKED ISSUE #1013: [Slot] Triggers a warning in dev mode about keys
### Current Behavior

Current warning is shown in the console:

```
Warning: Each child in a list should have a unique "key" prop.
Check the render method of `Slot`.
```

### Reproducible example

https://codesandbox.io/s/jolly-cdn-5wor8

(don't mind the tailwind classes)

### Suggested solution

I think the problem lies in here, no `key` attribute seems to appear:

https://github.com/radix-ui/primitives/blob/e9f43ed94b53a0c07279fddbe7371bd54a53cfd4/packages/react/slot/src/Slot.tsx#L17-L28

### Your environment

| Radix Package(s) |      tooltip   |     0.1.1    |

jjenzz:
Closes #1013

Admittedly, I don't seem to be able to reproduce the issue but makes sense so I've switched it to use `React.Children.map` to prevent any warnings like that regardless.

### DIFF
--- a/packages/react/slot/src/Slot.tsx
+++ b/packages/react/slot/src/Slot.tsx
@@ -11,12 +11,11 @@ interface SlotProps extends React.HTMLAttributes<HTMLElement> {
 
 const Slot = React.forwardRef<HTMLElement, SlotProps>((props, forwardedRef) => {
   const { children, ...slotProps } = props;
-  const childArray = React.Children.toArray(children);
 
-  if (childArray.some(isSlottable)) {
+  if (React.Children.toArray(children).some(isSlottable)) {
     return (
       <>
-        {childArray.map((child) => {
+        {React.Children.map(children, (child) => {
           return isSlottable(child) ? (
             <SlotClone {...slotProps} ref={forwardedRef}>
               {child.props.children}
--- a/.yarn/versions/4826891d.yml
+++ b/.yarn/versions/4826891d.yml
@@ -1,0 +1,40 @@
+releases:
+  "@radix-ui/react-accessible-icon": patch
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-announce": patch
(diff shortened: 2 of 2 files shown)


packages/react/slot/src/Slot.tsx
```diff
@@ -11,12 +11,11 @@ interface SlotProps extends React.HTMLAttributes<HTMLElement> {
...
 const Slot = React.forwardRef<HTMLElement, SlotProps>((props, forwardedRef) => {
   const { children, ...slotProps } = props;
-  const childArray = React.Children.toArray(children);
 
-  if (childArray.some(isSlottable)) {
+  if (React.Children.toArray(children).some(isSlottable)) {
     return (
       <>
-        {childArray.map((child) => {
+        {React.Children.map(children, (child) => {
```
benoitgrelard, line 18: TIL!

I didn't expect this would fix the issue, without setting `key`, is `React.Children.map` doing something I am unaware of?

I assumed it was equivalent to `React.Children.toArray(children).map`.

benoitgrelard (approved):
> Admittedly, I don't seem to be able to reproduce the issue but makes sense so I've switched it to use `React.Children.map` to prevent any warnings like that regardless.

It can be reproduced in cases where we have used `Slottable`, ie. when there are multiple children internally.
We actually have an example of it in the chromatic story of `Tooltip` and you can see the error in the console. 

I can confirm this fix resolves the issue.

packages/react/slot/src/Slot.tsx
jjenzz: React will always use the index as the key if one isn't provided, but it warns when one isn't provided because providing one is the preferred approach. Using `React.Children.map` just suppresses the warning AFAIK.

benoitgrelard: Awesome, TIL 🙂

PR #1018, opened 2021-12-07
[Tooltip] Fix trigger when another tooltip overlays it

### LINKED ISSUE #1001: [Tooltip] tooltip is not triggered when tooltip covers trigger
## Bug report

Tooltip is not triggered when tooltip covers trigger.

### Current Behavior

When you have a tooltip that covers a trigger, it is possible to end up hovering over a trigger and no tooltip displayed.

I tried using `pointerEvent: "none",` on the tooltip content, but that didn't seem to work.

### Expected behavior

When focusing on a trigger after a tooltip is hidden, it should show always show the tooltip.

### Reproducible example

https://codesandbox.io/s/dazzling-haze-4z9r5?file=/App.js

In the example sandbox:

1. Start with mouse off both buttons and no tooltip showing
1. Hover on button 1, wait untill tooltip is shown
1. Quickly move mouse onto button 2

The image below shows an icon hovered (red) and no tooltip.

image

### Your environment

| Radix Package(s) |     radix-ui/react-tooltip    |   0.1.1      |
| React            | n/a     |     17.0.2    |
| Browser          |     Chrome    |      96.0.4664.45    |
| Operating System |    Windows     |    10     |

jjenzz:
Fixes #1001 

I mentioned on that ticket that I would look into this as part of the tooltip work.

### DIFF
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -235,7 +235,7 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
           data-state={context.stateAttribute}
           {...triggerProps}
           ref={composedTriggerRef}
-          onMouseEnter={composeEventHandlers(props.onMouseEnter, context.onOpen)}
+          onMouseOver={composeEventHandlers(props.onMouseOver, context.onOpen)}
           onMouseLeave={composeEventHandlers(props.onMouseLeave, context.onClose)}
           onMouseDown={composeEventHandlers(props.onMouseDown, context.onClose)}
           onFocus={composeEventHandlers(props.onFocus, context.onFocus)}
--- a/.yarn/versions/8c144add.yml
+++ b/.yarn/versions/8c144add.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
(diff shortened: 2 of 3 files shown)


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -235,7 +235,7 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
           data-state={context.stateAttribute}
           {...triggerProps}
           ref={composedTriggerRef}
-          onMouseEnter={composeEventHandlers(props.onMouseEnter, context.onOpen)}
+          onMouseOver={composeEventHandlers(props.onMouseOver, context.onOpen)}
```
jjenzz, line 238: There is an issue with this... If a dialog close button (for example) is positioned on top of a tooltip trigger, when the dialog closes the tooltip will open:

https://user-images.githubusercontent.com/175330/145034878-49fa44d9-3076-4932-9f92-ee51eebac977.mp4

The alternative is to try adding `pointer-events: none` to the content parts but `Popover` has a wrapping div that cannot be styled.

benoitgrelard: I'm not sure it's an issue actually. To be honest, I kinda like it that it does that!

benoitgrelard (approved):

packages/react/tooltip/src/Tooltip.tsx
jjenzz: Furry muff 🎉

PR #1022, opened 2021-12-09
[Dialog][AlertDialog] Omit `asChild` from `Portal`

jjenzz:
I noticed when documenting the release that the prop types for `Dialog.Portal` have the `asChild` prop but it doesn't do anything at the moment. Either we remove it from types or pass it along?

I've gone with passing it along for now in case people want the portal to render a `div`, then they can pass `asChild={false}`. I dunno if that's likely to be a valid use-case though so happy to remove it.

### DIFF
--- a/.yarn/versions/a90db96c.yml
+++ b/.yarn/versions/a90db96c.yml
@@ -1,0 +1,6 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-dialog": patch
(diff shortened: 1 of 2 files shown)


benoitgrelard (commented):
I think I'd opt for removing it.

jjenzz:
@benoitgrelard Done 👍

benoitgrelard (approved):

PR #1034, opened 2021-12-10
[AlertDialog][Dialog] Fix `Portal` types

jjenzz:


packages/react/dialog/src/Dialog.tsx
```diff
@@ -139,23 +138,21 @@ interface DialogPortalProps extends Omit<PortalProps, 'asChild'> {
...
-  }
-);
+const DialogPortal: React.FC<DialogPortalProps> = (props: ScopedProps<DialogPortalProps>) => {
+  const { __scopeDialog, forceMount, children, container } = props;
+  const context = useDialogContext(PORTAL_NAME, __scopeDialog);
+  return (
+    <>
+      {React.Children.map(children, (child) => (
+        <Presence present={forceMount || context.open}>
+          <UnstablePortal asChild container={container}>
```
benoitgrelard, line 148: I guess that works because we know Portal only has one prop for now.
Shall we worry about it later if Portal ever takes another prop? (probably won't happen anyway right? 😀)

benoitgrelard (approved):

packages/react/dialog/src/Dialog.tsx
jjenzz: yeah any new optional props we add won't be flagged 🙁  but since we know this is always an `asChild` it does seem weird for it to accept and forward props to two different elements.

PR #1035, opened 2021-12-13
[Tooltip] Fix tooltip visibility logic

jjenzz:
A couple of fixes here so I've left comments inline to clarify what's what.

### DIFF
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -36,7 +36,7 @@ type TooltipProviderContextValue = {
   isOpenDelayed: boolean;
   delayDuration: number;
   onOpen(): void;
-  onTriggerLeave(): void;
+  onClose(): void;
 };
 
 const [TooltipProviderContextProvider, useTooltipProviderContext] =
@@ -44,7 +44,7 @@ const [TooltipProviderContextProvider, useTooltipProviderContext] =
     isOpenDelayed: true,
     delayDuration: DEFAULT_DELAY_DURATION,
     onOpen: () => {},
-    onTriggerLeave: () => {},
+    onClose: () => {},
   });
 
 interface TooltipProviderProps {
@@ -87,7 +87,7 @@ const TooltipProvider: React.FC<TooltipProviderProps> = (
         window.clearTimeout(skipDelayTimerRef.current);
         setIsOpenDelayed(false);
       }, [])}
-      onTriggerLeave={React.useCallback(() => {
+      onClose={React.useCallback(() => {
         window.clearTimeout(skipDelayTimerRef.current);
         skipDelayTimerRef.current = window.setTimeout(
           () => setIsOpenDelayed(true),
@@ -115,7 +115,6 @@ type TooltipContextValue = {
   trigger: TooltipTriggerElement | null;
   onTriggerChange(trigger: TooltipTriggerElement | null): void;
   onTriggerEnter(): void;
-  onTriggerLeave(): void;
   onOpen(): void;
   onClose(): void;
 };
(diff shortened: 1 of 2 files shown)


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -164,10 +163,8 @@ const Tooltip: React.FC<TooltipProps> = (props: ScopedProps<TooltipProps>) => {
         // we dispatch here so `TooltipProvider` isn't required to
         // ensure other tooltips are aware of this one opening.
         document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN));
-      } else {
-        isFocusOpenRef.current = false;
+        onOpen();
       }
-      window.clearTimeout(openTimerRef.current);
```
jjenzz, line 170: This was only clearing if the `open` value changed from `false` to `true` but not if the trigger was focused and then the mouse entered it because when it was focused, open will have been true and mousing over it would have tried to set open to true again which will have ran the open timer.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -241,12 +243,20 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
...
-          onBlur={composeEventHandlers(props.onBlur, context.onTriggerLeave)}
+          onMouseDown={composeEventHandlers(props.onMouseDown, () => {
+            isMouseDownRef.current = true;
+            context.onClose();
+            document.addEventListener('mouseup', () => (isMouseDownRef.current = false), {
+              once: true,
+            });
+          })}
+          onFocus={composeEventHandlers(props.onFocus, () => {
+            if (!isMouseDownRef.current) context.onOpen();
```
jjenzz, line 254: Explicitly only open if it isn't a mousedown focus.

packages/react/tooltip/src/Tooltip.tsx
benoitgrelard: Good catch!

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -164,17 +162,25 @@ const Tooltip: React.FC<TooltipProps> = (props: ScopedProps<TooltipProps>) => {
...
     },
   });
-
   const stateAttribute = React.useMemo(() => {
-    return open ? (wasOpenDelayed ? 'delayed-open' : 'instant-open') : 'closed';
-  }, [wasOpenDelayed, open]);
+    return open ? (wasOpenDelayedRef.current ? 'delayed-open' : 'instant-open') : 'closed';
+  }, [open]);
+
+  const handleOpen = React.useCallback(() => {
```
benoitgrelard, line 174: Could this be name `handleOpenInstantly` or something like that?
It's all getting a bit hard to follow now in `onTriggerEnter` with the branching + the external handler.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -241,12 +249,20 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
           ref={composedTriggerRef}
           onMouseEnter={composeEventHandlers(props.onMouseEnter, context.onTriggerEnter)}
           onMouseLeave={composeEventHandlers(props.onMouseLeave, context.onTriggerLeave)}
-          onMouseDown={composeEventHandlers(props.onMouseDown, context.onTriggerLeave)}
-          onFocus={composeEventHandlers(props.onFocus, context.onTriggerFocus)}
-          onBlur={composeEventHandlers(props.onBlur, context.onTriggerLeave)}
+          onMouseDown={composeEventHandlers(props.onMouseDown, () => {
+            context.onClose();
+            isMouseDownRef.current = true;
+            document.addEventListener('mouseup', () => (isMouseDownRef.current = false), {
```
benoitgrelard, line 255: I always get confused with those, what if the trigger unmounts between mousedown and mouseup? Then it's leaking no?

packages/react/tooltip/src/Tooltip.tsx
jjenzz: I feel like `handleOpen` and `handleClose` is already instant... otherwise should we rename all of our `handleOpen` handlers in our other comps?  It already implies that it's instant. 

If it helps, I've abstracted the delayed logic into a `handleDelayedOpen` so that you can see easily that sometimes it is delayed.

packages/react/tooltip/src/Tooltip.tsx
jjenzz: Good spot. I've added a clean up for this!

packages/react/tooltip/src/Tooltip.tsx
benoitgrelard: Yeah that works for me.

benoitgrelard (approved):

PR #1037, opened 2021-12-13
[All] Deprecate `IdProvider`

jjenzz:


benoitgrelard (commented):
I get why we're doing this but now we have to remember to remove that kind of stuff later.
And when would that be?
When we hit 1.0.0 in semver?

packages/react/id/src/id.tsx
```diff
@@ -14,4 +14,13 @@ function useId(deterministicId?: string): string {
...
 
-export { useId };
+function IdProvider({ children }: { children: React.ReactNode }) {
+  React.useEffect(() => {
+    console.warn(ID_PROVIDER_DEPRECATED);
+  }, []);
+  return <>{children}</>;
+}
+
+const ID_PROVIDER_DEPRECATED = `Warning: The \`IdProvider\` has been deprecated. For details, see https://radix-ui.com/docs/primitives/utilities/id-provider`;
```
benoitgrelard, line 24: Can probably inline it? Makes it even easier to delete later.

jjenzz:
> I get why we're doing this but now we have to remember to remove that kind of stuff later.

Answered in DM but for anyone else following, I agree and I don't know the answers to your questions unfortunately, but we can figure that out later I guess. Atm, it's breaking things for folks and we've guided people better in the past (e.g. our `as` prop warning).

> Can probably inline it? Makes it even easier to delete later.

I followed the format we used for `as` prop warning and all other messages that we log for consistency. 

I'm not sure it really matters if we're adding a comment for us to find+replace later.

benoitgrelard:
> I'm not sure it really matters if we're adding a comment for us to find+replace later, especially since putting it inline wraps n stuff and you know that makes me itch 😅

Definitely doesn't matter no, let's keep with what we have.

benoitgrelard (approved):

PR #1042, opened 2021-12-14
[Accordion] Add missing dependency

jjenzz:


jjenzz:
Looks like @andy-hook already did this 🎉

PR #1048, opened 2021-12-16
[Storybook] Update stitches + conventions

benoitgrelard:
🔥 **NOTE: I have `--no-verify`'ed this for now to put the PR up but, as discussed on Slack, the pre-push is not happy with the stitches types… However, everything is fine on CI, build, storybook build, chromatic, etc**  

I've called out the meaningful changes I've made, but apart from those, it's mainly just a lot of replace which I explain below.

This PR does a few things:
- Update `stitches` to latest (also switch to `@stitches/core` as we were only using the core for a while anyway)
  - biggest diff was that stitches class need to be called now `()`, thankfully we were consistent with our naming so I was able to do this with a regex replace 😀
  - tokens in theme shouldn't start with `$` anymore
  - `keyframes` is a new export
  - color tokens were used in `boxShadow` but aren't mapped now, so had to update to `$colors$…`
- Use dotted syntax and `import * as …` for all multi-part primitives consistently


packages/react/menu/src/Menu.stories.tsx
```diff
@@ -466,7 +455,7 @@ const itemClass = css({
 });
 
 const subTriggerClass = css(itemClass, {
-  '&[data-state="open"]': {
+  '&:not(:focus)[data-state="open"]': {
```
benoitgrelard, line 458: This is the only meaningful change I had to make after checking every single story manually. It seems we were relying on some specificity bug from stitches before.

packages/react/dialog/src/Dialog.stories.tsx
```diff
@@ -521,7 +512,7 @@ const animatedOverlayClass = css(overlayClass, {
 
 const animatedContentClass = css(contentDefaultClass, {
   '&[data-state="open"]': {
-    animation: `${fadeIn} ${scaleIn} 300ms ease-out`,
+    animation: `${fadeIn} 150ms ease-out, ${scaleIn} 200ms ease-out`,
```
benoitgrelard, line 515: This is another change I made, not sure if it ever worked, but I noticed it definitely didn't with this new version (ie. you need 2 separate animations, comma separated).

packages/react/context-menu/src/ContextMenu.stories.tsx
```diff
@@ -73,138 +59,160 @@ export const Modality = () => {
     <div
       style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', height: '110vh' }}
     >
-      <div style={{ display: 'grid', gridGap: 50 }}>
+      <div style={{ display: 'grid', gap: 50 }}>
```
andy-hook, line 62: `gripGap` was deprecated long ago and no longer renders with that syntax in latest Stitches

package.json
```diff
@@ -83,7 +83,7 @@
     "replace-in-files": "^3.0.0",
     "start-server-and-test": "^1.12.5",
     "ts-jest": "^26.2.0",
-    "typescript": "^3.9.7"
+    "typescript": "^4.6.3"
```
andy-hook, line 86: As suspected, this resolved the type issue.

package.json
```diff
@@ -43,9 +43,9 @@
     "@babel/template": "^7.12.13",
     "@parcel/babel-plugin-transform-runtime": "^2.0.0-nightly.1823",
     "@parcel/babel-preset-env": "^2.0.0-alpha.3",
-    "@stitches/react": "0.0.3-canary.4",
-    "@storybook/addon-storysource": "6.1.0-beta.4",
-    "@storybook/react": "6.1.0-beta.4",
+    "@stitches/core": "^1.2.7",
+    "@storybook/addon-storysource": "^6.4.21",
+    "@storybook/react": "^6.4.21",
```
andy-hook, line 48: https://github.com/styleguidist/react-docgen-typescript/issues/356

tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "paths": {
-      "*": ["node_modules/*"],
```
andy-hook, line 5: https://github.com/cypress-io/cypress/issues/16188

packages/react/toast/src/Toast.stories.tsx
```diff
@@ -549,7 +559,7 @@ const buttonClass = css({
   border: '1px solid black',
   borderRadius: 4,
   background: 'gainsboro',
-  font: 'inherit',
```
andy-hook, line 552: This is the cause of the chromatic diff prompt, for some reason `font` wasn't rendering in the previous Stitches version (we usually use `fontFamily`)

I've updated it to use `fontFamily` just for consistency with the other stories though it's functionally the same now anyway since it was seemingly fixed.

andy-hook (approved):

PR #1049, opened 2021-12-17
[Storybook] Fix Collection stories

benoitgrelard:
Whilst checking every story when working on #1048 I noticed all of the `Collection` stories were broken.
Probably have been broken since we use the scoping of contexts, the scope wasn't matching in `useCollection` usage.

### DIFF
--- a/packages/react/collection/src/Collection.stories.tsx
+++ b/packages/react/collection/src/Collection.stories.tsx
@@ -157,7 +157,7 @@ const MemoItem = React.memo(Item);
 const MemoItems = React.memo(WrappedItems);
 
 function LogItems({ name = 'items' }: { name?: string }) {
-  const getItems = useCollection({});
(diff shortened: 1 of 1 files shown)


packages/react/collection/src/Collection.stories.tsx
```diff
@@ -157,7 +157,7 @@ const MemoItem = React.memo(Item);
 const MemoItems = React.memo(WrappedItems);
 
 function LogItems({ name = 'items' }: { name?: string }) {
-  const getItems = useCollection({});
+  const getItems = useCollection(undefined);
```
benoitgrelard, line 160: Basically, we were passing `scope={undefined}` everywhere on the component props, but here we were passing an empty object.

andy-hook (approved):

jjenzz (approved):

PR #1053, opened 2021-12-23
Update Accordion.test.tsx

luisorbaiceta:
### Description

As seen in the described behaviour, clicking the trigger again hides the content

I don't know if this is a relevant PR, but just made sense to me when I was extending and testing the Accordion behaviour.

I haven't been able to install dependencies in my fork so I am not able to run the ci pipeline to check this.

### DIFF
--- a/packages/react/accordion/src/Accordion.test.tsx
+++ b/packages/react/accordion/src/Accordion.test.tsx
@@ -80,11 +80,11 @@ describe('given a single Accordion', () => {
         fireEvent.click(trigger);
       });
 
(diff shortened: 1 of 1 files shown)


jjenzz:
I believe these tests were correct. The trigger should only hide the content in a `type="single"` accordion when it has a `collapsible` prop. By default, it wouldn't.

luisorbaiceta:
I see. Thanks for pointing this out! I just thought that the example in the website was the default behaviour 😅

PR #1060, opened 2022-01-02
[Fix][Dialog] forceMount type for `Overlay` & `Content`

joshuaellis:
### Description

the `forceMount` prop on [`Dialog.Overlay`](https://www.radix-ui.com/docs/primitives/components/dialog#overlay) and [`Dialog.Content`](https://www.radix-ui.com/docs/primitives/components/dialog#content) both specify that the value should be `boolean` however the type signature has been declared as `true` therefore providing `false` creates a TS error. 

Maybe the initial idea was that `undefined` could be used instead, but if you pair it up with `onOpenChange` from `Dialog.Root` you're only provided `boolean` values.

### What

To solve this, I've replaced the `true` value with `false`.

### DIFF
--- a/.yarn/versions/6427e47a.yml
+++ b/.yarn/versions/6427e47a.yml
@@ -1,3 +1,6 @@
 releases:
   "@radix-ui/react-alert-dialog": patch
   "@radix-ui/react-dialog": patch
+
+declined:
+  - primitives
--- a/.yarn/versions/6427e47a.yml
+++ b/.yarn/versions/6427e47a.yml
@@ -1,0 +1,3 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-dialog": patch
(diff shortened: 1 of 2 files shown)


andy-hook:
Hi @joshuaellis, `forceMount` is not intended to be dynamic and the typings here are to indicate that. Components should only be force mounted for their lifetime. The prop is intended primarily for [usage with JS animation libs](https://www.radix-ui.com/docs/primitives/overview/animation#delegating-unmounting-for-javascript-animation) in order to handover mount handling. 

Is there a reason you need to be pairing it with `onOpenChange`?

joshuaellis:
I suppose my use-case is rather niche and potentially goes against what the library should be used for:

I'm using Remix but `stitches` doesn't collect the styles correctly unless the component is initially mounted i.e. if you mount after the initial render it doesn't work (see https://github.com/modulz/stitches/issues/908). As a work around I can use `useSpring` to animate the component in and out and dynamically set the mounting of the component.

It'd be good to understand (if not for my own understanding) why forceMount is not intended to by dynamic, are there performance issues etc.?

You can see my "hacky" implementation [here](https://github.com/pmndrs/react-spring/blob/docs/rebuild/docs-new/app/components/Header.tsx).

PR #1066, opened 2022-01-04
[usePrevious] Persist return value between renders

andy-hook:
Based on our discussion on the expected behaviour of `usePrevious`. The [typical approach](https://usehooks.com/usePrevious/) actually returns the value from the previous **render**, our thinking however is that this is unintuitive and unexpected. Our expectation is that the value should represent the previous `value` that was **different** e.g. the previous value after it was changed, independent of any re-rendering that may occur.

You can see the difference below, notice how the previous value in the "common" demo only meets our expectation for a single render while "Adjusted" consistently meets it no matter how many renders occur.
 
https://codesandbox.io/s/summer-dream-7qymy?file=/src/App.js

### DIFF
--- a/packages/react/use-previous/src/usePrevious.tsx
+++ b/packages/react/use-previous/src/usePrevious.tsx
@@ -1,8 +1,6 @@
 import * as React from 'react';
 
 function usePrevious<T>(value: T) {
-  // The ref object is a generic container whose current property is mutable ...
-  // ... and can hold any value, similar to an instance property on a class
   const ref = React.useRef({ value, previous: value });
 
   // We compare values before making an update to ensure that
--- a/packages/react/use-previous/src/usePrevious.tsx
+++ b/packages/react/use-previous/src/usePrevious.tsx
@@ -3,15 +3,18 @@ import * as React from 'react';
 function usePrevious<T>(value: T) {
   // The ref object is a generic container whose current property is mutable ...
   // ... and can hold any value, similar to an instance property on a class
-  const ref = React.useRef<T>(value);
+  const ref = React.useRef({ value, previous: value });
 
-  // Store current value in ref
-  React.useEffect(() => {
-    ref.current = value;
-  }, [value]); // Only re-run if value changes
-
-  // Return previous value (happens before update in useEffect above)
-  return ref.current;
+  // We compare values before making an update to ensure that
+  // a change has been made. This ensures the previous value is
+  // persisted correctly between renders.
+  return React.useMemo(() => {
+    if (ref.current.value !== value) {
+      ref.current.previous = ref.current.value;
+      ref.current.value = value;
+    }
+    return ref.current.previous;
+  }, [value]);
 }
 
 export { usePrevious };
--- a/.yarn/versions/ee709007.yml
+++ b/.yarn/versions/ee709007.yml
@@ -1,0 +1,10 @@
+releases:
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-radio-group": patch
+  "@radix-ui/react-slider": patch
+  "@radix-ui/react-switch": patch
+  "@radix-ui/react-tooltip": patch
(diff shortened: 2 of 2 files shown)


packages/react/use-previous/src/usePrevious.tsx
```diff
@@ -3,15 +3,18 @@ import * as React from 'react';
...
-    ref.current = value;
-  }, [value]); // Only re-run if value changes
-
-  // Return previous value (happens before update in useEffect above)
-  return ref.current;
+  // We compare values before making an update to ensure that
+  // a change has been made. This ensures the previous value is
+  // persisted correctly between renders.
+  return React.useMemo(() => {
+    if (ref.current.value !== value) {
```
andy-hook, line 12: Primarily to ensure [consistent behaviour on CSB](https://github.com/codesandbox/codesandbox-client/issues/6357)

benoitgrelard: Why redundant though? If we didn't check, then wouldn't the value be reassigned for any change?

andy-hook: In an alternative version we could get away without the check:

https://stackblitz.com/edit/react-ejzqys?file=src/App.js

but it was unreliable in CSB:

https://codesandbox.io/s/loving-lamport-fimun?file=/src/App.js

Which is why the code was adjusted the use the check.

packages/react/use-previous/src/usePrevious.tsx
```diff
@@ -3,15 +3,18 @@ import * as React from 'react';
 function usePrevious<T>(value: T) {
   // The ref object is a generic container whose current property is mutable ...
   // ... and can hold any value, similar to an instance property on a class
-  const ref = React.useRef<T>(value);
+  const ref = React.useRef({ value, previous: value });
```
andy-hook, line 6: Just thinking, `previous` should probably be initially `undefined`? the original hook doesn't account for that but I think that makes more sense?

packages/react/use-previous/src/usePrevious.tsx
benoitgrelard: Ah yeah I see, thanks for the clarification.
Man that's frustrating to have to add code in our lib to please CSB though…

packages/react/use-previous/src/usePrevious.tsx
jjenzz: We purposefully made it default to `value` so that hooks and things didn't run on initial render. The hook is usually used to signal a "value change" and on initial render, the value hasn't changed.

packages/react/use-previous/src/usePrevious.tsx
jjenzz: Yeah but there is also this:

> In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.
> https://reactjs.org/docs/hooks-reference.html#usememo

So, relying on deps to signify "value has changed" isn't entirely accurate with `useMemo`.

jjenzz (approved):
Thanks for doing this 💛

packages/react/use-previous/src/usePrevious.tsx
```diff
@@ -3,15 +3,18 @@ import * as React from 'react';
 function usePrevious<T>(value: T) {
   // The ref object is a generic container whose current property is mutable ...
   // ... and can hold any value, similar to an instance property on a class
```
jjenzz, line 5: Oh, one little favour while we're here if you don't mind... Can we remove this comment? 🙏 I've never understood why we have a comment explaining how refs work 😅

benoitgrelard: Smells like copy/paste 😀

benoitgrelard (approved):
Did we do a sweep to check that all the components using this hook still work reliably?

PR #1075, opened 2022-01-10
[Dialog][AlertDialog] Share `forceMount` from Portal

### LINKED ISSUE #1061: [Dialog] framer-motion exit animations for Portals require `forceMount` prop on descendants.
This bug report is inspired by discussion #1058, identified by @jjenzz 

### Current Behavior

Current way to enable exit animations with framer-motion is:
- wrap the `Portal` in `AnimatePresence`
- Make it a controlled component
- place `forceMount` prop on `<Portal />` as well as `<Overlay />` and `<Content />`.

### Expected behavior

`forceMount` prop should only be required on `<Portal />` and not it's direct descendants.

### Your environment

| Radix Package(s) |     @radix-ui/react-dialog    |    0.1.5     |
| framer-motion            | n/a     |     5.5.5     |
| Browser          |     Firefox    |    95      |
| Assistive tech   |    not used     |         |
| Node             | n/a     |     v16.13.0   |
| npm/yarn         |    yarn     |      1.22.17   |
| Operating System |    macOS(Apple Silicon)     |   Big Sur 11.6.2     |

jjenzz:
Closes #1061

Current implementation required `forceMount` to be added to all three parts which was unintuitive.

- Shares prop from portal
- Adds `ForceMount` story to Chromatic output to highlight breaking changes with `forceMount` in future

**Note: Not sure why it isn't force mounted in the Chromatic story lol, ack. I'll figure that out in a bit.**

### DIFF
--- a/packages/react/dialog/src/Dialog.stories.tsx
+++ b/packages/react/dialog/src/Dialog.stories.tsx
@@ -210,15 +210,16 @@ export const Animated = () => (
 export const ForcedMount = () => (
   <Dialog>
     <DialogTrigger>open</DialogTrigger>
-    <DialogPortal>
-      <DialogOverlay className={overlayClass} forceMount />
-      <DialogContent className={contentDefaultClass} forceMount>
+    <DialogPortal forceMount>
+      <DialogOverlay className={overlayClass} />
+      <DialogContent className={contentDefaultClass}>
         <DialogTitle>Title</DialogTitle>
         <DialogClose>close</DialogClose>
       </DialogContent>
     </DialogPortal>
   </Dialog>
 );
+ForcedMount.parameters = { chromatic: { disable: false } };
 
 export const AllowPinchZoom = () => (
   <div style={{ display: 'grid', placeItems: 'center', height: '200vh' }}>
--- a/packages/react/dialog/src/Dialog.stories.tsx
+++ b/packages/react/dialog/src/Dialog.stories.tsx
@@ -35,10 +35,10 @@ export const NonModal = () => (
         <DialogOverlay className={overlayClass} />
         <DialogContent
           className={contentSheetClass}
-          aria-describedby={undefined}
           onInteractOutside={(event) => event.preventDefault()}
         >
           <DialogTitle>Booking info</DialogTitle>
+          <DialogDescription>Description</DialogDescription>
           <DialogClose className={closeClass}>close</DialogClose>
         </DialogContent>
       </DialogPortal>
@@ -62,8 +62,9 @@ export const Controlled = () => {
       <DialogTrigger>{open ? 'close' : 'open'}</DialogTrigger>
       <DialogPortal>
         <DialogOverlay className={overlayClass} />
-        <DialogContent className={contentDefaultClass} aria-describedby={undefined}>
+        <DialogContent className={contentDefaultClass}>
           <DialogTitle>Title</DialogTitle>
+          <DialogDescription>Description</DialogDescription>
           <DialogClose>close</DialogClose>
         </DialogContent>
       </DialogPortal>
@@ -77,9 +78,10 @@ export const FocusTrap = () => (
       <DialogTrigger>open</DialogTrigger>
       <DialogPortal>
         <DialogOverlay className={overlayClass} />
-        <DialogContent className={contentDefaultClass} aria-describedby={undefined}>
+        <DialogContent className={contentDefaultClass}>
           <DialogClose>close</DialogClose>
           <DialogTitle>Title</DialogTitle>
+          <DialogDescription>Description</DialogDescription>
           <div>
             <label htmlFor="firstName">First Name</label>
             <input type="text" id="firstName" placeholder="John" />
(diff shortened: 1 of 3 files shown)


jjenzz:
Not sure why it isn't force mounted in the Chromatic story lol, ack. I'll figure that out in a bit.

packages/react/dialog/src/Dialog.stories.tsx
```diff
@@ -210,15 +210,16 @@ export const Animated = () => (
...
+    <DialogPortal forceMount>
+      <DialogOverlay className={overlayClass} />
+      <DialogContent className={contentDefaultClass}>
         <DialogTitle>Title</DialogTitle>
         <DialogClose>close</DialogClose>
       </DialogContent>
     </DialogPortal>
   </Dialog>
 );
+ForcedMount.parameters = { chromatic: { disable: false } };
```
benoitgrelard, line 222: What's the reason for enabling this story separately as opposed to adding a new thing in the Chromatic story as we usually do?

packages/react/dialog/src/Dialog.tsx
```diff
@@ -176,7 +181,8 @@ interface DialogOverlayProps extends DialogOverlayImplProps {
 
 const DialogOverlay = React.forwardRef<DialogOverlayElement, DialogOverlayProps>(
   (props: ScopedProps<DialogOverlayProps>, forwardedRef) => {
-    const { forceMount, ...overlayProps } = props;
+    const portalContext = usePortalContext(OVERLAY_NAME, props.__scopeDialog);
+    const { forceMount = portalContext.forceMount, ...overlayProps } = props;
```
benoitgrelard, line 185: Why are we keeping the props on the parts? Is it in case they don't use the portal in specific non-modal inline dialogs maybe?

benoitgrelard (commented):
Yeah not sure what's going on with the chromatic stuff 😀

packages/react/dialog/src/Dialog.stories.tsx
jjenzz: The dialog chromatic story is a PITA 😅  it's a bunch of absolutely positioned dialogs with 20% width so adding a 5th was a lot of manually fiddling hardcoded values to get things positioned correctly and just made things hard to visualise with 5 cramped dialogs in a row.

I contemplated updating the whole story to take advantage of the new `<Portal container={} />` API but that would have made the visual changes in this PR trickier to review. Happy to give it a go though if preferred?

benoitgrelard: Ah yes, of course, I forgot about that. That's fine then, I think at some point we could split back into stories given we have quite a lot of snapshots allowance.

packages/react/dialog/src/Dialog.tsx
jjenzz: > Is it in case they don't use the portal in specific non-modal inline dialogs maybe?
Yeah exactly, cos it's not a required part.

jjenzz:
I've updated the stories here to prevent errors/warnings after the changes we made to description/title as I thought the title error might be effecting this and also increased the story delay (to debug) but no luck.

I'm really stumped as to why this story isn't force mounting 🤔 😢

andy-hook:
> I'm really stumped as to why this story isn't force mounting in Chromatic but works fine locally

Turned out to be a visual issue, the logic does work in Chromatic it's just that the story wasn't preserving document space to show it all correctly in the rendered snapshot. I moved it to the main `Chromatic` story for consistency as that already allocates the space in layout.

packages/react/dialog/src/Dialog.stories.tsx
```diff
@@ -35,6 +35,7 @@ export const NonModal = () => (
         <DialogOverlay className={overlayClass} />
         <DialogContent
           className={contentSheetClass}
+          aria-describedby={undefined}
```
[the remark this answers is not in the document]
andy-hook: I agree, I can make this change.

packages/react/dialog/src/Dialog.tsx
```diff
@@ -486,11 +493,9 @@ const [WarningProvider, useWarningContext] = createContext(TITLE_WARNING_NAME, {
   docsSlug: 'dialog',
 });
 
-type WarningProps = {
-  contentRef: React.RefObject<DialogContentElement>;
-};
+type TitleWarningProps = { titleId?: string };
 
-const TitleWarning: React.FC<WarningProps> = ({ contentRef }) => {
+const TitleWarning: React.FC<TitleWarningProps> = ({ titleId }) => {
```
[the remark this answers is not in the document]
andy-hook: I was wondering the same, perhaps @jjenzz can provide some context.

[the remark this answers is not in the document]
benoitgrelard: Re-reading this comment here: https://github.com/radix-ui/primitives/pull/1075#issuecomment-1054372456
and looking at the sha before that comment: https://github.com/radix-ui/primitives/commit/b759d667d4dc094b9136cf3247571b51499f69c1 it seems it was intentional.

Looking at the actual changes there too they seem inline with the previous changes regarding title/description handling for accessibility and rely less on implementation detail but rather on the rules vs. escape hatches we have put in place.

So I think it's actually correct and good to go.

benoitgrelard (approved):

andy-hook (approved):

PR #1082, opened 2022-01-13
[DropdownMenu] Fix non-modal dropdown pointer-events

### LINKED ISSUE #1059: [DropdownMenu] Clicking on trigger when menu is open and `modal={false}` closes then immediately reopens menu
_Happy New Year, Radix! 🎉_

### Current Behavior

If you set `modal={true}` on `DropdownMenu.Root`, clicking on the trigger when the menu is open will close but then immediately reopen the menu.

CleanShot 2022-01-01 at 13 15 59

### Expected behavior

Clicking on the trigger should only ever close or open the menu (not both in quick succession), regardless of whether `modal={true}` or `modal={false}`.

CleanShot 2022-01-01 at 13 24 20

### Reproducible example

https://codesandbox.io/s/radix-dropdown-menu-modal-double-trigger-bug-un41f?file=/App.js

### Suggested solution

I'm not sure.

### Additional context

I found that this behavior is introduced in `0.1.2-rc.7`. It does not happen on `0.1.2-rc.6`.

### Your environment

| Radix Package(s) |     @radix-ui/react-dropdown-menu    |    0.1.4 (but bug introduced in 0.1.2-rc.7)     |
| Browser          |    Chrome     |     Version 97.0.4692.56 (Official Build) beta (arm64)    |
| Assistive tech   |     none    |         |
| Node             | n/a     |     10.20.1    |
| npm/yarn         |    whichever codesandbox uses by default     |     unknown    |
| Operating System |     macOS    |    12.1     |

jjenzz:
Fixes #1059

Also: 

- While fixing, I noticed that non-modal dropdowns don't focus the trigger when closing via trigger click so fixed that too
- Updated the other components where the `onClick` logic was added to ensure this only fires when it's not a mouse click (since these components already have `onMouseDown` or `onPointerDown` handlers for that).

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -175,15 +175,15 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
               // prevent trigger focusing when opening
               // this allows the content to be given focus without competition
               if (!context.open) event.preventDefault();
-              // trigger always opens, and content click outside closes. since trigger is
-              // considered outside content, clicking trigger when open will close the content.
-              context.onOpenChange(true);
+              context.onOpenToggle();
             }
           })}
           onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
-            if (!disabled && ['ArrowDown', 'Enter', ' '].includes(event.key)) {
+            if (disabled) return;
+            if (['Enter', ' '].includes(event.key)) context.onOpenToggle();
+            if (event.key === 'ArrowDown') {
               context.onOpenChange(true);
-              // Prevent keypresses from scrolling window
+              // prevent arrow keypress from scrolling window
               event.preventDefault();
             }
           })}
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -168,9 +168,6 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           disabled={disabled}
           {...triggerProps}
           ref={composeRefs(forwardedRef, context.triggerRef)}
-          // Handle anything that the browser considers a click for the element type if
-          // not using pointer e.g. Space keyup and Enter keydown
-          onClick={composeEventHandlers(props.onClick, () => context.onOpenChange(true))}
           onPointerDown={composeEventHandlers(props.onPointerDown, (event) => {
             // only call handler if it's the left button (mousedown gets triggered by all mouse buttons)
             // but not when the control key is pressed (avoiding MacOS right click)
@@ -178,8 +175,14 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
               // prevent trigger focusing when opening
               // this allows the content to be given focus without competition
               if (!context.open) event.preventDefault();
-              context.onOpenToggle();
+              context.onOpenChange(true);
             }
+          })}
+          onClick={composeEventHandlers(props.onClick, (event) => {
+            // Handle anything that the browser considers a click for the element type if
+            // not using pointer e.g. Space keyup and Enter keydown
+            const isPointerClick = event.detail > 0;
+            if (!isPointerClick) context.onOpenChange(true);
           })}
           onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
             if (!disabled && event.key === 'ArrowDown') {
@@ -256,8 +259,6 @@ const DropdownMenuRootContent = React.forwardRef<
   const { __scopeDropdownMenu, portalled = true, ...contentProps } = props;
   const context = useDropdownMenuContext(CONTENT_NAME, __scopeDropdownMenu);
   const menuScope = useMenuScope(__scopeDropdownMenu);
-  const hasInteractedOutsideRef = React.useRef(false);
-
   return context.isRootMenu ? (
     <MenuPrimitive.Content
       id={context.contentId}
@@ -268,35 +269,11 @@ const DropdownMenuRootContent = React.forwardRef<
       portalled={portalled}
       onCloseAutoFocus={(event) => {
         props.onCloseAutoFocus?.(event);
-
         if (!event.defaultPrevented) {
-          if (!hasInteractedOutsideRef.current) context.triggerRef.current?.focus();
+          if (context.modal) context.triggerRef.current?.focus();
           // Always prevent auto focus because we either focus manually or want user agent focus
           event.preventDefault();
         }
-
-        hasInteractedOutsideRef.current = false;
-      }}
-      onInteractOutside={(event) => {
-        props.onInteractOutside?.(event);
-
-        if (!event.defaultPrevented) {
-          const originalEvent = event.detail.originalEvent as PointerEvent;
-          const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
-          const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
-
-          if (!context.modal || isRightClick) hasInteractedOutsideRef.current = true;
-        }
-
-        // Prevent dismissing when clicking the trigger.
-        // As the trigger is already setup to close, without doing so would
-        // cause it to close and immediately open.
-        //
-        // We use `onInteractOutside` as some browsers also
-        // focus on pointer down, creating the same issue.
-        const target = event.target as HTMLElement;
-        const targetIsTrigger = context.triggerRef.current?.contains(target);
-        if (targetIsTrigger) event.preventDefault();
       }}
     />
   ) : null;
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -259,6 +259,7 @@ const DropdownMenuRootContent = React.forwardRef<
   const { __scopeDropdownMenu, portalled = true, ...contentProps } = props;
   const context = useDropdownMenuContext(CONTENT_NAME, __scopeDropdownMenu);
   const menuScope = useMenuScope(__scopeDropdownMenu);
+  const isRightClickOutsideRef = React.useRef(false);
   return context.isRootMenu ? (
     <MenuPrimitive.Content
       id={context.contentId}
@@ -267,14 +268,18 @@ const DropdownMenuRootContent = React.forwardRef<
       {...contentProps}
       ref={forwardedRef}
       portalled={portalled}
-      onCloseAutoFocus={(event) => {
-        props.onCloseAutoFocus?.(event);
-        if (!event.defaultPrevented) {
-          if (context.modal) context.triggerRef.current?.focus();
-          // Always prevent auto focus because we either focus manually or want user agent focus
-          event.preventDefault();
-        }
-      }}
+      onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
+        if (context.modal && !isRightClickOutsideRef.current) context.triggerRef.current?.focus();
+        isRightClickOutsideRef.current = false;
+        // Always prevent auto focus because we either focus manually or want user agent focus
+        event.preventDefault();
+      })}
+      onPointerDownOutside={composeEventHandlers(props.onPointerDownOutside, (event) => {
+        const originalEvent = event.detail.originalEvent as PointerEvent;
+        const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
+        const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
+        isRightClickOutsideRef.current = isRightClick;
+      })}
     />
   ) : null;
 });
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -175,19 +175,16 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
               // prevent trigger focusing when opening
               // this allows the content to be given focus without competition
               if (!context.open) event.preventDefault();
+              // trigger always opens, and content click outside closes. since trigger is
+              // considered outside content, clicking trigger when open will close the content.
               context.onOpenChange(true);
             }
           })}
-          onClick={composeEventHandlers(props.onClick, (event) => {
-            // Handle anything that the browser considers a click for the element type if
-            // not using pointer e.g. Space keyup and Enter keydown
-            const isPointerClick = event.detail > 0;
-            if (!isPointerClick) context.onOpenChange(true);
-          })}
           onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
-            if (!disabled && event.key === 'ArrowDown') {
-              event.preventDefault();
+            if (!disabled && ['ArrowDown', 'Enter', ' '].includes(event.key)) {
               context.onOpenChange(true);
+              // Prevent keypresses from scrolling window
+              event.preventDefault();
             }
           })}
         />
@@ -259,7 +256,7 @@ const DropdownMenuRootContent = React.forwardRef<
   const { __scopeDropdownMenu, portalled = true, ...contentProps } = props;
   const context = useDropdownMenuContext(CONTENT_NAME, __scopeDropdownMenu);
   const menuScope = useMenuScope(__scopeDropdownMenu);
-  const isRightClickOutsideRef = React.useRef(false);
+  const hasInteractedOutsideRef = React.useRef(false);
   return context.isRootMenu ? (
     <MenuPrimitive.Content
       id={context.contentId}
@@ -269,16 +266,16 @@ const DropdownMenuRootContent = React.forwardRef<
       ref={forwardedRef}
       portalled={portalled}
       onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
-        if (context.modal && !isRightClickOutsideRef.current) context.triggerRef.current?.focus();
-        isRightClickOutsideRef.current = false;
+        if (!hasInteractedOutsideRef.current) context.triggerRef.current?.focus();
+        hasInteractedOutsideRef.current = false;
         // Always prevent auto focus because we either focus manually or want user agent focus
         event.preventDefault();
       })}
-      onPointerDownOutside={composeEventHandlers(props.onPointerDownOutside, (event) => {
+      onInteractOutside={composeEventHandlers(props.onInteractOutside, (event) => {
         const originalEvent = event.detail.originalEvent as PointerEvent;
         const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
         const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
-        isRightClickOutsideRef.current = isRightClick;
+        if (!context.modal || isRightClick) hasInteractedOutsideRef.current = true;
       })}
     />
   ) : null;
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -181,11 +181,9 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
             if (disabled) return;
             if (['Enter', ' '].includes(event.key)) context.onOpenToggle();
-            if (event.key === 'ArrowDown') {
-              context.onOpenChange(true);
-              // prevent arrow keypress from scrolling window
-              event.preventDefault();
-            }
+            if (event.key === 'ArrowDown') context.onOpenChange(true);
+            // prevent keypresses from scrolling window
+            if ([' ', 'ArrowDown'].includes(event.key)) event.preventDefault();
           })}
         />
       </MenuPrimitive.Anchor>
--- a/packages/react/tabs/src/Tabs.tsx
+++ b/packages/react/tabs/src/Tabs.tsx
@@ -181,11 +181,8 @@ const TabsTrigger = React.forwardRef<TabsTriggerElement, TabsTriggerProps>(
               context.onValueChange(value);
             }
           })}
-          onClick={composeEventHandlers(props.onClick, (event) => {
-            // Handle anything that the browser considers a click for the element type if
-            // not using pointer e.g. Space keyup and Enter keydown
-            const isPointerClick = event.detail > 0;
-            if (!isPointerClick) context.onValueChange(value);
+          onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
+            if ([' ', 'Enter'].includes(event.key)) context.onValueChange(value);
           })}
           onFocus={composeEventHandlers(props.onFocus, () => {
             // handle "automatic" activation if necessary
--- a/packages/react/tabs/src/Tabs.tsx
+++ b/packages/react/tabs/src/Tabs.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { createContextScope } from '@radix-ui/react-context';
-import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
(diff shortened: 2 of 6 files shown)


packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -278,25 +281,16 @@ const DropdownMenuRootContent = React.forwardRef<
...
+        const target = event.target as HTMLElement;
+        const isTriggerTarget = !context.triggerRef.current?.contains(target);
         props.onInteractOutside?.(event);
 
         if (!event.defaultPrevented) {
           const originalEvent = event.detail.originalEvent as PointerEvent;
           const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
           const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
-
-          if (!context.modal || isRightClick) hasInteractedOutsideRef.current = true;
```
jjenzz, line 288: This was preventing the trigger from receiving focus when user clicks trigger to close because we have logic in `onCloseAutoFocus` that only focuses the trigger if it wasn't an outside interaction.

I have omitted the trigger from being considered "outside".

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -168,19 +168,22 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
...
               // prevent trigger focusing when opening
               // this allows the content to be given focus without competition
               if (!context.open) event.preventDefault();
-              context.onOpenToggle();
+              context.onOpenChange(true);
             }
           })}
+          onClick={composeEventHandlers(props.onClick, (event) => {
+            // Handle anything that the browser considers a click for the element type if
+            // not using pointer e.g. Space keyup and Enter keydown
```
jjenzz, line 183: I wonder if we really need this comment anymore? The code is kinda self-documenting now.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -268,35 +269,11 @@ const DropdownMenuRootContent = React.forwardRef<
       portalled={portalled}
       onCloseAutoFocus={(event) => {
         props.onCloseAutoFocus?.(event);
-
         if (!event.defaultPrevented) {
-          if (!hasInteractedOutsideRef.current) context.triggerRef.current?.focus();
+          if (context.modal) context.triggerRef.current?.focus();
```
jjenzz, line 273: I realised I could delete all of the `onInteractOutside` code if I just checked for modality here. If it's non-modal then we just want the browser to focus whatever they clicked which will automatically happen because we `event.preventDefault()` our `onCloseAutoFocus` below.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -168,19 +168,22 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
...
-          onClick={composeEventHandlers(props.onClick, () => context.onOpenChange(true))}
           onPointerDown={composeEventHandlers(props.onPointerDown, (event) => {
             // only call handler if it's the left button (mousedown gets triggered by all mouse buttons)
             // but not when the control key is pressed (avoiding MacOS right click)
             if (!disabled && event.button === 0 && event.ctrlKey === false) {
               // prevent trigger focusing when opening
               // this allows the content to be given focus without competition
               if (!context.open) event.preventDefault();
-              context.onOpenToggle();
+              context.onOpenChange(true);
```
jjenzz, line 178: This change allows us to remove all the code here https://github.com/radix-ui/primitives/pull/1082/files#diff-5e95ef32039f278b6120df2487c18e5ada5be130bb884c0d36d89c0f20ee41ddL291-L299

Instead this happens:

- Pointer down opens
- Pointer down to close fires open event again but;
- Pointer down bubbles to document which triggers `onInteractOutside`
- Outside interaction closes it

Sooo, pointer down on trigger always opens, outside interaction always closes regardless of modality.

packages/react/tabs/src/Tabs.tsx
```diff
@@ -157,7 +156,6 @@ const TabsTrigger = React.forwardRef<TabsTriggerElement, TabsTriggerProps>(
     const triggerId = makeTriggerId(context.baseId, value);
     const contentId = makeContentId(context.baseId, value);
     const isSelected = value === context.value;
-    const handleTabChange = useCallbackRef(() => context.onValueChange(value));
```
jjenzz, line 160: Remove the `useCallbackRef` dep while I was here as I couldn't work out why it was necessary?

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: > Is it an issue that it focuses?

My concern is how this affects the next tab position and interferes with text selection / cursor position inside inputs.

jjenzz: Ah, yes, for some reason in my mind I expected that the thing I right clicked on **wouldn't** be focused because it was inert when I clicked it. But I can see the content in the context menu relates to the the thing they right clicked so it should probably be focused!

Okay, I'll add some logic back in to handle this.

jjenzz: Hmmm, this is why I found it weird:

https://user-images.githubusercontent.com/175330/149388758-dffaf212-622e-494e-9fff-5ec383bd96ac.mp4

When it's modal and you **click** outside, the trigger focuses. But when it's modal and you **right click** outside, the textarea focuses. Why the difference in expectation here?

andy-hook: > Why the difference in expectation here?

This was modelled after native behaviour if I recall.

https://user-images.githubusercontent.com/11708259/149390858-9363d32d-a9df-472c-a25d-9d0bf56a1b73.mp4

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -266,38 +267,26 @@ const DropdownMenuRootContent = React.forwardRef<
...
+        const originalEvent = event.detail.originalEvent as PointerEvent;
+        const ctrlLeftClick = originalEvent.button === 0 && originalEvent.ctrlKey === true;
+        const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
+        if (context.modal && isRightClick) {
+          // close after the browser context menu handler so that
+          // context menu options are inert (for the body)
+          document.addEventListener(
+            'contextmenu',
+            () => window.setTimeout(() => context.onOpenChange(false), 0),
+            { once: true }
```
jjenzz, line 285: I know we have added a `document.removeListener` in a cleanup effect when we do this in the past but I don't think it's needed here because it'll all happen as part of the same event (the right click will instantly trigger context menu event).

packages/react/dropdown-menu/src/DropdownMenu.tsx
jjenzz: Which app is that? MacOS Notes doesn't behave like ours on left click let alone right click so I'm not sure what we should be following:

https://user-images.githubusercontent.com/175330/149395660-a254f1c3-d65b-450b-8877-d205754f31bc.mp4

When I left click the menubar, it focuses the textarea 😅

jjenzz: Okay lads, I've pushed code that makes it behave like the [first video here](https://github.com/radix-ui/primitives/pull/1082#discussion_r784226074). I'm confused but if that's how it was before and for good reason then l'm happy to stay confused 😅

andy-hook: > Which app is that?

Yeah MacOS Notes

Native select is probably a clearer example:

https://user-images.githubusercontent.com/11708259/149396585-a1c7257b-673e-4bf1-a5a8-f24fca62c922.mp4

I agree with the point about consistency, but it was definitely something we discussed at the time and was an intentional decision. As a user I do find it more helpful than not.

jjenzz: Ah, that `select` is a great example, cheers 👍

Code reflects that now 💛

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: I really like this approach, though the mechanics of closing are a bit more implicit than we usually have with triggers. Do you think the code is clear enough as is? or would an inline comment explaining help us in the future?

packages/react/dropdown-menu/src/DropdownMenu.tsx
andy-hook: Would the inverse `isKeyboardClick` be accurate? something like that would make this comment totally redundant imo.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -168,19 +168,22 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
...
               // this allows the content to be given focus without competition
               if (!context.open) event.preventDefault();
-              context.onOpenToggle();
+              context.onOpenChange(true);
             }
           })}
+          onClick={composeEventHandlers(props.onClick, (event) => {
+            // Handle anything that the browser considers a click for the element type if
+            // not using pointer e.g. Space keyup and Enter keydown
+            const isPointerClick = event.detail > 0;
```
andy-hook, line 184: This is neat! 🤯

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -266,38 +268,18 @@ const DropdownMenuRootContent = React.forwardRef<
...
-        // cause it to close and immediately open.
-        //
-        // We use `onInteractOutside` as some browsers also
-        // focus on pointer down, creating the same issue.
-        const target = event.target as HTMLElement;
-        const targetIsTrigger = context.triggerRef.current?.contains(target);
-        if (targetIsTrigger) event.preventDefault();
-      }}
+      onCloseAutoFocus={composeEventHandlers(props.onCloseAutoFocus, (event) => {
+        if (context.modal && !isRightClickOutsideRef.current) context.triggerRef.current?.focus();
```
andy-hook, line 272: I think we still need to account for closing via `esc` in the non-modal case.

packages/react/tabs/src/Tabs.tsx
andy-hook: Good spot, looks unnecessary to me as well 🤔

packages/react/tabs/src/Tabs.tsx
```diff
@@ -176,22 +174,25 @@ const TabsTrigger = React.forwardRef<TabsTriggerElement, TabsTriggerProps>(
...
-              handleTabChange();
+              context.onValueChange(value);
             }
           })}
+          onClick={composeEventHandlers(props.onClick, (event) => {
+            // Handle anything that the browser considers a click for the element type if
+            // not using pointer e.g. Space keyup and Enter keydown
+            const isPointerClick = event.detail > 0;
+            if (!isPointerClick) context.onValueChange(value);
+          })}
```
benoitgrelard, line 189: I'm really in 2 minds since we've added these `onClick` handlers to all these :(
I get the "why" but I keep having this niggling feeling that ultimately we end up with **more** code and code that is **less explicit** than just implementing the "positive" /explicit cases.

I've felt like that in my Select PR's too where I'm doing this but then having to undo some behaviour anyway.

Am I the only one who feels like this?

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: Just adding my 2 cents here that yes, this behaviour was on purpose as I had shown the same types of native examples to @andy-hook back then, right-click is treated differently.

packages/react/tabs/src/Tabs.tsx
andy-hook: > I get the "why" but I keep having this niggling feeling that ultimately we end up with more code and code that is less explicit than just implementing the "positive" /explicit cases.

Just to be clear, is your thought that it might be clearer to simply stick with the original explicit keyboard binding over usage of `onClick`?

  Are we functionally gaining anything by using `onClick` in place of? does it help us in any other ways beyond the keyboard stuff?

benoitgrelard: Yes that's what I meant, before introducing `onClick` we were reproducing the behaviour that `onClick` gives in some cases.

But the thing is I feel like over time we realising in multiple components that we don't always want "always all " that behaviour, so we end up guarding against certain conditions, or undoing certain behaviours.

So I feel like when I try to follow the code, it's actually less clear and less explicit in most cases now.

andy-hook: Right, yes, I see your point. I guess we need to step back and weigh the benefits to making `onClick` work for us vs the explicit binding stuff. For some reason I thought there was an element of [library interop involved](https://github.com/radix-ui/primitives/issues/980) which `onClick` would make easier. Perhaps @jjenzz has more thoughts.

jjenzz: > feeling that ultimately we end up with **more code**

I'm not getting that feeling personally. Perhaps it feels that way because our implementation before didn't match what `onClick` handles? 

> does it help us in any other ways beyond the keyboard stuff?

It mostly just helps because we rely on the browser handling instead of coding it ourselves (incorrectly). For example, we were opening on space **keydown** before, but browsers click buttons on space **keyup** and enter **keydown**. And this behaviour changes per node type, which is worth considering...

If someone changed the node to an anchor, space keypress doesn't click them because that's how the browser decides anchors should behave. So, do we want the behaviour to respect the underlying node type and use `onClick` or do we want to say that it should always open via space keypress regardless of node type?

I can put the following code in instead so we're ensuring the same `button` keyboard behaviour for all node types, but then does that mean we need to add `event.preventDefault()` back in to prevent windows from scrolling for all node types? We agreed we wouldn't add code to support a change in node type so where do we draw the line?

```jsx
onKeyDown={composeEventHandlers(props.onKeyDown, event => {
  if (event.key === 'Enter') context.onOpen();
})}
onKeyUp={composeEventHandlers(props.onKeyUp, event => {
  if (event.key === ' ') context.onOpen();
})}

// vs 
onClick={composeEventHandlers(props.onClick, event => {
  if (event.button === 0) context.onOpen();
})}
```

benoitgrelard: But that's where I'm not sure I agree, because ultimately we are not after a "click" behaviour on a button. If you look at native select, or native dropdown menus in MacOS, for example it's space keydown, and no enter key at all to open. The only reason we added enter key, is because some people asked for it.

So I just feel like somehow we're trying to fit in the native handling of a single button into completely different patterns with a different set of requirements.

benoitgrelard: > If someone changed the node to an anchor, space keypress doesn't click them because that's how the browser decides anchors should behave. So, do we want the behaviour to respect the underlying node type and use onClick or should it always open via space keypress regardless of node type?

I can see this for component that could end up with an anchor trigger, like `HoverCard`, I think yeah that's a valid point. But I don't see this applying to `DropdownMenu` for example, it just wouldn't make sense as part of the pattern.

jjenzz: > I can see this for component that could end up with an anchor trigger, like `HoverCard`, I think yeah that's a valid point. But I don't see this applying to `DropdownMenu` for example, it just wouldn't make sense as part of the pattern.

I agree, so perhaps we should evaluate here which components require which approach instead of making it consistent across the board. They have separate concerns.

- `Tooltip` - `onClick` so it closes when button or anchor is considered "clicked"
- `DropdownMenu` - `onKeyDown` so it behaves same for all node types
-  `Tabs` - ?

benoitgrelard: Yeah it seems we should handle them one by one for what they are, I agree.

> Also, what about the event.preventDefault() question?

You mean for preventing to scroll when pressing space?
I guess same, we should only do it when it's necessary?

jjenzz: > I guess same, we should only do it when it's necessary?

When is it necessary though? 

Right now it's not necessary in any of our components because the triggers are `button`s but if someone changes them to an `a` or a`<div role="button">` then it will be necessary. 

I recall we decided against adding code to support changes to the underlying node though so do we just leave it out?

benoitgrelard: I think is the use is normally a button then we shouldn't do anything (unless it's necessary to be patched in some browsers). We should only consider doing it for primitives where the trigger may not be a button I suppose.

jjenzz: Sweet, I'll mark this as resolved then. I've gone with `onKeyDown` for `Tabs`.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -168,23 +168,23 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
...
+              context.onOpenChange(true);
             }
           })}
           onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
-            if (!disabled && event.key === 'ArrowDown') {
-              event.preventDefault();
+            if (!disabled && ['ArrowDown', 'Enter', ' '].includes(event.key)) {
               context.onOpenChange(true);
+              // Prevent keypresses from scrolling window
+              event.preventDefault();
```
jjenzz, line 187: have to prevent default here for `ArrowDown` keypress mostly.

andy-hook (approved):

jjenzz:
@benoitgrelard I was about to merge this as I have an approval but wondered if you wanted to take one last look as well before I do?

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -266,38 +265,18 @@ const DropdownMenuRootContent = React.forwardRef<
...
-
-        // Prevent dismissing when clicking the trigger.
-        // As the trigger is already setup to close, without doing so would
-        // cause it to close and immediately open.
-        //
-        // We use `onInteractOutside` as some browsers also
-        // focus on pointer down, creating the same issue.
-        const target = event.target as HTMLElement;
-        const targetIsTrigger = context.triggerRef.current?.contains(target);
-        if (targetIsTrigger) event.preventDefault();
```
benoitgrelard, line 299: Going back to my earlier comment again (and it seems @andy-hook's comment too), I'm not sure this is the way to go.

I believe it comes down to whether we consider the trigger as only opening vs. toggling.
If I imagine a non-modal dropdown menu, or a non-modal popover, and I don't want clicking outside to close (say I disabled it), then, would I still want/expect clicking the trigger to close the menu?

I feel like I would and that's part of it's functionality.

benoitgrelard (commented):
I've left a few comments.
Also, depending on what we decide to do, wouldn't similar changes be needed in other components like `Popover` and `Dialog` as they can also be non-modal? (Although testing them currently doesn't reveal the same bug we're addressing here so maybe not)

jjenzz:
> Aren't we relying on the order of events then here when non-modal?

[Yep](https://github.com/radix-ui/primitives/pull/1082#discussion_r783902088), but now non-modal and modal will behave the same, whereas the logic we had could before behaved different depending on modality.

> Could there ever be a case where it's fragile and the events happens in a different order instead?

A pointer outside is a document event, so will always happen after the trigger events (bubble) regardless of whether it has `setTimeout`. 

The order would only be broken it we do `setTimeout` in the trigger pointer event, but then that's the same with any event handling, i.e, you wouldn't put a `setTimeout` in an `onPointerDown` if you want it to fire before focus.

benoitgrelard:
> > Aren't we relying on the order of events then here when non-modal?
> 
> [Yep](https://github.com/radix-ui/primitives/pull/1082#discussion_r783902088), but now non-modal and modal will behave the same (and the bug is fixed), whereas the logic we had before behaved different depending on modality.
> 
> > Could there ever be a case where it's fragile and the events happens in a different order instead?
> 
> A pointer outside is a document event, so will always happen after the trigger events (bubble) regardless of whether it has `setTimeout`.
> 
> The order would be broken it we do `setTimeout(() => {}, 1)` in the trigger pointer event, but then that's the same with any event handling, i.e, you wouldn't put a `setTimeout` in an `onPointerDown` if you want it to fire before focus.
> 
> > * it's open
> > * pointer down outside of content: `onOpenChange(false)`, closes
> > * pointer down on trigger: `onOpenChange(true)`, opens
> > * still opened
> 
> I'm not really following this example, it looks correct to me except last step would be "re-open".
> 
> If the pointer happens outside first it will have closed before they've had a chance to move the mouse and pointer down on the trigger unless we did something like `setTimeout(() => {}, 3000)` in `onPointerDownOutside` which we wouldn't.
> 
> > [...] and I don't want clicking outside to close (say I disabled it), then, would I still want/expect clicking the trigger to close the menu?
> 
> This is a good question 🤔 I'm not sure of the answer here. But I'll revisit this in a sec and see if I can achieve what you're asking without re-introducing the bug that was reported.

Yeah I think I'm with you regarding this part.
What do you think of this though: https://github.com/radix-ui/primitives/pull/1082#discussion_r789730996

jjenzz:
> What do you think of this though: #1082 (comment)

The last bit of my reply that you quoted addressed that bit, sorry, there was a lot of waffle there that drowned it out 😅

benoitgrelard:
>This is a good question 🤔 I'm not sure of the answer here. But I'll revisit this in a sec and see if I can achieve what you're asking without re-introducing the bug that was reported.

Cool, I know we had discussed this previously as it had come up for Dialog, and I know @colmtuite wanted them to be toggling. From what I can see in our codebase, all triggers are currently toggling.

I think imagining it from the perspective of a case where you've said clicking outside doesn't close, helps a bit.

packages/react/dropdown-menu/src/DropdownMenu.tsx
jjenzz, line 299: I've added the `onOpenToggle` handlers back in and everything seems to be working fine without this.

Can anyone recall which browser this was an issue in? I've tested both modal and non-modal in Firefox, Chrome, Brave, Safari, iOS Safari and Edge.

packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -182,9 +179,12 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
...
           })}
           onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
-            if (!disabled && event.key === 'ArrowDown') {
-              event.preventDefault();
+            if (disabled) return;
+            if (['Enter', ' '].includes(event.key)) context.onOpenToggle();
+            if (event.key === 'ArrowDown') {
               context.onOpenChange(true);
+              // prevent arrow keypress from scrolling window
+              event.preventDefault();
```
benoitgrelard, line 187: I think we need this for space as well as it looks like the page scrolls when using space in non-modal case (must be similar reason to what I was showing you in `Select`).

packages/react/dropdown-menu/src/DropdownMenu.tsx
benoitgrelard: I can't recall anything specific, but it seems to work well.

benoitgrelard (commented):
This seems to work well.
If this is the way to go we should do the same in other primitives, as I'm pretty sure we have that similar trigger check in onInteractOutside in other places.

jjenzz:
> I can't recall anything specific, but it seems to work well.

I think I've figured this out... it seems historically some of these components did use to use `onClick`, and `Popover` and `Dialog` still do (correctly). This `onInteractOutside` logic is necessary if the trigger uses `onClick` because `onInteractOutside` is pointer down:

- mouse down on trigger fires interact outside and closes
- mouse up on trigger fires onclick and opens

So, for these PRed cases where we no longer use `onClick` it is fine to remove the `onInteractOutside` workaround but for `Popover` and `Dialog` it will need to remain.

benoitgrelard (approved):

afrokick:
@jjenzz sorry for mention you, but maybe you had an issue with mobile users.

When a mobile user scroll the page, onPointerDown trigger select to be open=true and user will see opened dropdown. But the user just want to scroll page down.

Any suggestion how to solve this issue?

jjenzz:
@afrokick i'd suggest raising a separate issue with a codesandbox reproduction for the team to take a look at.

PR #1097, opened 2022-01-17
[DropdownMenu] Stop menu from preventing tab keypress

### LINKED ISSUE #1096: [Dropdown][AlertDialog][Dialog] Focus behavior is broken when AlertDialog/Dialog is used inside a Dropdown
### Current Behavior

When a `AlertDialog` or `Dialog` is inside a `Dropdown`, tabbing between focusable elements is broken. The last focusable element seems to be focused and you can only tab through the first and last elements.

CleanShot 2022-01-17 at 08 37 46

### Expected behavior

1. The first focusable element inside `<Dialog.Content>` or `<AlertDialog.Content>` should be focused when it opens
2. Tabbing through all focusable elements should work
3. Focus should be trapped to `.Content`

### Reproducible example

[Here's a CodeSandbox reproducing this issue.](https://codesandbox.io/s/gracious-snyder-i230o?file=/src/App.js)

[CodeSandbox Template](https://codesandbox.io/s/2r30e)

### Additional context

I've adapted this example from [Jenna's solution](https://twitter.com/jjenzz/status/1478375014474960897) for a [tweet I made a few weeks ago](https://twitter.com/triangulo_dev/status/1478366865428819970).

### Your environment

| Radix Package(s) |  @radix-ui/react-alert-dialog, @radix-ui/react-dialog, @radix-ui/react-dropdown-menu        |  0.1.5, 0.1.5, and 0.1.4       |
| React            | n/a     |  17.0.2       |
| Browser          |  Chrome       |  97.0.4692.71       |
| Browser          |  Firefox Developer Edition       |  96.0b10       |
| Browser          |  Firefox Developer Edition       |  97.0b4      |
| Browser          |  Safari       |  15.1 (17612.2.9.1.20)      |
| Operating System |  macOS       |  12.0.1 (21A559)       |

jjenzz:
Fixes #1096

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -608,8 +608,11 @@ const MenuContentImpl = React.forwardRef<MenuContentImplElement, MenuContentImpl
                       const target = event.target as HTMLElement;
                       const isKeyDownInside = event.currentTarget.contains(target);
                       const isModifierKey = event.ctrlKey || event.altKey || event.metaKey;
-                      if (isKeyDownInside && !isModifierKey && event.key.length === 1) {
-                        handleTypeaheadSearch(event.key);
+                      const isCharacterKey = event.key.length === 1;
+                      if (isKeyDownInside) {
+                        // menus should not be navigated using tab key so we prevent it
+                        if (event.key === 'Tab') event.preventDefault();
+                        if (!isModifierKey && isCharacterKey) handleTypeaheadSearch(event.key);
                       }
                       // focus first/last item based on key pressed
                       const content = contentRef.current;
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -611,8 +611,6 @@ const MenuContentImpl = React.forwardRef<MenuContentImplElement, MenuContentImpl
                       if (isKeyDownInside && !isModifierKey && event.key.length === 1) {
                         handleTypeaheadSearch(event.key);
                       }
-                      // menus should not be navigated using tab key so we prevent it
-                      if (event.key === 'Tab') event.preventDefault();
                       // focus first/last item based on key pressed
                       const content = contentRef.current;
                       if (event.target !== content) return;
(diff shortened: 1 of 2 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -611,8 +611,6 @@ const MenuContentImpl = React.forwardRef<MenuContentImplElement, MenuContentImpl
                       if (isKeyDownInside && !isModifierKey && event.key.length === 1) {
                         handleTypeaheadSearch(event.key);
                       }
-                      // menus should not be navigated using tab key so we prevent it
-                      if (event.key === 'Tab') event.preventDefault();
```
jjenzz, line 615: Because of events bubbling through portals, this was preventing tab keypress in `Dialog`s opened from within a dropdown.

I'm not sure why we ever needed this though because menus use `RovingFocusGroup` so everything will be `tabindex={-1}` apart from the active item so tab keypress wouldn't do anything anyway.

andy-hook: Good catch!

> I'm not sure why we ever needed this line?

> Menus use RovingFocusGroup so everything will be tabindex="-1" apart from the active item meaning tab keypress wouldn't do anything anyway.

This is only true if we’re trapping focus, otherwise tabbing will skip us to the next tabbable element outside the list e.g. when non-modal or in a submenu.

https://user-images.githubusercontent.com/11708259/149922836-2e58a4bf-de24-45d2-a4e3-90b6925248bb.mp4

benoitgrelard: > …but I'm not sure why we ever needed this line?

If my memory isn't failing me, I remember discussing this with you and we found said that if it was a requirement from the component to behave as a single tab-stop (and comparing to native ones) that tab doesn't do anything, even though it was kinda working out of the box with roving, we thought it was too implicit and therefore we decided to be explicit about it.

If we wanted to still stick with this rationale, we could still do that explicitly but with a DOM event rather than react event?

jjenzz: > we thought it was too implicit and therefore we decided to be explicit about it.

Your memory is so much better than mine. I don't recall any of that 😅  I'll add the `if (isKeyDownInside && event.key === 'Tab')` as mentioned above because that variable already exists.

benoitgrelard (approved):

PR #1098, opened 2022-01-17
[Dialog][AlertDialog] Add aria-describedby only when a description is provided

### LINKED ISSUE #1094: [Dialog] aria-describedby included when no description
### Current Behavior

When no `DialogPrimitiv.Description` and no `aria-describedby` on `Dialog.Content` there is a `aria-describedby`attribute is on the dialog root.

image
image

### Expected behavior

No `aria-describedby`attribute on dialog root.

### Reproducible example

[No description CodeSandbox Template](https://codesandbox.io/s/jolly-alex-iq1b3)

Or you can just open the CodeSandbox from https://www.radix-ui.com/docs/primitives/components/dialog and delete the `DialogPrimitiv.Description`.

### Additional context

It appears `aria-describedby` is optional.

[Dialog roles](https://www.w3.org/TR/wai-aria-practices-1.2/#dialog_roles_states_props)

> Optionally, the aria-describedby property is set on the element with the dialog role to indicate which element or elements in the dialog contain content that describes the primary purpose or message of the dialog.

### Your environment

| Radix Package(s) |    dialog     |  0.1.5       |

luisorbaiceta:
### Description

Fixes #1094

Right [now](https://github.com/radix-ui/primitives/blob/28c4fd981319992458757d860c26942fed703419/packages/react/dialog/src/Dialog.tsx#L82) the `dialog` component always generates a `descriptionId` even if no description component is registered.

This leads to [always](https://github.com/radix-ui/primitives/blob/28c4fd981319992458757d860c26942fed703419/packages/react/dialog/src/Dialog.tsx#L399) generate an `aria-describedby` even if no description has been added.

### TODO

- [x] Throw for missing label
- [x] Warn for missing description
- [x] Add test cases

### Notes for docs

- 🔥 `aria-label` will be ignored, use `Title` with `<VisuallyHidden asChild>`
- ✨  Use `aria-describedby={undefined}` to silence missing description warning

### DIFF
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -366,16 +366,7 @@ interface DialogContentImplProps extends Omit<DismissableLayerProps, 'onDismiss'
 
 const DialogContentImpl = React.forwardRef<DialogContentImplElement, DialogContentImplProps>(
   (props: ScopedProps<DialogContentImplProps>, forwardedRef) => {
-    const {
-      __scopeDialog,
-      'aria-label': ariaLabel,
-      'aria-labelledby': ariaLabelledBy,
-      'aria-describedby': ariaDescribedBy,
-      trapFocus,
-      onOpenAutoFocus,
-      onCloseAutoFocus,
-      ...contentProps
-    } = props;
+    const { __scopeDialog, trapFocus, onOpenAutoFocus, onCloseAutoFocus, ...contentProps } = props;
     const context = useDialogContext(CONTENT_NAME, __scopeDialog);
     const contentRef = React.useRef<HTMLDivElement>(null);
     const composedRefs = useComposedRefs(forwardedRef, contentRef);
@@ -396,21 +387,17 @@ const DialogContentImpl = React.forwardRef<DialogContentImplElement, DialogConte
           <DismissableLayer
             role="dialog"
             id={context.contentId}
-            aria-describedby={ariaDescribedBy || context.descriptionId}
-            // If `aria-label` is set, ensure `aria-labelledby` is undefined as to avoid confusion.
-            // Otherwise fallback to an explicit `aria-labelledby` or the ID used in the
-            // `DialogTitle`
-            aria-labelledby={ariaLabel ? undefined : ariaLabelledBy || context.titleId}
-            aria-label={ariaLabel || undefined}
+            aria-describedby={context.descriptionId}
+            aria-labelledby={context.titleId}
             data-state={getState(context.open)}
             {...contentProps}
             ref={composedRefs}
             onDismiss={() => context.onOpenChange(false)}
           />
         </FocusScope>
-        {process.env.NODE_ENV === 'development' && (
+        {['development', 'test'].includes(process.env.NODE_ENV!) && (
           <>
-            <LabelWarning contentRef={contentRef} />
+            <TitleWarning contentRef={contentRef} />
             <DescriptionWarning contentRef={contentRef} />
           </>
         )}
@@ -491,9 +478,9 @@ function getState(open: boolean) {
   return open ? 'open' : 'closed';
 }
 
-const LABEL_WARNING_NAME = 'DialogLabelWarning';
+const TITLE_WARNING_NAME = 'DialogTitleWarning';
 
-const [WarningProvider, useWarningContext] = createContext(LABEL_WARNING_NAME, {
+const [WarningProvider, useWarningContext] = createContext(TITLE_WARNING_NAME, {
   contentName: CONTENT_NAME,
   titleName: TITLE_NAME,
   docsSlug: 'dialog',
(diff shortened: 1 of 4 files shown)


luisorbaiceta:
Another solution would be to query for a predefined attribute in the Description component and generate the `aria-describedBy` if there is a match... imo this is not as clean as the suggested above because we could have naming collisions if the developer used that attribute name...

jjenzz:
This is a tricky one because the proposed solutions depend on a fix happening client-side. Ideally, the a11y would be available server-side for optimum time to interactive for screen readers.

I'm not sure if that's possible though? 😬

benoitgrelard:
> This is a tricky one because the proposed solutions depend on a fix happening client-side. Ideally, the a11y would be available server-side for optimum time to interactive for screen readers.
> 
> I'm not sure if that's possible though? 😬

Yeah I was thinking about SSR too…
Maybe if we add it by default, but then check in an effect and remove it if the description isn't there?

luisorbaiceta:
Yep, inverting the logic should work fine

luisorbaiceta:
Seems that removing the `aria-describedBy` when no description has been added has exposed an a11y violation when no `Title` component is registered. I'll expose the problem in this push. Should this be enforced somehow or documented to prevent users from accessibility violations? Something like:

> ## Title
An accessible name to be announced when the dialog is opened. Alternatively, you can provide aria-label or aria-labelledby to Dialog.Content and exclude this component. Note that one of the above should be provided in order to make the dialog accesible

benoitgrelard:
> Seems that removing the `aria-describedBy` when no description has been added has exposed an a11y violation when no `Title` component is registered. I'll expose the problem in this push. Should this be enforced somehow or documented to prevent users from accessibility violations? Something like:
> 
> > ## Title
> > An accessible name to be announced when the dialog is opened. Alternatively, you can provide aria-label or aria-labelledby to Dialog.Content and exclude this component. Note that one of the above should be provided in order to make the dialog accesible

We already provide an explicit warning if there is no title or no `aria-label` I think.

jjenzz:
> Maybe if we add it by default, but then check in an effect and remove it if the description isn't there?

But that doesn't fix the original bug report then? SSR without a `Description` would incorrectly have the attribute.

luisorbaiceta:
> Welcome to the world of Radix Primitives @luisorbaiceta ... seemingly simple things never are 😅

😂

IMHO It would be good to enforce/encourage users to use the components `Title` and `Description` rather than adding any logic to the current implementation. It might be less flexible but ensures a correct a11y strategy

jjenzz:
> IMHO It would be good to enforce/encourage users to use the components Title and Description rather than adding any logic to the current implementation.

I was thinking about this too but `Description` isn't required for a11y. We could definitely throw for `Title` though and allow them to silence it with a `visuallyHidden` prop or something. We can remove our `aria-label` logic as well then.

The `Description` is trickier.

jjenzz:
Perhaps it's worth testing how SRs behave if someone did the following? 😬

```jsx
<Description srOnly /> {/* empty */}
```

luisorbaiceta:
> Perhaps it's worth testing how SRs behave if someone did the following? 😬
> 
> ```js
> <Description srOnly /> {/* empty */}
> ```
> 
> Then we could error if description isn't mounted, and they can hide it. But that doesn't feel very intuitive 🙁

Maybe it's a good move that if the Description isn't required for a11y, we leave it's implementation to the user, and we avoid dangling auto generated ids

jjenzz:
Perhaps we leave things as they are but we do something like this in dev mode:

```jsx
React.useEffect(() => {
  if (process.env.NODE_ENV !== 'development') {
    const content = document.getElementById(contentId);
    if (!content) {
      console.warn("If you do not need a `Description`, please pass `aria-describedby={undefined}` to `Dialog.Content` to ensure correct accessibility");
    }
  }
});
```

benoitgrelard:
> Perhaps we leave things as they are for description but tweak `Title` as described above, and then do something like this in `Dialog.Content` in dev mode:
> 
> ```js
> React.useEffect(() => {
>   if (process.env.NODE_ENV !== 'development') {
>     const description = document.getElementById(descriptionId);
>     if (!description) {
>       console.warn("Warning: Missing `Description` or `aria-describedby={undefined}` for `Dialog.Content`.");
>     }
>   }
> }, []);
> ```
> 
> So these are the accessible approaches:
> 
> ```js
> <>
>   <AlertDialog.Content aria-describedby={undefined}>
>     <AlertDialog.Title>Are you sure?</AlertDialog.Title>
>   </AlertDialog.Content>
> 
>   <AlertDialog.Content>
>     <AlertDialog.Title srOnly>Are you sure?</AlertDialog.Title>
>     <AlertDialog.Description>Your changes cannot be undone.</AlertDialog.Description>    
>   </AlertDialog.Content>
> 
>   <AlertDialog.Content>
>     <AlertDialog.Title>Are you sure?</AlertDialog.Title>
>     <AlertDialog.Description>Your changes cannot be undone.</AlertDialog.Description>
>   </AlertDialog.Content>
> </>
> ```

Yeah I think that makes sense.
The only thing I'm not sure about is the `srOnly` prop when we expose `VisuallyHidden` publicly already.
We could just as well have a section about labelling in the docs, and show that for this use case (no visual title, but an accessible one) they can just wrap the `Title` part in `VisuallyHidden` themselves.
It's also better for bundle-size as most people won't do this.

jjenzz:
> We could just as well have a section about labelling in the docs, and show that for this use case (no visual title, but an accessible one) they can just wrap the `Title` part in `VisuallyHidden` themselves.
It's also better for bundle-size as most people won't do this.

The `srOnly` prop is to silence our internal warnings about a missing title and and then add `VisuallyHidden` for them.

jjenzz:
> We could just as well have a section about labelling in the docs, and show that for this use case (no visual title, but an accessible one) they can just wrap the `Title` part in `VisuallyHidden` themselves.
It's also better for bundle-size as most people won't do this.

Good point. Let's go with that. We can update the missing title warning to explain that they can wrap it in a `VisuallyHidden` if they don't want it to be visible.

luisorbaiceta:
Great! I'll go for it then!

luisorbaiceta:
I'm getting some weird behavior with tests. If no description has been provided, `axe` does not complain at all for the lack of a valid label. If a `Dialog.Description` is registered, then I get the `Invalid ARIA attribute value: aria-labelledby="radix-N"`

packages/react/dialog/src/Dialog.tsx
```diff
@@ -520,6 +525,23 @@ For more information, see https://radix-ui.com/primitives/docs/components/${labe
...
+  const descriptionWarningContext = useWarningContext(DESCRIPTION_WARNING_NAME);
+
+  const MESSAGE = `Warning: Missing \`Description\` or \`aria-describedby={undefined}\` for {${descriptionWarningContext.contentName}}.`;
+
+  React.useEffect(() => {
+    const hasDescription = document.getElementById(
+      contentRef.current?.getAttribute('aria-describedby')!
+    );
+    if (!hasDescription) console.warn(MESSAGE);
+  }, [MESSAGE, contentRef]);
```
jjenzz, line 540: It looks like this would still warn if they passed `aria-describedby={undefined}`. Maybe need an extra check?

```suggestion
  React.useEffect(() => {
    const describedById = contentRef.current?.getAttribute('aria-describedby');
    const hasDescription = document.getElementById(describedById);
    if (describedById && !hasDescription) console.warn(MESSAGE);
  }, [MESSAGE, contentRef]);
```

jjenzz (changes requested):
Thanks for this. Love that you've included tests! 

I'm not entirely sure if it's ready yet but here are some initial bits... 🙂

packages/react/dialog/src/Dialog.tsx
```diff
@@ -396,19 +387,20 @@ const DialogContentImpl = React.forwardRef<DialogContentImplElement, DialogConte
...
+            aria-describedby={context.descriptionId}
+            aria-labelledby={context.titleId}
             data-state={getState(context.open)}
             {...contentProps}
             ref={composedRefs}
             onDismiss={() => context.onOpenChange(false)}
           />
         </FocusScope>
-        {process.env.NODE_ENV === 'development' && <LabelWarning contentRef={contentRef} />}
+        {['development', 'test'].includes(process.env.NODE_ENV!) && (
```
jjenzz, line 398: I've always wondered about this. We do `process.env.NODE_ENV === 'development'` everywhere but should we be doing `process.env.NODE_ENV !== 'production'` instead?

packages/react/dialog/src/Dialog.tsx
```diff
@@ -486,35 +478,51 @@ function getState(open: boolean) {
...
+  return null;
+};
+
+const DESCRIPTION_WARNING_NAME = 'DialogDescriptionWarning';
+
+const DescriptionWarning: React.FC<WarningProps> = ({ contentRef }) => {
+  const descriptionWarningContext = useWarningContext(DESCRIPTION_WARNING_NAME);
+
+  const MESSAGE = `Warning: Missing \`Description\` for {${descriptionWarningContext.contentName}}.`;
+
```
jjenzz, line 519: Looks like your changes to this message have disappeared? It's missing the instructions you had before for adding `aria-labelledby={undefined}` which I think we still need for users that don't want to render a description.

jjenzz (approved):

luisorbaiceta:
> Thanks @luisorbaiceta 🎉

Anytime!

PR #1101, opened 2022-01-19
[Tooltip] if skipDelayDuration is 0 don't ever skip delay

hipstersmoothie:
### Description

Before this change the `TooltipProvider` would skip the open delay even though we set `skipDelayDuration` to 0.

Before (with 1000ms `delayDuration`):

https://user-images.githubusercontent.com/1192452/150194900-d807f8b1-830f-42f9-a412-a1a37e9d46eb.mp4

After:

https://user-images.githubusercontent.com/1192452/150195130-5d60aaa5-1f66-43b4-a640-9a2864bd3bdd.mp4

### DIFF
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -85,7 +85,10 @@ const TooltipProvider: React.FC<TooltipProviderProps> = (
       delayDuration={delayDuration}
       onOpen={React.useCallback(() => {
         window.clearTimeout(skipDelayTimerRef.current);
-        setIsOpenDelayed(false);
+
+        if (skipDelayDuration !== 0) {
+          setIsOpenDelayed(false);
+        }
       }, [])}
       onClose={React.useCallback(() => {
         window.clearTimeout(skipDelayTimerRef.current);


jjenzz:
I think the **after** video is correct so we must have had a bug before.

The `skipDelayDuration` is:

> How much time a user has to enter another trigger without incurring a delay again.

So, if it was set to `1000` and moved to the next trigger within one second, it would instantly open. If they moved to the next trigger after one second, it would incur the delay duration again.

In your case, you're giving the user `0` milliseconds before incurring the delay again so they will always incur the delay.

I think that's correct, but @benoitgrelard can confirm.

benoitgrelard:
> I think the **after** video is correct so we must have had a bug before.
> 
> The `skipDelayDuration` is:
> 
> > How much time a user has to enter another trigger without incurring a delay again.
> 
> So, if it was set to `1000` and user moved to the next trigger within one second, it would instantly open. If they moved to the next trigger after one second, it would incur the delay duration again.
> 
> In your case, you're giving the user `0` milliseconds to move to the next trigger before incurring the delay again so they will always get the delay.
> 
> I think that's correct, but @benoitgrelard can confirm.

Yes, @jjenzz is correct, this is what `skipDelayDuration` controls.

Furthermore, I'm not sure which "before" version you are referring to, but that functionality has always been the same as far as I know.

Perhaps the confusion was that in previous version there wasn't a `Tooltip.Provider` part, and instead `skipDelayDuration` lived on each instance of `Tooltip.Root`. I went back to a version without the provider, and set it to 0 on all instances, and it still works the same way as now (ie. if `0`, you're telling it there is no "grace" time to get to another trigger without having the delay again).

I will close the PR now, but feel free to keep discussing here.

hipstersmoothie:
> In your case, you're giving the user 0 milliseconds to move to the next trigger before incurring the delay again so they will always get the delay.

Yeah this is exactly what I want, to always incur the delay.

[Here is a sandbox](https://codesandbox.io/s/lucid-tree-fz6gp?file=/App.js) that displays this is not true if the triggers are siblings with no space between.

jjenzz:
Ooooh apologies, I thought your **after** video was the result of `skipDelayDuration={0}`.

I can see the issue now.

jjenzz:
It's interesting that it behaves correctly when there is a spacer between but not when they're right next to each other. This makes me think there is something wrong with our events and that the proposed fixed might just be a workaround to an underlying issue 🤔

jjenzz:
Yep, I can see the issue I think. 

The proposed fix would not work if someone set `skipDelayDuration={0.3}` or something small. Obviously, people aren't likely to do that but it highlights that the fix isn't fixing the underlying cause. It takes 0.4ms to move in this video but I don't incur the delay:

https://user-images.githubusercontent.com/175330/150522090-01df491f-7ed0-4bb6-9111-0c3020f95737.mp4

The issue is that we rely on a re-render to happen before the other tooltips can know if they need to delay opening or not, but when tooltips are right next to each other, the `open` event could fire on the next tooltip before the `setIsOpenDelayed(true)` has been set and re-rendered. We somehow need to fix that.

@benoitgrelard I've tested this in the old version and it seems it was an issue there too so need to get our thinking caps on for this one.

jjenzz:
@hipstersmoothie if you're happy to see if you can figure a way to solve the events issue, we'd be super grateful. Otherwise, feel free to close this PR and open a Github Issue for us to investigate later.

hipstersmoothie:
I'll close. Getting low values for setTimeout to work accurately is gonna be a challenge! Excited to see what you all come up with.

https://github.com/radix-ui/primitives/issues/1110

jjstobbe:
@jjenzz  I'm experiencing this same issue, and I'm not sure there is a valid workaround described here? Is there any chance we can re-open this PR and get this merged?

I've noticed that the `NavigationMenu` has incredibly similar logic to what this PR proposes [[here](https://github.com/radix-ui/primitives/blob/eca6babd188df465f64f23f3584738b85dba610e/packages/react/navigation-menu/src/NavigationMenu.tsx#L125-L129)]. If it works for the `NavigationMenu`, I'm not sure why it wouldn't also be fine for the `Tooltip` as well.

jjenzz:
@jjstobbe the team cannot merge this pr bcos it does not fix the underlying cause as [pointed out here](https://github.com/radix-ui/primitives/pull/1101#issuecomment-1018442666).

the necessary fix may well be something similar to what the `NavigationMenu` but there hasn't been any resources allocated to investigating it. feel free to raise a pr yourself if you're able to find an appropriate fix that handles small delays.

PR #1107, opened 2022-01-21
[Presence] Account for early node removal

andy-hook:
I ran across this issue when using the render prop to pass `present` to a node outside of the `<Presence/>` tree. If the node has a long running animation and is removed early i.e. its parent unmounts, then `present` remains `true`. My expectation is that it would flip to `false` when the node is removed.

The `animationcancel` event does fire on node removal, but as we cleanup those handlers inside the effect they don't appear to run when the node is cleared.

https://codesandbox.io/s/youthful-feather-ifsvk?file=/src/App.js

### DIFF
--- a/packages/react/presence/src/Presence.tsx
+++ b/packages/react/presence/src/Presence.tsx
@@ -118,7 +118,8 @@ function usePresence(present: boolean) {
         node.removeEventListener('animationend', handleAnimationEnd);
       };
     } else {
-      // Transition to the unmounted state if the node is removed prematurely
+      // Transition to the unmounted state if the node is removed prematurely.
+      // We avoid doing so during cleanup as the node may change but still exist.
       send('ANIMATION_END');
     }
   }, [node, send]);
--- a/packages/react/presence/src/Presence.tsx
+++ b/packages/react/presence/src/Presence.tsx
@@ -117,6 +117,9 @@ function usePresence(present: boolean) {
         node.removeEventListener('animationcancel', handleAnimationEnd);
         node.removeEventListener('animationend', handleAnimationEnd);
       };
+    } else {
+      // Transition to the unmounted state if the node is removed prematurely
+      send('ANIMATION_END');
     }
   }, [node, send]);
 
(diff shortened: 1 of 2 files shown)


packages/react/presence/src/Presence.tsx
```diff
@@ -117,6 +117,9 @@ function usePresence(present: boolean) {
         node.removeEventListener('animationcancel', handleAnimationEnd);
         node.removeEventListener('animationend', handleAnimationEnd);
       };
+    } else {
+      // Transition to the unmounted state if the node is removed prematurely
+      send('ANIMATION_END');
```
jjenzz, line 122: Reading your description of the issue, I was sorta expecting to see:

```jsx
return () => {
  node.removeEventListener('animationstart', handleAnimationStart);
  node.removeEventListener('animationcancel', handleAnimationEnd);
  node.removeEventListener('animationend', handleAnimationEnd);
  // Transition to the unmounted state if the node is removed prematurely
  send('ANIMATION_END');
};
```

Since the cleanup is in a `if (node) {}` check, it will cleanup when `node` becomes `null`. 

Would that work?

jjenzz (commented):
Good catch on this!!

packages/react/presence/src/Presence.tsx
benoitgrelard: It seems like it would be equivalent to me but perhaps I'm missing a case?

benoitgrelard (approved):
Great catch! 🎉

packages/react/presence/src/Presence.tsx
andy-hook: Yeah I started here and could have gone either way, though after working on https://github.com/radix-ui/primitives/pull/1108 I realised they’re not technically equivalent as the node could change or new deps could be introduced in the future.

benoitgrelard: Haha yeah good point, teaches us something 😀

jjenzz: Yeah, definitely a good catch. Worth adding a code comment here too about that perhaps?

jjenzz (approved):

PR #1108, opened 2022-01-21
[useSize] Avoid `undefined` when switching nodes

andy-hook:
Raising for discussion. Setting `undefined` in the cleanup means we can't smoothly transition between values when switching nodes.

Contrived example:
https://codesandbox.io/s/charming-glitter-2fx1q?file=/src/App.js

### DIFF
--- a/.yarn/versions/573f03d4.yml
+++ b/.yarn/versions/573f03d4.yml
@@ -1,0 +1,16 @@
+releases:
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-hover-card": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-popper": patch
+  "@radix-ui/react-radio-group": patch
+  "@radix-ui/react-slider": patch
+  "@radix-ui/react-switch": patch
+  "@radix-ui/react-tooltip": patch
+  "@radix-ui/react-use-size": patch
+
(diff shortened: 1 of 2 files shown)


jjenzz:
wait whut, I thought the old and new here would be equivalent? what's going on here that is making them behave differently? 🤯

benoitgrelard:
> Raising for discussion. Setting undefined in the cleanup means we can't smoothly transition between values when switching nodes.

I'm not surprised by the premise, because I never originally intended for these to be switched at runtime. The cases where we were using it so far were just to measure an element that never changes. But that's a great catch either way!

> wait whut, I thought the old and new here would be equivalent? what's going on here that is making them behave differently? 🤯

Yeah that surprised me too!
It looks like basically the cleanup will run in both cases because `element` is changing but in the updated one, the `else` case (ie. no element) in the body of the hook won't execute because there is an element still.

benoitgrelard (approved):
Mindbending for sure, but looks like a great change to me! 👍

jjenzz:
Ooooh yeah okay, I see that now. Can we add a code comment to explain that please? Something like:

> we only set to undefined when the element becomes `null` and not when it changes to another element.

jjenzz (approved):

andy-hook:
> because I never originally intended for these to be switched at runtime

Totally, It's not very typical tbh. I only stumbled across it while testing some cool code suggestions from @jjenzz in another PR. Had me scratching my head for a bit though.

> It looks like basically the cleanup will run in both cases because element is changing but in the updated one, the else case (ie. no element) in the body of the hook won't execute because there is an element still.

Bang on 👍

PR #1117, opened 2022-01-25
[ToggleGroup] Allow item with empty string value

jjenzz:
Fixes [issue raised in Discord](https://discord.com/channels/752614004387610674/803656530259738674/935449963687706624).

### DIFF
--- a/packages/react/toggle-group/src/ToggleGroup.tsx
+++ b/packages/react/toggle-group/src/ToggleGroup.tsx
@@ -98,9 +98,9 @@ const ToggleGroupImplSingle = React.forwardRef<
   return (
     <ToggleGroupValueProvider
       scope={props.__scopeToggleGroup}
-      value={value ? [value] : []}
+      value={value !== undefined ? [value] : []}
       onItemActivate={setValue}
-      onItemDeactivate={React.useCallback(() => setValue(''), [setValue])}
+      onItemDeactivate={React.useCallback(() => setValue(undefined), [setValue])}
     >
       <ToggleGroupImpl {...toggleGroupSingleProps} ref={forwardedRef} />
     </ToggleGroupValueProvider>
--- a/.yarn/versions/d5024a5f.yml
+++ b/.yarn/versions/d5024a5f.yml
@@ -1,0 +1,6 @@
+releases:
+  "@radix-ui/react-toggle-group": patch
+  "@radix-ui/react-toolbar": patch
+
+declined:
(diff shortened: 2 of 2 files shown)


packages/react/toggle-group/src/ToggleGroup.tsx
```diff
@@ -98,7 +98,7 @@ const ToggleGroupImplSingle = React.forwardRef<
   return (
     <ToggleGroupValueProvider
       scope={props.__scopeToggleGroup}
-      value={value ? [value] : []}
+      value={value !== undefined ? [value] : []}
```
benoitgrelard, line 101: Is this correct?
I feel like we did this on purpose no because we set `''` when deactivating an item.
🤔 

Wouldn't this make it that you can't ever have nothing selected in a single one?

jjenzz: Hmm that was working when i tested it but I think you've uncovered another issue because we do definitely `setValue('')` to deactivate but for some reason that didn't select the empty one so I thought this was all working.

Let me tweak this.

benoitgrelard: Yeah but I'm pretty sure there was a reason we did this. Probably to do with controlled/uncontrolled? Because setting the value to `undefined` changes that no?

jjenzz: Interesting because our stories don't have a default string value so either we missed that there or we should also make multiple return `['', '', '']` .

benoitgrelard: Yeah it's a weird one.

`select` gets away with it for an option with `value === ""` because a value cannot be `undefined` anyway as if it is the `value` will default to the `textContent` instead.

Personally, I feel like what we have is fine and of a consistent type too.

There's nothing blocking for the issue raised, you have full control of everything, you can just put `"empty"` as a value and be done with it. To me, it's still an actual value rather than the lack of a value. It's a toggle group with 3 options to pick from, having "nothing" selected is different from selecting that "empty" item…

jjenzz: Funny you mention the `"empty"` value. I had considered that too so I'm happy to propose that as the solution here.

benoitgrelard: Cool, I've updated them on Discord.

PR #1118, opened 2022-01-25
[ToggleGroup] Use radio role for single toggle group

### LINKED ISSUE #951: [ToggleGroup] Use `role=radio` when `type=single`
Originally spotted here https://github.com/radix-ui/primitives/issues/949

The WAI ARIA toolbar example shows the expectation https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html

jjenzz:
Fixes #951

### DIFF
--- a/packages/react/toggle-group/src/ToggleGroup.tsx
+++ b/packages/react/toggle-group/src/ToggleGroup.tsx
@@ -53,6 +53,7 @@ ToggleGroup.displayName = TOGGLE_GROUP_NAME;
 /* -----------------------------------------------------------------------------------------------*/
 
 type ToggleGroupValueContextValue = {
+  type: 'single' | 'multiple';
   value: string[];
   onItemActivate(value: string): void;
   onItemDeactivate(value: string): void;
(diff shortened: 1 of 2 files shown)


packages/react/toggle-group/src/ToggleGroup.tsx
```diff
@@ -275,8 +278,12 @@ const ToggleGroupItemImpl = React.forwardRef<ToggleGroupItemImplElement, ToggleG
   (props: ScopedProps<ToggleGroupItemImplProps>, forwardedRef) => {
     const { __scopeToggleGroup, value, ...itemProps } = props;
     const valueContext = useToggleGroupValueContext(ITEM_NAME, __scopeToggleGroup);
+    const pressed = valueContext.value.includes(value);
```
andy-hook, line 281: Just curious about declaring `pressed` again vs taking from `props`. Is that to keep the prims props distinct from the underlying `Toggle`?

benoitgrelard: Ah yeah good catch, it seems at this point it will be a prop that will have been passed down already internally right?

jjenzz: good catch indeed 👍

andy-hook (approved):

benoitgrelard (approved):

PR #1119, opened 2022-01-25
[Menu][RadioGroup][ScrollArea][Slider][Tabs][ToggleGroup][Toolbar] Direction provider

### LINKED ISSUE #749: [useDirection] Eager hook is called on every state change.
### Current Behavior

Several components (such as the [`DropdownMenu`](https://github.com/radix-ui/primitives/blob/main/packages/react/dropdown-menu/src/DropdownMenu.tsx) component) use the [`useDirection`](https://github.com/radix-ui/primitives/blob/main/packages/react/use-direction/src/useDirection.tsx#L15) hook. 

If this hook does not receive a `dir` prop, the hook [reads the element's computed styles](https://github.com/radix-ui/primitives/blob/main/packages/react/use-direction/src/useDirection.tsx#L1) by pulling the `direction` property from the element's computed style. 

This can be expensive. As none of the elements provide a default for this prop, if used on an element that might changes very often for whatever reason, this will cause the hook to run on every frame. This was happening to me: see a before and after before I went source-code spelunking.

Before (with `<Dropdown.Root>`): [Video Link](
https://www.dropbox.com/s/dbl3huzumayhifz/Kapture%202021-06-27%20at%2019.18.12.mp4?dl=0)

After (with `<Dropdown.Root dir="ltr">`): [Video Link](
https://www.dropbox.com/s/u51np3gdflkpijo/Kapture%202021-06-27%20at%2019.25.02.mp4?dl=0)

### Expected behavior

If no property is provided, the hook should stick with its first result rather than repeating the expensive job of getting the element's computed styles.

### Reproducible example

Extremely contrived
https://codesandbox.io/s/nervous-jackson-plg87?file=/src/App.tsx

### Suggested solution

At the risk of being a touch imperialist, I'd expect `ltr` to be the default for the `dir` prop. 

If you do want to read the element itself to get the direction, then I'd expect that the hook did not refresh when the element changes. Specifically, I'd suggest that `element` be dropped from [this dependency array](https://github.com/radix-ui/primitives/blob/main/packages/react/use-direction/src/useDirection.tsx#L19).

Otherwise, if you want to foreground the `ltr` or `rtl` choice, make it a required prop.

If none of the above are possible, I'd expect to see this flagged as a potential performance issue when leaving it undefined for components that may change often.

### Your environment

| Radix Package(s) |   @radix-ui/react-dropdown      |   0.0.22      |
| React            | n/a     |   Latest / see sandbox      |
| Browser          |         |      `   |
| Assistive tech   |         |    `     |
| Node             | n/a     |     `    |
| npm/yarn         |         |     `    |
| Operating System |         |    `     |

jjenzz:
Closes #749


packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -87,14 +89,13 @@ const RadioGroup = React.forwardRef<RadioGroupElement, RadioGroupProps>(
...
           orientation={orientation}
-          dir={dir}
+          dir={direction}
           loop={loop}
         >
           <Primitive.div
             role="radiogroup"
             aria-orientation={orientation}
             aria-labelledby={labelledBy}
-            dir={dir}
```
jjenzz, line 97: This isn't needed because `RovingFocusGroup` adds it. We haven't added it in other places for that reason but not sure what is the best approach. I've gone with removing it throughout this PR for now but happy to put back in if we feel that's best.

packages/react/tabs/src/Tabs.tsx
```diff
@@ -227,6 +227,7 @@ const TabsContent = React.forwardRef<TabsContentElement, TabsContentProps>(
         hidden={!isSelected}
         id={contentId}
         tabIndex={0}
+        dir={context.dir}
```
jjenzz, line 230: Right-to-left tab content was broken without this

packages/react/direction/src/Direction.tsx
```diff
@@ -0,0 +1,33 @@
...
+  dir: Direction;
+}
+const DirectionProvider: React.FC<DirectionProviderProps> = (props) => {
+  const { dir, children } = props;
+  return <DirectionContext.Provider value={dir}>{children}</DirectionContext.Provider>;
+};
+
+/* -----------------------------------------------------------------------------------------------*/
+
+function useDirection(directionProp?: Direction) {
```
benoitgrelard, line 20: Some initial feedback:

Not sure if I'm missing the point here, but it seems strange to me to pass in the dir prop inside the hook, just to use that prop if defined instead of the context, rather than do that `||` on the usage side.

What's the advantage?

We're usually checking the `dir` prop in one place at the top anyway and then exposing that through context, so really the fallback check would happen only in one place anyway. Seems to make more sense that way.

I'm probably missing something obvious… 😀

packages/react/tabs/src/Tabs.tsx
```diff
@@ -125,7 +130,6 @@ const TabsList = React.forwardRef<TabsListElement, TabsListProps>(
         <Primitive.div
           role="tablist"
           aria-orientation={context.orientation}
-          dir={context.dir}
```
jjenzz, line 128: Moved this to `Tabs.Root` div instead so that right-to-left content works.

packages/react/menu/src/Menu.tsx
```diff
@@ -453,6 +446,11 @@ interface MenuContentImplProps
    * (default: `true`)
    */
   portalled?: boolean;
+  disableOutsidePointerEvents?: boolean;
+  onEscapeKeyDown?: DismissableLayerProps['onEscapeKeyDown'];
+  onPointerDownOutside?: DismissableLayerProps['onPointerDownOutside'];
+  onFocusOutside?: DismissableLayerProps['onFocusOutside'];
+  onInteractOutside?: DismissableLayerProps['onInteractOutside'];
```
jjenzz, line 453: Put these inline instead because the props aren't spread on the `DismissableLayer` part and meant I didn't need to omit `dir` prop from two places.

packages/react/menu/src/Menu.tsx
```diff
@@ -453,6 +446,11 @@ interface MenuContentImplProps
    * (default: `true`)
    */
   portalled?: boolean;
+  disableOutsidePointerEvents?: DismissableLayerProps['disableOutsidePointerEvents'];
+  onEscapeKeyDown?: DismissableLayerProps['onEscapeKeyDown'];
+  onPointerDownOutside?: DismissableLayerProps['onPointerDownOutside'];
+  onFocusOutside?: DismissableLayerProps['onFocusOutside'];
+  onInteractOutside?: DismissableLayerProps['onInteractOutside'];
```
jjenzz, line 453: Put these inline instead because the props aren't spread on the `DismissableLayer` part and meant I didn't need to omit `dir` prop from two places.

benoitgrelard:
Looking at chromatic, it looks like the scrollarea one is broken in the inherited scenario?

jjenzz:
> Looking at chromatic, it looks like the scrollarea one is broken in the inherited scenario?

I thought that too but it's not, it's because the direction is added to the scrollarea div now from context instead of to a 100% width div that wraps it.

packages/react/menu/src/Menu.tsx
```diff
@@ -91,12 +91,12 @@ interface MenuProps {
 }
 
 const Menu: React.FC<MenuProps> = (props: ScopedProps<MenuProps>) => {
-  const { __scopeMenu, open = false, children, onOpenChange, modal = true } = props;
+  const { __scopeMenu, open = false, children, dir, onOpenChange, modal = true } = props;
```
benoitgrelard, line 94: What do you think about doing `dir = 'ltr'` here rather than having the double `||` later?

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -61,14 +62,15 @@ const RadioGroup = React.forwardRef<RadioGroupElement, RadioGroupProps>(
       value: valueProp,
       required = false,
       orientation,
-      dir = 'ltr',
+      dir,
```
benoitgrelard, line 65: that's different from the other place now, no double `||`, but equally removed the default too, I'm curious why.

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -99,8 +99,8 @@ const RovingFocusGroupImpl = React.forwardRef<
   const {
     __scopeRovingFocusGroup,
     orientation,
-    dir = 'ltr',
     loop = false,
+    dir,
```
benoitgrelard, line 103: that one seems consistent with above one, so there must be a reason 😀

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -80,13 +86,13 @@ const ScrollArea = React.forwardRef<ScrollAreaElement, ScrollAreaProps>(
...
     const composedRefs = useComposedRefs(forwardedRef, (node) => setScrollArea(node));
-    const computedDirection = useDirection(scrollArea, scrollAreaProps.dir);
+    const direction = useDirection() || dir;
 
     return (
       <ScrollAreaProvider
         scope={__scopeScrollArea}
         type={type}
-        dir={computedDirection}
+        dir={direction || 'ltr'}
```
benoitgrelard, line 95: That one's different again with an inline default here

packages/react/slider/src/Slider.tsx
```diff
@@ -259,7 +260,7 @@ const SliderHorizontal = React.forwardRef<SliderHorizontalElement, SliderHorizon
           }}
           onSlideEnd={() => (rectRef.current = undefined)}
           onStepKeyDown={(event) => {
-            const isBackKey = BACK_KEYS[direction].includes(event.key);
+            const isBackKey = BACK_KEYS[direction || 'ltr'].includes(event.key);
```
benoitgrelard, line 263: that one's default is only at a single use case here.

benoitgrelard (commented):
It seems a bit all over in terms of conventions with regards to the `dir` prop's default value.

It looks like you intentionally removed the default value in destructuring in places, but then sometimes a default is added inline after the `useDirection()` hook, sometimes not, and other times a default is added inline for specific use-cases.

There's probably a reason behind all that, but after reading this I'm wondering why we can't always default to `ltr` in destructuring, thus insuring there will always be a value no matter what and the convention is the same everywhere.

As to the versioning I'm still unclear what we should do here, probably ok for now to go with patches…

packages/react/direction/src/Direction.tsx
```diff
@@ -0,0 +1,33 @@
...
+const DirectionProvider: React.FC<DirectionProviderProps> = (props) => {
+  const { dir, children } = props;
+  return <DirectionContext.Provider value={dir}>{children}</DirectionContext.Provider>;
+};
+
+/* -----------------------------------------------------------------------------------------------*/
+
+function useDirection(localDir?: Direction) {
+  const globalDir = React.useContext(DirectionContext);
+  return localDir || globalDir || 'ltr';
```
andy-hook, line 22: I moved the fallback logic into the hook as I found it clearer to see the cascade preference.

benoitgrelard: That's funny cause Jenna had done that originally which I asked about: https://github.com/radix-ui/primitives/pull/1119#discussion_r792541890

> @benoitgrelard re your last round of comments – I couldn't see any reason not to provide the default ltr value so I resolved them and went for this, let me know what you think though.
Edit – I think chromatic has just shown me the reason 😅

Can you provide more detail about this bit, i'm not following.

andy-hook: We checked in offline about this, the primary reason that I went back to passing an arg to the hook is for style. Because we can’t conditionally call hooks it’d be more verbose when done inline:

```tsx
const contextDir = useDirection();
const direction = dirContext || dirProp || 'ltr' ;
```

vs  

```tsx
const direction = useDirection(dirProp)
```

The latter of which we seem to prefer on balance.

> Can you provide more detail about this bit, i'm not following.

Seems that the original changes in this PR removed default values in certain places due to how some primitives get recomposed and the cascading effect of always setting a `dir` attribute from a default value. This is made clear by a preexisting bug in `Toolbar` where `ToolbarToggleGroup` doesn't respect `dir="rtl"` when added to `Root`, I think the changes were to remedy these issues.

We decided it would be clearer overall to provide a default in all cases and ensure we consistency pass `dir` values down the tree where necessary.

packages/react/toolbar/src/Toolbar.tsx
```diff
@@ -177,6 +176,7 @@ const ToolbarToggleGroup = React.forwardRef<
     return (
       <ToggleGroupPrimitive.Root
         data-orientation={context.orientation}
+        dir={context.dir}
```
andy-hook, line 179: We need to pass this down to `ToggleGroupPrimitive` as we now consistent provide a fallback `dir` value. Without doing so `ToolbarToggleGroup` renders `dir="ltr"` even when `<Toolbar.Root dir="rtl"/>`, thus we pass it explicitly via context to prevent the fallback behaviour.

benoitgrelard (approved):

PR #1123, opened 2022-01-29
[Tabs] Add renderAll prop

AhmedAF13:
Renders all tab content for SEO purposes regardless of tab active state


AhmedAF13:
already mentioned here #855

Issue #1074, opened 2022-01-10
[Accordion] Close animation flickering in React 18


tomkuehl:
## Bug report

The Accordion close animation flickers when rendered using the upcoming concurrent mode for React. (ReactDOM.createRoot)

### Current Behavior

https://user-images.githubusercontent.com/14299145/148775476-a662ac9d-ada1-4068-b3a3-fa33d576a677.mp4

### Expected behavior

It shouldn't flicker when a closing animation is set and the accordion finished the closing animation.

### Reproducible example

[CodeSandbox](https://codesandbox.io/s/hungry-chatterjee-k5q7k)

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |    @radix-ui/react-accordion     |     0.1.5 (bug also exists in latest rc)    |
| React            | n/a     |     18.0.0-rc.0    |
| Browser          |   Chrome     |     97.0.4692.71    |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |     macOS (M1)    |    12.1     |

dlehmhus:
I'm encountering the same flickering issue with the [@radix-ui/react-collapsible](https://www.radix-ui.com/docs/primitives/components/collapsible) component.

benoitgrelard:
@dlehmhus, yeah that would make sense as `Accordion` is built on top of `Collapsible`.

devianllert:
@jjenzz 
Encountering the same flickering issue with the [@radix-ui/react-dialog](https://www.radix-ui.com/docs/primitives/components/dialog) component.
It seems like all components that supports animations (Presence) have this issue (only with react@18.0.0, react@17.0.2 works fine).  Adding `animation-fill-mode: forwards` fixes that issue but it feels like a hack.

benoitgrelard:
Hi, at the moment this is the solution, but we do agree that it kinda feels like a patch rather than an actual solution.
We will try and look into it more when we have some time.

rickyzhangca:
still happens today with radix collapsible, once `hidden` is set on the content it repeats the close animation

andypandy-tappit:
Came across this thread while investigating why my accordion element kept flickering closed and open while using drag and drop functionality.
![accordion_animation_flicker](https://github.com/radix-ui/primitives/assets/153526083/b7dedbcd-92f3-46dc-bba9-b0aa97f09ad7)

[...]

PR #1125, opened 2022-01-30
[Collapsible][Accordion] Fix content animation of collapsible

dlehmhus:
Fixes https://github.com/radix-ui/primitives/issues/1074 by allowing to add an animation-fill-mode. 

When using the [Collapsible](https://github.com/radix-ui/primitives/tree/main/packages/react/collapsible) with React 18 Concurrent Mode, the close animation flickers. This is due to the fact, that it takes too long (not quite sure why) for the content to be hidden/unmounted. Since the [animation](https://www.radix-ui.com/docs/primitives/components/collapsible#animating-content-size) doesn't specify any fill-mode, the animation styles will only be applied during the execution of the animation. Therefore the content height is animated to 0 during the animation but jumps back to its original size afterwards, before beeing unmounted.  Adding `animation-fill-mode: forwards` fixes that issue (visually). 

### Notes for docs

- Update [Animation page](https://www.radix-ui.com/docs/primitives/overview/animation) to include `forwards`
- Update all component demos to include `forwards`
- Update all component docs examples to include `forwards`

### DIFF
--- a/.yarn/versions/9b75b880.yml
+++ b/.yarn/versions/9b75b880.yml
@@ -1,0 +1,6 @@
+releases:
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-collapsible": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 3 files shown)


jjenzz:
@dlehmhus would you mind running `yarn bump:check -i` to select the versioning strategy for this PR? you'll have to do it twice annoyingly to reject the main workspace in order for the PR checks to pass.

we need to update our contributing guidelines to explain all this, apologies it isn't clear anywhere atm 🙂

jjenzz (approved):
Thanks for this 🙏

PR #1140, opened 2022-02-07
[DropdownMenu][ContextMenu] Bind keyboard checks only when necessary

### LINKED ISSUE #1139: On mount, Menu adds 3 event listeners which can be expensive
## Question

@benoitgrelard I know we've talked about menus before haha. Anyway, we are doing more stress testing with our application in 6x slowdown and we noticed that there are numerous `pointermove` events firing in the dev tools:

image

I've tracked it down to when the [`Menu` component mounts](https://github.com/radix-ui/primitives/blob/58486382a2afa83e9727d49f2fae857335be6d0f/packages/react/menu/src/Menu.tsx#L106), it adds **3** event listeners to the document _for each instance of a menu_:

I can see that the callbacks for these are just so that you can assign a value to `isUsingKeyboardRef` which is being used in a few places to apply focus if the user is currently using their keyboard.. This seems like a pretty hefty perf trade-off just to handle some focus states for a keyboard user. Any thoughts? In our application, we have lots of interactions where users are dragging items and if there are a lot of these menus on the page, with 6x slowdown, they can really clog up the main thread and result in a janky experience for the user dragging.

Just _one_ of these `pointermove` events on 6x slowdown is almost **10ms**:
image

Have you ever heard of the [idle-until-urgent pattern](https://philipwalton.com/articles/idle-until-urgent/) before? It's a pattern used by Google for handling expensive performance operations and might be something you could utilize in this case?

jjenzz:
Closes #1139

I was about to write up a response to the latest comments on the ticket but had an idea which should solve this without getting into another perf discussion.

### DIFF
--- a/cypress/integration/ContextMenu.spec.ts
+++ b/cypress/integration/ContextMenu.spec.ts
@@ -140,6 +140,13 @@ describe('ContextMenu', () => {
         cy.findByText('New Window').should('be.visible');
       });
 
+      it('should focus first item when pressing right arrow key after opening submenu with mouse', () => {
+        pointerOver('Bookmarks →');
+        cy.findByText('Inbox').should('be.visible');
+        cy.findByText('Bookmarks →').type('{rightarrow}');
+        cy.findByText('Inbox').should('be.focused');
+      });
+
       it('should close all menus when pressing escape, enter or space key on any item', () => {
         // Test close on root menu
         cy.findByText('Bookmarks →').type('{esc}').should('not.exist');
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -99,13 +99,15 @@ const Menu: React.FC<MenuProps> = (props: ScopedProps<MenuProps>) => {
   const computedDirection = useDirection(content, props.dir);
 
   React.useEffect(() => {
-    const handleKeyDown = () => (isUsingKeyboardRef.current = true);
-    const handlePointer = () => (isUsingKeyboardRef.current = false);
     // Capture phase ensures we set the boolean before any side effects execute
     // in response to the key or pointer event as they might depend on this value.
+    const handleKeyDown = () => {
+      isUsingKeyboardRef.current = true;
+      document.addEventListener('pointerdown', handlePointer, { capture: true, once: true });
+      document.addEventListener('pointermove', handlePointer, { capture: true, once: true });
+    };
+    const handlePointer = () => (isUsingKeyboardRef.current = false);
     document.addEventListener('keydown', handleKeyDown, { capture: true });
-    document.addEventListener('pointerdown', handlePointer, { capture: true });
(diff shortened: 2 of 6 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -368,7 +348,10 @@ const MenuSubContent = React.forwardRef<MenuSubContentElement, MenuSubContentPro
         trapFocus={false}
         onOpenAutoFocus={(event) => {
           // when opening a submenu, focus content for keyboard users only
-          if (context.isUsingKeyboardRef.current) ref.current?.focus();
+          const handleKeyDown = () => ref.current?.focus();
+          const handlePointerDown = () => document.removeEventListener('keydown', handleKeyDown);
+          document.addEventListener('keydown', handleKeyDown, { once: true });
+          document.addEventListener('pointerdown', handlePointerDown, { once: true });
```
dwilt, line 354: love it.

dwilt:
@jjenzz any reason why you were able to ditch `pointermove`? That was the big thread hog so I'm happy but just curious how you completely ditched it altogether

packages/react/menu/src/Menu.tsx
```diff
@@ -517,6 +501,13 @@ const MenuContentImpl = React.forwardRef<MenuContentImplElement, MenuContentImpl
       }
     };
 
+    React.useEffect(() => {
+      const handleKeyDown = () => (isKeyboardEntryRef.current = true);
+      const handlePointerDown = () => document.removeEventListener('keydown', handleKeyDown);
+      document.addEventListener('keydown', handleKeyDown, { once: true });
+      document.addEventListener('pointerdown', handlePointerDown, { once: true });
+    }, []);
```
jjenzz, line 509: realised this is wrong because `onEntryFocus` fires when the content receives focus which isn't necessarily only when it first opens. 

will look into this.

cypress/integration/ContextMenu.spec.ts
```diff
@@ -133,21 +133,21 @@ describe('ContextMenu', () => {
...
         // Test close on root menu
-        cy.findByText('Bookmarks →').type('{esc}').should('not.exist');
+        cy.findByText('Bookmarks →').focus().type('{esc}').should('not.exist');
 
         // Reopen menu and test keys from within the submenu
         cy.findByText('Right Click Here').rightclick();
-        cy.findByText('Bookmarks →').type('{enter}');
-        cy.findByText('Inbox').type('{esc}').should('not.exist');
+        cy.findByText('Bookmarks →').focus().type('{enter}');
+        cy.findByText('Inbox').focus().type('{esc}').should('not.exist');
```
jjenzz, line 150: there were a bunch of failing tests on CI because `.type` API expects the element to have `tabindex` https://docs.cypress.io/api/commands/type#Input-Textarea

triggering focus ensures `tabindex="0"`.

andy-hook (approved):

PR #1143, opened 2022-02-09
[DropdownMenu] Focus last Item on ArrowUp on Trigger

luisorbaiceta:
### Description

As described in the WAI-Aria pattern, dropdown should open and focus last item, when `ArrowUp` on `Trigger`

### DIFF
--- a/.yarn/versions/63e43016.yml
+++ b/.yarn/versions/63e43016.yml
@@ -1,6 +1,12 @@
 releases:
+  "@radix-ui/react-context-menu": patch
   "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-radio-group": patch
+  "@radix-ui/react-roving-focus": patch
+  "@radix-ui/react-tabs": patch
+  "@radix-ui/react-toggle-group": patch
+  "@radix-ui/react-toolbar": patch
 
 declined:
   - primitives
-  - "@radix-ui/react-menu"
--- a/.yarn/versions/63e43016.yml
+++ b/.yarn/versions/63e43016.yml
@@ -1,0 +1,6 @@
+releases:
+  "@radix-ui/react-dropdown-menu": patch
+
+declined:
+  - primitives
+  - "@radix-ui/react-menu"
--- a/cypress/integration/DropdownMenu.spec.ts
+++ b/cypress/integration/DropdownMenu.spec.ts
@@ -1,14 +1,26 @@
 describe('DropdownMenu', () => {
+  beforeEach(() => {
+    cy.visitStory('dropdownmenu--submenus');
+  });
+
+  describe('when trigger is focused', () => {
+    it('should open and focus first item when pressing down arrow', () => {
+      cy.findByText('Open').trigger('keydown', { key: 'ArrowDown' });
+      cy.findByText('New Tab').should('be.focused');
+    });
+
+    it('should open and focus last item when pressing up arrow', () => {
+      cy.findByText('Open').trigger('keydown', { key: 'ArrowUp' });
+      cy.findByText('Find…').should('be.focused');
+    });
+  });
+
   describe('given submenu user', () => {
     beforeEach(() => {
-      cy.visitStory('dropdownmenu--submenus');
+      cy.findByText('Open').click();
     });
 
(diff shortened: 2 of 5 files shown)


luisorbaiceta:
Tests were failing for `RTL` mode. I've fixed that and refactored a bit too. Hope this is fine

luisorbaiceta:
I've provided a `focusLast` option to the `RovingFocus` component and used that to implement the desired behaviour in `Menu`. For this to take effect I have finally added the `ArrowUp` to the `DropdownMenu` component `keyDown` event. Hope this makes sense

benoitgrelard (commented):
Hey @luisorbaiceta,

Thanks for looking into this! It's been on our list for a while but we hadn't had a chance to get to it. Great job on the tests too! ♥️ it!

Whilst this works, there are a few things that bother me:
- I don't know if it should be implemented at the `RovingFocusGroup` level, given that it isn't really a requirement for itself. In my mind, it's more a requirement of `DropdownMenu` itself so I feel like it would be better implemented there if we could.
- Doing so right now sort of mutates what `onEntryFocus` does depending on that new `focusLast` prop, so we end up with this comment for example:
```tsx
onEntryFocus={(event) => {
  // only focus first item when using keyboard
```
even though in reality, it might have been a focus last…

What do you think? Did you have any of these nagging feelings in the back of your mind when implementing it?

luisorbaiceta:
> What do you think? Did you have any of these nagging feelings in the back of your mind when implementing it?

Totally. I was hoping for some direction here, but didn't seem polite to ask before trying.

I went for the `RovingFocusGroup` because the focus of the first item in the group was happening there and it seemed correct to not separate them. Maybe focus first item behaviour should be removed too and provide a way to implement that by the consumer?

Do you think this should be implemented in the `DropdownMenu` (right now it only handles the opening part), or should this work be done in the underlying `Menu` component?

Thanks a lot for the feedback @benoitgrelard!

benoitgrelard:
> I went for the RovingFocusGroup because the focus of the first item in the group was happening there and it seemed correct to not separate them.

Yeah that makes sense as a first instinct, my brain went there first too.

> Maybe focus first item behaviour should be removed too

This one is really part of our `RovingFocusGroup` implementation because that's how we manage re-directing focus from the group wrapper to the first item, which is what should happen (ie. the group should not be focused).

> Do you think this should be implemented in the DropdownMenu (right now it only handles the opening part), or should this work be done in the underlying Menu component?

My gut feel is telling me it should belong to `DropdownMenu` because it only is a feature there when considering `DropdownMenu` and `ContextMenu`. That would work if we then build `Menubar` from `DropdownMenu` but it wouldn't work if we build `Menubar` from `Menu` directly. So I guess it depends on the direction we end up taking there.

jjenzz:
@luisorbaiceta Benoit has reminded me separately that the [up arrow functionality is optional](https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-13) in the spec and as you have discovered, it's not a 5 min task to figure out how to tackle this problem which is why we had omitted it initially. 

Perhaps best to leave this out for now and we can tackle it later when we have some more time to think about all of this.

luisorbaiceta:
Great! I'll leave this for now.

Just one last word. I was thinking that maybe, to avoid separating this behaviour, `RovingFocusGroup` could provide the list of focus candidates to the `onEntryFocus` function so people could prevent the original focus first and customize its behaviour

benoitgrelard:
> Just one last word. I was thinking that maybe, to avoid separating this behaviour, `RovingFocusGroup` could provide the list of focus candidates to the `onEntryFocus` function so people could prevent the original focus first and customize its behaviour

That's not a bad shout conceptually. Then potentially we could expose `onEntryFocus` to `Menu.Content` and implement only in `DropdownMenu`.

We'll do a bit more thinking about this, for now yeah we can ignore and keep going with the `Menubar` stuff and layer that one after.

benoitgrelard:
Hi @luisorbaiceta we'll close this PR for now as we're not focusing on this.

jjenzz:
curious why radix team is closing things that aren't being focused on? should consumers avoid raising feature requests?

benoitgrelard:
Re-captured as a ticket here: #1319

Issue #1088, opened 2022-01-14
[DismissableLayer] Layering breaks with different component versions


jjenzz:
Initially raised [on Discord](https://discord.com/channels/752614004387610674/803656530259738674/931602628327145483).

If two components (e.g. `DropdownMenu` and `Dialog`) are using different versions of `DismissableLayer` internally, then the layering breaks because of the different context instances.

We never added scoping to `DismissableLayer` since it should always cross component boundaries, however, I wonder if scoping could help here since it effectively passes context through a private prop that it expects the component to use. It might be worth looking into as a solution.

benoitgrelard:
Hmmm yeah that's where these multi-package distributions become complicated sometimes, these types of issues are really hard to stay on top of.

I'm worried the scoped context would kinda break the intention of `DismissableLayer` though.

jjenzz:
could this be solved by changing the `"@radix-ui/react-dismissable-layer": "workspace:*"` dependency in each of the `package.json`s to `"@radix-ui/react-dismissable-layer": "latest"` ? 🤔

benoitgrelard:
I don't think that would help as ultimately this is already the case, when a new change is made in `@radix-ui/react-dismissable-layer` the other packages would have a new version with the version of `@radix-ui/react-dismissable-layer` being latest. But it doesn't prevent anybody from installing versions that use different version of `@radix-ui/react-dismissable-layer`. Hardcoding to `"latest"` instead of a said version would kinda solve that, but I'm not sure it's viable or even makes sense from a code perspective as potentially changes could be introduced and older versions of popover for example might not be compatible with those changes forced by the latest.

Interestingly though, I had a somewhat similar thought process a few weeks back. I was thinking potentially we could "solve" this issue by making the incompatibility visible at install time by making `@radix-ui/react-dismissable-layer` a peer dependency.

jjenzz:
> Hardcoding to `"latest"` instead of a said version would kinda solve that, but [...] changes could be introduced and older versions of popover for example might not be compatible with those changes forced by the latest.

i had considered this but since `DismissableLayer` is a "global" component I had thought maybe changes to its apis could be handled withy deprecation warnings and things. given that it's a pretty stable comp at this point and not likely to change much, that might not be so bad. it seems better than the current state of things anyway where things just completely break for folk.

making it a peer dependency would surely suffer the same issues except radix would be adding an extra step for consumer by requiring them to manage the dependency.

pedromcunha:
[...]

PR #1147, opened 2022-02-11
[DismissableLayer] Add sibling support

jjenzz:
Our current implementation assumes that layers will always be opened from within another layer so escape keypress would close based on React tree. However, with `Toast`, they're opened as siblings (not from inside each other) and escape should close one at a time. 

These changes update `DismissableLayer` to order layers based on creation order rather than the React tree order.

Here's some manual testing showing existing expectations are still working:

https://user-images.githubusercontent.com/175330/153633217-94e58f05-b4ae-4572-877b-a2b777e42542.mp4

https://user-images.githubusercontent.com/175330/153633388-d5a4b385-d9d8-42f1-9633-f5d3ba344de9.mp4

### DIFF
--- a/packages/react/dismissable-layer/src/DismissableLayer.tsx
+++ b/packages/react/dismissable-layer/src/DismissableLayer.tsx
@@ -19,7 +19,7 @@ const FOCUS_OUTSIDE = 'dismissableLayer.focusOutside';
 
 const DismissableLayerContext = React.createContext({
   layers: new Set<DismissableLayerElement>(),
-  modals: new Set<DismissableLayerElement>(),
+  layersWithOutsidePointerEventsDisabled: new Set<DismissableLayerElement>(),
 });
 
 type DismissableLayerElement = React.ElementRef<typeof Primitive.div>;
@@ -73,13 +73,15 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
     const [node, setNode] = React.useState<DismissableLayerElement | null>(null);
     const [, force] = React.useState({});
     const composedRefs = useComposedRefs(forwardedRef, (node) => setNode(node));
-    const [highestModal] = [...context.modals].slice(-1);
-    const highestModalIndex = [...context.layers].indexOf(highestModal);
-    const layerIndex = node ? [...context.layers].indexOf(node) : -1;
-    const hasPointerEvents = layerIndex >= highestModalIndex;
+    const layers = Array.from(context.layers);
+    const [highestLayerWithOutsidePointerEventsDisabled] = [...context.layersWithOutsidePointerEventsDisabled].slice(-1); // prettier-ignore
+    const highestLayerWithOutsidePointerEventsDisabledIndex = layers.indexOf(highestLayerWithOutsidePointerEventsDisabled); // prettier-ignore
+    const index = node ? layers.indexOf(node) : -1;
+    const isBodyPointerEventsDisabled = context.layersWithOutsidePointerEventsDisabled.size > 0;
+    const isPointerEventsEnabled = index >= highestLayerWithOutsidePointerEventsDisabledIndex;
 
     const pointerDownOutside = usePointerDownOutside((event) => {
-      if (!hasPointerEvents) return;
+      if (!isPointerEventsEnabled) return;
       onPointerDownOutside?.(event);
       onInteractOutside?.(event);
       if (!event.defaultPrevented) onDismiss?.();
@@ -92,7 +94,7 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
     });
 
     useEscapeKeydown((event) => {
-      const isHighestLayer = layerIndex === context.layers.size - 1;
+      const isHighestLayer = index === context.layers.size - 1;
       if (!isHighestLayer) return;
       onEscapeKeyDown?.(event);
       if (!event.defaultPrevented) onDismiss?.();
@@ -102,16 +104,22 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
 
     React.useEffect(() => {
       if (!node) return;
-      if (disableOutsidePointerEvents) context.modals.add(node);
+      if (disableOutsidePointerEvents) context.layersWithOutsidePointerEventsDisabled.add(node);
       context.layers.add(node);
       dispatchUpdate();
     }, [node, disableOutsidePointerEvents, context]);
 
+    /**
+     * We purposefully prevent combining this effect with the `disableOutsidePointerEvents` effect
+     * because a change to `disableOutsidePointerEvents` would remove this layer from the stack
+     * and add it to the end again so the layering order wouldn't be _creation order_.
+     * We only want them to be removed from context stacks when unmounted.
+     */
     React.useEffect(() => {
       return () => {
         if (!node) return;
         context.layers.delete(node);
-        context.modals.delete(node);
+        context.layersWithOutsidePointerEventsDisabled.delete(node);
         dispatchUpdate();
       };
     }, [node, context]);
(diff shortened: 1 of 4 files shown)


packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -1,241 +1,167 @@
...
+    }, [node, disableOutsidePointerEvents, context]);
+
+    React.useEffect(() => {
+      return () => {
+        if (!node) return;
+        context.layers.delete(node);
+        context.modals.delete(node);
+        dispatchUpdate();
+      };
+    }, [node, context]);
```
jjenzz, line 117: There is no `DismissableLayer.Provider` anymore otherwise layers have to be part of the same React tree. Instead I mutate the `defaultContext` in `React.createContext` and dispatch an event on the document so that other dismissable layers know when to trigger a re-render.

This avoids us needing to expose a new `DismissableLayerProvider` part that consumers would need to wrap around their applications and I also think it could potentially lead to a solution to #1088 but will need to explore that separately.

andy-hook: Curious why we need the context at all now. Could we track layer state locally and sync that via the event?

jjenzz: ooooo interesting 👀

jjenzz: looked into this and i'm not sure if it's worth it... i think the way it is currently makes it clearer that there is some state shared across primitives. thoughts?

CleanShot 2022-02-15 at 12 06 17

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -1,241 +1,167 @@
...
 import { Primitive } from '@radix-ui/react-primitive';
-import { createContext } from '@radix-ui/react-context';
+import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { useBodyPointerEvents } from '@radix-ui/react-use-body-pointer-events';
 import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
 import { useEscapeKeydown } from '@radix-ui/react-use-escape-keydown';
 
 import type * as Radix from '@radix-ui/react-primitive';
 
-// We need to compute the total count of layers AND a running count of all layers
```
benoitgrelard, line 11: Happy to see all this go TBH 😀

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -1,241 +1,167 @@
...
+    useBodyPointerEvents({ disabled: disableOutsidePointerEvents });
+
+    React.useEffect(() => {
+      if (!node) return;
+      if (disableOutsidePointerEvents) context.modals.add(node);
+      context.layers.add(node);
+      dispatchUpdate();
+    }, [node, disableOutsidePointerEvents, context]);
+
+    React.useEffect(() => {
```
benoitgrelard, line 110: What is preventing these 2 effects be combined? as the second should just be the cleanup right, ie. when a layer is removed, or `disableOutsidePointerEvents` is changed.

packages/react/dismissable-layer/src/DismissableLayer.tsx
benoitgrelard: Yeah this feels a bit strange because then each instance of `DismissableLayer` is effectively tracking the state locally, duplicating it all over, and syncing it with the event. The context one feels more like a singleton which feels more appropriate here somehow.

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -1,241 +1,167 @@
...
+      ...layerProps
+    } = props;
+    const context = React.useContext(DismissableLayerContext);
+    const [node, setNode] = React.useState<DismissableLayerElement | null>(null);
+    const [, force] = React.useState({});
+    const composedRefs = useComposedRefs(forwardedRef, (node) => setNode(node));
+    const [highestModal] = [...context.modals].slice(-1);
+    const highestModalIndex = [...context.layers].indexOf(highestModal);
+    const layerIndex = node ? [...context.layers].indexOf(node) : -1;
+    const hasPointerEvents = layerIndex >= highestModalIndex;
```
benoitgrelard, line 79: `hasPointerEvents` confuses me and I keep having to double take it, because I don't know if it means `pointerEvents: 'auto'` or `pointerEvents: 'none'`.

Would renaming to `allowsPointerEvents` or `enablePointerEvents` clear that up?

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -1,241 +1,167 @@
...
-    );
-  }
-);
-
-DismissableLayer.displayName = DISMISSABLE_LAYER_NAME;
-
-/* -----------------------------------------------------------------------------------------------*/
+const DismissableLayerContext = React.createContext({
+  layers: new Set<DismissableLayerElement>(),
+  modals: new Set<DismissableLayerElement>(),
```
benoitgrelard, line 22: It really bothers me that we use "modality" throughout to refer to the layers that disable outside pointer events because at this stage it's not yet about modality.

At the same time, I can see why you went for that because it's so much smaller than `layersWithDisabledOutsidePointerEvents` which is more the language I was using previously…

I'm torn!

benoitgrelard: Maybe we could use a different term like `blockingLayers` and refer to "blocking" throughout?

packages/react/dismissable-layer/src/DismissableLayer.tsx
jjenzz: A change to `disableOutsidePointerEvents` would remove from the stack and add it to the end again meaning the layering order wouldn't be "creation order" anymore. I only want them to be removed from layer/modal stacks when unmounted.

benoitgrelard: Ah that's because `Set` guarantees insertion order?

packages/react/dismissable-layer/src/DismissableLayer.tsx
andy-hook: Yeah I see your point/s, losing that understanding for the sake of avoiding `force` probably isn’t worth it.

packages/react/dismissable-layer/src/DismissableLayer.tsx
jjenzz: it's just the same as an array right, but prevents duplicates

benoitgrelard: Yeah I always forget this somehow!

packages/react/dismissable-layer/src/DismissableLayer.tsx
jjenzz: I read it like "this layer _hasPointerEvents_", so, pointer events are enabled. I've changed to `isPointerEventsEnabled` instead, if that helps?

benoitgrelard: Yeah that works better for me, thanks!

packages/react/dismissable-layer/src/DismissableLayer.tsx
jjenzz: i find `blocking` more confusing tbh, blocking what? 😬

the term modality encapsulates that things cannot be clicked outside of them so seems clearer there but i get that it could imply more than that too.

i'll see if i can come up with something else.

benoitgrelard: Yeah it's one of the criteria of modality, but not the entirety of it I guess that's what is bothering me.

benoitgrelard (approved):

PR #1157, opened 2022-02-15
[ContextMenu][DropdownMenu] Prevent submenu click from firing twice

### LINKED ISSUE #1145: TriggerItem.onClick fires twice
### Current Behavior

Check console in sandbox. 

https://codesandbox.io/s/pedantic-burnell-mse7l?file=/App.js

### Expected behavior

onClick only fires once

jjenzz:
Fixes #1145


packages/react/menu/src/Menu.tsx
```diff
@@ -800,7 +801,7 @@ const MenuSubTrigger = React.forwardRef<MenuSubTriggerElement, MenuSubTriggerPro
     }, [pointerGraceTimerRef, onPointerGraceIntentChange]);
 
     return context.isSubmenu ? (
-      <MenuAnchor asChild {...props}>
+      <MenuAnchor asChild {...scope}>
```
jjenzz, line 804: props were being spread here and on `MenuItemImpl` so we were composing the consumer click event twice.

benoitgrelard: Yeah that's surprising 🤔

benoitgrelard (approved):

PR #1161, opened 2022-02-16
[DismissableLayer] Allow external nodes to be considered inside

jjenzz:
Needed for `Toast`. When a `DismissableLayer.Branch` is rendered, it will force any rendered `DismissableLayer.Root`s to consider the branch node inside the `DismissableLayer`.

In other words, pointer outside and focus outside etc. events will not fire when interacting with a branch.

### DIFF
--- a/packages/react/dismissable-layer/src/DismissableLayer.tsx
+++ b/packages/react/dismissable-layer/src/DismissableLayer.tsx
@@ -20,6 +20,7 @@ const FOCUS_OUTSIDE = 'dismissableLayer.focusOutside';
 const DismissableLayerContext = React.createContext({
   layers: new Set<DismissableLayerElement>(),
   layersWithOutsidePointerEventsDisabled: new Set<DismissableLayerElement>(),
+  branches: new Set<DismissableLayerBranchElement>(),
 });
 
 type DismissableLayerElement = React.ElementRef<typeof Primitive.div>;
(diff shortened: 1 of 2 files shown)


packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -155,6 +161,40 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
...
+>((props, forwardedRef) => {
+  const context = React.useContext(DismissableLayerContext);
+  const ref = React.useRef<DismissableLayerBranchElement>(null);
+  const composedRefs = useComposedRefs(forwardedRef, ref);
+
+  React.useEffect(() => {
+    const node = ref.current;
+    if (node) {
+      context.branches.add(node);
+      dispatchUpdate();
```
benoitgrelard, line 185: Why do we need to trigger an update here in this case?
This doesn't affect the layering does it?

benoitgrelard (commented):
Just one small question!

packages/react/dismissable-layer/src/DismissableLayer.tsx
jjenzz: If a `Branch` is rendered after a `DismissableLayer` it should be considered inside all of the dismissable layers currently rendered. I did this to account for generic use-cases, not just `Toast`.

benoitgrelard (approved):

PR #1165, opened 2022-02-18
[Toast] New primitive

jjenzz:


packages/react/toast/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "@babel/runtime": "^7.13.10",
+    "@radix-ui/primitive": "workspace:*",
+    "@radix-ui/react-compose-refs": "workspace:*",
+    "@radix-ui/react-context": "workspace:*",
+    "@radix-ui/react-portal": "workspace:*",
+    "@radix-ui/react-presence": "workspace:*",
+    "@radix-ui/react-primitive": "workspace:*",
+    "@radix-ui/react-use-callback-ref": "workspace:*",
+    "@radix-ui/react-use-controllable-state": "workspace:*",
+    "@radix-ui/react-use-escape-keydown": "workspace:*",
```
andy-hook, line 29: This one looks redundant as we've since moved to `dismissableLayer` ? which looks like is also missing here.

andy-hook: @jjenzz We should add `dismissable-layer` to the list as well

benoitgrelard (approved):

PR #1169, opened 2022-02-22
[Select] New primitive

benoitgrelard:


packages/react/select/src/Select.tsx
```diff
@@ -0,0 +1,1359 @@
...
+                    let candidateNodes = items.map((item) => item.ref.current!);
+
+                    if (['prev', 'last'].includes(focusIntent)) {
+                      candidateNodes = candidateNodes.slice().reverse();
+                    }
+                    if (['prev', 'next'].includes(focusIntent)) {
+                      const currentElement = event.target as SelectItemElement;
+                      const currentIndex = candidateNodes.indexOf(currentElement);
+                      candidateNodes = candidateNodes.slice(currentIndex + 1);
+                    }
```
jjenzz, line 655: i had commented on @andy-hook's navigation stuff about the focus intent abstractions here, i've been finding it difficult to pick apart. 

we changed andy's navigation to check the keys inline, so something along these lines in your case:

```jsx
if (['ArrowUp', 'ArrowDown', 'Home', 'End'].includes(event.key)) {
  const items = getItems().filter((item) => !item.disabled);
  let candidateNodes = items.map((item) => item.ref.current!);
  if (['ArrowUp', 'End'].includes(event.key)) {
    candidateNodes = candidateNodes.slice().reverse();
  }
  if (['ArrowUp', 'ArrowDown'].includes(event.key)) {
    const currentElement = event.target as SelectItemElement;
    const currentIndex = candidateNodes.indexOf(currentElement);
    candidateNodes = candidateNodes.slice(currentIndex + 1);
  }
  event.preventDefault();
}
```

thoughts?

jjenzz (approved):

packages/react/select/src/Select.tsx
benoitgrelard: Ah yeah, the abstraction came from `RovingFocusGroup` I think, but probably isn't needed here as there's no special handling for orientation or RTL. It's just a bit trickier inline when there's these 2 cases to handle.

I'll take a look at what you guys did and will update here.

benoitgrelard: Addressed here: #1179

PR #1170, opened 2022-02-22
[Toast] Close focused toast on escape keypress

jjenzz:
Escape keypress will close the most recent toast unless a toast is focused, in which case, it will close the focused one.

https://user-images.githubusercontent.com/175330/155144594-fa47582a-71b1-4322-836e-fc9a55131691.mp4

### DIFF
--- a/packages/react/toast/src/Toast.stories.tsx
+++ b/packages/react/toast/src/Toast.stories.tsx
@@ -478,7 +478,7 @@ const viewportClass = css({
 });
 
 const rootClass = css({
-  all: 'unset',
+  listStyle: 'none',
   width: 230,
   borderRadius: 4,
   border: '1px solid black',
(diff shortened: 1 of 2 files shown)


packages/react/toast/src/Toast.stories.tsx
```diff
@@ -478,7 +478,7 @@ const viewportClass = css({
 });
 
 const rootClass = css({
-  all: 'unset',
+  listStyle: 'none',
```
jjenzz, line 481: @benoitgrelard this wasn't what was making the styles super bad (a `yarn reset` mostly fixed that) but i realised the `all: unset` was breaking some other styles i missed so just resetting the necessary bits only.

benoitgrelard: Yeah it's an issue with this version of stitches we have in, that's why I tried to upgrade it a while ago to avoid that.

benoitgrelard (approved):

PR #1173, opened 2022-02-22
[NavigationMenu] Avoid focus loss without viewport

andy-hook:
The "basic" usage of the content (without viewport) had an issue where if you animated the content and tabbed quickly enough after dismissing, you'd lose focus. This is because the content is still in tab order during its transitional exit.

https://user-images.githubusercontent.com/11708259/155197388-ac14a1e3-37d4-4dab-a5c4-6be782578486.mp4

To solve this we unify the tab management for viewport and non viewport variants so that they both benefit from tab removal.

### DIFF
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -339,7 +339,8 @@ type NavigationMenuItemContextValue = {
   focusProxyRef: React.RefObject<FocusProxyElement>;
   onEntryKeyDown(): void;
   onFocusProxyEnter(side: 'start' | 'end'): void;
-  onViewportContentFocusOutside(): void;
+  onRootContentClose(): void;
+  onContentFocusOutside(): void;
 };
 
 const [NavigationMenuItemContextProvider, useNavigationMenuItemContext] =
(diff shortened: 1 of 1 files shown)


packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -649,6 +653,9 @@ const NavigationMenuContent = React.forwardRef<
   const commonProps = {
     value: itemContext.value,
     triggerRef: itemContext.triggerRef,
+    focusProxyRef: itemContext.focusProxyRef,
+    onContentFocusOutside: itemContext.onContentFocusOutside,
+    onRootContentClose: itemContext.onRootContentClose,
```
andy-hook, line 658: Since I've moved the tab handling into the common `impl` I decided that passing the methods verbatim as props was clearer than trying to split logic across both the `impl` and the `itemContext` methods.

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -778,12 +765,13 @@ const NavigationMenuContentImpl = React.forwardRef<
     if (context.isRootMenu && content) {
       const handleClose = () => {
         onItemDismiss();
+        onRootContentClose();
```
andy-hook, line 768: We need to remove from tab order on dismiss in addition to focus outside as the latter doesn't fire unless you're already inside the content when it's dismissed.

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -462,8 +464,8 @@ const NavigationMenuTrigger = React.forwardRef<
         </FocusGroupItem>
       </Collection.ItemSlot>
 
-      {/* Proxy tab order between trigger and content when viewport is in use */}
-      {open && context.viewport && (
+      {/* Proxy tab order between trigger and content */}
+      {open && (
```
jjenzz, line 468: can you help me understand why the proxy is available for non-portalled content as well now?

andy-hook: For re-enabling tabbables and entering the content. We could use a different method for non-viewport (which doesn't use the proxy element) but that seemed more complex than simply applying the same solution for both.

jjenzz (approved):

PR #1174, opened 2022-02-22
[Toast] Pointer capture on swipe only

jjenzz:
I noticed that pointer down on `Action` and moving mouse away to release (cancel your action) would still click the action because I had set pointer capture regardless of swipe direction.

This ensures pointer is only captured when swiping in the specified `swipeDirection`.


benoitgrelard:
Needs to be rebased I think, I've approved though cause if I get it right, that's the only change? https://github.com/radix-ui/primitives/pull/1174/commits/1c1438bf2801db49e74b6f644ba21644308b933d

benoitgrelard (approved):

jjenzz:
Love your proactivity there ha, yes, that's the only change :+1:

PR #1176, opened 2022-02-23
[NavigationMenu] Ignore mouse intent on touch

andy-hook:
Previously pointer events intended for mouse hover intent were firing on touch, causing unexpected dismissals. This change ignores those hover intent interactions on touch.

### DIFF
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -434,14 +434,20 @@ const NavigationMenuTrigger = React.forwardRef<
             onPointerEnter={composeEventHandlers(props.onPointerEnter, () => {
               wasClickCloseRef.current = false;
             })}
-            onPointerMove={composeEventHandlers(props.onPointerMove, () => {
-              if (disabled || wasClickCloseRef.current) return;
-              context.onItemOver(itemContext.value);
-            })}
-            onPointerLeave={composeEventHandlers(props.onPointerLeave, () => {
-              if (disabled) return;
-              context.onItemLeave();
-            })}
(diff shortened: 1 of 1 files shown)


packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -838,7 +844,7 @@ const NavigationMenuContentImpl = React.forwardRef<
         })}
         onPointerDownOutside={composeEventHandlers(props.onPointerDownOutside, (event) => {
           const target = event.target as HTMLElement;
-          const isTrigger = triggerRef.current?.contains(target);
+          const isTrigger = getItems().some((item) => item.ref.current?.contains(target));
```
andy-hook, line 847: Prevent `pointerDown` on other triggers from dismissing then reopening, lets us smoothly move between them.

benoitgrelard (approved):
How does it behave on touch devices then?

jjenzz:
I'm jus testing on touch and realise the size calculation for viewport height on initial open doesn't seem quite right (not tall enough):

https://user-images.githubusercontent.com/175330/155336551-e8e73e3c-f957-491f-b944-dd9c752ffe5d.MP4

jjenzz (approved):

andy-hook:
> How does it behave on touch devices then?

(emulated here) but on device it's essentially:

https://user-images.githubusercontent.com/11708259/155338463-509b7387-e201-433f-a1cc-0e6e907af253.mp4


> I'm jus testing on touch and realised the size calculation for viewport height on initial open doesn't seem quite right (not tall enough):

Thanks, I spotted this too. I think it's because the size query is happening during an animation. Pretty sure we've solved for that in `Collapsible` so the same could apply here.

PR #1184, opened 2022-02-23
[Select] SSR

benoitgrelard:


packages/react/select/src/Select.tsx
```diff
@@ -297,21 +297,25 @@ interface SelectContentProps extends SelectContentImplProps {}
...
       <SelectContentImpl {...props} ref={forwardedRef} />
-    ) : (
+    ) : fragment ? (
       ReactDOM.createPortal(
-        <Collection.Slot scope={props.__scopeSelect}>
-          {/* @ts-ignore: This is to avoid the "SelectViewport must be inside "SelectContent" error */}
-          <SelectContentContextProvider scope={props.__scopeSelect}>
+        // @ts-ignore: This is to avoid the "SelectViewport must be inside "SelectContent" error
+        <SelectContentContextProvider scope={props.__scopeSelect}>
+          <Collection.Slot scope={props.__scopeSelect}>
```
jjenzz, line 312: i had to reorder these otherwise the collection slot was trying to slot onto the provider

jjenzz (approved):

packages/react/select/src/Select.tsx
benoitgrelard: Nicely caught!

benoitgrelard (commented):
Thanks for the help @jjenzz, I've just added a comment.

PR #1185, opened 2022-02-24
[NavigationMenu] Fix `Viewport` size calculation in Safari

andy-hook:
Safari was having problems applying the correct `Viewport` dimensions if using `scale/rotate` animations on open. This is because it was falling back to `getBoundingClientRect` inside `useSize` as it [doesn't support `borderBoxSize` yet](https://caniuse.com/mdn-api_resizeobserverentry_borderboxsize). `getBoundingClientRect` returns dimensions **after** transform, hence the issue.

Switching to trusty `offsetWidth/height` works as expected. I considered making this change at the `useSize` level but I think it warrants further testing across other consumers first.

Before:

https://user-images.githubusercontent.com/11708259/155436308-96e656a9-cb57-4877-8579-4f1bacdf307e.mp4

After:

https://user-images.githubusercontent.com/11708259/155436327-efd16efa-5b4b-4093-a015-8547d0696984.mp4

### DIFF
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -925,7 +925,9 @@ const NavigationMenuViewportImpl = React.forwardRef<
 
   /**
    * Update viewport size to match the active content node.
-   * We prefer offset dimensions over `getBoundingClientRect` as the latter respects CSS transform. For example, if content animates in from `scale(0.5)` the dimensions would be anything from `0.5` to `1` of the intended size.
+   * We prefer offset dimensions over `getBoundingClientRect` as the latter respects CSS transform.
+   * For example, if content animates in from `scale(0.5)` the dimensions would be anything
+   * from `0.5` to `1` of the intended size.
    */
   const handleSizeChange = () => {
     if (content) setSize({ width: content.offsetWidth, height: content.offsetHeight });
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -915,7 +915,7 @@ const NavigationMenuViewportImpl = React.forwardRef<
     props.__scopeNavigationMenu
   );
   const [size, setSize] = React.useState<{ width: number; height: number } | null>(null);
-  const [contentNode, setContentNode] = React.useState<NavigationMenuContentElement | null>(null);
+  const [content, setContent] = React.useState<NavigationMenuContentElement | null>(null);
   const viewportWidth = size ? size?.width + 'px' : undefined;
   const viewportHeight = size ? size?.height + 'px' : undefined;
   const open = Boolean(context.value);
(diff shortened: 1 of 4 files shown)


packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -915,17 +914,24 @@ const NavigationMenuViewportImpl = React.forwardRef<
...
     props.__scopeNavigationMenu
   );
-  const [activeContent, setActiveContent] = React.useState<NavigationMenuContentElement | null>(
-    null
-  );
-  const contentSize = useSize(activeContent);
-  const viewportWidth = contentSize ? contentSize?.width + 'px' : undefined;
-  const viewportHeight = contentSize ? contentSize?.height + 'px' : undefined;
+  const [size, setSize] = React.useState<{ width: number; height: number } | null>(null);
+  const [contentNode, setContentNode] = React.useState<NavigationMenuContentElement | null>(null);
```
andy-hook, line 918: Renamed to avoid awkward wrapping, if we prefer the originals then that's fine by me, I just found it irritating.

benoitgrelard: To be honest in most component I think we call these just `[content, setContent]`.

benoitgrelard (approved):

jjenzz:
when reading your PR description i thought "but `getBoundingClientRect` is the border box size no?" and then read this:

> `getBoundingClientRect` returns dimensions **after** transform, hence the issue.

can you help me understand this part? 🙏  what transform? what's the order of things here?

andy-hook:
> can you help me understand this part? 🙏 what transform? what's the order of things here?

A transform applied via css will affect the value returned by `getBoundingClientRect`, so the bug goes:

- Open menu
- `Viewport` animates in from `scale(0.5)` to `scale(1)`
- During this time, `getBoundingClientRect` reports on the size of `Content` and returns its dimensions **including** this transformation (this is anywhere between 0.5 and 1 of the original size).
- The resulting `Viewport` size is a scaled version of the `Content` dimensions

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -915,17 +914,24 @@ const NavigationMenuViewportImpl = React.forwardRef<
...
+  const viewportHeight = size ? size?.height + 'px' : undefined;
   const open = Boolean(context.value);
   // We persist the last active content value as the viewport may be animating out
   // and we want the content to remain mounted for the lifecycle of the viewport.
   const activeContentValue = open ? context.value : context.previousValue;
 
+  /**
+   * Update viewport size to match the active content node.
+   * We prefer offsetWidth and height over `getBoundingClientRect` as the latter returns values after transform.
+   */
```
jjenzz, line 929:
```suggestion
  /**
   * Update viewport size to match the active content node.
   * We prefer offset dimensions over `getBoundingClientRect` as the latter respects CSS transform. For example, if content animates in from `scale(0.8)` the dimensions here would be anything from `0.8` to `1` scale.
   */
```

jjenzz (approved):
thanks for clearing that up

PR #1189, opened 2022-02-24
refactor(checkbox): import primitives from one statement

dheerajsinghnagdali:


benoitgrelard (commented):
These are 2 different packages, so that would fail.

dheerajsinghnagdali:
My bad

PR #1191, opened 2022-02-24
[NavigationMenu] Fix SSR autoValue mismatch

andy-hook:
Switch to `useId` and provide deterministic non empty string ID for initial render to avoid open flash.

### DIFF
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -94,25 +94,23 @@ const NavigationMenu = React.forwardRef<NavigationMenuElement, NavigationMenuPro
     const composedRef = useComposedRefs(forwardedRef, (node) => setNavigationMenu(node));
 
     return (
-      <Collection.Provider scope={__scopeNavigationMenu}>
-        <NavigationMenuProvider
-          scope={__scopeNavigationMenu}
-          isRootMenu={true}
-          value={value}
-          onValueChange={onValueChange}
-          defaultValue={defaultValue}
-          dir={useDirection(navigationMenu, props.dir)}
-          orientation={orientation}
-          rootNavigationMenu={navigationMenu}
-        >
-          <Primitive.nav
-            aria-label="Main"
-            data-orientation={orientation}
-            {...NavigationMenuProps}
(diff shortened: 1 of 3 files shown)


packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -354,8 +328,10 @@ interface NavigationMenuItemProps extends PrimitiveListItemProps {
 const NavigationMenuItem = React.forwardRef<NavigationMenuItemElement, NavigationMenuItemProps>(
   (props: ScopedProps<NavigationMenuItemProps>, forwardedRef) => {
     const { __scopeNavigationMenu, value: valueProp, ...itemProps } = props;
-    const autoValue = React.useMemo(() => String(Math.random()), []);
-    const value = valueProp || autoValue;
+    const autoValue = useId();
+    // We need to provide an initial deterministic value as `useId` will return
+    // empty string on the first render and we don't want to match our internal "closed" value.
+    const value = valueProp || autoValue || 'LEGACY_REACT_AUTO_VALUE';
```
andy-hook, line 334: Couldn't pass to `useId` here as that would always use the ID in place of the generated version (so all items would be the same). Instead we want to use the deterministic fallback for the first render only.

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -285,9 +257,11 @@ const NavigationMenuProvider: React.FC<NavigationMenuProviderProps> = (
         });
       }, [])}
     >
-      <ViewportContentProvider scope={scope} items={viewportContent}>
-        {children}
-      </ViewportContentProvider>
+      <Collection.Provider scope={scope}>
```
andy-hook, line 260: Removed `SubImpl` part as it was only needed to give us access to this collection, since we don't need to access it anymore I moved it back into the common provider.

packages/react/navigation-menu/src/NavigationMenu.tsx
benoitgrelard: Ah yeah good point, otherwise it always uses the one you provide as an argument.

benoitgrelard (approved):

Issue #1214, opened 2022-03-01
[ContextMenu][DropdownMenu] right click item should select


BrianHung:
### Current Behavior

Right clicking on a context menu opens the browser context menu.

### Expected behavior

Right clicking on a context menu should behave as a normal (left) click that acts as a select and does not open the browser context menu.

### Reproducible example

https://www.radix-ui.com/docs/primitives/components/context-menu

### Suggested solution

Modify menu item in https://github.com/radix-ui/primitives/blob/main/packages/react/menu/src/Menu.tsx
such that it behaves like a left click via `handleSelect`.

```js
        onClick={composeEventHandlers(props.onClick, handleSelect)}
        onContextMenu={composeEventHandlers(props.onContextMenu, handleSelect)}
```

If developers want to change this behavior, they could call `event.preventDefault()` because of `composeEventHandlers`.

```js
<ContextMenu.Item
  onContextMenu={e => e.preventDefault()}
/>
```

### Additional context

Right clicking on browser context menu behaves like a normal click. Another example is Figma's context menu which behaves the same.

benoitgrelard:
Hi @BrianHung, thanks for reporting this.
This is correct, and is something we took into account when building the `Select`.
We have a raft of pointer event enhancements we need to bring to the other menu components after learning things on `Select`. If that's ok with you I will close your PR for now and we will address it as part of bigger pointer event improvements.

benoitgrelard:
See #657

benoitgrelard:
Closing as duplicate, will be handled by #657

PR #1215, opened 2022-03-01
Add `onContextMenu` prop to `MenuItem` (solves #1214)

BrianHung:


benoitgrelard:
Closing as we will address as part of larger improvements to menus.
more details here: https://github.com/radix-ui/primitives/issues/1214#issuecomment-1081739517

PR #1217, opened 2022-03-03
remove style outline:none from menu primitive

BrianHung:
### Description

Removes `outline: none` from menu primitive style.

Using this library with tailwindcss, and noticing I can't override the outline style unless I use the important property since menu primitive ships with `outline: none`.

```jsx
    <ContextMenu.Content className="outline outline-1 outline-sky-500">
```

shows up as

```html
<div 
  role="menu"
  aria-orientation="vertical"
  data-state="open" dir="ltr"
  class="outline outline-1 outline-sky-500" 
  tabindex="-1" 
  data-orientation="vertical" 
  style="outline: none;--radix-context-menu-content-transform-origin:var(--radix-popper-transform-origin);pointer-events: auto;" 
  data-side="right" 
  data-align="start"
>
```

with style taking precedence over class.

Tailwind workaround right now is

```jsx
    <ContextMenu.Content className="!outline !outline-1 !outline-sky-500">
```

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -604,7 +604,7 @@ const MenuContentImpl = React.forwardRef<MenuContentImplElement, MenuContentImpl
                     {...popperScope}
                     {...contentProps}
                     ref={composedRefs}
-                    style={{ outline: 'none', ...contentProps.style }}
+                    style={{ ...contentProps.style }}
                     onKeyDown={composeEventHandlers(contentProps.onKeyDown, (event) => {
                       // submenu key events bubble through portals. We only care about keys in this menu.
                       const target = event.target as HTMLElement;


benoitgrelard:
Yeah we've debated many times whether we should do this or not.
I guess this is another reason why we perhaps shouldn't…

packages/react/menu/src/Menu.tsx
```diff
@@ -604,7 +604,7 @@ const MenuContentImpl = React.forwardRef<MenuContentImplElement, MenuContentImpl
                     {...popperScope}
                     {...contentProps}
                     ref={composedRefs}
-                    style={{ outline: 'none', ...contentProps.style }}
+                    style={{ ...contentProps.style }}
```
benoitgrelard, line 607: This can be removed as props are spread anyway right above.

jjenzz: i think benoit is proposing this change:

```diff
-                     style={{ ...contentProps.style }}
```

benoitgrelard:
Hi @BrianHung regarding this, we have given it a bit more thought and we will keep things as is for now.
Here are some of the reasons:

- Due to our focus management implementation for accessibility in these components, the focus may jump from an item to the content and back, causing unnecessary and jarring visual changes for users. This is why we disabled the default browser focus `outline` by default
- 99% of the time this means that it is undesirable to style the `:focus` state of the `Content` as you would generally want it to be consistent regardless where focus is
- In your case, it seems you are not even concerned with focus state at all and are trying to style the `Content`'s border in its default state. In which case there are lots of different ways to go about it: `border` or `box-shadow` for example on top of the workaround you have provided.

Hope this makes sense and sheds a bit of light on our thought process.

✌️

PR #1224, opened 2022-03-04
Fixes #1223

### LINKED ISSUE #1223: [Select] Underlying native `select` has wrong default value
### Current Behavior

When you set defaultValue in Select component, hidden html select has first option as value instead of defaultValue. So, when you submit uncontrolled form with defaultValue form sends first option of the select as value.
This happens too when you select currently selected option, It sets first option as value.

### Expected behavior

Hidden html select should have defaultValue as value if you do not change the selected option, and if you select the currently selected option, it should not change value.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/select-radix-wrong-value-gl0i4w?file=/App.js)

frgarciames:
Closes #1223

### DIFF
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -1219,7 +1219,7 @@ const BubbleSelect = React.forwardRef<HTMLSelectElement, React.ComponentPropsWit
      */
     return (
       <VisuallyHidden asChild>
-        <select {...selectProps} ref={composedRefs} defaultValue={value} />
+        <select {...selectProps} ref={composedRefs} value={value} defaultValue={value} />
       </VisuallyHidden>
     );
   }
--- a/.yarn/versions/e69e53ef.yml
+++ b/.yarn/versions/e69e53ef.yml
@@ -1,0 +1,2 @@
+releases:
+  "@radix-ui/react-select": patch
--- a/.yarn/versions/e69e53ef.yml
+++ b/.yarn/versions/e69e53ef.yml
@@ -1,2 +1,5 @@
 releases:
   "@radix-ui/react-select": patch
+
+declined:
+  - primitives


packages/react/select/src/Select.tsx
```diff
@@ -1219,7 +1219,7 @@ const BubbleSelect = React.forwardRef<HTMLSelectElement, React.ComponentPropsWit
      */
     return (
       <VisuallyHidden asChild>
-        <select {...selectProps} ref={composedRefs} defaultValue={value} />
+        <select {...selectProps} ref={composedRefs} value={value} defaultValue={value} />
```
jjenzz, line 1222: The `defaultValue` should be fine but the `value` will break form `onChange` event. See the comment above:

>      * We purposefully do not add the `value` attribute here to allow the value
>      * to be set programatically and bubble to any parent form `onChange` event.
>      * Adding the `value` will cause React to consider the programatic
>      * dispatch a duplicate and it will get swallowed.

jjenzz (changes requested):

packages/react/select/src/Select.tsx
frgarciames: I'm blind, sorry!
So, what would be the best way to fix this ? 
https://github.com/radix-ui/primitives/issues/1223

jjenzz: I think the `defaultValue` is all that's needed

jjenzz: perhaps something like `<select key={children} ....>` would solve it?

benoitgrelard: That's a cool idea, but it doesn't work because there are no `children` as they are portalled in.
Even if trying with an arbitrary key update at any moment in time, it doesn't seem to work either, guessing because the children are not in there rendered.

It seems to me that the native select will only work with the `defaultValue` if they are rendered at the same time, for example this doesn't work either:

https://codesandbox.io/s/serverless-architecture-e77e62?file=/src/App.js

jjenzz: scrap that, `MutationObserver` could work maybe? observe `childList` and change `key`.

jjenzz: seems the only way would be to render `BubbleSelect` in viewport so children can be passed directly to it and provide some context that `item` can use to determine what to render based on whether it's native select or not 😬

benoitgrelard: I don't think that would work as you still need both the custom and native select. Not sure how you would otherwise redirect the native `option` inside the native `select`, unless I'm not understanding what you're suggesting.

jjenzz: suggesting something like this:

```jsx
const Viewport = () => {
  return (
    <Collection.Slot>
      <Primitive.div>{children}</Primitive.div>

      <IsInsideNativeSelect.Provider value={true}>
        <BubbleInput>{children}</BubbleInput>
      </IsInsideNativeSelect.Provider>
    </Collection.Slot>
  );
};

const Item = () => {
  const isInsideNativeSelect = React.useContext(IsInsideNativeSelect)
  return isInsideNativeSelect ? <option /> : <Item />
};

```

benoitgrelard: Oh I see, duplicating the children basically. hmmm, man that seems rather heavy 😀

The React tree would now be double the size, and also not sure about the `select` being inside the viewport, although we could potentially portal the whole subtree higher up maybe?

jjenzz: how has that doubled in size? it's rendering the same number of `options` as we are currently. it's doing the exact same thing but without portalling.

benoitgrelard: > how has that doubled in size? it's rendering the same number of options as it is currently. it's doing the same thing but without portalling.

In terms of DOM yes, but not in terms of React tree, because you now get double amounts of `SelectItem`s in the tree with half of them rendering to different elements in DOM tree.

> i don't think it really matters where the select renders given that it's hidden? oh but do you mean flex styling n stuff is impacted perhaps? hmm, is it if it's hidden?

I was more referring to the fact that if it was in the viewport, it would be in the "collapsed" part. I know we do render it in a document fragment for typeahead, but I'm not sure if everything would work in a fragment if we need more control, like focus events, etc

Worth a shot anyway.

benoitgrelard:
I'll close the PR for now, but I've linked the issue where we can continue the discussion and we can look back at this linked PR for posterity.

PR #1245, opened 2022-03-15
[Select] Fix SelectContent asChild prop

### LINKED ISSUE #1244: [Select] asChild breaks `SelectContent` positioning
The `SelectContent` has a `SelectContentContextProvider` that wraps its `children` inside the primitive part, therefore, positioning is breaking because `asChild` is replacing the provider.

The children need to be wrapped in `Slottable` or the provider needs to be moved to wrap the primitive.

https://codesandbox.io/s/crimson-meadow-65bzrl?file=/App.js

jjenzz:
Fixes #1244

The `asChild` would strip out the context provider because it was inside the `Primitive.div`. This moves it outside.

Easier to review with hidden whitespace changes.

### DIFF
--- a/.yarn/versions/f27979d2.yml
+++ b/.yarn/versions/f27979d2.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-select": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


JeromeIllgner:
Hey @jjenzz, awesome work on the quick turn around on the patch. Do you have any idea what a temporary fix is in the mean-time?

benoitgrelard (approved):
Curious why you chose to lift it rather than use `Slottable` like you suggested in the issue?

I guess, it won't cause extra re-rendering given it's the provider we're moving and only the components using the context hook would potentially re-render on changes.

jjenzz:
> I guess, it won't cause extra re-rendering given it's the provider we're moving and only the components using the context hook would potentially re-render on changes.

yep, exactly. the other primitives only really use `Slottable` when `children` has a sibling inside the `Primitive.*` part (which was the original reason we created `Slottable`) so it didn't feel required in this case.

PR #1252, opened 2022-03-16
Modify Menu to only register 1 listener instead of N

### LINKED ISSUE #1239: Each Menu adds its own document event listeners for detecting keyboard state
## Bug report

This is basically a reopening of https://github.com/radix-ui/primitives/issues/1139 since we're still seeing considerable slow down even after updating past the fix.

### Current Behavior

Radix Dropdown Menus and Context Menus each add their own document event listeners for keydown and pointerdown which can cause a slowdown in the handling of the corresponding events

### Expected behavior

Handling of keydown and pointerdown don't slowdown as a result of additional menu components on the page.

### Reproducible example

Reproduction is to add n menus to a react app and observe that there are n document event listeners for either keydown or pointerdown

My Profile demonstrating consistent slowness with handling mousedown events:
[Profile-20220312T073820.json.zip](https://github.com/radix-ui/primitives/files/8237628/Profile-20220312T073820.json.zip)

### Suggested solution

Provide a global keyboard state context which registers a single document event listener and all menus can share.

### Your environment

| Radix Package(s) |   @radix-ui/react-context-menu, @radix-ui/react-dropdown-menu      |    0.1.5     |
| React            | n/a     |     16.13    |
| Browser          |    Chrome     |     99    |
| Assistive tech   |    n/a     |   n/a      |
| npm/yarn         |    yarn     |     1.22    |
| Operating System |     MacOS    |    12.2.1     |

hipstersmoothie:
With the current `Menu` component every time one is rendered it will add a `pointerdown`, `pointermove`, and `keydown` listener. This can lead to lots of listeners and considerable input lag when rendering lots of menus on a page. This PR pulls the effect that attaches those listeners into a hook that will only attach them once. Subsequent calls to the hook will will just reuse the data from the first invocation. I could have used `React.Context` to accomplish this but then users would have to use that context at the root of their app (like Tooltip.Provider). With this PR users will not have to modify their code at all but will still receive the perf bump.

## Tests

To test this I verified that in the radix storybook the listeners only attach once and in my app. Perf of keydown in the profiler went from 1s to 9ms 🎉 

Fixes #1239

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -38,6 +38,60 @@ const SUB_CLOSE_KEYS: Record<Direction, string[]> = {
 };
 
 /* -------------------------------------------------------------------------------------------------
+ * useInputMethod
+ * -----------------------------------------------------------------------------------------------*/
+
+const createInputMethodProvider = () => {
+  const listeners: ((newValue: boolean) => void)[] = [];
+  let mounted = false;
+  let isUsingKeyboard = false;
+
+  return function useInputMethod() {
+    const isUsingKeyboardRef = React.useRef<boolean>(false);
+
+    React.useEffect(() => {
+      if (mounted) {
+        return;
+      }
+
+      mounted = true;
+      // Capture phase ensures we set the boolean before any side effects execute
+      // in response to the key or pointer event as they might depend on this value.
+      const handleKeyDown = () => {
+        isUsingKeyboard = true;
+        listeners.map((listener) => listener(isUsingKeyboard));
+        document.addEventListener('pointerdown', handlePointer, { capture: true, once: true });
+        document.addEventListener('pointermove', handlePointer, { capture: true, once: true });
+      };
+
+      const handlePointer = () => {
+        isUsingKeyboard = false;
+        listeners.map((listener) => listener(isUsingKeyboard));
+      };
+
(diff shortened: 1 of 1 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -37,6 +37,68 @@ const SUB_CLOSE_KEYS: Record<Direction, string[]> = {
...
+        }
+      };
+    }, [listenerId]);
+
+    React.useEffect(() => {
+      if (mounted) {
+        return;
+      }
+
+      mounted = true;
```
hipstersmoothie, line 92: The first component to use the hook will mount the listeners

packages/react/menu/src/Menu.tsx
```diff
@@ -37,6 +37,68 @@ const SUB_CLOSE_KEYS: Record<Direction, string[]> = {
...
+    React.useEffect(() => {
+      listeners.set(listenerId, (newValue) => {
+        isUsingKeyboardRef.current = newValue;
+      });
+
+      return () => {
+        listeners.delete(listenerId);
+
+        if (listeners.size === 0) {
+          unmount();
```
hipstersmoothie, line 82: The last component to use the hook will remove the listeners.

benoitgrelard:
Closing this for the reasons explained in #1239

Issue #1266, opened 2022-03-24
Menu: When exit animation is running pointer events still fire 


hipstersmoothie:
### Current Behavior

When using an exit animation with a menu the pointer events to move focus still fire. This can make it hard to focus on something other than the menu/trigger when clicking a menu item

### Expected behavior

Once I click on a menu item the events should no longer fire to manage focus

### Reproducible example

Notice how logs still happen while the menu is closing

https://codesandbox.io/s/cool-feather-p4mqz5?file=/App.js

### Suggested solution

We have a pretty common pattern in our app where we open/focus on something other than the trigger when clicking a menu item. This could be a popover, modal, or really anything else. In our code right now we constantly have to battle the menu's focus management to make it work for these cases. Two things I would want:

1. Events to move focus don't fire as the menu is running an exit animation (I can make a PR for this)
2. A prop on the menu item to _not_ return focus to the trigger. I know I can wire this up through existing props but it's clunky and I have to repeat the same pattern everywhere

benoitgrelard:
This looks like something we may want to handle generally for all components that can "close" via unmount animation.
`Presence` might be the right candidate here.

PR #1267, opened 2022-03-24
don't run pointerMove when menu is closing

hipstersmoothie:
### Description

Don't run pointerMove events while closing. If you have an animation these events will still fire and potentially steal focus away from where you actually want it.

Partially addresses #1266


hipstersmoothie:
On further thought this might not work for triggers for nested menus

benoitgrelard:
Hi @hipstersmoothie,

Thanks for the contribution.
This is probably the wrong level we would want to make a change. As we've done some research in #1266, we should be able to handle that issue more generally.

✌️

PR #1269, opened 2022-03-24
[Select] Fix defaultValue not being recognised

anthify:
### Description
I've been using Select a lot recently in forms and noticed `defaultValue` wasn't being reflected when the element's value was checked; e.g. in a form submission event with an uncontrolled Select. Instead the first option was defaulting as the value. 

Here is an issue covering the same thing https://github.com/radix-ui/primitives/issues/1223

* Removed the `defaultValue` prop from the `<select />` inside `BubbleSelect` and just added `value={value}`
* Ensured this change fixed uncontrolled behaviour with unit and integration tests
* Ensured this change didn't break controlled behaviour with unit and integration tests

### DIFF
--- a/cypress/integration/Select.spec.ts
+++ b/cypress/integration/Select.spec.ts
@@ -1,0 +1,49 @@
+describe('Select', () => {
+  describe('with in a form', () => {
+    beforeEach(() => {
+      cy.visitStory('select--within-form');
+    });
+
+    describe('with defaultValue', () => {
+      it('is present in form response', () => {
+        cy.get('pre').contains('fr').should('not.exist');
+        cy.get('input').type('Testing'); // 👈 Triggers form response to be printed to the DOM
+        cy.get('pre').contains('fr');
+      });
+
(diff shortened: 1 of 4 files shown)


jbw (approved):

jjenzz:
Did you see the comment in the code above the change you made? 

>  
     * We purposefully do not add the `value` attribute here to allow the value
     * to be set programatically and bubble to any parent form `onChange` event.
     * Adding the `value` will cause React to consider the programatic
     * dispatch a duplicate and it will get swallowed.

frgarciames:
@anthify See https://github.com/radix-ui/primitives/pull/1224#discussion_r821566517

benoitgrelard:
Yeah as others have pointed at, that's what we've been discussing over in #1224.
I can't immediately see a way to fix it.
I can think of a way to get around it but it would probably break the initial render without JS (that said, the portalling probably means it's broken already…)

benoitgrelard:
Closing this as duplicate of #1224

PR #1285, opened 2022-03-30
[Toast] Fix MutationObserver on key change

### LINKED ISSUE #1283: [Toast] MutationObserver doesn't reorder DOM if toast `key` changes
## Bug report

When the `key` changes for a toast that is already in the viewport alongside some other toasts, it will get portalled to the bottom of the viewport and stay there.

### Current Behavior

The `MutationObserver` logic is not recognising a `childList` change so it does not get re-ordered to the top.

### Expected behavior

When the `key` changes, the expectation was that it would get portalled back into the viewport at the top as if it were a new toast because `key` causes it to reinstantiate. 

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/laughing-voice-8w10br?file=/src/App.js)

- Open both toasts
- While both open, click to open the top one that is already open
- It will get moved to the bottom of the stack

https://user-images.githubusercontent.com/175330/160805058-c109b9c0-0e4e-4c60-81e9-3d5bbca1b762.mp4

### Suggested solution

not the best code but this forces the toast to "re-open" if it has been re-initialised. maybe this can help spur some ideas for a better solution 😬

image

### Additional context

N/A

### Your environment

jjenzz:
Fixes #1283

### DIFF
--- a/.yarn/versions/efafef3a.yml
+++ b/.yarn/versions/efafef3a.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-toast": patch
+
+declined:
+  - primitives
--- a/packages/react/toast/src/Toast.stories.tsx
+++ b/packages/react/toast/src/Toast.stories.tsx
@@ -128,6 +128,32 @@ export const Promise = () => {
   );
 };
 
+export const KeyChange = () => {
+  const [toastOneCount, setToastOneCount] = React.useState(0);
+  const [toastTwoCount, setToastTwoCount] = React.useState(0);
+
+  return (
+    <Toast.Provider>
+      <button onClick={() => setToastOneCount((count) => count + 1)}>Open toast one</button>
+      <button onClick={() => setToastTwoCount((count) => count + 1)}>Open toast two</button>
+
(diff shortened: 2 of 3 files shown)


benoitgrelard (commented):
The code looks good to me 👍 

However, the change seemed to have had unintended effect in stacking order: https://www.chromatic.com/build?appId=601857eb614bae0021c9b9dd&number=1574

I'm not actually sure which order is correct.

In the story in JSX, we have in that order:
- foreground toast
- background toast

On main, the resulting stack in the viewport is in that order:
- foreground toast
- background toast

After this change, the resulting stack in the viewport is in that order:
- background toast
- foreground toast

Any idea which is the correct order?

jjenzz:
@benoitgrelard the new one is correct because the background toast gets mounted after the foreground toast (so it is the most recent toast). most recent toast should be at top. i hadn't noticed that was broken before 😬  two birds with one stone!

benoitgrelard:
Cool, I've accepted the change in Chromatic as the new baseline 👍

benoitgrelard (approved):

Issue #1287, opened 2022-03-30
[ContextMenu] Native context menu incorrectly appears after second click


davidkpiano:
### Current Behavior

Right-clicking a trigger twice will cause inconsistent behavior:

1. On first right click, the custom context menu appears 👍 
2. On second right click, the native context menu appears instead ❌ 

https://user-images.githubusercontent.com/1093738/160920104-22efa939-7225-4c5f-9f9e-5945554ea377.mp4

### Expected behavior

For the custom context menu to appear just like it does in the Radix documentation on _every_ right click:

Uploading CleanShot 2022-03-30 at 15.58.34.mp4…

### Reproducible example

https://codesandbox.io/s/quizzical-gwen-y1vsly?file=/src/App.tsx

### Suggested solution

I'm not sure; is there a fundamental difference between the CodeSandbox repro (very minimal custom styling) and what Radix has in the example?

### Additional context

We've seen this surface in stately.ai/editor (try right-clicking a state multiple times)

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |    `@radix-ui/react-context-menu`     |     `0.1.6`     |
| React            | n/a     |         |
| Browser          |     Chrome    |    100     |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

benoitgrelard:
Hey @davidkpiano 👋 

Thanks for reporting this, it seems like this is a new bug since React 18.
If I downgrade to React 17 in your sandbox, it all works like in our docs too.

davidkpiano:
Yes, we are on React 18.

benoitgrelard:
Thanks for confirming, I've created a new label for `react-18` related issues.
We'll try and get to it as quickly as we can.

🙏

benoitgrelard:
Now that we've updated the entire project to React 18, our cypress test actually caught it, so for now we've had to skip that specific test, but we should aim to re-enable it once fixed: https://github.com/radix-ui/primitives/pull/1048/files#diff-e2c4e750ca3d92eba090c7fafa585dc6c3d8063c3f5818dc94b99a8b73dffed4

PR #1292, opened 2022-03-31
Wrap custom events dispatching with `flushSync`

Andarist:
fixes https://github.com/radix-ui/primitives/issues/1287

Ok, so this is what I got from a quick debugging session:
1. it's because in React 18 the `contextmenuchange` isn't dispatched at all in this case and thus you can't prevent it and fire a custom logic for it
2. this in turn is caused by the fact that you set `pointer-events: none` on the body, [here](https://github.com/radix-ui/primitives/blob/250ed1cd6a6b62bc353e1a8e448046a2f3c2ba20/packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx#L23)
3. you need to set this to handle some requirements and that is fine, the problem is that in this scenario [`resetPointerEvents`](https://github.com/radix-ui/primitives/blob/250ed1cd6a6b62bc353e1a8e448046a2f3c2ba20/packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx#L53) is not called "soon enough" to unblock the mentioned `contextmenuchange` event
4. In React 17 it actually happens "soon enough" so the whole thing works

I've attached `onOpenChange` to the `ContextMenu.Root` and just logged the received value together with some other stuff.

This is the order of the things happening under the hood:
React 17: `resetPointerEvents()` -> `onOpenChange(false)`
React 18: `onOpenChange(false)` -> `resetPointerEvents()`

As we can see... this somewhat matches the info that I've gathered when debugging. The order is flipped and this leads to a special sort of race condition.

Based on my intuition I've just deduced that this might behave slightly differently in terms of "flushing" the updates as there were some changes to that in React 18. I was looking for a rogue `setState`, to wrap it with `flushSync` but then I've realized that during debugging I've seen some `CustomEvent` stuff and since React is using `window.event` to determine the update priority... it made sense to just wrap **all** of those custom events with `flushSync`. You might want to be more granular, but I think that it's safer to just merge this and then reexamine if some of your custom events should actually not be wrapped with this.

### DIFF
--- a/packages/react/dismissable-layer/src/DismissableLayer.tsx
+++ b/packages/react/dismissable-layer/src/DismissableLayer.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { flushSync } from 'react-dom';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { Primitive } from '@radix-ui/react-primitive';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
@@ -281,7 +282,7 @@ function dispatchCustomEvent<E extends CustomEvent, OriginalEvent extends Event>
   const target = detail.originalEvent.target as HTMLElement;
   const event = new CustomEvent(name, { bubbles: false, cancelable: true, detail });
   if (handler) target.addEventListener(name, handler as EventListener, { once: true });
-  return !target.dispatchEvent(event);
+  return flushSync(() => !target.dispatchEvent(event));
 }
 
 const Root = DismissableLayer;
--- a/packages/react/dismissable-layer/src/DismissableLayer.tsx
+++ b/packages/react/dismissable-layer/src/DismissableLayer.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { flushSync } from 'react-dom';
+import * as ReactDOM from 'react-dom';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { Primitive } from '@radix-ui/react-primitive';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
@@ -282,7 +282,7 @@ function dispatchCustomEvent<E extends CustomEvent, OriginalEvent extends Event>
   const target = detail.originalEvent.target as HTMLElement;
   const event = new CustomEvent(name, { bubbles: false, cancelable: true, detail });
   if (handler) target.addEventListener(name, handler as EventListener, { once: true });
-  return flushSync(() => !target.dispatchEvent(event));
+  return ReactDOM.flushSync(() => !target.dispatchEvent(event));
 }
 
 const Root = DismissableLayer;
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
 import { composeEventHandlers } from '@radix-ui/primitive';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { createContextScope } from '@radix-ui/react-context';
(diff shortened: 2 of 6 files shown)


benoitgrelard:
Hi @Andarist,

Thanks for this. I am not aware of `flushSync` and have been researching it a bit.
How much do you know about it?
Looking at [this here](https://github.com/facebook/react/issues/11527) this statement scares me a bit:

> You also have an option to flush the entire tree if you know what you’re doing. The API is called ReactDOM.flushSync(fn). I don’t think we have documented it yet, but we definitely will do so at some point during the 16.x release cycle. **Note that it actually forces complete re-rendering for updates that happen inside of the call, so you should use it very sparingly**. This way it doesn’t break the guarantee of internal consistency between props, state, and refs.

I will look into the original issue a bit more too.

Andarist:
You can check out the Twitter thread where I've linked to this PR: https://twitter.com/AndaristRake/status/1509537858423566344 This might have some additional interesting info in it. At no point, any React team member has suggested there that this is the wrong solution.

I understand why you might be worried about this here but I would roughly classify this as equivalent to an inferred default priority for all discrete events. It just happens that you are using custom events and thus the builtin detection in React can't classify this automatically as triggered by the user. Note that there might be some subtle differences as this one really flushes the whole update synchronously whereas maybe for discrete events things can be batched in a better way. I think both would flush in the same synchronous frame - but perhaps with `flushSync` you can end up rerendering multiple times during that synchronous frame (one update would come from within `flushSync` and another one would have to come from some other place, so the question is also - can another update be scheduled in the same synchronous frame **outside** of the `flushSync` "scope"?). 

Either way - I think it's worth landing this ASAP and researching the performance implications separately as this should fix bugs for people that are hard to diagnose and I don't think it should come with a severe performance cost, maybe just with a minor one (if at all).

@jjenzz might also have some context for this function as I remember she was asking about it in the React repo quite some time ago

benoitgrelard:
Thanks for the added details and the link to the tweet.
As you said, it seems the React team agrees that's the way to go.

It might make sense to apply the same change to other places we use custom events (a quick search shows 3 files: `DismissableLayer.tsx`, `Toast.tsx` and `Tooltip.tsx`).

One thing I am unclear about, is whether this only applies to custom events created using the `CustomEvent` constructor, or also to custom events created using the `Event` constructor. I'm asking because we use those a fair few also.

I appreciate your help on this 🙏

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { flushSync } from 'react-dom';
```
benoitgrelard, line 2: We usually do this:

```suggestion
import * as ReactDOM from 'react-dom';
```

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -281,7 +282,7 @@ function dispatchCustomEvent<E extends CustomEvent, OriginalEvent extends Event>
   const target = detail.originalEvent.target as HTMLElement;
   const event = new CustomEvent(name, { bubbles: false, cancelable: true, detail });
   if (handler) target.addEventListener(name, handler as EventListener, { once: true });
-  return !target.dispatchEvent(event);
+  return flushSync(() => !target.dispatchEvent(event));
```
benoitgrelard, line 285:
```suggestion
  return ReactDOM.flushSync(() => !target.dispatchEvent(event));
```

benoitgrelard (commented):
We will also need to add `react-dom` as a peer dependency of any package that now uses it and also to any of the packages that depend on one of those.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -157,13 +158,15 @@ const Tooltip: React.FC<TooltipProps> = (props: ScopedProps<TooltipProps>) => {
...
     defaultProp: defaultOpen,
     onChange: (open) => {
-      if (open) {
-        // we dispatch here so `TooltipProvider` isn't required to
-        // ensure other tooltips are aware of this one opening.
-        document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN));
-        onOpen();
-      }
-      onOpenChange?.(open);
+      ReactDOM.flushSync(() => {
```
Andarist, line 161: In here, I've decided to wrap the whole `onChange` handler to "batch" all updates scheduled by within this callback

Andarist:
I guess I forgot to update the lock file - just fixed that

benoitgrelard:
Can you run `yarn version check --interactive` locally and patch them. May need to run a second time to decline the main workspace after too.

🙏

Andarist:
@benoitgrelard I've just done this - please verify if this is how you want to bump packages based on this change

benoitgrelard:
It's all good apart from the main workspace which should be declined instead of patch 👍

benoitgrelard:
> I believe that this would apply to Event constructor as well. It's just that React is internally determining the priority of the update based on window.event.type and thus if you use custom event names/types your updates might not be treated as coming from "discrete" events.

Do you think we should do those too then?

andy-hook:
Closing in favour of https://github.com/radix-ui/primitives/pull/1378, thanks again for your help on this 🙏

PR #1315, opened 2022-04-08
[Tooltip] Better "close on scroll" implementation

### LINKED ISSUE #937: [Tooltip] Scaling the trigger whilst trying to open prevents opening
### Current Behavior

Originally from Discord (https://discord.com/channels/752614004387610674/803656530259738674/902977153715544137)

> hi all! i've got a issue with React.Tooltip when its container is animated (e.g. with framer-motion). repro here: https://codesandbox.io/s/proud-wave-tjovb?file=/src/App.tsx 
> seems like the tooltip flashes and when the container scales it disappears, if i mouse out and mouse back when the button is scaled up, the tooltip will now display properly

### Expected behavior

The tooltip should open

### Reproducible example

https://codesandbox.io/s/proud-wave-tjovb?file=/src/App.tsx

benoitgrelard:
Fixes #937

**Note:** We had discussed in the issue potentially implementing this at the `Popper` layer as it could be useful for other components that open and are attached to a trigger. However, it's not possible there because the notion of "open/closed" doesn't exist at that level but is implemented in each primitive.

For that reason, I implemented it only in `Tooltip` for now so it's easy to review, and we can create a separate issue to maybe abstract that in its own package and implement within others.

### DIFF
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -11,8 +11,6 @@ import { Presence } from '@radix-ui/react-presence';
 import { Primitive } from '@radix-ui/react-primitive';
 import { Slottable } from '@radix-ui/react-slot';
 import { useControllableState } from '@radix-ui/react-use-controllable-state';
-import { usePrevious } from '@radix-ui/react-use-previous';
-import { useRect } from '@radix-ui/react-use-rect';
 import * as VisuallyHiddenPrimitive from '@radix-ui/react-visually-hidden';
 
 import type * as Radix from '@radix-ui/react-primitive';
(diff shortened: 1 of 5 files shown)


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -331,15 +329,27 @@ const TooltipContentImpl = React.forwardRef<TooltipContentImplElement, TooltipCo
...
 
+    // Close this tooltip if another one opens
     React.useEffect(() => {
-      // Close this tooltip if another one opens
       document.addEventListener(TOOLTIP_OPEN, onClose);
       return () => document.removeEventListener(TOOLTIP_OPEN, onClose);
     }, [onClose]);
 
+    // Close the tooltip if the trigger is scrolled
+    const { trigger } = context;
```
andy-hook, line 339: Shall we avoid destructuring here? we don't tend to do so unless we're accessing context methods in our effects.

andy-hook (approved):

PR #1316, opened 2022-04-08
Don't minify build output

hipstersmoothie:
### Description

Radix minifies it's output making applying patches and debugging code very hard. This PR simply stops minify the build output so it's easy for consumers to debug and apply patches

Discussed in https://github.com/radix-ui/primitives/discussions/1092


benoitgrelard:
Hey @hipstersmoothie,

Thanks for making the change.
Whilst I would still like to offer both dev and prop builds, I think this is a good first step either way as it at least means our users can benefit from our dev checks/warnings and it's likely they are minifying their apps anyway.

Would you be able to run `yarn version check --interactive` locally? I believe you'll be presented with only the root package, in which case you can "decline" the change.

benoitgrelard:
Closing this in favour of #1332 as parcel was updated and so were the options.
Thanks for initiating the change 🙏

PR #1332, opened 2022-04-14
Prevent minification

benoitgrelard:
Supersedes #1316 as we have updated parcel since and the option has changed.

### DIFF
--- a/.yarn/versions/c123ddeb.yml
+++ b/.yarn/versions/c123ddeb.yml
@@ -1,0 +1,2 @@
+declined:
+  - primitives
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "// build": "For context on tsconfig replacements in build scripts, see https://github.com/radix-ui/primitives/pull/361#discussion_r555004944",
     "build": "yarn build:config && yarn build:packages && yarn build:cleanup",
     "build:config": "mv tsconfig.json tsconfig.tmp.json && mv tsconfig.production.json tsconfig.json",
-    "build:packages": "parcel build 'packages/*/*/' --no-cache && yarn build:fix-type-defs",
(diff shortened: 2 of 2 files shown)


benoitgrelard:
This doesn't seem to cover everything yet still, because it still strips out our dev checks.
I think this is because parcel sets `NODE_ENV` to production by default, as mentioned here: https://parceljs.org/features/production/#development-branch-removal

I still need to find how we do this.

benoitgrelard:
I'm not even sure that `--no-minimize` option is working at all as I can't see any difference between the dist before and after that change 🤔

jjenzz:
just been fiddling with this on a side-project and i'm going `NODE_ENV=development parcel build 'packages/*/*'` for now (in case it helps)

benoitgrelard:
https://parceljs.org/features/node-emulation/#node_env

>  The NODE_ENV environment variable is automatically set by Parcel depending on the mode. When running parcel build, NODE_ENV is set to production by default, otherwise it is set to development. **This can be overridden by setting NODE_ENV yourself (e.g. in your shell).**

So I can override that in our build command

```diff
- "build:packages": "parcel build 'packages/*/*/' --no-cache --no-optimize && yarn build:fix-type-defs",
+ "build:packages": "NODE_ENV=development parcel build 'packages/*/*/' --no-cache --no-optimize && yarn build:fix-type-defs",
```

and it does changes the output, namely this:

image

isn't compiled away, and we can find the warnings in the output.

However, the check is stripped too so that's still not exactly what we want, otherwise when people do end up building their apps, they have no way to get rid of them and they will always remain…

Still looking for a solution…

benoitgrelard:
I've tried this: https://parceljs.org/features/node-emulation/#disabling-these-features
But that doesn't seem to work either.

benoitgrelard:
Created a reduced test-case and opened an issue with parcel: https://github.com/parcel-bundler/parcel/issues/8025

PR #1344, opened 2022-04-26
[Presence] React 18 – Fix animation frame sync flicker 

andy-hook:
closes https://github.com/radix-ui/primitives/issues/1074
related https://github.com/radix-ui/primitives/pull/1292 https://github.com/radix-ui/primitives/pull/1125

The introduction of batched updates across all handlers in 18 concurrent mode introduces a frame sync issue where the state transition (`unmount`) is communicated a frame after animation has completed.

Currently we are recommending `animation-fill-mode: forwards` to cover this gap. While it does work, it [creates other problems](https://github.com/radix-ui/primitives/discussions/1311).

I'm still getting my head around the internal mechanics of react scheduling so feedback here would be very welcome, but as it stands I feel that this change is a superior option vs the css we are recommending.

⚠️  Note for docs: Remove `forwards` recommendation from demos if accepted?

Before:

https://user-images.githubusercontent.com/11708259/165278293-c86c3658-f133-4b05-a11a-f7f24ff5ca2b.mp4

After:

https://user-images.githubusercontent.com/11708259/165278317-0b396a47-615b-4d65-9110-8f28cddbf70a.mp4

### DIFF
--- a/packages/react/presence/package.json
+++ b/packages/react/presence/package.json
@@ -21,7 +21,8 @@
     "@radix-ui/react-use-layout-effect": "workspace:*"
   },
   "peerDependencies": {
-    "react": "^16.8 || ^17.0 || ^18.0"
+    "react": "^16.8 || ^17.0 || ^18.0",
+    "react-dom": "^16.8 || ^17.0 || ^18.0"
   },
   "homepage": "https://radix-ui.com/primitives",
   "repository": {
--- a/packages/react/presence/src/Presence.tsx
+++ b/packages/react/presence/src/Presence.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { useLayoutEffect } from '@radix-ui/react-use-layout-effect';
 import { useStateMachine } from './useStateMachine';
@@ -100,7 +101,7 @@ function usePresence(present: boolean) {
         const currentAnimationName = getAnimationName(stylesRef.current);
         const isCurrentAnimation = currentAnimationName.includes(event.animationName);
         if (event.target === node && isCurrentAnimation) {
-          send('ANIMATION_END');
+          ReactDOM.flushSync(() => send('ANIMATION_END'));
         }
       };
       const handleAnimationStart = (event: AnimationEvent) => {
--- a/packages/react/presence/src/Presence.tsx
+++ b/packages/react/presence/src/Presence.tsx
@@ -101,6 +101,9 @@ function usePresence(present: boolean) {
         const currentAnimationName = getAnimationName(stylesRef.current);
         const isCurrentAnimation = currentAnimationName.includes(event.animationName);
         if (event.target === node && isCurrentAnimation) {
+          // With React 18 concurrency this update is applied
+          // a frame after the animation ends, creating a flash of visible content.
+          // By manually flushing we ensure they sync within a frame, removing the flash.
           ReactDOM.flushSync(() => send('ANIMATION_END'));
         }
       };
--- a/.yarn/versions/8edec830.yml
+++ b/.yarn/versions/8edec830.yml
@@ -1,0 +1,20 @@
+releases:
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-collapsible": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-hover-card": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-navigation-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-presence": patch
+  "@radix-ui/react-radio-group": patch
+  "@radix-ui/react-scroll-area": patch
+  "@radix-ui/react-toast": patch
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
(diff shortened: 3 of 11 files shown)


packages/react/presence/src/Presence.tsx
```diff
@@ -100,7 +101,10 @@ function usePresence(present: boolean) {
         const currentAnimationName = getAnimationName(stylesRef.current);
         const isCurrentAnimation = currentAnimationName.includes(event.animationName);
         if (event.target === node && isCurrentAnimation) {
-          send('ANIMATION_END');
+          // With React 18 concurrency this update is applied
+          // a frame after the animation ends, creating a flash of visible content.
+          // By manually flushing we ensure they sync within a frame, removing the flash.
+          ReactDOM.flushSync(() => send('ANIMATION_END'));
```
benoitgrelard, line 107: Interesting…
I'm no unclear as to when `flushSync` should be used or not. 

As far as I understand, we use `flushSync` to opt-out of automatic batching.

But what's not clear to me here, is that if it was automatically batching and that was the issue, why would it be delayed by a frame? Unless it is batching 2 state updates "later", and opting out results in applying the first one at the correct time for us?

In which case that would kinda make sense.

packages/react/presence/package.json
```diff
@@ -21,7 +21,8 @@
     "@radix-ui/react-use-layout-effect": "workspace:*"
   },
   "peerDependencies": {
-    "react": "^16.8 || ^17.0 || ^18.0"
+    "react": "^16.8 || ^17.0 || ^18.0",
+    "react-dom": "^16.8 || ^17.0 || ^18.0"
```
benoitgrelard, line 25: Because of this, we will need to track all of the dependents of `Presence` and their dependends subsequently and add that peer-dependency too.

Sometimes I wonder if it would be easier to always add them both as peer-dep anyway haha, would there be any harm considering that they are meant to be used with the DOM anyway?

benoitgrelard (commented):
I do feel like it's a better fix too, I've left a few comments inline.

packages/react/presence/src/Presence.tsx
andy-hook: > But what's not clear to me here, is that if it was automatically batching and that was the issue, why would it be delayed by a frame?

This isn't entirely clear to me either, my current thought is that React is pushing this to the next tick as part of the batching mechanism, I saw this eluded to in the [wg post](https://github.com/reactwg/react-18/discussions/21):

> React doesn’t render the result of the first setState synchronously—the render occurs during the next browser tick. So the render hasn’t happened yet

Which I think is why we're seeing the state change trail the native event, here's a reduced example:

https://codesandbox.io/s/elated-snow-x36ypm?file=/src/App.js

benoitgrelard: This is what I'm confused about though with that reduced example.
Once you isolate just the bad case (the red box), there's only really 1 setState right? So where's the batching? There's no batching in this case, or are we saying that React delays it always regardless of whether there are more than 1 updates?

packages/react/presence/package.json
andy-hook: Good point!

> Sometimes I wonder if it would be easier to always add them both as peer-dep anyway haha, would there be any harm considering that they are meant to be used with the DOM anyway?

Practically speaking I don't think it'd be a problem, technically inaccurate in terms of the dependency contracts but maybe that's ok in this case? at least we'd never forget in the future.

benoitgrelard: Yeah exactly, I'm mostly worried that it's hard to track through multiple levels of dependencies, if it was just flat, it might be ok, but it becomes a bit harder when there's more than 1 level.

packages/react/presence/src/Presence.tsx
benoitgrelard: Man yeah definitely seems to be the case, it's pretty clear here.
It does make me think we probably need to do it in other places too then, as we use things like resize observer too…
The worse part in all that really, the part that bothers me is that it seems like there's no way to tell where it should be applied until you see an issue really…

benoitgrelard (approved):
Approving for the code change, but still needs to deps changes.

packages/react/presence/src/Presence.tsx
andy-hook: I hear you, though I do think we’ll begin to see best practices emerge over time through guidance from the core team and community (as we did with state, FCs, hooks etc). I’d like to better understand the underlying mechanics better though, right now it’s not particularly intuitive when to reach for certain tools.

Something that crossed my mind is where we explicitly opt out of batching using `setTimeout` when programatically focusing, nothing has come up yet though so we're probably ok. We might need to think about how to test for 17 if we decide to make changes in any spicy areas.

jjenzz:
super educational following along here 👀 

i just wondered if this would work in React 16 tho and my quick hacky try yielded a "no":

image
https://codesandbox.io/s/react-16-8-0-forked-1vey3y?file=/src/components/App.js

😬

benoitgrelard:
Yikes!!
How are we supposed to support both??! 😬

andy-hook:
> i just wondered if this would work in React 16.8 tho and my quick hacky try yielded a "no":

Interesting, wasn’t aware of this limitation in 16 but worth keeping in mind! I know that's contrived but `useLayoutEffect` is already sync so not a valid case.

Given that syncing state should be avoided I can't imagine many cases where this would bite us. Providing we call it outside of render we should be fine:

https://codesandbox.io/s/react-16-8-0-forked-t6fhlc?file=/src/components/App.js

Thanks for mentioning, underscores this point:

> We might need to think about how to test in 17 if we decide to make changes in any spicy areas.

benoitgrelard:
oh right, I missed the `useLayoutEffect` I thought simply that calling React.flushSync in 16 was an issue.
So I think we're ok to merge then?

jjenzz:
Ah my bad, i saw that `Presence` calls it in a `useLayoutEffect` so tested the same but missed the fact that it actually happens in an event handler inside the `useLayoutEffect`. that's good to know!

benoitgrelard:
Yeah in your sandbox in 17 it doesn't error but it has a warning instead, seems they changed that in 17.

Issue #860, opened 2021-09-13
[Tabs] Add `Presence` to `TabContent` to support unmount animation


jjenzz:
Looks like we forgot to add `Presence` to this component.

asherccohen:
Hi, I'm not sure what the above means but I would also suggest to not render the content of an inactive tab, or at least make it configurable.

This is a pretty common case for tabs to enhance performance of heavy sub components.

In my code I have to do:

```
        <Tabs
              defaultValue={selectedTab}
            >
              <TabsList>
                <TabsTrigger value="0">Tab 1</TabsTrigger>
                <TabsTrigger value="1">Tab 2</TabsTrigger>
              </TabsList>

              {selectedTab === "0" ? (
                <TabsContent value="0">
                  <MyHeavySubComponent1 />
                </TabsContent>
              ) : null}

              {selectedTab === "1" ? (
                <TabsContent value="1">
                  <MyHeavySubComponent2 />
                </TabsContent>
              ) : null}

            </Tabs>
```

benoitgrelard:
Hi @asherccohen, if I follow you correctly, what you're talking about is the same as described in this issue? #813 
If so, we have already addressed that in #859 and it's released in the latest release candidates. I believe it's available since  [0.1.1-rc.6](https://www.npmjs.com/package/@radix-ui/react-tabs/v/0.1.1-rc.6).

benoitgrelard:
@asherccohen yes that's exactly what we did: https://github.com/radix-ui/primitives/pull/859/files#diff-270239882d345dfbd3b8e546c353876c6d801c253a85b21254b7395792715f26R200

PR #1346, opened 2022-04-27
[Tabs] Add animation support for content

andy-hook:
closes https://github.com/radix-ui/primitives/issues/860

Very similar to our approach for [`Collapsible`](https://github.com/radix-ui/primitives/blob/415a3e1173c81829adcda88c0f02e57df3dedbab/packages/react/collapsible/src/Collapsible.tsx#L182-L185) with a couple of modifications (we don't need to measure in this case). I assume we have historic / a11y reasons for controlling render of `children` rather than on wrapping element, hence my use of the render prop in this instance.

### DIFF
--- a/packages/react/tabs/src/Tabs.stories.tsx
+++ b/packages/react/tabs/src/Tabs.stories.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { DirectionProvider } from '@radix-ui/react-direction';
-import { css } from '../../../../stitches.config';
+import { css, keyframes } from '../../../../stitches.config';
 import * as Tabs from '@radix-ui/react-tabs';
 
 export default { title: 'Components/Tabs' };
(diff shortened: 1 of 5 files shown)


packages/react/tabs/src/Tabs.stories.tsx
```diff
@@ -213,6 +277,38 @@ export const Chromatic = () => (
       </Tabs.Root>
     </DirectionProvider>
 
+    <h1>Animated</h1>
```
andy-hook, line 280: Bit weird to see an animated story in chromatic but the long duration is to catch via diff when there's a simple regression in mount behaviour. Under normal conditions it should obviously not be animating at all.

packages/react/tabs/src/Tabs.stories.tsx
```diff
@@ -308,21 +405,32 @@ const triggerClass = css({
 
 const RECOMMENDED_CSS__TABS__CONTENT = {
   flexGrow: 1,
-};
-
-const contentClass = css({
-  ...RECOMMENDED_CSS__TABS__CONTENT,
-
   padding: '1em',
```
benoitgrelard, line 408: rather than moving all this to the "recommended" css, I think instead what you wanted to do was to extend contentClass in animatedContentClass.

packages/react/tabs/src/Tabs.tsx
```diff
@@ -224,21 +227,52 @@ const TabsContent = React.forwardRef<TabsContentElement, TabsContentProps>(
...
+
+        node.style.animationDuration = animationPrevented ? '0s' : originalDuration;
+        isMountAnimationPreventedRef.current = false;
+      }
+
+      /**
+       * Depends on `isSelected` as we want to restore the original
+       * animation when navigating away from an initially open tab.
+       */
+    }, [isSelected]);
```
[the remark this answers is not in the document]
benoitgrelard: Yeah I see what you mean, the alternative isn't better though…

packages/react/tabs/src/Tabs.tsx
```diff
@@ -224,21 +227,52 @@ const TabsContent = React.forwardRef<TabsContentElement, TabsContentProps>(
...
+        originalStylesRef.current = originalStylesRef.current || {
+          animationDuration: node.style.animationDuration,
+        };
+
+        // Prevent animations from running if open when initially mounted
+        const originalDuration = originalStylesRef.current.animationDuration;
+        const animationPrevented = isMountAnimationPreventedRef.current;
+
+        node.style.animationDuration = animationPrevented ? '0s' : originalDuration;
+        isMountAnimationPreventedRef.current = false;
```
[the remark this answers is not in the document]
benoitgrelard: It seems that came from here: https://github.com/radix-ui/primitives/pull/1030. Do you think we should do it the same way to be consistent / avoid similar regressions?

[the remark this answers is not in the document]
andy-hook: Do you mean align this change with the existing RAF approach?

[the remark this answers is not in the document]
benoitgrelard: Yeah as I'm guessing there was a reason we did it this way in the other one.
Maybe @jjenzz can shed some light on this?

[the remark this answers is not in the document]
andy-hook: It's actually a lot simpler with a RAF, not sure why I didn't go here originally. Unless i'm missing something this seems to work well https://github.com/radix-ui/primitives/pull/1346/commits/e0f1b40ea7b5c0ecc3579dbf898688b5d6c11166

benoitgrelard:
The chromatic diff is weird:
CleanShot 2022-05-20 at 14 34 02@2x

First time I ever see everything moved a tiny bit, there must be a reason 🤔

andy-hook:
> First time I ever see everything moved a tiny bit, there must be a reason

I'll check it out, this branch is quite a bit behind now so needs a refresh anyway

packages/react/tabs/src/Tabs.stories.tsx
benoitgrelard: Does my comment make sense here? I'm not sure I made it very clear 😀  Feel free to ping me if that's the case.

packages/react/tabs/src/Tabs.stories.tsx
```diff
@@ -253,6 +349,7 @@ const RECOMMENDED_CSS__TABS__ROOT = {
 
 const rootClass = css({
   ...RECOMMENDED_CSS__TABS__ROOT,
+  border: '1px solid #eee',
```
andy-hook, line 352: @benoitgrelard I remembered that the diff change is because I moved this border to the outer wrapper so that the animations look better in the story.

packages/react/tabs/src/Tabs.stories.tsx
```diff
@@ -312,17 +409,25 @@ const RECOMMENDED_CSS__TABS__CONTENT = {
 
 const contentClass = css({
   ...RECOMMENDED_CSS__TABS__CONTENT,
-
   padding: '1em',
-  border: '1px solid #eee',
```
benoitgrelard, line 317: oh that would be why it's shifted no? 🙂

andy-hook: https://github.com/radix-ui/primitives/pull/1346#discussion_r878163522 yes, I forgot I did that haha

Issue #1301, opened 2022-04-04
[NavigationMenu] Open state functionality


haniotis:
### Overview

Add `open` state to `NavigationMenu.Root` so we can override opening/closing states for use with external lib (e.g. Framer Motion).

Same functionality as found in `Collapsible`, `DropdownMenu`, `Dialog`, etc.

Edit: There's also no way of closing the menu when navigation between pages in a Next.js app and the menu stays open (located in global layout outside page content)

Thank you!

andy-hook:
Hi @weskor , just to clarify, is your issue that the menu isn't closing when navigating client side to another route in your app? i.e. the `open` state of the menu is persisting on route change?

weskor:
Hi @andy-hook, yes that seems to be the case. I could not find anything in the documentation regarding closing. Not sure if this the intended behavior and that I have to implement something for this.

andy-hook:
> I could not find anything in the documentation regarding closing. Not sure if this the intended behavior and that I have to implement something for this.

@weskor This should be being handled for you out of the box to be honest so seems like an oversight on our part.

We'll raise a fix soon but in the meantime you can control the value yourself and set it to an empty string in order to close the menu manually

https://codesandbox.io/s/distracted-gwen-domugj?file=/pages/_app.js

Hope this helps 🙏

weskor:
Hi @andy-hook, thank you so much 🙏 Really love what Radix is doing and thanks for the quick reaction.

andy-hook:
> Add open state to NavigationMenu.Root so we can override opening/closing states for use with external lib 

@haniotis Just to clarify on this point, the functionality for managing the "open" state does exist but rather than a `boolean` value per `Collapsible`, `DropdownMenu`, `Dialog` this menu uses a `value` prop which controls the currently active `Item`. Out of the box we generate these but if you need to control the currently active item you can assign your own to each item.

https://www.radix-ui.com/docs/primitives/components/navigation-menu#item

Happy to provide an example of animating this with framer if you're still having issues.

haniotis:
@andy-hook Yup! That's what I ended up figuring out therefore closed the issue.

Wouldn't mind an example with framer-motion though - that would be greatly appreciated

samburgers:
thanks for this! i implemented a slight change to @andy-hook example, hooking into route changes rather than click events - meaning it should catch things like browser fwd/back buttons and keyboard shortcuts etc too

```
  const [value, setValue] = React.useState("");
[...]

PR #1347, opened 2022-04-27
[NavigationMenu] Close on `Link` click

andy-hook:
closes https://github.com/radix-ui/primitives/issues/1301

Specifically the uncontrolled confusion:

> I could not find anything in the documentation regarding closing. Not sure if this the intended behaviour and that I have to implement something for this.

### DIFF
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -496,6 +496,8 @@ interface NavigationMenuLinkProps extends PrimitiveLinkProps {
 const NavigationMenuLink = React.forwardRef<NavigationMenuLinkElement, NavigationMenuLinkProps>(
   (props: ScopedProps<NavigationMenuLinkProps>, forwardedRef) => {
     const { __scopeNavigationMenu, active, ...linkProps } = props;
+    const ref = React.useRef<NavigationMenuLinkElement>(null);
+    const composedRefs = useComposedRefs(ref, forwardedRef);
 
     return (
       <FocusGroupItem asChild>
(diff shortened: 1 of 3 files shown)


packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -496,14 +496,23 @@ interface NavigationMenuLinkProps extends PrimitiveLinkProps {
...
           aria-current={active ? 'page' : undefined}
           {...linkProps}
-          ref={forwardedRef}
+          ref={composedRefs}
+          onClick={composeEventHandlers(props.onClick, () => {
+            const rootContentDismissEvent = new Event(ROOT_CONTENT_DISMISS, {
+              bubbles: true,
+              cancelable: true,
+            });
+            ref.current?.dispatchEvent(rootContentDismissEvent);
```
benoitgrelard, line 514: Could we do `event.target.dispatchEvent` to avoid having to setup the extra ref?

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -504,6 +504,13 @@ const NavigationMenuLink = React.forwardRef<NavigationMenuLinkElement, Navigatio
           aria-current={active ? 'page' : undefined}
           {...linkProps}
           ref={forwardedRef}
+          onClick={composeEventHandlers(props.onClick, (event) => {
+            const rootContentDismissEvent = new Event(ROOT_CONTENT_DISMISS, {
```
benoitgrelard, line 508: Given the changes in #1292 I wonder if it'll be worth at some point creating a reusable dispatch event function which uses `flushSync` or if it's better to keep the usage local, and case by case when we see issues. This is my biggest struggle, not knowing if it's needed or not 😀

benoitgrelard (approved):

Issue #1354, opened 2022-04-29
[Tooltip] Regression: sometimes opens on touch devices


benoitgrelard:
## Bug report

See https://github.com/radix-ui/primitives/issues/1351#issuecomment-1113096620

PR #1358, opened 2022-04-29
[Tooltip] Standardise on `pointerEvent`, don't open on touch

andy-hook:
Fixes https://github.com/radix-ui/primitives/issues/1354

Also moved mouse events to pointer to align with recent approaches. This is a breaking change for tests triggering these events or consumers preventing behaviour.

### DIFF
--- a/.yarn/versions/bb0efa48.yml
+++ b/.yarn/versions/bb0efa48.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -48,7 +48,7 @@ const [TooltipProviderContextProvider, useTooltipProviderContext] =
 interface TooltipProviderProps {
   children: React.ReactNode;
   /**
-   * The duration from when the mouse enters the trigger until the tooltip gets opened.
+   * The duration from when the pointer enters the trigger until the tooltip gets opened.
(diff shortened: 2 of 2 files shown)


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -246,14 +246,16 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
           data-state={context.stateAttribute}
           {...triggerProps}
           ref={composedTriggerRef}
-          onMouseEnter={composeEventHandlers(props.onMouseEnter, context.onTriggerEnter)}
-          onMouseLeave={composeEventHandlers(props.onMouseLeave, context.onClose)}
-          onMouseDown={composeEventHandlers(props.onMouseDown, () => {
-            isMouseDownRef.current = true;
-            document.addEventListener('mouseup', handleMouseUp, { once: true });
+          onPointerEnter={composeEventHandlers(props.onPointerEnter, (event) => {
+            if (event.pointerType !== 'touch') context.onTriggerEnter();
```
andy-hook, line 250: Is `pen` relevant here? I assumed not as you can "hover" with these?

.yarn/versions/bb0efa48.yml
```diff
@@ -0,0 +1,5 @@
+releases:
+  "@radix-ui/react-tooltip": patch
```
andy-hook, line 2: Given that we will be adjusting these in the release I just marked as patch, is that correct?

benoitgrelard (approved):

PR #1366, opened 2022-05-03
Fix React resolution issue in SSR test app

andy-hook:
Noticed the upgrade to 18 created a mismatch in our deps for the SSR playground.

As we import code directly via `externalDir`, aligning the versions resolves it.

image

### DIFF
--- a/ssr-testing/package.json
+++ b/ssr-testing/package.json
@@ -8,13 +8,13 @@
     "start": "next start"
   },
   "dependencies": {
-    "next": "12.1.0",
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1"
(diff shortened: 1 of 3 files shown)


ssr-testing/package.json
```diff
@@ -8,13 +8,13 @@
     "start": "next start"
   },
   "dependencies": {
-    "next": "12.1.0",
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1"
+    "next": "^12.1.6",
```
andy-hook, line 11: https://github.com/vercel/next.js/issues/36103

benoitgrelard (approved):

PR #1370, opened 2022-05-04
[Select] Fix alignment issues

### LINKED ISSUE #1357: [Select] Alignment is broken after upgrading to React 18

benoitgrelard:
Closes #1357

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -4,6 +4,10 @@ module.exports = {
   stories: ['../packages/react/**/*.stories.tsx', '../packages/core/**/*.stories.tsx'],
   addons: ['@storybook/addon-storysource'],
 
+  reactOptions: {
+    strictMode: true,
+  },
+
   // we need to add aliases to webpack so it knows how to follow
   // to the source of the packages rather than the built version (dist)
   webpackFinal: async (config) => ({
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -4,8 +4,6 @@ module.exports = {
   stories: ['../packages/react/**/*.stories.tsx', '../packages/core/**/*.stories.tsx'],
   addons: ['@storybook/addon-storysource'],
 
-  reactOptions: { legacyRootApi: true },
-
   // we need to add aliases to webpack so it knows how to follow
   // to the source of the packages rather than the built version (dist)
   webpackFinal: async (config) => ({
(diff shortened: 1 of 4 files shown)


benoitgrelard:
Green!!! ✅

packages/react/select/src/Select.tsx
```diff
@@ -1031,7 +1042,7 @@ const SelectScrollUpButton = React.forwardRef<
   const [canScrollUp, setCanScrollUp] = React.useState(false);
   const composedRefs = useComposedRefs(forwardedRef, contentContext.onScrollButtonChange);
 
-  React.useEffect(() => {
+  useLayoutEffect(() => {
```
benoitgrelard, line 1045: I believe these were the only thing needed to make it go green and the reason I saw the jumps and misalignments in React 18.

packages/react/select/src/Select.tsx
```diff
@@ -475,20 +480,25 @@ const SelectContentImpl = React.forwardRef<SelectContentImplElement, SelectConte
         const willAlignWithoutTopOverflow = contentTopToItemMiddle <= topEdgeToTriggerMiddle;
 
         if (willAlignWithoutTopOverflow) {
+          const isLastItem = selectedItem === items[items.length - 1].ref.current;
           contentWrapper.style.bottom = 0 + 'px';
+          const scrollClearance = isLastItem ? viewportPaddingBottom : 0;
           const viewportOffsetBottom =
-            content.clientHeight - viewport.offsetTop - viewport.offsetHeight;
+            content.clientHeight - viewport.offsetTop - viewport.offsetHeight + scrollClearance;
```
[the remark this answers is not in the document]
benoitgrelard: Also, not super happy with the new variable names (`scrollClearance` and the now overloaded `viewportOffsetBottom`).

Perhaps, I shouldn't bundle them as now it's not truly `viewportOffsetBottom`…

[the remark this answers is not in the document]
benoitgrelard: Unfortunately no, we'd have to do single stories for it to be caught, let's catchup on a call.

.storybook/main.js
```diff
@@ -4,8 +4,6 @@ module.exports = {
   stories: ['../packages/react/**/*.stories.tsx', '../packages/core/**/*.stories.tsx'],
   addons: ['@storybook/addon-storysource'],
 
-  reactOptions: { legacyRootApi: true },
```
andy-hook, line 7: Are we planning to enable `stictMode` separately?

andy-hook (approved):

.storybook/main.js
andy-hook: Spoke too soon 😀

Issue #1364, opened 2022-05-02
[NavigationMenu] Menu should close when navigating to a Link 


fscgustavo:
### Current Behavior

The menu root remains open even if the route changes. 

There was a merged solution that fixes this problem, but for some reason, it still does not work on the next projects. Have I made a mistake?

![navigation_link](https://user-images.githubusercontent.com/48841071/166306208-ba1e3af3-4a77-4e63-9f88-f0ef876e46a2.gif)

### Expected behavior

The menu root should close after a click on the navigation link

### Reproducible example

Visit the `day` route to see the menu

/day/any_term

[Stackblitz Template](https://stackblitz.com/edit/nextjs-lt13ab?file=pages%2Fday%2F[slug].jsx)

benoitgrelard:
Hi @fscgustavo, this is a duplicate of #1301.
It was already addressed in #1347.
If you don't want to wait for the stable release, you can update to the latest rc to get the new behaviour.

fscgustavo:
@benoitgrelard The `0.1.3-rc.10` version is being used on the template

https://stackblitz.com/edit/nextjs-lt13ab?file=package-lock.json

PR #1372, opened 2022-05-05
[NavigationMenu] Improve router link event composition

andy-hook:
See https://github.com/radix-ui/primitives/issues/1364#issuecomment-1116216373

- Adds an `onSelect` custom event so that we can maintain consistent "out of the box" close behaviour when composing with `next/link` (and others)
- Provides a mechanism to opt out of close on select without affecting default navigation behaviour (as they would if they simply `onClick={e => e.preventDefault()}`)

### DIFF
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -506,22 +506,24 @@ const NavigationMenuLink = React.forwardRef<NavigationMenuLinkElement, Navigatio
           aria-current={active ? 'page' : undefined}
           {...linkProps}
           ref={forwardedRef}
-          onClick={(event) => {
-            props.onClick?.(event);
+          onClick={composeEventHandlers(
+            props.onClick,
+            (event) => {
+              const target = event.target;
+              const linkSelectEvent = new Event(LINK_SELECT, { bubbles: true, cancelable: true });
+              target.addEventListener(LINK_SELECT, (event) => onSelect?.(event), { once: true });
+              target.dispatchEvent(linkSelectEvent);
 
-            const target = event.target;
-            const linkSelectEvent = new Event(LINK_SELECT, { bubbles: true, cancelable: true });
-            target.addEventListener(LINK_SELECT, (event) => onSelect?.(event), { once: true });
-            target.dispatchEvent(linkSelectEvent);
-
-            if (!linkSelectEvent.defaultPrevented) {
-              const rootContentDismissEvent = new Event(ROOT_CONTENT_DISMISS, {
-                bubbles: true,
-                cancelable: true,
-              });
-              target.dispatchEvent(rootContentDismissEvent);
-            }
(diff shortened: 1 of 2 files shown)


packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -504,13 +506,22 @@ const NavigationMenuLink = React.forwardRef<NavigationMenuLinkElement, Navigatio
...
-            });
-            event.target.dispatchEvent(rootContentDismissEvent);
-          })}
+          onClick={(event) => {
+            props.onClick?.(event);
+
+            const target = event.target;
+            const linkSelectEvent = new Event(LINK_SELECT, { bubbles: true, cancelable: true });
+            target.addEventListener(LINK_SELECT, (event) => onSelect?.(event), { once: true });
+            target.dispatchEvent(linkSelectEvent);
```
andy-hook, line 515: Felt clearer to use an additional event vs piggybacking `rootContentDismissEvent`

benoitgrelard: It makes me wonder if it should just be an `onDismiss` custom event then given that the linking behaviour wouldn't be attached to `onSelect` like it is on menus?

On menus, `onSelect` makes sense, because people attach strictly onto this for behaviour.

Although, I like the consistency, not sure which way around it should be…

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -504,13 +506,22 @@ const NavigationMenuLink = React.forwardRef<NavigationMenuLinkElement, Navigatio
...
           {...linkProps}
           ref={forwardedRef}
-          onClick={composeEventHandlers(props.onClick, (event) => {
-            const rootContentDismissEvent = new Event(ROOT_CONTENT_DISMISS, {
-              bubbles: true,
-              cancelable: true,
-            });
-            event.target.dispatchEvent(rootContentDismissEvent);
-          })}
+          onClick={(event) => {
```
benoitgrelard, line 509: Why did we move away from `composeEventHandlers` here?

benoitgrelard (approved):

packages/react/navigation-menu/src/NavigationMenu.tsx
andy-hook: Because I forgot we had the `checkForDefaultPrevented` flag 😄

benoitgrelard: We really need to refactor to these 2 methods at some point 😀

packages/react/navigation-menu/src/NavigationMenu.tsx
andy-hook: Caught up offline, decided that although it doesn’t feel great to have to provide this in the context of a `Link`, `onSelect` does map neatly with existing DX and the user interaction itself.

PR #1373, opened 2022-05-05
[Select] Add more chromatic coverage

benoitgrelard:


benoitgrelard:
Looks like I need to tweak the last few, one sec.

benoitgrelard:
This is now ready @andy-hook

packages/react/select/src/Select.stories.tsx
```diff
@@ -430,7 +430,104 @@ export const WithinDialog = () => (
...
+export const ChromaticBottomLastPaddedViewport = () => (
+  <ChromaticStoryBottomLast padding="viewport" />
+);
+ChromaticBottomLastPaddedViewport.parameters = { chromatic: { disable: false } };
+
+type PaddedElement = 'content' | 'viewport';
+
+interface ChromaticSelectProps extends React.ComponentProps<typeof Select.Trigger> {
+  count?: number;
+  padding?: PaddedElement;
```
andy-hook, line 473: Would it be clearer to name the prop `paddedElement` as well? with `padding` as a prop name I instinctively expect it to directly accept a style value for some reason.

andy-hook (approved):

PR #1376, opened 2022-05-06
[Slot] Fix children expectation when using `Slottable`

### LINKED ISSUE #1261: [Slot] `Slottable` not working as expected
See https://github.com/radix-ui/primitives/discussions/1255

benoitgrelard:
Fixes #1261

### DIFF
--- a/packages/react/slot/src/Slot.stories.tsx
+++ b/packages/react/slot/src/Slot.stories.tsx
@@ -57,22 +57,22 @@ export const ButtonAsLink = () => (
   <>
     <h1>Button with left/right icons</h1>
     <Button
-      iconLeft={<FakeIcon color="tomato" />}
-      iconRight={<FakeIcon color="royalblue" />}
+      iconLeft={<MockIcon color="tomato" />}
+      iconRight={<MockIcon color="royalblue" />}
       ref={console.log}
     >
-      Click <em>me</em>!
+      Button <em>text</em>
     </Button>
 
     <h1>Button with left/right icons as link (asChild)</h1>
     <Button
       asChild
-      iconLeft={<FakeIcon color="tomato" />}
-      iconRight={<FakeIcon color="royalblue" />}
+      iconLeft={<MockIcon color="tomato" />}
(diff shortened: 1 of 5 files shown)


benoitgrelard:
Looks like that broke a few chromatic tests, looking into it.

packages/react/slot/src/Slot.tsx
```diff
@@ -11,20 +11,29 @@ interface SlotProps extends React.HTMLAttributes<HTMLElement> {
...
+
+  if (slottable) {
+    const newElement = slottable.props.children;
+    const newChildren = childrenArray.map((child) => {
+      if (child === slottable) {
+        const slottableChildren = slottable.props.children as React.ReactNode;
+        if (React.Children.count(slottableChildren) > 1) return React.Children.only(null);
+        return React.isValidElement(slottableChildren)
+          ? (slottableChildren.props.children as React.ReactNode)
+          : null;
```
andy-hook, line 25: Is it worth us providing some comments to explain the flow here? the children of children is a bit of a mind bender without context. Hard to know if I'm finding it easier to follow simply because I already have that context.

benoitgrelard: I've added comment but also re-wrote it a bit which I think makes it easier to read.
Let me know your thoughts: https://github.com/radix-ui/primitives/pull/1376/commits/cb531c0c3faf85ae0e6efa4d872fcc4d9f88475d

packages/react/slot/src/Slot.stories.tsx
```diff
@@ -53,6 +53,31 @@ export const WithComposedEvents = () => (
   </>
 );
 
+export const ButtonAsLink = () => (
+  <>
+    <h1>Button with left/right icons</h1>
+    <Button
+      iconLeft={<FakeIcon color="tomato" />}
+      iconRight={<FakeIcon color="royalblue" />}
+      ref={console.log}
```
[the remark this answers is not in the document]
andy-hook: Ah, makes sense 👍

andy-hook (approved):

PR #1378, opened 2022-05-09
[core/primitive] React 18 – Flush discrete `CustomEvent` dispatch calls

andy-hook:
Fixes https://github.com/radix-ui/primitives/issues/1287
Supersedes https://github.com/radix-ui/primitives/pull/1292

Big thanks to @Andarist for providing great insight into a resolution. This change is mostly equivalent though with the additional goal of providing a clear heuristic for when this is necessary and applying the change to all applicable code.

### Context
One of the core changes in React 18 is the introduction of consistent batching across all event handlers, this differs from 17 where only certain events would be subject to batching. As a result, certain state updates are now batched where before they were not, creating issues like those seen in https://github.com/radix-ui/primitives/issues/1287

### Event priorities
React [assigns priorities](https://github.com/facebook/react/blob/a8a4742f1c54493df00da648a3f9d26e3db9c8b5/packages/react-dom/src/events/ReactDOMEventListener.js#L294-L350) to updates based on the type of user interaction (`discrete`, `continuous`, `default`), `discrete` is an important distinction as it allows updates from those handlers to be made immediately. As stated in https://github.com/radix-ui/primitives/pull/1292#issuecomment-1091983795 the way that React determines this priority is by usage of `window.event`.

Because custom event types are unable to be assessed in this way, they fall through to `default`. In most cases this wouldn’t be a problem, but in the case of `ContextMenu` our code structure is such that there is an implicit expectation for one update to commit before another. 

### Why this can be problematic
In the case of `ContextMenu`, the incorrect behaviour is caused by the unexpected batching of updates in `pointerdown` and `contextmenu`, while the execution order remains correct (`pointerdown` -> `contextmenu`), the resulting state change is batched in 18, resulting in no update  to `open`. As the [effect cleanup](https://github.com/radix-ui/primitives/blob/f2d50b0d310fbe2168333ceff95c7f43133fe7f0/packages/react/use-body-pointer-events/src/useBodyPointerEvents.tsx#L32-L57) is no longer cycled by moving through `false` -> `true` `pointer-events: none;` does not get removed and our handler in  `contextmenu` is blocked.

You can see and compare the behaviour in [this example](https://codesandbox.io/s/batching-simple-euprqh?file=/src/App.js).  

### A heuristic for consistency
Given our understanding of how React splits priorities for updates, and that the intention for immediate user interaction i.e. `discrete` events is to not defer, `dispatchDiscreteCustomEvent` should be used whenever a custom event type is dispatched from a `discrete` handler. This is the [recommendation](https://twitter.com/dan_abramov/status/1509543312427991046?s=20&t=Uzu2oiNruibt3QnSvK_Xvw) from the core team.

Internal p riorities for these types [can be found here](https://github.com/facebook/react/blob/a8a4742f1c54493df00da648a3f9d26e3db9c8b5/packages/react-dom/src/events/ReactDOMEventListener.js#L294-L350)

### Use of `Event` vs `CustomEvent`

I’ve changed every usage of the the `Event` constructor to be `CustomEvent` when a custom type is provided, this is a style change as either constructor is susceptible (it’s important to remember priority is inferred by the type passed, not which constructor used) the hope being that staying consistent will make deciding when to use `dispatchDiscreteCustomEvent` easier.

  e.g. 
```jsx
//   dispatching a click
  target.dispatchEvent(new Event(‘click’))
  
  //     dispatching a custom type
  target.dispatchEvent(new CustomEvent(‘customType’))
  
  //       dispatching a custom type within a discrete handler
  onPointerDown={(e) => dispatchDiscreteCustomEvent(e.target, new CustomEvent(‘customType’))}
```

### DIFF
--- a/.yarn/versions/ce619930.yml
+++ b/.yarn/versions/ce619930.yml
@@ -1,24 +1,33 @@
 releases:
-  "@radix-ui/primitive": patch
+  "@radix-ui/react-accessible-icon": patch
   "@radix-ui/react-accordion": patch
   "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-announce": patch
+  "@radix-ui/react-arrow": patch
+  "@radix-ui/react-aspect-ratio": patch
   "@radix-ui/react-avatar": patch
   "@radix-ui/react-checkbox": patch
   "@radix-ui/react-collapsible": patch
+  "@radix-ui/react-collection": patch
   "@radix-ui/react-context-menu": patch
   "@radix-ui/react-dialog": patch
   "@radix-ui/react-dismissable-layer": patch
   "@radix-ui/react-dropdown-menu": patch
   "@radix-ui/react-focus-scope": patch
   "@radix-ui/react-hover-card": patch
+  "@radix-ui/react-label": patch
   "@radix-ui/react-menu": patch
   "@radix-ui/react-navigation-menu": patch
   "@radix-ui/react-popover": patch
   "@radix-ui/react-popper": patch
+  "@radix-ui/react-portal": patch
+  "@radix-ui/react-primitive": patch
+  "@radix-ui/react-progress": patch
   "@radix-ui/react-radio-group": patch
   "@radix-ui/react-roving-focus": patch
   "@radix-ui/react-scroll-area": patch
   "@radix-ui/react-select": patch
+  "@radix-ui/react-separator": patch
   "@radix-ui/react-slider": patch
   "@radix-ui/react-switch": patch
   "@radix-ui/react-tabs": patch
@@ -27,6 +36,8 @@ releases:
   "@radix-ui/react-toggle-group": patch
   "@radix-ui/react-toolbar": patch
   "@radix-ui/react-tooltip": patch
+  "@radix-ui/react-visually-hidden": patch
 
 declined:
   - primitives
+  - "@radix-ui/primitive"
--- a/.yarn/versions/ce619930.yml
+++ b/.yarn/versions/ce619930.yml
@@ -1,0 +1,32 @@
+releases:
+  "@radix-ui/primitive": patch
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-avatar": patch
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-collapsible": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
+  "@radix-ui/react-dismissable-layer": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-focus-scope": patch
+  "@radix-ui/react-hover-card": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-navigation-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-popper": patch
+  "@radix-ui/react-radio-group": patch
+  "@radix-ui/react-roving-focus": patch
+  "@radix-ui/react-scroll-area": patch
+  "@radix-ui/react-select": patch
+  "@radix-ui/react-slider": patch
+  "@radix-ui/react-switch": patch
+  "@radix-ui/react-tabs": patch
+  "@radix-ui/react-toast": patch
+  "@radix-ui/react-toggle": patch
+  "@radix-ui/react-toggle-group": patch
+  "@radix-ui/react-toolbar": patch
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
--- a/cypress/integration/ContextMenu.spec.ts
+++ b/cypress/integration/ContextMenu.spec.ts
@@ -6,7 +6,7 @@ describe('ContextMenu', () => {
     });
 
     describe('when using pointer', () => {
-      it.skip('should close open submenus and reopen the root menu when right clicking trigger', () => {
+      it('should close open submenus and reopen the root menu when right clicking trigger', () => {
         pointerOver('Bookmarks →');
         cy.findByText('Inbox').should('be.visible');
         cy.findByText('Right Click Here').rightclick({ force: true });
--- a/packages/core/primitive/src/primitive.tsx
+++ b/packages/core/primitive/src/primitive.tsx
@@ -6,7 +6,7 @@ function composeEventHandlers<E>(
   return function handleEvent(event: E) {
     originalEventHandler?.(event);
 
-    if (checkForDefaultPrevented === false || !(event as unknown as Event).defaultPrevented) {
+    if (checkForDefaultPrevented === false || !((event as unknown) as Event).defaultPrevented) {
       return ourEventHandler?.(event);
     }
   };
--- a/packages/core/primitive/src/primitive.tsx
+++ b/packages/core/primitive/src/primitive.tsx
@@ -1,3 +1,9 @@
+import * as ReactDOM from 'react-dom';
+
+function dispatchDiscreteCustomEvent(target: Document | HTMLElement, event: CustomEvent) {
+  ReactDOM.flushSync(() => target.dispatchEvent(event));
+}
+
 function composeEventHandlers<E>(
   originalEventHandler?: (event: E) => void,
   ourEventHandler?: (event: E) => void,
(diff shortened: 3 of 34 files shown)


packages/core/primitive/src/primitive.tsx
```diff
@@ -1,3 +1,5 @@
+import * as ReactDOM from 'react-dom';
```
andy-hook, line 1: I put this inside of the `core/primitive` package to colocate with the other event helper, do we feel that's the right place for it given the new deps which are react specific? would `react/primitive` be better?

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -269,19 +273,19 @@ function useFocusOutside(onFocusOutside?: (event: FocusOutsideEvent) => void) {
 }
 
 function dispatchUpdate() {
-  const event = new Event(CONTEXT_UPDATE);
+  const event = new CustomEvent(CONTEXT_UPDATE);
```
andy-hook, line 276: > I’ve changed every usage of the the Event constructor to be CustomEvent when a custom type is provided, this is a style change as either constructor is susceptible (it’s important to remember priority is inferred by the type passed, not which constructor used) the hope being that staying consistent will make deciding when to use dispatchDiscreteCustomEvent easier.

Interested for feedback on this, my understand is that these are equivalent aside from the optional `detail`, so I figured using `CustomEvent` for anything custom would make reasoning easier.

packages/core/primitive/src/primitive.tsx
benoitgrelard: Yeah I think `react/primitive` might make more sense for that reason.

packages/core/primitive/src/primitive.tsx
```diff
@@ -18,4 +14,37 @@ function composeEventHandlers<E>(
...
 
+/**
+ * Flush custom event dispatch
+ * https://github.com/radix-ui/primitives/pull/1378
+ *
+ * React batches *all* event handlers since version 18, this introduces certain considerations when using custom event types.
+ *
+ * Internally, React prioritises events in the following order:
+ *  - discrete
+ *  - continous
```
benoitgrelard, line 25:
```suggestion
 *  - continuous
```

packages/core/primitive/src/primitive.tsx
```diff
@@ -18,4 +14,37 @@ function composeEventHandlers<E>(
...
+ * `discrete` is an  important distinction as updates within these events are applied immediately.
+ * React however, is not able to infer the priority of custom event types due to how they are detected internally.
+ * Because of this, it's possible for updates from custom events to be unexpectedly batched when
+ * dispatched by another `discrete` event.
+ *
+ * In order to ensure that updates from custom events are applied predictably, we need to manually flush the batch.
+ * This utility should be used when dispatching a custom event from within another `discrete` event, for example:
+ *
+ * dispatching a click event:
+ * target.dispatchEvent(new Event(‘click’))
```
benoitgrelard, line 37: Wouldn't this work automatically given that React would find that native type?

cypress/integration/ContextMenu.spec.ts
```diff
@@ -6,7 +6,7 @@ describe('ContextMenu', () => {
     });
 
     describe('when using pointer', () => {
-      it.skip('should close open submenus and reopen the root menu when right clicking trigger', () => {
+      it('should close open submenus and reopen the root menu when right clicking trigger', () => {
```
benoitgrelard, line 9: what happened here?

packages/react/avatar/package.json
```diff
@@ -23,7 +23,8 @@
     "@radix-ui/react-use-layout-effect": "workspace:*"
   },
   "peerDependencies": {
-    "react": "^16.8 || ^17.0 || ^18.0"
+    "react": "^16.8 || ^17.0 || ^18.0",
+    "react-dom": "^16.8 || ^17.0 || ^18.0"
```
benoitgrelard, line 27: Will have to re-review these with the change of package so I'm skipping for now.

packages/react/toast/src/Toast.tsx
```diff
@@ -471,10 +471,18 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
                   const eventDetail = { originalEvent: event, delta };
                   if (hasSwipeMoveStarted) {
                     swipeDeltaRef.current = delta;
-                    dispatchCustomEvent(TOAST_SWIPE_MOVE, onSwipeMove, eventDetail);
+                    handleAndDispatchDiscreteCustomEvent(
```
benoitgrelard, line 474: haha, annoying 😀

packages/core/primitive/src/primitive.tsx
andy-hook: Yup, I was trying to show an example of when you would / wouldn't use this util

For example
- A known type 👎 
- A custom type called by a non-discrete event 👎 
- A custom type called by a discrete event 👍 

Perhaps I can make this clearer in the comment some how 🤔

cypress/integration/ContextMenu.spec.ts
andy-hook: We fixed the issue so the test can be re-introduced.

packages/core/primitive/src/primitive.tsx
benoitgrelard: oh right!!! I understand now.
Yeah I think we should re-write this a bit then because when reading this:

> This utility should be used when dispatching a custom event from within another `discrete` event, for example: 

I understood that all these should use it.

Perhaps as simple as being explicit and saying here's when you shouldn't do it, show the 2 first examples. Here's when you should do it (last example) with the descriptions you provided it's even better:

>A known type 👎
A custom type called by a non-discrete event 👎
A custom type called by a discrete event 👍

andy-hook: Hopefully this is a clearer now https://github.com/radix-ui/primitives/pull/1378/commits/9d35751ca74ca99444709bab9e4fc83e7b7c4e5d

.yarn/versions/ce619930.yml
```diff
@@ -0,0 +1,43 @@
+releases:
+  "@radix-ui/react-accessible-icon": patch
```
benoitgrelard, line 2: Just checking if you've re-run this from scratch after the updates on the packages?

packages/core/primitive/src/primitive.tsx
```diff
@@ -6,7 +6,7 @@ function composeEventHandlers<E>(
   return function handleEvent(event: E) {
     originalEventHandler?.(event);
 
-    if (checkForDefaultPrevented === false || !((event as unknown) as Event).defaultPrevented) {
+    if (checkForDefaultPrevented === false || !(event as unknown as Event).defaultPrevented) {
```
benoitgrelard, line 9: Ah I guess this change will trigger the changes in versions anyway so your versions are probably correct.

packages/react/collection/package.json
```diff
@@ -23,7 +23,8 @@
     "@radix-ui/react-slot": "workspace:*"
   },
   "peerDependencies": {
-    "react": "^16.8 || ^17.0 || ^18.0"
+    "react": "^16.8 || ^17.0 || ^18.0",
+    "react-dom": "^16.8 || ^17.0 || ^18.0"
```
benoitgrelard, line 27: Why does this one need it?

yarn.lock
```diff
@@ -3051,6 +3051,7 @@ __metadata:
     "@radix-ui/react-visually-hidden": "workspace:*"
   peerDependencies:
     react: ^16.8 || ^17.0 || ^18.0
+    react-dom: ^16.8 || ^17.0 || ^18.0
```
benoitgrelard, line 3054: re-ran `yarn` too?

packages/react/collection/package.json
andy-hook: It consumes types from `@radix-ui/react-primitive`, does it still apply in that case?

packages/core/primitive/src/primitive.tsx
andy-hook: In fact I'm going to revert it just to keep things simple.

packages/react/collection/package.json
benoitgrelard: I guess it still does by the definition of peer deps sadly…

benoitgrelard (approved):

Andarist:
A great level of detail has been added here in the description and in the comment in the code 👍

PR #1379, opened 2022-05-09
[Select] When there's no `value` or `defaultValue`, align with first non-disabled item

### LINKED ISSUE #1324: [Select] Select content is misaligned when defaultValue or value is not passed
## Bug report

When `value` or `defaultValue` is not passed `SelectContent` is misaligned.

### Current Behavior

Screenshot 2022-04-13 at 10 31 46

### Expected behavior

I think the behavior of the custom select should replicate the behavior of the native select and the first available option should be automatically selected in this case.

### Reproducible example

https://codesandbox.io/s/happy-leaf-oscbxo?file=/App.js

### Suggested solution

Automatically select first available option if `defaultValue` or `value` is not passed. Basically, the same behavior as with a native select.

### Your environment

| Radix Package(s) |  react-select       |   0.1.1      |

benoitgrelard:
Fixes #1324 

I've managed to do even slightly better than we had tested. It will now align not with the first item, but the first non-disabled item. This gives a better experience if some of the first items are disabled.

I've added a chromatic story for it too.

> **Note**: Best to review individual commits

### DIFF
--- a/packages/react/select/src/Select.stories.tsx
+++ b/packages/react/select/src/Select.stories.tsx
@@ -150,6 +150,42 @@ export const Position = () => (
   </div>
 );
 
+export const NoDefault = () => (
+  <div style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', height: '100vh' }}>
+    <Label>
+      Choose a number:
+      <Select.Root>
+        <Select.Trigger className={triggerClass()}>
+          <Select.Value />
+          <Select.Icon />
+        </Select.Trigger>
+        <Select.Content className={contentClass()}>
+          <Select.Viewport className={viewportClass()}>
+            <Select.Item className={itemClass()} value="one" disabled>
+              <Select.ItemText>One</Select.ItemText>
+              <Select.ItemIndicator className={indicatorClass()}>
+                <TickIcon />
+              </Select.ItemIndicator>
+            </Select.Item>
+            <Select.Item className={itemClass()} value="two">
+              <Select.ItemText>Two</Select.ItemText>
+              <Select.ItemIndicator className={indicatorClass()}>
+                <TickIcon />
+              </Select.ItemIndicator>
+            </Select.Item>
+            <Select.Item className={itemClass()} value="three">
+              <Select.ItemText>Three</Select.ItemText>
+              <Select.ItemIndicator className={indicatorClass()}>
+                <TickIcon />
+              </Select.ItemIndicator>
+            </Select.Item>
+          </Select.Viewport>
+        </Select.Content>
+      </Select.Root>
+    </Label>
+  </div>
+);
+
 export const Typeahead = () => (
   <div style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', height: '300vh' }}>
     <Label>
--- a/packages/react/select/src/Select.stories.tsx
+++ b/packages/react/select/src/Select.stories.tsx
@@ -118,7 +118,7 @@ export const Position = () => (
   >
     <Label>
       Choose an item:
-      <Select.Root defaultValue="item-25">
+      <Select.Root>
         <Select.Trigger className={triggerClass()}>
           <Select.Value />
           <Select.Icon />
@@ -133,7 +133,7 @@ export const Position = () => (
                   key={value}
                   className={itemClass()}
                   value={value}
-                  disabled={i > 5 && i < 9}
+                  disabled={i >= 0 && i < 9}
                 >
                   <Select.ItemText>item {i + 1}</Select.ItemText>
                   <Select.ItemIndicator className={indicatorClass()}>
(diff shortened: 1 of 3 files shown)


packages/react/select/src/Select.tsx
```diff
@@ -624,7 +625,25 @@ const SelectContentImpl = React.forwardRef<SelectContentImplElement, SelectConte
       }
     });
 
+    const handleItemSeen = React.useCallback(
+      (node: SelectItemElement | null, value: string, disabled: boolean) => {
+        const isFirstSeenValidItem = !firstItemRef.current && !disabled;
+        const isSelectedItem = context.value !== undefined && context.value === value;
```
andy-hook, line 631: Is `context.value !== undefined` so that we can support empty strings as values for items?

packages/react/select/src/Select.tsx
```diff
@@ -380,6 +380,7 @@ const SelectContentImpl = React.forwardRef<SelectContentImplElement, SelectConte
     const [isPositioned, setIsPositioned] = React.useState(false);
     const shouldRepositionRef = React.useRef(true);
     const shouldExpandOnScrollRef = React.useRef(false);
+    const firstItemRef = React.useRef(false);
```
andy-hook, line 383: As we're using this to short-circuit perhaps something like `hasFoundFirstSelectableItemRef` (or a less verbose alternative) might be better? At first glance this reads like it was the `ref` of the first item regardless of validity.

Also, is this something we need to cleanup after closing?

packages/react/select/src/Select.tsx
```diff
@@ -624,7 +625,25 @@ const SelectContentImpl = React.forwardRef<SelectContentImplElement, SelectConte
...
+          if (isFirstSeenValidItem) firstItemRef.current = true;
+        }
+      },
+      [context.value]
+    );
     const handleItemLeave = React.useCallback(() => content?.focus(), [content]);
+    const handleItemTextSeen = React.useCallback(
+      (node: SelectItemTextElement | null, value: string) => {
+        const isSelectedItemText = context.value !== undefined && context.value === value;
+        if (isSelectedItemText || !firstItemRef?.current) setSelectedItemText(node);
```
andy-hook, line 643: `!firstItemRef?.current` feels incidental based on React rendering, will the if block run for every iteration up until the parent item passes the validity check? would that be expected?

packages/react/select/src/Select.tsx
```diff
@@ -624,7 +625,25 @@ const SelectContentImpl = React.forwardRef<SelectContentImplElement, SelectConte
...
+        const isSelectedItem = context.value !== undefined && context.value === value;
+        if (isSelectedItem || isFirstSeenValidItem) {
+          setSelectedItem(node);
+          if (isFirstSeenValidItem) firstItemRef.current = true;
+        }
+      },
+      [context.value]
+    );
     const handleItemLeave = React.useCallback(() => content?.focus(), [content]);
+    const handleItemTextSeen = React.useCallback(
```
andy-hook, line 640: I'm not sure about usage of `seen` in the naming, I feel like we should be more specific as relates to the internals. Struggling for alternatives though but is it right to say we want to convey this handles when items are "available" or have transitioned into "actionable" / "rendered" state?

packages/react/select/src/Select.tsx
```diff
@@ -625,7 +625,25 @@ const SelectContentImpl = React.forwardRef<SelectContentImplElement, SelectConte
...
+      (node: SelectItemTextElement | null, value: string) => {
+        const isSelectedItemText = context.value !== undefined && context.value === value;
+        if (isSelectedItemText || !firstItemRef?.current) setSelectedItemText(node);
+      },
+      [context.value]
+    );
```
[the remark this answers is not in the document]
andy-hook: I like the colocation 👍

packages/react/select/src/Select.stories.tsx
```diff
@@ -470,6 +506,37 @@ export const ChromaticBottomLastPaddedViewport = () => (
...
+                <TickIcon />
+              </Select.ItemIndicator>
+            </Select.Item>
+          ))}
+        </Select.Viewport>
+        <Select.ScrollDownButton className={scrollDownButtonClass()}>▼</Select.ScrollDownButton>
+      </Select.Content>
+    </Select.Root>
+  </div>
+);
```
benoitgrelard, line 538:
```suggestion
);
ChromaticNoDefaultValue.parameters = { chromatic: { disable: false } };
```

packages/react/select/src/Select.tsx
```diff
@@ -317,7 +317,6 @@ const SelectContent = React.forwardRef<SelectContentElement, SelectContentProps>
       <SelectContentImpl {...props} ref={forwardedRef} />
     ) : fragment ? (
       ReactDOM.createPortal(
-        // @ts-ignore: This is to avoid the "SelectViewport must be inside "SelectContent" error
```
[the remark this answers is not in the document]
benoitgrelard: I think it was just me being lazy and never getting back to it.

packages/react/select/src/Select.tsx
benoitgrelard: Yes exactly.

packages/react/select/src/Select.tsx
benoitgrelard: How about `firstValidItemFoundRef`?

packages/react/select/src/Select.tsx
benoitgrelard: I don't think that is incidental, because React always renders the children first. So it will let any item text through until it finds its parent item which is the first non-disabled one.

packages/react/select/src/Select.tsx
benoitgrelard: I'm not sure I follow here.

andy-hook: What does “seen” refer to in this context?

benoitgrelard: I wanted to say it's been seen by the "content" part, as in it knows it mounted through the callback ref.
It's making me question now though where I need to guard it with `if (node)` too? 🤔

packages/react/select/src/Select.tsx
andy-hook: If I have a list of 4 items and the first 3 are disabled, `setSelectedItemText` be called incorrectly 3 times followed by a valid fourth, this is in contrast to `setSelectedItem` which is only called once.  

Is that what we want? shouldn't the behaviour match?

benoitgrelard: oh you mean specifically because of the `disabled` check? Yeah I guess we could add a disabled check in the item text too to make it equivalent if that's what you mean?

andy-hook: Yes, sorry, in effect that’s what I meant. My bigger point was that the code expects the parent `Item` to run before `ItemText`, this doesn't happen because React renders children first, and so one behaves differently until the flag is updated by the parent.

packages/react/select/src/Select.tsx
andy-hook: Gotcha, is it clearer to keep it simple `handleItemMount`? I don't feel super strongly tbh, just that "seen" makes me think of visibility logic or something with overflow vs lifecycle.

> It's making me question now though where I need to guard it with if (node) too?

hmm, not sure it makes much difference, `position` still has to guard anyway right?

andy-hook (approved):

MatheusPires99:
Hey @benoitgrelard , this issues is still happening on the CodeSandbox link example of the `Select` component.

https://codesandbox.io/s/5exdpd?module=App.js

If you remove the `defaultValue` the alignment is off.

MatheusPires99:
Oh nice to hear that. I will wait to check the v1 then. Thanks a lot @benoitgrelard

PR #1384, opened 2022-05-10
[Select] Add placeholder support

### LINKED ISSUE #1235: [Select] Provide a Placeholder component
### Overview

I'm working on migrating parts of a project from `react-aria` to Radix. Structurally, not every select menu in my project can have a `defaultValue` -- it would be overreaching in the application to provide a default selection for some form fields. In #1231, it's somewhat clear that the Radix Select primitive doesn't have first-class support for an unselected state -- a `value` or a `defaultValue` has to be supplied in order the menu to line up with its trigger.

I imagine something like a `Placeholder` component could support this case well in a Radix-y fashion. `Placeholder` would conditionally render its children when `value` and `defaultValue` are `null | undefined` (or perhaps when value and defaultValue have no corresponding `Item` (?)).

Usage could look like

```jsx
<Select.Root value={null}>
  <Select.Trigger>
    <Select.Value /> {/* Not rendered? */}
    <Select.Placeholder>
      Select an option {/* Rendered! */}
    </Select.Placeholder>
    <Select.Icon />
  </Select.Trigger>

  {/* ... */}
</Select.Root>
```

[@react-aria/select](https://react-spectrum.adobe.com/react-aria/useSelect.html)
```tsx
<span {...valueProps}>
  {state.selectedItem
    ? state.selectedItem.rendered
    : 'Select an option'}
</span>
```

[react-select](https://react-select.com/props) (docs in-page links are broken but search "placeholder")
```tsx
<Select
  // ...
  placeholder="Select an option"
/>
```

Everybody using the Radix Select primitive.

### Additional context

I recognize that the native `<select>` element doesn't support empty values and folks have to provide some sort of default selection and disable the option as shown in [answers to this SO post](https://stackoverflow.com/questions/5805059/how-do-i-make-a-placeholder-for-a-select-box). I don't think this is a very savory solution (and judging by spotty browser behaviors in the comments on that post, I don't think other folks do either). I think it's likely a reason folks reach for a JS library for select menus in the first place, so I think Radix could embrace that here.

Thank you by the way! Loving this library so far.

benoitgrelard:
Closes #1235

### DIFF
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -197,8 +197,10 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
         aria-autocomplete="none"
         aria-labelledby={labelledBy}
         dir={context.dir}
+        data-state={context.open ? 'open' : 'closed'}
         disabled={disabled}
         data-disabled={disabled ? '' : undefined}
+        data-placeholder={context.value === undefined ? '' : undefined}
         {...triggerProps}
         ref={composedRefs}
         onPointerDown={composeEventHandlers(triggerProps.onPointerDown, (event) => {
@@ -271,6 +273,30 @@ const SelectValue = React.forwardRef<SelectValueElement, SelectValueProps>(
 );
 
 SelectValue.displayName = VALUE_NAME;
+
+/* -------------------------------------------------------------------------------------------------
+ * SelectPlaceholder
+ * -----------------------------------------------------------------------------------------------*/
+
+const PLACEHOLDER_NAME = 'SelectPlaceholder';
+
+interface SelectPlaceholderProps {
+  children?: React.ReactNode;
+}
+
+const SelectPlaceholder: React.FC<SelectPlaceholderProps> = (
+  props: ScopedProps<SelectPlaceholderProps>
+) => {
+  const { __scopeSelect, children } = props;
+  const context = useSelectContext(PLACEHOLDER_NAME, __scopeSelect);
+
+  /* Portal the placeholder into the trigger value node when there is no value */
+  return context.value === undefined && context.valueNode && !context.valueNodeHasChildren
+    ? ReactDOM.createPortal(children, context.valueNode)
+    : null;
+};
+
+SelectPlaceholder.displayName = PLACEHOLDER_NAME;
 
 /* -------------------------------------------------------------------------------------------------
  * SelectIcon
@@ -1341,6 +1367,7 @@ function wrapArray<T>(array: T[], startIndex: number) {
 const Root = Select;
 const Trigger = SelectTrigger;
 const Value = SelectValue;
+const Placeholder = SelectPlaceholder;
 const Icon = SelectIcon;
 const Content = SelectContent;
 const Viewport = SelectViewport;
@@ -1359,6 +1386,7 @@ export {
   Select,
   SelectTrigger,
   SelectValue,
+  SelectPlaceholder,
   SelectIcon,
   SelectContent,
   SelectViewport,
@@ -1374,6 +1402,7 @@ export {
   Root,
   Trigger,
   Value,
+  Placeholder,
   Icon,
   Content,
   Viewport,
@@ -1390,6 +1419,7 @@ export type {
   SelectProps,
   SelectTriggerProps,
   SelectValueProps,
+  SelectPlaceholderProps,
   SelectIconProps,
   SelectContentProps,
   SelectViewportProps,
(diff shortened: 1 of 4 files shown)


packages/react/select/src/Select.tsx
```diff
@@ -272,6 +274,30 @@ const SelectValue = React.forwardRef<SelectValueElement, SelectValueProps>(
...
+
+const SelectPlaceholder: React.FC<SelectPlaceholderProps> = (
+  props: ScopedProps<SelectPlaceholderProps>
+) => {
+  const { __scopeSelect, children } = props;
+  const context = useSelectContext(PLACEHOLDER_NAME, __scopeSelect);
+
+  /* Portal the placeholder into the trigger value node when there is no value */
+  return context.value === undefined && context.valueNode && !context.valueNodeHasChildren
+    ? ReactDOM.createPortal(children, context.valueNode)
```
jjenzz, line 295: given this doesn't render anything except the children, i'm curious what the thinking is behind making this a part instead of a prop on root?

benoitgrelard: That's a good question. It started as a part originally. That said, I still feel like it's better as a component for a few reasons:
- "rendering" it inside the `Trigger` feels good seeing as that's where you want it
- It can be more than a string, and I feel like anything but a simple string reads pretty awkward in JSX as a prop

Arguments against:
- usually parts render an element
- it'll probably end up as a prop in people's abstractions anyway

Curious about other pros/cons to help make a decision!

jjenzz: ah yeah, good points. 

i guess the prop could be on the `Value` part regarding your first point (to colocate it better), but yeah, if people want more than a string i agree with the "JSX in a prop" niggle. 

i wonder tho if most of the time it would be a string and this is potentially overheard for a stylistic preference that's not the 90% use-case?

benoitgrelard: > i guess the prop could be on the Value part regarding your first point (to colocate it better)

That could be nice potentially

> but yeah, if people want more than a string i agree with the "JSX in a prop" niggle.

haha yeah, I hate to see JSX in props 😀, but again likely that's what it will become in all abstractions of the pattern… so? 🤷‍♂️ 

> i wonder tho if most of the time it would be a string and this is potentially overheard for a stylistic preference that's not the 90% use-case?

Yeah that's a good point too.

> only other point i can think of is SSR, prop could render placeholder from server.

How does a prop vs. part+`children` prop changes things here?

cc @andy-hook, would love to get your input here too so we can make a call.

benoitgrelard: We are all leaning towards adding placeholder prop to Value part now, so we'll do that.

andy-hook (approved):
Simple but effective 👌 

Should we start adding simple stories to Chromatic for things like this?

benoitgrelard:
> Should we start adding simple stories to Chromatic for things like this?

We do, I had a story for no value already, and I have added a example with the placeholder to it. So we have both cases.

Let me know if I misunderstood though.

andy-hook:
> Let me know if I misunderstood though.

No you're right, I misread the story location when scanning the diff 👍

drewkos:
@benoitgrelard trying to implement this placeholder value attr, but it's not working for me. is there somethign else I need to do besides add `placeholder={placeholderString}` to `<Select.Value />`?

benoitgrelard:
> @benoitgrelard trying to implement this placeholder value attr, but it's not working for me. is there somethign else I need to do besides add `placeholder={placeholderString}` to `<Select.Value />`?

No that's all, but maybe you are using the latest public rather than the latest release candidate?
This change isn't yet available in the public version.

MatheusPires99:
Hey @benoitgrelard , which version exactly the placeholder prop is available?

packages/react/select/src/Select.tsx
```diff
@@ -243,18 +245,20 @@ const VALUE_NAME = 'SelectValue';
...
 
 const SelectValue = React.forwardRef<SelectValueElement, SelectValueProps>(
   (props: ScopedProps<SelectValueProps>, forwardedRef) => {
     // We ignore `className` and `style` as this part shouldn't be styled.
-    const { __scopeSelect, className, style, ...valueProps } = props;
+    const { __scopeSelect, className, style, children, placeholder, ...valueProps } = props;
     const context = useSelectContext(VALUE_NAME, __scopeSelect);
     const { onValueNodeHasChildrenChange } = context;
-    const hasChildren = props.children !== undefined;
+    const hasChildren = children !== undefined;
```
LexSwed, line 258: One of the difficult things for me to figure out was that `value` has to be `undefined` for the `placeholder` to be shown. I used to use `null` or `''` for empty (reset) value, and not just missing. I understand it's a wrong habit to have, just want others to be clear about it.

aromeronavia: This is a veeeery good point, and I just spent 1 hour debugging why my placeholder was not showing up...

PR #1389, opened 2022-05-11
[RovingFocusGroup][Toast][Tooltip] Avoid flushing dispatch when called in lifecycle

andy-hook:
Some wider changes in https://github.com/radix-ui/primitives/pull/1378 were a little overzealous as `flushSync` was unexpectedly being called during lifecycle in some cases.

In `Tooltip` this was clear because `onChange` is [called during lifecycle](https://github.com/radix-ui/primitives/blob/c92d43343f1233a94a41d12c5eb56a42b72dc98e/packages/react/use-controllable-state/src/useControllableState.tsx#L49), `RovingFocusGroup` and `Toast` were due to `focus` / `focusout` being called either during mount or during teardown.

Interesting to note the implicitness of the `focusout` case due to the browser removing the nodes and then firing the event.

### DIFF
--- a/packages/react/primitive/src/Primitive.tsx
+++ b/packages/react/primitive/src/Primitive.tsx
@@ -98,8 +98,9 @@ const AS_ERROR = `Warning: The \`as\` prop has been removed in favour of \`asChi
  * dispatching a custom type within a `discrete` event 👍
  * onPointerDown={(event) => dispatchDiscreteCustomEvent(event.target, new CustomEvent(‘customType’))}
  *
- * Note: care should be taken with the `focusout` event as it's possible for handlers to be called
- * implicitly during lifecycle, e.g. when focus is within the component as it is unmounted.
+ * Note: though React classifies `focus`, `focusin` and `focusout` events as `discrete`, it's  not recommended to use
+ * this utility with them. This is because it's possible for those handlers to be called implicitly during render
+ * e.g. when focus is within a component as it is unmounted, or when managing focus on mount.
  */
 
 function dispatchDiscreteCustomEvent<E extends CustomEvent>(target: E['target'], event: E) {
--- a/packages/react/primitive/src/Primitive.tsx
+++ b/packages/react/primitive/src/Primitive.tsx
@@ -79,6 +79,8 @@ const AS_ERROR = `Warning: The \`as\` prop has been removed in favour of \`asChi
  *  - continuous
  *  - default
  *
+ * https://github.com/facebook/react/blob/a8a4742f1c54493df00da648a3f9d26e3db9c8b5/packages/react-dom/src/events/ReactDOMEventListener.js#L294-L350
+ *
  * `discrete` is an  important distinction as updates within these events are applied immediately.
  * React however, is not able to infer the priority of custom event types due to how they are detected internally.
  * Because of this, it's possible for updates from custom events to be unexpectedly batched when
--- a/packages/react/primitive/src/Primitive.tsx
+++ b/packages/react/primitive/src/Primitive.tsx
@@ -97,6 +97,9 @@ const AS_ERROR = `Warning: The \`as\` prop has been removed in favour of \`asChi
  *
  * dispatching a custom type within a `discrete` event 👍
  * onPointerDown={(event) => dispatchDiscreteCustomEvent(event.target, new CustomEvent(‘customType’))}
+ *
+ * Note: care should be taken with the `focusout` event as it's possible for handlers to be called
+ * implicitly during lifecycle, e.g. when focus is within the component as it is unmounted.
  */
 
 function dispatchDiscreteCustomEvent<E extends CustomEvent>(target: E['target'], event: E) {
--- a/packages/react/roving-focus/src/RovingFocusGroup.tsx
+++ b/packages/react/roving-focus/src/RovingFocusGroup.tsx
@@ -159,10 +159,6 @@ const RovingFocusGroupImpl = React.forwardRef<
 
           if (event.target === event.currentTarget && isKeyboardFocus && !isTabbingBackOut) {
             const entryFocusEvent = new CustomEvent(ENTRY_FOCUS, EVENT_OPTIONS);
-
-            // we avoid dispatching via `dispatchDiscreteCustomEvent` as it's possible
-            // for focus to be triggered during lifecycle, this is most common when
-            // focus is triggered by `FocusScope` and the `onMountAutoFocus` prop
             event.currentTarget.dispatchEvent(entryFocusEvent);
 
             if (!entryFocusEvent.defaultPrevented) {
--- a/packages/react/roving-focus/src/RovingFocusGroup.tsx
+++ b/packages/react/roving-focus/src/RovingFocusGroup.tsx
@@ -4,7 +4,7 @@ import { createCollection } from '@radix-ui/react-collection';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { createContextScope } from '@radix-ui/react-context';
 import { useId } from '@radix-ui/react-id';
-import { Primitive, dispatchDiscreteCustomEvent } from '@radix-ui/react-primitive';
+import { Primitive } from '@radix-ui/react-primitive';
 import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
 import { useControllableState } from '@radix-ui/react-use-controllable-state';
 import { useDirection } from '@radix-ui/react-direction';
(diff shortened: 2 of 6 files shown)


benoitgrelard (commented):
> For now it’s safer to revert changes outside the original scope, we can apply localised testing in these areas in the future if needed.

What do you mean by the original scope? I've noticed that a lot of these are clearly custom events from discrete events, so it's a bit confusing why those wouldn't work/shouldn't be discrete custom events.

andy-hook:
> What do you mean by the original scope?

I should have said “issue” to be honest. I agree that we should use the heuristic in all places where it applies and exceptions clearly stated / documented. I’ll update the PR to reflect this.

packages/react/toast/src/Toast.tsx
```diff
@@ -471,18 +485,14 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
                   const eventDetail = { originalEvent: event, delta };
                   if (hasSwipeMoveStarted) {
                     swipeDeltaRef.current = delta;
-                    handleAndDispatchDiscreteCustomEvent(
-                      TOAST_SWIPE_MOVE,
-                      onSwipeMove,
-                      eventDetail
-                    );
+                    dispatchCustomEvent(TOAST_SWIPE_MOVE, onSwipeMove, eventDetail, {
+                      discrete: false,
```
andy-hook, line 489: `onPointerMove` doesn't fit the heuristic and it makes sense to batch these anyway.

packages/react/primitive/src/Primitive.tsx
```diff
@@ -97,6 +97,9 @@ const AS_ERROR = `Warning: The \`as\` prop has been removed in favour of \`asChi
  *
  * dispatching a custom type within a `discrete` event 👍
  * onPointerDown={(event) => dispatchDiscreteCustomEvent(event.target, new CustomEvent(‘customType’))}
+ *
+ * Note: care should be taken with the `focusout` event as it's possible for handlers to be called
+ * implicitly during lifecycle, e.g. when focus is within the component as it is unmounted.
```
benoitgrelard, line 102: 👍

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -159,7 +159,11 @@ const RovingFocusGroupImpl = React.forwardRef<
 
           if (event.target === event.currentTarget && isKeyboardFocus && !isTabbingBackOut) {
             const entryFocusEvent = new CustomEvent(ENTRY_FOCUS, EVENT_OPTIONS);
-            dispatchDiscreteCustomEvent(event.currentTarget, entryFocusEvent);
+
+            // we avoid dispatching via `dispatchDiscreteCustomEvent` as it's possible
+            // for focus to be triggered during lifecycle, this is most common when
+            // focus is triggered by `FocusScope` and the `onMountAutoFocus` prop
+            event.currentTarget.dispatchEvent(entryFocusEvent);
```
benoitgrelard, line 166: 👍

packages/react/toast/src/Toast.tsx
benoitgrelard: Great point!

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -158,7 +158,10 @@ const Tooltip: React.FC<TooltipProps> = (props: ScopedProps<TooltipProps>) => {
       if (open) {
         // we dispatch here so `TooltipProvider` isn't required to
         // ensure other tooltips are aware of this one opening.
-        dispatchDiscreteCustomEvent(document, new CustomEvent(TOOLTIP_OPEN));
+        //
+        // as `onChange` is called within a lifecycle method we
+        // avoid dispatching via `dispatchDiscreteCustomEvent`.
+        document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN));
```
benoitgrelard, line 164: 👍

packages/react/toast/src/Toast.tsx
```diff
@@ -680,18 +686,21 @@ ToastClose.displayName = CLOSE_NAME;
 
 /* ---------------------------------------------------------------------------------------------- */
 
-function handleAndDispatchDiscreteCustomEvent<
-  E extends CustomEvent,
-  ReactEvent extends React.SyntheticEvent
->(
+function dispatchCustomEvent<E extends CustomEvent, ReactEvent extends React.SyntheticEvent>(
```
benoitgrelard, line 689: Did you rename because of the wrapping?
Didn't we have a similar function in another component? I'd rather keep consistent if they essentially do the same thing (minus the discrete boolean caveat)

andy-hook: > Didn't we have a similar function in another component

I kept the naming here due to the additional arg vs the other case which always dispatches `discrete`.

Perhaps they should align them both to be `handleAndDispatchCustomEvent`?

benoitgrelard: Yeah I think so, as that's what they're both doing.

packages/react/primitive/src/Primitive.tsx
```diff
@@ -97,6 +97,10 @@ const AS_ERROR = `Warning: The \`as\` prop has been removed in favour of \`asChi
  *
  * dispatching a custom type within a `discrete` event 👍
  * onPointerDown={(event) => dispatchDiscreteCustomEvent(event.target, new CustomEvent(‘customType’))}
+ *
+ * Note: though React classifies `focus`, `focusin` and `focusout` events as `discrete`, it's  not recommended to use
+ * this utility with them. This is because it's possible for those handlers to be called implicitly during render
+ * e.g. when focus is within a component as it is unmounted, or when managing focus on mount.
```
andy-hook, line 103: I updated this comment and removed others as they should be covered by this now.

andy-hook:
@benoitgrelard I've updated this from our findings and reviewed relevant stories.

packages/react/toast/src/Toast.tsx
```diff
@@ -157,14 +157,16 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
           const pauseEvent = new CustomEvent(VIEWPORT_PAUSE);
-          dispatchDiscreteCustomEvent(viewport, pauseEvent);
+          viewport.dispatchEvent(pauseEvent);
           context.isClosePausedRef.current = true;
         };
+
         const handleResume = () => {
           const resumeEvent = new CustomEvent(VIEWPORT_RESUME);
-          dispatchDiscreteCustomEvent(viewport, resumeEvent);
+          viewport.dispatchEvent(resumeEvent);
```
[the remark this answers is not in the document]
benoitgrelard: Just so I understand, the pointerenter/pointerleave ones are excluded too because they are not considered discrete by React?

btw, it might be nice to add a link to https://github.com/facebook/react/blob/a8a4742f1c54493df00da648a3f9d26e3db9c8b5/packages/react-dom/src/events/ReactDOMEventListener.js#L294-L350 directly in that big comment in the primitive file.

benoitgrelard (approved):

packages/react/toast/src/Toast.tsx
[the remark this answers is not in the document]
andy-hook: Exactly

> it might be nice to add a link

Good shout

jjenzz:
i've been following this all for educational purposes and woah, 🤯 this is a complex issue ha.

i wonder if it would have been simpler to create a generic utility that is used for all custom events but it reads the original triggering `event.type` to determine whether it should be discrete instead?

```jsx
const handleFocus = (event) => {
  dispatchCustomEvent({ originalEvent: event, event: CUSTOM_EVENT, handler, detail }); 
}
```

that way the logic that determines when things should be discrete or not is not littered across the codebase.

benoitgrelard:
I like the idea! Although seeing how things have been going with this subject, I'd be worried of locking the decision purely on the original even type for a few reasons:
- a bunch of times we might not have access to it as it happens as a series of events (like the focus redirect here with `FocusScope` + `RovingFocusGroup`
- the handling needs to remain separate from the dispatching because only a few places handle it as a one off, sometimes the listening happens somewhere less in the file, so you need to construct the event separately anyway
- we're still not 100% sure and are still learning about when we should/shouldn't flush so it feels better to remain "open" for now in how we handle it, perhaps once we're more sure of what we're doing we can try and formalise more.

andy-hook:
> i wonder if it would have been simpler to have a generic utility that is used for all custom events

We started here but for the reasons described we decided to forego a higher abstraction, there is still nuance to vet first.

Issue #1050, opened 2021-12-20
[ScrollArea] Add `Presence` to `ScrollThumb`


tawhidurrahman:
## Feature request
`Thumb` component's unmounting state of `Scroll Area` primitives to be animatable with stitches.

### Overview
I tried to animate the `opacity` of `Thumb` component's unmount state with stitches but couldn't. So one of your maintainers suggested me to add an issue.

```
import * as ScrollAreaPrimitive from '@radix-ui/react-scroll-area';
const ScrollAreaScrollbar = styled(ScrollAreaPrimitive.Scrollbar, {});

const show = keyframes({ '0%': { opacity: 0 }, '100%': { opacity: 1 } });

const fade = keyframes({ '100%': { opacity: 1 }, '0%': { opacity: 0 } });

export const ScrollAreaThumb = styled(ScrollAreaPrimitive.Thumb, {
  flex: 1,
  borderRadius: 6,
  backgroundColor: '#aaa',
  position: 'relative',
  [`${ScrollAreaScrollbar}[data-state="visible"] &`]: {
    animation: `${show} .2s forwards linear`,
  },

  //I want something like this

  [`${ScrollAreaScrollbar}[data-state="hidden"] &`]: {
    animation: `${fade} .2s forwards linear`,
  },
});
```

### Examples in other libraries
Checkout the thumb animation here. They are using `perfect-scrollbar`.

[https://demos.creative-tim.com/black-dashboard-pro-react/?_ga=2.257667595.1079478053.1638938139-485400365.1601043380#/admin/dashboard](url)

### Who does this impact? Who is this for?
Anyone specially me who want a beautiful smooth scrollbar.

jjenzz:
@tawhidurrahman I know I requested this ticket because I forgot to add `Presence` to the thumb but, in case it helps your use-case, you can animate the whole scrollbar instead https://codesandbox.io/s/laughing-darwin-9sdcd?file=/src/App.js

tawhidurrahman:
That'll do for now. Thank you. But whenever you can please try to add this feature.

PR #1392, opened 2022-05-12
[ScrollArea] Add `Presence` to `ScrollThumb`

bukinoshita:
### Description

Added `Presence` to `ScrollThumb`. Closes https://github.com/radix-ui/primitives/issues/1050

### DIFF
--- a/.yarn/versions/0ca5f5ea.yml
+++ b/.yarn/versions/0ca5f5ea.yml
@@ -1,5 +1,5 @@
 releases:
   "@radix-ui/react-scroll-area": patch
 
-decline:
+declined:
   - primitives
--- a/.yarn/versions/0ca5f5ea.yml
+++ b/.yarn/versions/0ca5f5ea.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-scroll-area": patch
+
+decline:
+  - primitives
--- a/packages/react/scroll-area/src/ScrollArea.stories.tsx
+++ b/packages/react/scroll-area/src/ScrollArea.stories.tsx
@@ -95,13 +95,11 @@ export const ContentChange = () => {
 
 export const Animated = () => {
   return (
-    <>
-      <ScrollAreaStory scrollAreaClassName={animatedThumbClass} style={{ width: 800, height: 800 }}>
-        {Array.from({ length: 30 }).map((_, index) => (
-          <Copy key={index} />
-        ))}
-      </ScrollAreaStory>
-    </>
+    <ScrollAreaStory scrollAreaClassName={animatedThumbClass} style={{ width: 800, height: 800 }}>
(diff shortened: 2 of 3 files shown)


benoitgrelard:
I think here you can simply add the new animation code directly to `thumbClass`.

Would you mind updating the name whilst you're here too?

```tsx
- const THUMB_NAME = 'ScrollbarThumb';
+ const THUMB_NAME = 'ScrollAreaThumb';
```

🙏

benoitgrelard:
> For the Storybook, we use [this component](https://github.com/radix-ui/primitives/blob/main/packages/react/scroll-area/src/ScrollArea.stories.tsx#L358-L377). I want to pass the animated class only for the [ScrollThumb](https://github.com/radix-ui/primitives/blob/main/packages/react/scroll-area/src/ScrollArea.stories.tsx#L367). What's recommended here? Add a new prop like scrollThumbClass and then <ScrollArea.Thumb className={scrollThumbClass ? scrollThumbClass() : thumbClass()} />?

Ah, I see what you're saying now, it might be good to only have the animation in the `Animated` story as otherwise all the chromatic stories won't see the thumb as it starts invisible because of the animation.

We usually don't have abstractions like this in stories, but it can happen.

In this case, I would say pass a `animated` prop to `ScrollAreaStory` and default it to `false` and only pass it to `true` in the `Animated` story.

bukinoshita:
> > For the Storybook, we use [this component](https://github.com/radix-ui/primitives/blob/main/packages/react/scroll-area/src/ScrollArea.stories.tsx#L358-L377). I want to pass the animated class only for the [ScrollThumb](https://github.com/radix-ui/primitives/blob/main/packages/react/scroll-area/src/ScrollArea.stories.tsx#L367). What's recommended here? Add a new prop like scrollThumbClass and then <ScrollArea.Thumb className={scrollThumbClass ? scrollThumbClass() : thumbClass()} />?
> 
> Ah, I see what you're saying now, it might be good to only have the animation in the `Animated` story as otherwise all the chromatic stories won't see the thumb as it starts invisible because of the animation.
> 
> We usually don't have abstractions like this in stories, but it can happen.
> 
> In this case, I would say pass a `animated` prop to `ScrollAreaStory` and default it to `false` and only pass it to `true` in the `Animated` story.

Done!

jjenzz:
@benoitgrelard usually when we wrapped `Presence` we would create an `Impl` part to prevent all the hook logic etc. running unnecessarily when it isn't mounted. it looks like that was forgotten here.

benoitgrelard:
That's a good point. Looking at the codebase, it seems there are other places we haven't done it (like `RadioGroupIndicator`, though arguably simple/nothing running there) but `ScrollAreaThumb` is reasonably complex that it should do it 👍

Issue #1309, opened 2022-04-05
[DropdownMenu][ContextMenu] Provide explicit `Sub` part to avoid nesting issues


benoitgrelard:
See #1016 for an example of problems.
Also, we have a couple precedents:
- in `Menu` itself we have the user use a `SubTrigger`
- in `NavigationMenu` we have distinct parts

PR #1394, opened 2022-05-12
[DropdownMenu][ContextMenu] Provide explicit `Sub` parts

andy-hook:
Closes https://github.com/radix-ui/primitives/issues/1309

- [x] https://github.com/radix-ui/primitives/pull/1393
- [x] https://github.com/radix-ui/primitives/pull/1396
- [x] https://github.com/radix-ui/primitives/pull/1406

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -155,7 +155,8 @@ const CONTENT_NAME = 'ContextMenuContent';
 
 type ContextMenuContentElement = React.ElementRef<typeof MenuPrimitive.Content>;
 type MenuContentProps = Radix.ComponentPropsWithoutRef<typeof MenuPrimitive.Content>;
-interface ContextMenuContentProps extends Omit<MenuContentProps, 'portalled' | 'side' | 'align'> {}
+interface ContextMenuContentProps
+  extends Omit<MenuContentProps, 'portalled' | 'side' | 'sideOffset' | 'align'> {}
 
 const ContextMenuContent = React.forwardRef<ContextMenuContentElement, ContextMenuContentProps>(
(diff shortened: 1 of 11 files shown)


benoitgrelard:
When doing stacked like this I think we need to do the version only in the top-level PR as otherwise it produces multiple version files and it's not happy about it.

andy-hook:
> When doing stacked like this I think we need to do the version only in the top-level PR as otherwise it produces multiple version files and it's not happy about it.

Indeed, i'll fix it when all the changes are in.

benoitgrelard:
Playing with the stories, I noticed that the Menu>submenus story is broken:
CleanShot 2022-05-20 at 10 15 17@2x

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -134,15 +120,17 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
...
             handleOpen(event);
+            event.preventDefault();
           })}
           onPointerDown={composeEventHandlers(
             props.onPointerDown,
             whenTouchOrPen((event) => {
               // clear the long press here in case there's multiple touch points
               clearLongPress();
-              longPressTimerRef.current = window.setTimeout(() => handleOpen(event), 700);
+              longPressTimerRef.current = window.setTimeout(() => {
```
benoitgrelard, line 131: nit: was there a reason for this change?

benoitgrelard (approved):
All looks good for me apart from that broken story.

packages/react/context-menu/src/ContextMenu.tsx
andy-hook: Formatting hangover from a previous change I think

PR #1401, opened 2022-05-16
[DismissableLayer] Fix pointer-events issues

### LINKED ISSUE #1241: [Dialog] body pointer-events: none remains after closing
### Current Behavior

With `Dialog` version 0.1.7 the pointer-events seems to remain on the `body` with the Chrome device emulator using a mobile device  (e.g. iPad Air), while in the 0.1.6 it works fine.

### Expected behavior

The pointer-events should always be removed from the body after the closing animation has finished.

### Reproducible example

The 2 repos are identical, except the `@radix-ui/react-dialog` version:
- [Version 0.1.7](https://codesandbox.io/s/dialog-0-1-7-pointer-events-bug-jhp0nd)
- [Version 0.1.6](https://codesandbox.io/s/dialog-0-1-6-pointer-events-bug-zgt5um)

### Additional context

I forked the `radix-ui/design-system` repo on codesandbox to reproduce this bug keeping the original code but changing just the version of the `@radix-ui/react-dialog`.

**Note**: In the version 0.1.7, the problem should be solved by removing the `close` animation from the Content

### LINKED ISSUE #1263: [Dialog] Expected clickable elements behind overlay not to trigger on clickOutside in mobile mode.
## Bug report

When using a portalled dialog, I was expecting elements not to be clickable underneath the overlay.

I was having trouble fixing it until I set the body to be:

```<body style={{ pointerEvents: 'auto' }}>```

Now it works like I anticipated.

### Current Behavior

Clicking elements under a Dialog.Overlay are triggered.

### Expected behavior

Elements under a Dialog.Overlay should not be clickable until Dialog is closed.

### Reproducible example

[Official Dialog sample](https://www.radix-ui.com/docs/primitives/components/dialog)

1. Click 'Edit profile'
2. Click 'Show Code' underneath overlay.
3. Observe that button was clicked and now says 'Hide code'

### Suggested solution

Not sure, but something is messing with pointerEvents on the body which is causing this.

### Your environment

| Radix Package(s) |    react-dialog     |      0.1.7   |
| React            | n/a     |   17.0.2      |
| Browser          |     Chrome    |         |

### PR DESCRIPTION
**Pull request by benoitgrelard**
Fixes #1241
Fixes #1263

### CODE REVIEW

File: packages/react/dismissable-layer/src/DismissableLayer.tsx [hunk 1]
```diff
@@ -106,14 +113,21 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
...
+          originalBodyPointerEvents = document.body.style.pointerEvents;
+          document.body.style.pointerEvents = 'none';
+        }
+        context.layersWithOutsidePointerEventsDisabled.add(node);
+      }
       context.layers.add(node);
       dispatchUpdate();
-    }, [node, disableOutsidePointerEvents, context]);
+      return () => {
+        resetBodyPointerEvents();
```
andy-hook, line 128: I'm wondering whether it would be easier to reason by colocating the logic of applying **and** removing in the same place? either inline here, or with two methods e.g.  `disableBodyPointerEvents` `restoreBodyPointerEvents`

andy-hook, line 128: I'm wondering whether it would be easier to reason by colocating the logic of applying **and** removing in the same place? either inline here, or with two methods where we have declared e.g.  `disableBodyPointerEvents` `restoreBodyPointerEvents`

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -106,14 +113,21 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
...
+      };
+    }, [node, disableOutsidePointerEvents, resetBodyPointerEvents, context]);
```
andy-hook, line 130: Same question as before about the context obj

andy-hook, line 130: Same question as before about the context obj

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -206,14 +220,41 @@ type FocusOutsideEvent = CustomEvent<{ originalEvent: FocusEvent }>;
...
+            eventDetail,
+            { discrete: true }
+          );
+        }
+
+        /**
+         * On touch devices, we need to wait for a click event because browsers implement
+         * a ~350ms delay between the time the user stops touching the display and when the
+         * browser executres events. We need to ensure we don't reactivate pointer-events within
+         * this timeframe otherwise the browser may execute events that should have been prevented.
```
andy-hook, line 243: Is this part still accurate? or is it simply an implicit outcome now? only looking at the code we now dispatch the event based on `click` rather than calling `resetPointerEvents`.

andy-hook, line 243: Is this part still accurate? or is simply more of an implicit outcome now? only looking at the code we now dispatch the event based on `click` rather than calling `resetPointerEvents`.

jjenzz: yeah still true, it's the same logic as before except it's listening for click in `pointerdown` instead of effect cleanup.

cypress/integration/Dialog.spec.ts
```diff
@@ -1,11 +1,198 @@
...
+  function shouldAllowOutsideInteraction(action: 'realClick' | 'realTouch') {
+    cy.findByLabelText('count up')
+      .invoke('text')
+      .then((count) => {
+        if (action === 'realTouch') {
+          cy.findByLabelText('count up').realClick();
+        } else {
+          cy.findByLabelText('count up').realTouch();
+        }
+        cy.findByLabelText('count up').should('not.eq', count);
```
andy-hook, line 33: Is this asserting what we expect? `count` is `text` but is `cy.findByLabelText('count up')` a button? will they always **not** be equal?

benoitgrelard: hmm, good shout, let me check on that.

benoitgrelard: You were 💯 on the money here, great catch! I made it fail and it was still passing. Fixed now: https://github.com/radix-ui/primitives/pull/1401/commits/4d941df0a6aa728c0f031cd7df76883465861966

cypress/integration/Dialog.spec.ts
```diff
@@ -1,11 +1,198 @@
...
+
+      // using mouse on an input outside an animated dialog
+      cy.findByText('open').click();
+      shouldBeOpen();
+      cy.findByPlaceholderText('name').realClick();
+      cy.findByPlaceholderText('name').should('not.be.focused');
+      shouldBeClosed();
+
+      // finally, make sure that we have no pointer-events regression
+      shouldAllowOutsideInteraction('realClick');
```
andy-hook, line 102:
```suggestion
      // finally, ensure that pointer-events have been reset and interactions restored
      shouldAllowOutsideInteraction('realClick');
```

Should we be clearer here? is this just to ensure that `pointer-events: none` is no longer set on the body?

Also, is there any chance of a footgun when the counter is incremented as part of this assertion?

benoitgrelard: > is this just to ensure that pointer-events: none is no longer set on the body?

Yes.

> Also, is there any chance of a footgun when the counter is incremented as part of this assertion?

Sorry, what do you mean?

andy-hook: > Sorry, what do you mean?

I was just wondering whether we need to increment again given the context, or if leaning on Cypress to tell us that `pointer-events: none` was still set (via a normal `.click`) would be clearer.

No biggie tbh, in some ways testing the cause and effect with the increment is better.

andy-hook (approved):

cypress/integration/Dialog.spec.ts
benoitgrelard: Yeah I feel like testing the actual result is stronger.

brandalx:
i had same error so i tried to solve my way and this actually work 
i did this by removing the style of pointer events like so:

`  onClose: () => {
    set({ type: null, isOpen: false });

    // Delay restoring pointer events for 2 seconds
    setTimeout(() => {
      document.body.style.pointerEvents = "auto";
    }, 1000);
  },
}));`

Note, i did it with small timeout to make sure that first all modal scripts are completed and then im manually removing that. Hope it helps!

PR #1404, opened 2022-05-17
[ScrollArea] Add `ScrollAreaThumbImpl` part

bukinoshita:
### Description

This PR adds the `Impl` part to separate the hook logic in the `ScrollArea` component. It addresses https://github.com/radix-ui/primitives/pull/1392#issuecomment-1126938126

### DIFF
--- a/packages/react/scroll-area/src/ScrollArea.tsx
+++ b/packages/react/scroll-area/src/ScrollArea.tsx
@@ -747,8 +747,8 @@ interface ScrollAreaThumbProps extends ScrollAreaThumbImplProps {
 
 const ScrollAreaThumb = React.forwardRef<ScrollAreaThumbElement, ScrollAreaThumbProps>(
   (props: ScopedProps<ScrollAreaThumbProps>, forwardedRef) => {
-    const { forceMount, __scopeScrollArea, ...contentProps } = props;
-    const scrollbarContext = useScrollbarContext(THUMB_NAME, __scopeScrollArea);
+    const { forceMount, ...thumbProps } = props;
+    const scrollbarContext = useScrollbarContext(THUMB_NAME, props.__scopeScrollArea);
     return (
       <Presence present={forceMount || scrollbarContext.hasThumb}>
         <ScrollAreaThumbImpl ref={forwardedRef} {...contentProps} />
--- a/packages/react/scroll-area/src/ScrollArea.tsx
+++ b/packages/react/scroll-area/src/ScrollArea.tsx
@@ -751,7 +751,7 @@ const ScrollAreaThumb = React.forwardRef<ScrollAreaThumbElement, ScrollAreaThumb
     const scrollbarContext = useScrollbarContext(THUMB_NAME, props.__scopeScrollArea);
     return (
       <Presence present={forceMount || scrollbarContext.hasThumb}>
-        <ScrollAreaThumbImpl ref={forwardedRef} {...contentProps} />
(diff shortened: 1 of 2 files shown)


packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -747,7 +747,22 @@ interface ScrollAreaThumbProps extends PrimitiveDivProps {
 
 const ScrollAreaThumb = React.forwardRef<ScrollAreaThumbElement, ScrollAreaThumbProps>(
   (props: ScopedProps<ScrollAreaThumbProps>, forwardedRef) => {
-    const { __scopeScrollArea, forceMount, style, ...thumbProps } = props;
+    const { forceMount, __scopeScrollArea, ...contentProps } = props;
+    const scrollbarContext = useScrollbarContext(THUMB_NAME, __scopeScrollArea);
```
andy-hook, line 751:
```suggestion
    const { forceMount, ...thumbProps } = props;
    const scrollbarContext = useScrollbarContext(THUMB_NAME, props.__scopeScrollArea);
```

We should ensure  `__scopeScrollArea` gets spread into the `impl` part as well

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -747,7 +747,22 @@ interface ScrollAreaThumbProps extends PrimitiveDivProps {
...
+    return (
+      <Presence present={forceMount || scrollbarContext.hasThumb}>
```
andy-hook, line 753: Just curious, I took a quick peek at the animated story and couldn't see the exit animation firing. When is `hasThumb` being updated? or is what I see in the story just because the parent scrollbar isn't also animated?

benoitgrelard: The best way to test it in the Content Change story, basically it's when a scrollArea becomes scrollable or non-scrollable, so when content gets added/removed, or it gets resized to the point a thumb isn't necessary (but depending on the `type`, the track might still be displayed).

andy-hook: Thanks! yeah looks like we’d need an additional `visible` check or something to support an exit animation as `hasThumb` persists `true`. That would expose other issues though e.g. if the scrollbar is animated but the thumb is not, the thumb would unmount before the scrollbar had finished its exit (which would look a bit janky).

So perhaps it’s fine as is for now.

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -747,7 +747,22 @@ interface ScrollAreaThumbProps extends PrimitiveDivProps {
 
 const ScrollAreaThumb = React.forwardRef<ScrollAreaThumbElement, ScrollAreaThumbProps>(
   (props: ScopedProps<ScrollAreaThumbProps>, forwardedRef) => {
-    const { __scopeScrollArea, forceMount, style, ...thumbProps } = props;
+    const { forceMount, ...thumbProps } = props;
+    const scrollbarContext = useScrollbarContext(THUMB_NAME, props.__scopeScrollArea);
+    return (
+      <Presence present={forceMount || scrollbarContext.hasThumb}>
+        <ScrollAreaThumbImpl ref={forwardedRef} {...contentProps} />
```
andy-hook, line 754:
```suggestion
        <ScrollAreaThumbImpl ref={forwardedRef} {...thumbProps} />
```

andy-hook (approved):

PR #1406, opened 2022-05-18
[DropdownMenu][ContextMenu] Improve `SubContent` types

andy-hook:
- Tidies up nesting logic
- Exposes `SubContent` part with accurate prop typings

I haven't re-ordered anything in the file yet to make the diff easier to read, not sure if we think it would be best to move them to the bottom of the file as well.

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -59,17 +59,22 @@ const usePopperScope = createPopperScope();
 const useRovingFocusGroupScope = createRovingFocusGroupScope();
 
 type MenuContextValue = {
-  isUsingKeyboardRef: React.RefObject<boolean>;
-  dir: Direction;
   open: boolean;
   onOpenChange(open: boolean): void;
   content: MenuContentElement | null;
   onContentChange(content: MenuContentElement | null): void;
-  onRootClose(): void;
-  modal: boolean;
 };
 
 const [MenuProvider, useMenuContext] = createMenuContext<MenuContextValue>(MENU_NAME);
+
+type MenuRootContextValue = {
+  onClose(): void;
+  isUsingKeyboardRef: React.RefObject<boolean>;
+  dir: Direction;
+  modal: boolean;
+};
+
+const [MenuRootProvider, useMenuRootContext] = createMenuContext<MenuRootContextValue>(MENU_NAME);
 
 interface MenuProps {
   children?: React.ReactNode;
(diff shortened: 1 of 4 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -165,21 +164,24 @@ const MenuSub: React.FC<MenuSubProps> = (props: ScopedProps<MenuSubProps>) => {
...
         open={open}
         onOpenChange={handleOpenChange}
         content={content}
         onContentChange={setContent}
         onRootClose={parentMenuContext.onRootClose}
-        contentId={useId()}
-        trigger={trigger}
-        onTriggerChange={setTrigger}
-        triggerId={useId()}
         modal={false}
```
benoitgrelard, line 174: This isn't used by the submenu right? Do you think it's worth splitting out a general menu context from a root context / sub context?

Not sure if there are other things like this one which only are used by the root?

packages/react/menu/src/Menu.tsx
```diff
@@ -1260,6 +1257,7 @@ const Sub = MenuSub;
 const Anchor = MenuAnchor;
 const SubTrigger = MenuSubTrigger;
 const Content = MenuContent;
+const SubContent = MenuSubContent;
```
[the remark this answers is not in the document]
benoitgrelard: I think we should mirror what we did in the other 2, so put the sub stuff at the bottom. Thanks for not doing it straight away though, it's already pretty tricky to review this PR 😅

benoitgrelard (approved):
I think this is all correct 😅, pretty hard to follow the changes I have to admit though. Just the nature of it.

Left a few minor comments inline.

packages/react/menu/src/Menu.tsx
andy-hook: That’s a good point, in fact I don’t think we even need to recursively drill some of this stuff down anymore, we can just get it from root directly now.

  How about this? https://github.com/radix-ui/primitives/pull/1406/commits/13749a630b565c468cd604c6ddbb6120e60f5d72

benoitgrelard: I'm not sure about this. I feel like it should be the other way around, ie:
- `MenuContext` contains the things that are needed/used by the entire tree
- `MenuRootContext` contains the things that are only needed by the root
- `MenuSubContext` contains the things that are only needed by a sub

So for example `modal` would be only on `MenuRootContext`, but `dir` would be on `MenuContext`.

Does that make sense to you?

andy-hook: Works for me, is this what you meant? https://github.com/radix-ui/primitives/pull/1406/commits/006b3109548227b1a09a38613e4a7f02355a8106

PR #1415, opened 2022-05-23
Add aria-required attribute

SteveEdson:

### DIFF
--- a/.yarn/versions/4a91906d.yml
+++ b/.yarn/versions/4a91906d.yml
@@ -1,5 +1,6 @@
 releases:
-  "@radix-ui/react-tabs": patch
+  "@radix-ui/react-radio-group": patch
 
 declined:
   - primitives
+  - "@radix-ui/react-tabs"
--- a/packages/react/radio-group/src/Radio.tsx
+++ b/packages/react/radio-group/src/Radio.tsx
@@ -57,6 +57,7 @@ const Radio = React.forwardRef<RadioElement, RadioProps>(
         <Primitive.button
           type="button"
           role="radio"
+          aria-required={required}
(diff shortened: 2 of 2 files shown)


benoitgrelard (approved):
Thanks for the contribution!
Would you be able to run `yarn version check --interactive` and set it as a `patch`.
You may also need to run `yarn version check --interactive` a second time to set the overall project as `decline`.
Then you can commit the version file which should clear the CI checks.

🙏

SteveEdson:
@benoitgrelard no problem, updated. Thanks

.yarn/versions/4a91906d.yml
```diff
@@ -1,5 +1,6 @@
 releases:
-  "@radix-ui/react-tabs": patch
+  "@radix-ui/react-radio-group": patch
 
 declined:
   - primitives
+  - "@radix-ui/react-tabs"
```
benoitgrelard, line 6: I'm confused why `tabs` is here? Did it show up in the list to set versions when you ran the command? 🤔

benoitgrelard: On closer inspection, it looks like you have modified an existing version file. Do you know what happened?

SteveEdson: Not sure to be honest! I just ran the version command and it came up, so I set it to decline

benoitgrelard: hmmm, that's really odd, we'll take a look.

benoitgrelard:
Hey I raised a separate PR to fix the version issue, but thanks for your contribution!
You'll see in the description I did change one thing.
I'll close your PR in favour of #1422 now.

PR #1416, opened 2022-05-23
[Primitive] Improve debugging

benoitgrelard:
> **Note**: easier to review without whitespace

I was debugging some stuff in `Select` in react developer tools and noticed a lot of `Anonymous` components in the tree.
It was mostly all of the `Primitive.x` nodes, so I've added a display name to them to reflect it, makes it much more obvious what you're looking at.

I've noticed a few other mismatch related to display names that I might try to look at separately but this was the biggest one.


andy-hook (approved):
This is waaay better now 🙏

PR #1421, opened 2022-05-25
[Select] Fix mismatch with hidden native select

### LINKED ISSUE #1223: [Select] Underlying native `select` has wrong default value

benoitgrelard:
Closes #1223

> **Note**: easier to review without whitespace

### DIFF
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -335,9 +335,9 @@ const SelectContent = React.forwardRef<SelectContentElement, SelectContentProps>
         }, [])}
         onNativeOptionRemove={React.useCallback((option) => {
           setNativeOptionsSet((prev) => {
-            const itemsSet = new Set(prev);
-            itemsSet.delete(option);
-            return itemsSet;
+            const optionsSet = new Set(prev);
+            optionsSet.delete(option);
+            return optionsSet;
           });
         }, [])}
       >
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -293,11 +293,11 @@ SelectIcon.displayName = ICON_NAME;
 
 const CONTENT_NAME = 'SelectContent';
 
-type ReactOption = React.ReactElement<React.ComponentProps<'option'>>;
+type NativeOption = React.ReactElement<React.ComponentProps<'option'>>;
 
 type SelectNativeOptionsContextValue = {
-  onNativeOptionAdd(option: ReactOption): void;
-  onNativeOptionRemove(option: ReactOption): void;
+  onNativeOptionAdd(option: NativeOption): void;
+  onNativeOptionRemove(option: NativeOption): void;
 };
 const [SelectNativeOptionsProvider, useSelectNativeOptionsContext] =
   createSelectContext<SelectNativeOptionsContextValue>(CONTENT_NAME);
@@ -311,9 +311,14 @@ const SelectContent = React.forwardRef<SelectContentElement, SelectContentProps>
     const [fragment, setFragment] = React.useState<DocumentFragment>();
     // We set this to true by default so that events bubble to forms without JS (SSR)
     const isFormControl = context.trigger ? Boolean(context.trigger.closest('form')) : true;
-    const [nativeOptionsSet, setNativeOptionsSet] = React.useState(new Set<ReactOption>());
+    const [nativeOptionsSet, setNativeOptionsSet] = React.useState(new Set<NativeOption>());
 
-    const nativeSelectKey = Array.from(nativeOptionsSet.values())
+    // The native `select` only associates the correct default value if the corresponding
+    // `option` is rendered as a child **at the same time** as itself.
+    // Because it might take a few renders for our items to gather the information to build
+    // the native `option`(s), we generate a key on the `select` to make sure React re-builds it
+    // each time the options change.
+    const nativeSelectKey = Array.from(nativeOptionsSet)
       .map((option) => option.props.value)
       .join(';');
 
@@ -347,7 +352,7 @@ const SelectContent = React.forwardRef<SelectContentElement, SelectContentProps>
             // enable form autofill
             onChange={(event) => context.onValueChange(event.target.value)}
           >
-            {Array.from(nativeOptionsSet.values())}
+            {Array.from(nativeOptionsSet)}
           </BubbleSelect>
         ) : null}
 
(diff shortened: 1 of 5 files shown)


packages/react/select/src/Select.tsx
```diff
@@ -1010,11 +1044,31 @@ const SelectItemText = React.forwardRef<SelectItemTextElement, SelectItemTextPro
...
+
+    const textContent = itemTextNode?.textContent;
+    const nativeOption = React.useMemo(
+      () => (
+        <option key={itemContext.value} value={itemContext.value} disabled={itemContext.disabled}>
+          {textContent}
+        </option>
+      ),
+      [itemContext.disabled, itemContext.value, textContent]
     );
```
benoitgrelard, line 1064: I am building the option element directly here, this has 2 advantages in my opinion:
- it makes it more clear/explicit what this is used for, seeing the native option collocated here
- it means I don't need to invent a new data structure like `{ value, label: textContent, disabled }` just to pass it up and interpolate it higher up.

packages/react/select/src/Select.tsx
```diff
@@ -306,34 +293,83 @@ SelectIcon.displayName = ICON_NAME;
...
   (props: ScopedProps<SelectContentProps>, forwardedRef) => {
     const context = useSelectContext(CONTENT_NAME, props.__scopeSelect);
     const [fragment, setFragment] = React.useState<DocumentFragment>();
+    // We set this to true by default so that events bubble to forms without JS (SSR)
+    const isFormControl = context.trigger ? Boolean(context.trigger.closest('form')) : true;
+    const [nativeOptionsSet, setNativeOptionsSet] = React.useState(new Set<ReactOption>());
+
+    const nativeSelectKey = Array.from(nativeOptionsSet.values())
+      .map((option) => option.props.value)
+      .join(';');
```
andy-hook, line 318: We should try to explain why we need to do this, something like this? (probably bad first attempt):

```
In order for the native select value to update correctly, we need to track and re-render the entire element whenever it's options are changed
```

packages/react/select/src/Select.tsx
```diff
@@ -306,34 +293,83 @@ SelectIcon.displayName = ICON_NAME;
...
+            key={nativeSelectKey}
+            aria-hidden
+            tabIndex={-1}
+            name={context.name}
+            autoComplete={context.autoComplete}
+            value={context.value}
+            // enable form autofill
+            onChange={(event) => context.onValueChange(event.target.value)}
+          >
+            {Array.from(nativeOptionsSet.values())}
```
andy-hook, line 350: Just curious, is it required to explicitly access `values()`? or is `Array.from(nativeOptionsSet)`  equivalent?

packages/react/select/src/Select.tsx
```diff
@@ -1010,11 +1044,31 @@ const SelectItemText = React.forwardRef<SelectItemTextElement, SelectItemTextPro
     const context = useSelectContext(ITEM_TEXT_NAME, __scopeSelect);
     const contentContext = useSelectContentContext(ITEM_TEXT_NAME, __scopeSelect);
     const itemContext = useSelectItemContext(ITEM_TEXT_NAME, __scopeSelect);
-    const ref = React.useRef<SelectItemTextElement | null>(null);
-    const composedRefs = useComposedRefs(forwardedRef, ref, itemContext.onItemTextChange, (node) =>
-      contentContext.itemTextRefCallback?.(node, itemContext.value, itemContext.disabled)
+    const itemsContext = useSelectNativeOptionsContext(ITEM_TEXT_NAME, __scopeSelect);
```
andy-hook, line 1047:
```suggestion
    const nativeOptionsContext = useSelectNativeOptionsContext(ITEM_TEXT_NAME, __scopeSelect);
```

Easier to distinguish perhaps?

ssr-testing/pages/select.tsx
```diff
@@ -29,30 +29,32 @@ export default function SelectPage() {
...
-            </Select.Item>
-            <Select.Item value="3">
-              <Select.ItemText>Item 3</Select.ItemText>
-              <Select.ItemIndicator>✔</Select.ItemIndicator>
-            </Select.Item>
-          </Select.Viewport>
-          <Select.ScrollDownButton>▼</Select.ScrollDownButton>
-        </Select.Content>
-      </Select.Root>
+      <form>
```
andy-hook, line 32: Do you think there's value in rendering both cases for quick manual testing?

packages/react/select/src/Select.tsx
```diff
@@ -1010,11 +1044,31 @@ const SelectItemText = React.forwardRef<SelectItemTextElement, SelectItemTextPro
...
+    const [itemTextNode, setItemTextNode] = React.useState<SelectItemTextElement | null>(null);
+    const composedRefs = useComposedRefs(
+      forwardedRef,
+      (node) => setItemTextNode(node),
+      itemContext.onItemTextChange,
+      (node) => contentContext.itemTextRefCallback?.(node, itemContext.value, itemContext.disabled)
+    );
+
+    const textContent = itemTextNode?.textContent;
```
andy-hook, line 1056: 👍

packages/react/select/src/Select.tsx
andy-hook: haha yeah.. my initial gut reaction was to ask why we don't render closer, I think I'm just conditioned by typically seeing these built closer to the render. But:

> it means I don't need to invent a new data structure like { value, label: textContent, disabled } just to pass it up and interpolate it higher up.

Actually makes much more sense, it's easier to follow when you realise the alternative is just going around the houses for no great reason.

andy-hook (commented):
I like this, seems clear enough to me given our shared understanding of the problem. The comments are mostly nits, I think we should add a bit more context for the future with a comment or two though 🙏

packages/react/select/src/Select.tsx
benoitgrelard: Oh good point, now that it's a set it's probably equivalent, I'll check.

ssr-testing/pages/select.tsx
andy-hook: > I just thought we should have one with and one without a form

Ah! that's exactly what I was asking for before I took a second to expand the diff and saw that's what's been done here 😅 

All good!

packages/react/select/src/Select.tsx
```diff
@@ -306,34 +293,88 @@ SelectIcon.displayName = ICON_NAME;
...
+      <SelectNativeOptionsProvider
+        scope={props.__scopeSelect}
+        onNativeOptionAdd={React.useCallback((option) => {
+          setNativeOptionsSet((prev) => new Set(prev).add(option));
+        }, [])}
+        onNativeOptionRemove={React.useCallback((option) => {
+          setNativeOptionsSet((prev) => {
+            const itemsSet = new Set(prev);
+            itemsSet.delete(option);
+            return itemsSet;
```
andy-hook, line 340:
```suggestion
            const optionsSet = new Set(prev);
            optionsSet.delete(option);
            return optionsSet;
```

andy-hook (approved):

PR #1422, opened 2022-05-25
[RadioGroup] Add `aria-required`

benoitgrelard:
Continuation of #1415 which had issues with versions.

The attribute was moved to the radiogroup rather that each radio though as on the radio eslint was showing this:
https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/112261cbc84f5b7d74de9b427b529a10b41faece/docs/rules/role-supports-aria-props.md

### DIFF
--- a/.yarn/versions/8ccec28f.yml
+++ b/.yarn/versions/8ccec28f.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-radio-group": patch
+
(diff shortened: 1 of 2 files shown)


andy-hook (approved):

andy-hook:
@benoitgrelard Is this worth pulling into the milestone and documenting in the release?

Issue #1420, opened 2022-05-24
[DropdownMenu] Escape key exits fullscreen on Firefox & Safari


cpmsmith:
### Current Behavior

* Enter fullscreen mode
* Open dropdown menu
* Press escape
* Menu closes and browser exits fullscreen mode

### Expected behavior

Menu closes, browser remains in fullscreen mode

### Reproducible example

Reproducible on the [basic demo](https://www.radix-ui.com/docs/primitives/components/dropdown-menu) on Radix's site, or [this CodeSandbox](https://codesandbox.io/s/confident-goldstine-ctqfd4?file=/src/App.tsx)

### Suggested solution

Presumably, `preventDefault` isn't being called somewhere when it should be.

### Additional context

This works in Chrome, and is definitely possible—it works correctly in other libraries, e.g. [React Spectrum](https://react-spectrum.adobe.com/react-spectrum/Menu.html). Also, Dropdown Menu is the first component on which I noticed this, but it affects many components: Alert Dialog, Context Menu, Dialog, Navigation Menu, Popover, Select, and Tooltip. I'm guessing that means the issue actually lies in DismissableLayer.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | `react-dropdown-menu` | 0.1.6 |
| React            |      | 17.0.1 |
| Browser          | Firefox | 100.0.2 |
|  | Safari | 15.4 |
| Node             |      | ? |
| npm/yarn         |         | ? |
| Operating System | macOS | 12.3.1 |

<small>Note: I'm not sure what version of npm/Node CodeSandbox is using, but I don't think it's relevant. </small>

benoitgrelard:
Hi @cpmsmith,

Thanks for reporting this. This is interesting!
The reason is works on Chrome is because `Escape` isn't the shortcut there so that makes sense.

We'll take a look!

✌️

PR #1423, opened 2022-05-25
[DismissableLayer] Don't exit fullscreen when esc-ing out of a layer

cpmsmith:
### Description

Basically just call `preventDefault`! I’m not sure about the semantics of `onDismiss` being undefined, but I _think_ if it’s undefined, we shouldn’t be preventing the default—I’m open to input on that.

Fixes radix-ui/primitives#1420

### DIFF
--- a/.yarn/versions/aa7247ea.yml
+++ b/.yarn/versions/aa7247ea.yml
@@ -1,0 +1,16 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
+  "@radix-ui/react-dismissable-layer": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-hover-card": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-navigation-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-select": patch
+  "@radix-ui/react-toast": patch
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


.yarn/versions/aa7247ea.yml
```diff
@@ -0,0 +1,16 @@
...
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-navigation-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-select": patch
+  "@radix-ui/react-toast": patch
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
+  - "@radix-ui/react-hover-card"
```
benoitgrelard, line 16: Looks like a slip-up here, but `HoverCard` should be part of the version change too as it depends on `DismissableLayer`.

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -104,7 +104,10 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
       const isHighestLayer = index === context.layers.size - 1;
       if (!isHighestLayer) return;
       onEscapeKeyDown?.(event);
-      if (!event.defaultPrevented) onDismiss?.();
+      if (!event.defaultPrevented && onDismiss) {
```
benoitgrelard, line 107: 👍

benoitgrelard (approved):
Thanks for your contribution!
It's all good, just needs the version tweak and then I'll merge 👍

.yarn/versions/aa7247ea.yml
cpmsmith: I didn't bump hover-card because it's not possible to open a hover card with the keyboard, but I suppose it is possible to close them. I'll update that now.

benoitgrelard: Thanks! Yeah it has the same close on escape behaviour.

Issue #1386, opened 2022-05-10
[Portal] Hydration failure in SSR when initially rendered


andy-hook:
This manifests itself in `Dialog` and `AlertDialog` when using `Portal` with `DefaultOpen` in SSR env.

https://codesandbox.io/s/friendly-morning-3ny0r5?file=/pages/index.tsx

benoitgrelard:
It seems the original portal had some code to bail out of ssr.
Do we need similar for the new one?

andy-hook:
> It seems the original portal had some code to bail out of ssr.

Do you mean [this](https://github.com/radix-ui/primitives/commit/3d7300931d46c9306b7ce5bf47601d85e3249842)?

> Do we need similar for the new one?

I think this is slightly different as hydration expects matching server / client. If we want to continue injecting into `body` by default we’ll probably have to add another render and do it as another pass on client only.

laozhu:
The same issue here when working with next.js, any plan to fix this? Now my workaround is set the `open` to `false`, then set it `true` in `useEffect` when page has mounted.

andy-hook:
Another solution is to pass a wrapper node to [`container`](https://www.radix-ui.com/docs/primitives/components/dialog#portal)

gabrielmlinassi:
I'd like to know if this hydration error also relates to this issue. I'm conditionally rendering a Dialog or Button based on viewport size. If it's a small window, I'd like to redirect the user to the registration page, otherwise, show a register dialog.
This is a contrived example showing the error:
https://stackblitz.com/edit/nextjs-hahx6j?file=pages/index.js

Error: `Warning: Text content did not match. Server: "redirect to page.." Client: "open me dialog"`

andy-hook:
> I'd like to know if this hydration error also relates to this issue. I'm conditionally rendering a Dialog or Button based on viewport size.

Not directly, no. The server has no concept of screen media size until it hits the client which is why is falls into the `false` block on server and hydrates `true` on desktop and reports a mismatch.

ibnumusyaffa:
> Another solution is to pass a wrapper node to [`container`](https://www.radix-ui.com/docs/primitives/components/dialog#portal)

I've tried this, but popover not showing in the first render, which is open=true

ibnumusyaffa:
simple workaround iam using
```javascript
function Content({ children }) {
  const { open } = usePopover();

  //workaround for radix bug
  const [portalKey, setPortalKey] = useState(0);
[...]

Issue #1086, opened 2022-01-13
[Dialog] No longer computes and assigns a z-index


colepeters:
### Current Behavior

Since 0.1.5, the Dialog primitive no longer computes and assigns a z-index to the component (as it did in 0.1.1).

### Expected behavior

The maximum z-index should be calculated and assigned to the component, as it was before. Or, if this is an intentional change, documentation should be updated to account for this.

### Reproducible example

[See this demo on CodeSandbox](https://codesandbox.io/s/admiring-grass-ws06k?file=/src/App.js)

When opening the Dialog, the H1 element will show as overlaying the Dialog's content.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |  react-dialog       |   0.1.5      |
| React            |      |   17.x      |
| Browser          |   Chrome      |     latest    |
| Assistive tech   |   n/a      |         |
| Node             | n/a     |         |
| npm/yarn         |   npm      |    7.20     |
| Operating System |   macOS      |    11.6     |

benoitgrelard:
Hi @colepeters,

It is a change and the documentation has been updated in the release notes as well as each component pages.
This is because we now expose a new `Dialog.Portal` part so you have full access to the portal and how you want to position/layer things. Which means you now have full access to set your own z-index (if required).

This not only enables uses to fix layering issues they might have had when using these primitives in conjunction with other libs, but also enables some use-cases that weren't possible before like nesting the content inside the overlay to have an outer-scroll.

I'll close the issue for now, but if that isn't satisfactory, feel free to keep discussing here.

👍

colepeters:
Thanks for the info, @benoitgrelard. I suppose the release notes and the component documentation weren't super clear to me that I'd now have to set my own z-index, but maybe that's because I'm coming from a previous version where this was managed by the component. I did see the note about the new Portal component (which I used in my sandbox), but the fact of that component being exposed didn't really make the think that I'd now need to handle the z-index myself. (To be clear, I'm not complaining about having to handle it, this was just unexpected.)

Maybe this is just a case of being biased by a previous version. :)

jjenzz:
We definitely need to get better at documenting the migration path, we're aware and hoping to improve there going forward 🙏

colepeters:
Thanks @jjenzz, good to know! Documentation can definitely be a tough one. For what it's worth, I generally find the Radix docs to be some of the best out there!

PR #1425, opened 2022-05-26
[Popover] Expose new `Portal` part

### LINKED ISSUE #1306: [Popover] Expose `Portal` part like in `Dialog`
Wait on #1075 to be completed first.

benoitgrelard:
Closes #1306

### DIFF
--- a/.yarn/versions/b02f6991.yml
+++ b/.yarn/versions/b02f6991.yml
@@ -1,0 +1,11 @@
+releases:
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-hover-card": patch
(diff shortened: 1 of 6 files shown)


andy-hook (approved):
> ultimately we took an idea from ariakit's playbook which is to copy the zIndex from content onto the wrapper which means users only have to set it once and it just works.

Nice, yeah I think this is the best of the bunch 👍 

Changes look good, my only addition was to add it to the SSR test page and give it a rattle. The issue described in https://github.com/radix-ui/primitives/issues/1386 will of course still apply to all of these updates, but that's a separate issue.

Only other thing is that we should be sure to document the switch to manual z-index management per https://github.com/radix-ui/primitives/issues/1086#issuecomment-1013283785.

PR #1433, opened 2022-05-29
[Dialog] Allow multiple Dialog triggers with focus restoration

DylanVann:
Before this change if multiple dialog triggers were used the one that focus would be restored to would be somewhat random, because it would be whichever set its ref last.

After this change we get the `HTMLButtonElement` in the `Trigger`s `onClick` handler and add it to the context, so that the focus can be restored to the correct trigger element.

I added two extra trigger elements in the `Dialog/Styled` story to test with. If this seems like a good approach I can add another story for this.

### DIFF
--- a/packages/react/dialog/src/Dialog.stories.tsx
+++ b/packages/react/dialog/src/Dialog.stories.tsx
@@ -7,6 +7,8 @@ export default { title: 'Components/Dialog' };
 export const Styled = () => (
   <Dialog.Root>
     <Dialog.Trigger className={triggerClass()}>open</Dialog.Trigger>
+    <Dialog.Trigger className={triggerClass()}>open</Dialog.Trigger>
+    <Dialog.Trigger className={triggerClass()}>open</Dialog.Trigger>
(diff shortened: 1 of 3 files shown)


benoitgrelard:
Hey @DylanVann, I will echo what Andy was talking about too. The general architecture of Radix is centred around component re-use rather than instance re-use. For that reason we do not consider multiple trigger in a single instance as something we should support.

Trying to do so adds unnecessary complexity that we try to avoid.

Additionally, it might not look problematic for `Dialog`, but think about it for something like `Popover` or anything that positions next to a trigger/reference element. Sure, you could try and do it on the fly and say that the trigger interacted with should be the one used as the reference for position, but what happens if someone is controlling the component and opening it programatically or from another place? Where do we align it to?

These are all things we have already thought about / discussed internally.

We really appreciate contributions though in this case I will close the PR as this isn't something we will implement for the reasons stated above.

Hopefully that makes sense and feel free to keep discussing with us here too!

Thanks 🙏

PR #1455, opened 2022-06-08
[ContextMenu] Add `data-state` to `Trigger`

benoitgrelard:
Reviewing the new data-attributes PR on the docs, I realised that we never added `data-state` to `ContextMenu.Trigger` even though we specifically were controlling it in all stories to achieve that style 😀 Not sure why we never did, probably just an omission.


andy-hook (approved):
Documentation driven development 😄

PR #1460, opened 2022-06-09
[Portal] Kill old portal and replace with new one

benoitgrelard:
> **Note**: needs to be rebased once #1459 is merged

---

> **Warning**
> Breaking changes in docs:
> - Breaking changes for `Portal` util: manual z-index management + `containerRef` to `container` prop
> - update docs page for `Portal`

### DIFF
--- a/packages/react/toast/src/Toast.tsx
+++ b/packages/react/toast/src/Toast.tsx
@@ -4,7 +4,7 @@ import { composeEventHandlers } from '@radix-ui/primitive';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { createContextScope } from '@radix-ui/react-context';
 import * as DismissableLayer from '@radix-ui/react-dismissable-layer';
-import { UnstablePortal } from '@radix-ui/react-portal';
+import { Portal } from '@radix-ui/react-portal';
 import { Presence } from '@radix-ui/react-presence';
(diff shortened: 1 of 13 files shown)


jjenzz (approved):
;)

packages/react/toast/src/Toast.tsx
```diff
@@ -4,7 +4,7 @@ import { composeEventHandlers } from '@radix-ui/primitive';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { createContextScope } from '@radix-ui/react-context';
 import * as DismissableLayer from '@radix-ui/react-dismissable-layer';
-import { UnstablePortal } from '@radix-ui/react-portal';
+import { Portal } from '@radix-ui/react-portal';
```
andy-hook, line 7: Is there a reason we alias in some places but not others?

benoitgrelard: Yes, because a lot of the new components now expose their own `Portal` part so it would clash.

andy-hook: Was just about to update my comment, should we be consistent? does it matter?

benoitgrelard: I don't think it matter much, and generally I think we only alias when the need is there.

andy-hook (approved):

Issue #1211, opened 2022-03-01
[Toast] Toast doesn't play nice with Framer Motion


JuanM04:
### Current Behavior

When using the new Toast with Framer Motion, it does some wierd animation. I suppose it's because the Toast element appears in the document root for a few seconds and then it's moved inside the `<Toast.Viewport />`.

https://user-images.githubusercontent.com/16712703/156083401-bcb0ca14-c2b1-41db-9e85-f543a642dcb8.mp4

### Reproducible example

<details>
  <summary>This is my exact setup. It's a global zustand store that adds and removes notifications. I had it working with divs and I tried to migrate it to Radix</summary>

```tsx
import {
  CheckCircleIcon,
  ExclamationCircleIcon,
  InformationCircleIcon,
  XCircleIcon,
} from '@heroicons/react/outline'
import { XIcon } from '@heroicons/react/solid'
import * as Toast from '@radix-ui/react-toast'
import clsx from 'clsx'
import { AnimatePresence, motion } from 'framer-motion'
import { nanoid } from 'nanoid'
import { Except, Merge } from 'type-fest'
import createStore from 'zustand'

import { Spinner } from '@/components/Spinner'

import { NormalizedError } from './normalizedError'

export interface Notification {
  id: string
  type: 'loading' | 'success' | 'warning' | 'error' | 'info'
  title: string
  body?: string
  timeout: number
  preventClose?: boolean
}

interface NotifyStore {
  notifications: Notification[]
  notify: (data: Merge<Except<Notification, 'id'>, { timeout?: number | null }>) => string
  closeNotification: (id: string) => void
}

const useStore = createStore<NotifyStore>(set => ({
  notifications: [],
  notify: data => {
    const id = nanoid()

    set(state => ({
      notifications: [
        {
          ...data,
          id,
          timeout:
            typeof data.timeout === 'number'
              ? data.timeout
              : data.timeout === null
              ? Infinity
              : data.type === 'loading'
              ? Infinity
              : 5000,
          preventClose: data.preventClose ?? data.type === 'loading',
        },
        ...state.notifications,
      ],
    }))

    return id
  },
  closeNotification: (id: string) =>
    set(({ notifications }) => {
      const i = notifications.findIndex(item => item.id === id)
      if (i === -1) return { notifications }

      return {
        notifications: [...notifications.slice(0, i), ...notifications.slice(i + 1)],
      }
    }),
}))

export const notify = (...args: Parameters<NotifyStore['notify']>) =>
[...]

PR #1468, opened 2022-06-13
[Toast] Fix layout animation in Framer Motion

andy-hook:
Fixes https://github.com/radix-ui/primitives/issues/1211

> The Announce implementation does look to be the problem, motion.xx is passed as child when composing with asChild so this along with it's id and associated props get rendered in both places, seems that this causes framer-motion to try to interpolate between the two layout positions.

@jjenzz I tested in NVDA and VoiceOver, everything seems to function per the previous implementation but without the animation issue. Let me know if you spot anything that I'm not aware of though 🙏 

https://user-images.githubusercontent.com/11708259/173424282-53d4b07f-aa49-4742-aa6d-0d81f0c4927c.mp4


packages/react/toast/src/Toast.tsx
```diff
@@ -512,7 +519,10 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
                     });
                   }
                 })}
-              />
+              >
+                <VisuallyHidden>{renderLabel && context.label}</VisuallyHidden>
```
andy-hook, line 523: We now trigger the announce in NVDA by updating this string in next tick, this means we can avoid rendering `children` in two places and solves our issue.

benoitgrelard (approved):
Pending @jjenzz's feedback, but this looks all good to me! 🎉

jjenzz:
sorry that i've only just tested this but the toast stuff is a bit broken now 😬 

the separate `ToastAnnounce` was there so that SRs were not read aloud "Close" or "Try again" (text from close and action buttons) because they cannot easily goto those buttons within the time constraints.

instead, the `ToastAnnounce` would read out the `altText` which would tell them how to achieve the "action" somewhere else in the application... so if the button was "try again", with the `altText` the SR might hear "There was an error. Resubmit the contact form to try again" instead of "There was an error Close Try Again".

or, if the action read "Update" for a software update, with the `altText` it might read aloud "We've just released version 2. Goto your account settings to upgrade" to guide the SR user where they need to go for the upgrade instead of "We've just released version 2 Close Upgrade".

from what I can see, the `altText` is no longer used.

Issue #1322, opened 2022-04-12
[Toast] Re-ordering DOM is making JS animation difficult


jjenzz:
## Bug report

I'm trying to use `framer-motion` to animate the layout of toasts as they enter/leave but `framer-motion` is struggling to keep track of the elements. My guess is that Radix is disconnecting the DOM from the React tree by moving the toasts via the `MutationObserver` which causes issues for libs relying on React tree (e.g. `key`).

Here is an example of the animation I am trying to achieve: https://codesandbox.io/s/framer-motion-2-animating-shared-layouts-forked-w4ezvn?file=/src/App.js. Notice how previously opened toasts transition upwards as new toasts are added.

Here is what happens with Radix Toast: https://codesandbox.io/s/framer-motion-2-animating-shared-layouts-forked-n81u8m?file=/src/App.js. Previously opened toasts snap upwards when new toasts are added.

### Suggested solution

I'm thinking the `MutationObserver` probably isn't a good idea because it is changing render order outside of React-land. Instead, perhaps toast could manually manage focus to fix the tab order issue that this `MutationObserver` was solving? If focus enters the region, it can focus the last `li` first, and hijack tab keypress to move backwards through the items (similar to focus scope and NavigationMenu solutions).

benoitgrelard:
@jjenzz just so I know what we are talking about, I'm familiar with how we hijack tab keypress in `FocusScope` but what is the thing you mention about `NavigationMenu`?

andy-hook:
In `NavigationMenu` we route focus by manually [controlling it within the content](https://github.com/radix-ui/primitives/blob/23261ba58c2696406bda6ac0f3937237d8d9ad96/packages/react/navigation-menu/src/NavigationMenu.tsx#L821-L842) and moving it a proxy element when required. The Proxy then [determines the entrance / exit direction by the origin](https://github.com/radix-ui/primitives/blob/23261ba58c2696406bda6ac0f3937237d8d9ad96/packages/react/navigation-menu/src/NavigationMenu.tsx#L461-L470).

Just so I'm understanding the original issue, is the problem that we need to be able to tab through toast items from most recent the least?

jjenzz:
Would there be any objection to me removing this `MutationObserver` for now (so tab order wouldn't be right but not the end of the world) and then a separate PR can be raised to solve tab order?

benoitgrelard:
Similar to #1211

PR #1469, opened 2022-06-13
[Toast] Manually reverse tab order

andy-hook:
fixes https://github.com/radix-ui/primitives/issues/1322
builds on https://github.com/radix-ui/primitives/pull/1468

> Most recent should be tabbed to first and unfortunately, most recent will be at the bottom of the viewport DOM.

As described in https://github.com/radix-ui/primitives/issues/1322 – We use portals to move `Toasts` into the `Viewport`, out of the box this causes issues with keyboard accessibility as the append only nature of portals means our source order places most recent toasts last. We were previously using mutation observer to re-order the DOM and solve for it, unfortunately this causes issues with animation libraries which depend on a stable DOM order.

After much investigation there appears to be no fool proof way to re-order DOM without causing issues downstream. Instead this change manages focus programatically so that the tab order reflects newest -> oldest toast without touching the DOM structure.

### DIFF
--- a/cypress/integration/Toast.spec.ts
+++ b/cypress/integration/Toast.spec.ts
@@ -1,0 +1,98 @@
+describe('Toast', () => {
+  beforeEach(() => {
+    cy.visitStory('toast--cypress');
+  });
+
+  function actionIsFocused(identifier: number) {
+    cy.findByText(`Toast ${identifier} action`).should('be.focused');
+  }
+
+  function closeIsFocused(identifier: number) {
+    cy.findByText(`Toast ${identifier} close`).should('be.focused');
+  }
+
+  function toastIsFocused(identifier: number) {
+    cy.get(`[data-cy="toast-${identifier}"]`).should('be.focused');
+  }
+
+  describe('given zero toasts', () => {
+    it('should not interrupt natural tab order in the document', () => {
+      cy.findByText('Focusable before viewport').focus();
+
+      cy.realPress('Tab');
+      cy.findByText('Focusable after viewport').should('be.focused');
+
+      cy.realPress(['Shift', 'Tab']);
+      cy.findByText('Focusable before viewport').should('be.focused');
+    });
+  });
+
+  describe('given multiple toasts', () => {
+    beforeEach(() => {
+      cy.findByText('Add toast').click();
+      cy.findByText('Add toast').click();
+    });
+
+    it('should reverse tab order from most recent to least', () => {
+      cy.findByText('Focusable before viewport').focus();
+
+      cy.realPress('Tab');
+      toastIsFocused(2);
+
+      // Forward tab
+      cy.realPress('Tab');
+      closeIsFocused(2);
+
+      cy.realPress('Tab');
+      actionIsFocused(2);
+
+      cy.realPress('Tab');
+      toastIsFocused(1);
+
+      cy.realPress('Tab');
+      closeIsFocused(1);
+
+      cy.realPress('Tab');
+      actionIsFocused(1);
+
+      // End of viewport
+      cy.realPress('Tab');
+      cy.findByText('Focusable after viewport').should('be.focused');
+
+      // Backwards tab
+      cy.realPress(['Shift', 'Tab']);
+      actionIsFocused(1);
+
+      cy.realPress(['Shift', 'Tab']);
+      closeIsFocused(1);
+
+      cy.realPress(['Shift', 'Tab']);
+      toastIsFocused(1);
+
+      cy.realPress(['Shift', 'Tab']);
+      actionIsFocused(2);
+
+      cy.realPress(['Shift', 'Tab']);
+      closeIsFocused(2);
+
+      cy.realPress(['Shift', 'Tab']);
+      toastIsFocused(2);
+
+      // Start of viewport
+      cy.realPress(['Shift', 'Tab']);
+      cy.findByText('Focusable before viewport').should('be.focused');
+    });
+
+    it('should tab forwards from viewport to latest toast or backwards into the document', () => {
+      // Tab forward from viewport
+      cy.realPress('F8');
+      cy.realPress('Tab');
+      toastIsFocused(2);
+
+      // Tab backward from viewport
+      cy.realPress('F8');
+      cy.realPress(['Shift', 'Tab']);
+      cy.findByText('Focusable before viewport').should('be.focused');
+    });
+  });
+});
--- a/cypress/integration/Toast.spec.ts
+++ b/cypress/integration/Toast.spec.ts
@@ -3,16 +3,12 @@ describe('Toast', () => {
     cy.visitStory('toast--cypress');
   });
 
-  function actionIsFocused(identifier: number) {
-    cy.findByText(`Toast ${identifier} action`).should('be.focused');
-  }
-
-  function closeIsFocused(identifier: number) {
-    cy.findByText(`Toast ${identifier} close`).should('be.focused');
+  function buttonIsFocused(identifier: number) {
+    cy.findByText(`Toast button ${identifier}`).should('be.focused');
   }
 
   function toastIsFocused(identifier: number) {
-    cy.get(`[data-cy="toast-${identifier}"]`).should('be.focused');
+    cy.findByTestId(`toast-${identifier}`).should('be.focused');
   }
 
   describe('given zero toasts', () => {
@@ -41,19 +37,19 @@ describe('Toast', () => {
 
       // Forward tab
       cy.realPress('Tab');
-      closeIsFocused(2);
+      buttonIsFocused(2.1);
 
       cy.realPress('Tab');
-      actionIsFocused(2);
+      buttonIsFocused(2.2);
 
       cy.realPress('Tab');
       toastIsFocused(1);
 
       cy.realPress('Tab');
-      closeIsFocused(1);
+      buttonIsFocused(1.1);
 
       cy.realPress('Tab');
-      actionIsFocused(1);
+      buttonIsFocused(1.2);
 
       // End of viewport
       cy.realPress('Tab');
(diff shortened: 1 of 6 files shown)


packages/react/toast/src/Toast.tsx
```diff
@@ -233,20 +271,67 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
         tabIndex={-1}
         // incase list has size when empty (e.g. padding), we remove pointer events so
         // it doesn't prevent interactions with page elements that it overlays
-        style={{ pointerEvents: context.toastCount > 0 ? undefined : 'none' }}
+        style={{ pointerEvents: hasToasts ? undefined : 'none' }}
       >
+        {hasToasts && (
+          <FocusProxy
+            ref={headFocusProxyRef}
+            onEntranceFocus={() => focusFirst(getSortedCandidates({ direction: 'backwards' }))}
```
andy-hook, line 279: On the way in these proxies focus via the sorted toast order, on the way out they implicitly "pass through" focus to the next valid element in the document, that latter is very helpful as we don't need to query for anything outside of our control. It'll even pass focus to browser controls at the very edges of the document.

andy-hook:
> Looks like there's a regression in the toast order between foreground/background caught by Chromatic.

That’s expected right? we are choosing not to reorder which is why this is a breaking change.

benoitgrelard:
haha of course! 🤦‍♂️ I thought somehow `foreground`/`background` had a particular ordering thing, but they don't really right? It's more about how they get announced/prioritised?

andy-hook:
> I thought somehow `foreground`/`background` had a particular ordering thing, but they don't really right

Not in the DOM sense no, they dictate the nature of the announcement. I assume that chromatic story is intended to validate DOM order though so perhaps it could be updated to make that clearer.

benoitgrelard:
Yeah I think that's what confused me.

cypress/integration/Toast.spec.ts
```diff
@@ -0,0 +1,98 @@
+describe('Toast', () => {
```
benoitgrelard, line 1: ♥️ these!

packages/react/toast/src/Toast.stories.tsx
```diff
@@ -195,6 +195,49 @@ export const Animated = () => {
...
+        {[...Array(count)].map((_, index) => {
+          const identifier = index + 1;
+          return (
+            <Toast.Root key={index} open className={rootClass()} data-cy={`toast-${identifier}`}>
+              <Toast.Title className={titleClass()}>Toast {identifier} title</Toast.Title>
+              <Toast.Description className={descriptionClass()}>
+                Toast {identifier} description
+              </Toast.Description>
+
+              <Toast.Close className={buttonClass()} aria-label="Close">
```
benoitgrelard, line 219: We'd have to update the tests, but reading the tests it was weird to me why the close appeared before the action. Is there a particular reason? It made me question whether we were doing some reordering there too.

packages/react/toast/src/Toast.tsx
```diff
@@ -195,34 +204,63 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
 
+    const getSortedCandidates = React.useCallback(
+      ({ direction }: { direction: 'forwards' | 'backwards' }) => {
+        const items = getItems();
+        const candidateGroups = items.map((item) => {
+          const itemNode = item.ref.current!;
+          const itemGroup = [itemNode, ...getTabbableCandidates(itemNode)];
+          return direction === 'forwards' ? itemGroup.reverse() : itemGroup;
+        });
+        return (direction === 'forwards' ? candidateGroups : candidateGroups.reverse()).flat();
```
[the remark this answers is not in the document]
benoitgrelard: You mean because it depends if you're thinking about the user's tabbing direction, vs the direction focus is actually moving through elements? forwards/backwards in time in terms of toast creation?

Perhaps, if we qualify it a bit more like `tabbingDirection` that would make it clearer?

packages/react/toast/src/Toast.tsx
```diff
@@ -195,34 +204,63 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
       }
     }, [context.isClosePausedRef]);
 
+    const getSortedCandidates = React.useCallback(
```
benoitgrelard, line 207:
```suggestion
    const getSortedTabbableCandidates = React.useCallback(
```

to keep context

packages/react/toast/src/Toast.tsx
```diff
@@ -195,34 +204,63 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
             }
-          });
-        });
-        observer.observe(viewport, { childList: true });
-        return () => observer.disconnect();
+          }
+        };
+
+        // Toasts are not in the viewport React tree so we need to bind DOM events
+        viewport.addEventListener('keydown', handleKeyDown);
```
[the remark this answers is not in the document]
benoitgrelard: I was going to ask, why is it a native event rather than an `onKeyDown` on `Viewport`?
The keydown even would bubble through the portal no?

packages/react/toast/src/Toast.tsx
benoitgrelard, line 215: it might help here to be more specific:

```suggestion
        const toastItems = getItems();
        const tabbableCandidates = toastItems.map((toastItem) => {
          const toastNode = toastItem.ref.current!;
          const toastTabbableCandidates = [toastNode, ...getTabbableCandidates(toastNode)];
          return direction === 'forwards' ? toastTabbableCandidates.reverse() : toastTabbableCandidates;
        });
        return (direction === 'forwards' ? tabbableCandidates : tabbableCandidates.reverse()).flat();
```

benoitgrelard, line 215: Reading this confuses me a bit.
Why do we need to reverse the internal nodes when tabbing forward?
Once we're inside a toast, tabbing forward should be going natural order right?

Similarly, then the whole array is reverse when moving backwards this time, when to me it feels like it should be the opposite.

What am I not following here?

packages/react/toast/src/Toast.tsx
```diff
@@ -195,34 +204,63 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
-            if (!prepended.has(node)) {
-              viewport.prepend(node);
-              prepended.add(node);
+        const handleKeyDown = (event: KeyboardEvent) => {
+          const isMetaKey = event.altKey || event.ctrlKey || event.metaKey;
+          const isTabKey = event.key === 'Tab' && !isMetaKey;
+
+          if (isTabKey) {
+            const focusedElement = document.activeElement;
+            const isMovingBackwards = event.shiftKey;
```
benoitgrelard, line 232:
```suggestion
            const isTabbingBackwards = event.shiftKey;
```

packages/react/toast/src/Toast.tsx
```diff
@@ -195,34 +204,63 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
+
+            // If we're back tabbing after jumping to the viewport then we simply
+            // proxy focus out to the preceding document
+            if (targetIsViewport && isMovingBackwards) {
+              headFocusProxyRef.current?.focus();
+              return;
+            }
+
+            // Otherwise manually reverse the tab order
+            const direction = isMovingBackwards ? 'forwards' : 'backwards';
```
benoitgrelard, line 243: This will now read nicer after chaning the param name in `getSortedCandidates`:

```suggestion
            const tabbingDirection = isTabbingBackwards ? 'forwards' : 'backwards';
```

benoitgrelard, line 243: Ok that's maybe why the orders read backwards in the util for me, this reads pretty weird: `isMovingBackwards ? 'forwards' : 'backwards'`

packages/react/toast/src/Toast.tsx
```diff
@@ -233,20 +271,67 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
 ToastViewport.displayName = VIEWPORT_NAME;
 
+/* -----------------------------------------------------------------------------------------------*/
+
+const FOCUS_PROXY_NAME = 'ToastFocusProxy';
+
+type FocusProxyElement = React.ElementRef<typeof VisuallyHidden>;
+type VisuallyHiddenProps = Radix.ComponentPropsWithoutRef<typeof VisuallyHidden>;
+interface FocusProxyProps extends VisuallyHiddenProps {
+  onEntranceFocus(): void;
```
benoitgrelard, line 309: So `onEntranceFocus` means, if focus comes from outside, rather than from inside?

How about `onFocusFromOutside`?

packages/react/toast/src/Toast.tsx
benoitgrelard: It's not super clear to me how the "way out" is handled?

packages/react/toast/src/Toast.tsx
```diff
@@ -233,20 +271,67 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
+
+const FocusProxy = React.forwardRef<FocusProxyElement, ScopedProps<FocusProxyProps>>(
+  (props, forwardedRef) => {
+    const { __scopeToast, onEntranceFocus, ...proxyProps } = props;
+    const context = useToastProviderContext(FOCUS_PROXY_NAME, __scopeToast);
+
+    return (
+      <VisuallyHidden
+        aria-hidden
+        tabIndex={0}
```
benoitgrelard, line 320: should these have similar styles to focus guards to ensure no involuntary page scrolls (`position: fixed` primarily)?

packages/react/toast/src/Toast.tsx
```diff
@@ -697,6 +791,44 @@ function useNextFrame(callback = () => {}) {
...
+function focusFirst(candidates: HTMLElement[]) {
+  const previouslyFocusedElement = document.activeElement;
+  return candidates.some((candidate) => {
+    // if focus is already where we want to go, we don't want to keep going through the candidates
+    if (candidate === previouslyFocusedElement) return true;
+    candidate.focus();
+    return document.activeElement !== previouslyFocusedElement;
+  });
+}
+
```
benoitgrelard, line 831: Becoming more and more pressing to pull these out in a package soon.

packages/react/toast/src/Toast.tsx
andy-hook: Good spot!

benoitgrelard (approved):

benoitgrelard (approved):

PR #1474, opened 2022-06-17
Fix tree shaking

### LINKED ISSUE #1428: Tree-shaking is broken in "next" release
https://github.com/radix-ui/primitives/discussions/1424

Following the above discussion, I was doing some testing around tree-shaking and it looks like the `@latest` is able to tree-shake a bit, but the `@next` isn't so we must have broken something.

I suspect perhaps it's got something to do to the parcel upgrade.

benoitgrelard:
Fixes #1428


benoitgrelard:
Ok, this isn't as simple to change as I thought…

Mostly because of all the tsconfig mapping, but now also turns out the jest mapping too for tests.
They're all really geared better towards solutions with `index` files.

So we have a few options:
- try and update the jest mapping (`moduleNameMapper`) to make it work, so explicitly listing each package one by one like we have to do in the tsconfig (I've tried that quickly and still had issues)
- name all of the source files `index.ts(x)` so that the index is the source file directly. The thing I don't love about this is the experience in the editor when everything is called index. That said, I tried that quickly too and it looks like vscode now shows the "folder" that isn't in common next to the name, so combined with the fuzzy search, maybe this isn't as much as a big deal as I think?

Any other ideas? Obviously, the best thing would be that the issue gets fixed quickly 😀

Issue #1453, opened 2022-06-08
[Toast] Add onPause / onResume props


nielsvdpol:
## Feature request
Add onPause / onResume props to Toast component so that when a progress bar is used, the animation can be paused.

_Ps. Couldn't find feature request jjenzz talked about in this issue: https://github.com/radix-ui/primitives/issues/1362#issuecomment-1126935808_

la55u:
or something like `pauseOnHover` maybe?

kylemh:
I saw in #1520 this was discusses as something unlikely to be done. I saw that #1477 doesn't quite resolve this. May I open a PR to sort it out?

jjenzz:
@kylemh which part are you requesting to work on? benoit has asked in #1477 that the props be moved onto `Toast.Root` so when that feedback is addressed, you should be able to provide loading ui for your toasts.

kylemh:
I'm hoping to address this issue (pause timers if a user is focused or has cursor hovering over a toast in the provider area).

Thinking about it more now that you've commented, I'm left wondering if global may actually be ideal. If a user is pausing one toast in a provider only showing one toast, global seems overkill; however, if we render multiple toasts, it wouldn't make sense to pause one toast and not the others.

Basically, I'd like to recreate Vercel's toast UI, but with some tweaks: https://vercel.com/design/toast

- Multiple toasts can render and create a stack
- There is a progress UI on the foreground toast to reflect the timer.
- If focus is on any toast or if the cursor is hovering any toast, the timer should be paused.

jjenzz:
> If focus is on any toast or if the cursor is hovering any toast, the timer should be paused. (why this PR is needed)

the toast does already do this. please raise a reproducible GH issue separately if you are experiencing issues there tho.

this ticket is for exposing `onPause` and `onResume` props so that consumers can hook onto this behaviour when it happens in order to render timer UIs.

kylemh:
amazing! 🎉 

> Pauses closing on hover, focus and window blur.

Should've seen that in the docs. Ignore me.

PR #1477, opened 2022-06-18
Add onPause and onResume props to Toast

innub:
Addresses #1453

### DIFF
--- a/.yarn/versions/f4e1423e.yml
+++ b/.yarn/versions/f4e1423e.yml
@@ -1,0 +1,3 @@
+declined:
+  - primitives
+  - "@radix-ui/react-toast"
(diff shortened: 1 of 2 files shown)


innub:
I'm not sure if the version should have been patched or not so I just declined it

benoitgrelard (commented):
It looks like `onPause`/`onResume` are global at the moment?
Given the feature request is to be able to show a timer/progress bar per toast or something like that, it looks like this isn't going to work but rather would need to be per toast right?

I would expect the `onPause`, `onResume` callbacks to be new props on `Toast.Root`.

andy-hook:
Hey @innub thanks a lot for your effort here, I went ahead and made the requested changes via https://github.com/radix-ui/primitives/pull/1669 so will close this out in favour 🙏

PR #1478, opened 2022-06-18
[NavigationMenu] add delay duration skip delay duration props

### LINKED ISSUE #1457: [NavigationMenu] Add `delayDuration` and `skipDelayDuration` props
### Overview

Provide `delayDuration` and `skipDelayDuration` for managing trigger interactions. Mirroring the behaviour from [Tooltip](delayDuration) would be ideal.

Consumers who'd like more control of hover interactions / timing

innub:
Closes  #1457

### DIFF
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -28,6 +28,7 @@ type Direction = 'ltr' | 'rtl';
  * -----------------------------------------------------------------------------------------------*/
 
 const NAVIGATION_MENU_NAME = 'NavigationMenu';
+const DEFAULT_DELAY_DURATION = 700;
 
 const [Collection, useCollection, createCollectionScope] = createCollection<
   NavigationMenuTriggerElement,
@@ -81,6 +82,8 @@ interface NavigationMenuProps
     Omit<PrimitiveNavProps, 'defaultValue'> {
   dir?: Direction;
   orientation?: Orientation;
+  delayDuration?: number;
+  skipDelayDuration?: number;
 }
 
 const NavigationMenu = React.forwardRef<NavigationMenuElement, NavigationMenuProps>(
@@ -91,6 +94,8 @@ const NavigationMenu = React.forwardRef<NavigationMenuElement, NavigationMenuPro
       onValueChange,
       defaultValue,
       orientation = 'horizontal',
+      delayDuration,
+      skipDelayDuration,
       dir,
       ...NavigationMenuProps
     } = props;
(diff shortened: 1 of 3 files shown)


benoitgrelard:
Not sure what's going on here, but it seems the CI checks haven't run and I don't have the usual button to allow them to run.
You might need to force push or add a new commit or something to trigger it again.

✌️

innub:
Here's to hoping you can run the workflows now. Let me know if something is wrong.

I just want to take a moment to say that I love this project and would definitely try to contribute more if my college workload wasn't killer.

:heart:

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -197,6 +210,8 @@ const NavigationMenuProvider: React.FC<NavigationMenuProviderProps> = (
     defaultValue,
     dir,
     orientation,
+    delayDuration = DEFAULT_DELAY_DURATION,
+    skipDelayDuration = 300,
```
benoitgrelard, line 214: I think I'd prefer seeing the defaults being applied at the props level in the components, rather than on the provider.
I know this means multiple places but I think it's better that way so we can always assume a value internally anywhere.

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -208,9 +223,16 @@ const NavigationMenuProvider: React.FC<NavigationMenuProviderProps> = (
     onChange: onValueChange,
     defaultProp: defaultValue,
   });
+  const [shouldSkipDelay, setShouldSkipDelay] = React.useState(false);
```
benoitgrelard, line 226: It would be good if we could follow what was done in `Tooltip` for consistency purpose.
We track rather whether open should be delayed or not with then a separate `handleOpen` and `handleDelayedOpen`.
If we could stay as close as possible to this implementation that would be great 🙏

benoitgrelard (commented):
Hey @innub,

Thanks so much for working on this!
Let me know if the feedback makes sense.

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -28,6 +28,7 @@ type Direction = 'ltr' | 'rtl';
  * -----------------------------------------------------------------------------------------------*/
 
 const NAVIGATION_MENU_NAME = 'NavigationMenu';
+const DEFAULT_DELAY_DURATION = 700;
```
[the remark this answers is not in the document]
benoitgrelard: >  If we're too slow here it will exacerbate the issue of users thinking it's a click interaction and immediately closing the menu.

Yes, that's exactly what I thought too.

andy-hook:
I've addressed this feedback plus other issues I noticed via https://github.com/radix-ui/primitives/pull/1716 so will close this in favour.

Thanks again for your efforts and contributions to Radix @innub , much appreciated.

Issue #1470, opened 2022-06-14
[HoverCard] Text selection improvements


vladmoroz:
## Feature request

Improve text selection usability/accessibility in the Hover Card:
- When selecting text in the Hover Card, don't allow the selection to spill out of the Hover Card bounds.
- Don't close Hover Card on mouse out if there's an active text selection; require an additional pointer down instead.

### Examples in other libraries

VS Code hover card

innub:
Hey @vladmoroz and to anyone else who may be working on this. 

It's quite simple to implement a lock when there is an active text selection, but disallowing the selection to spill out is difficult.
- Using e.preventDefault() on any event other than 'selectstart' does not prevent the text from being highlighted (as such, removing the selection through the Selection API looks wacky because it appears then disappears)
- The mouse location is not included with the 'selectstart' event object and even if its was, the event only fired when the selection starts
- I got desperate so I tried adding an element similar to the one described by [::backdrop](https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop). I don't know why, but if you start highlighting in the hover card and slide your mouse outside the card... you can highlight text which is hidden beneath another element: https://codesandbox.io/s/gallant-cloud-xlkn57?file=/App.js

I looked into [VS Code's hover cards](https://github.com/microsoft/vscode/blob/b3ffbd0333df05cf091539c2d703285f95b7e095/src/vs/base/browser/ui/hover/hoverWidget.ts), but it didn't help much.

PR #1479, opened 2022-06-19
Fix Text Selection of HoverCard

innub:
Addresses: #1470

Implemented: 
- When selecting text in the Hover Card, don't allow the selection to spill out of the Hover Card bounds.
- Don't close Hover Card on mouse out if there's an active text selection; require an additional pointer down instead.

I implemented a selection lock on outside text when the card is open by settings the user-select on document.body to 'none' and on the interior content to 'text'. This method is pretty hacky and doesn't work on force mounts. I'm also not sure if this has some unknown side effects. As such, I made the selection lock an opt-in prop. 

Honestly, the first feature could just be implemented by the developer, but the second one seems like it needs to be implemented internally.

### DIFF
--- a/.yarn/versions/4606939a.yml
+++ b/.yarn/versions/4606939a.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-hover-card": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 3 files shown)


andy-hook:
Thanks for your work and research on this @innub, I’ve used parts of it to inform [this change](https://github.com/radix-ui/primitives/pull/1692) while also expanding some of the handling.  

I'll close this PR in favour and attribute you in the newest change.

Issue #620, opened 2021-04-26
[Tooltip] WCAG 2.1 AA Hover Content Disappearing


okry:
### Current Behavior

Tooltip content disappears when a user tries to hover over it. This violates WCAG 2.1 AA Success Criterion 1.4.13 Content on Hover or Focus.

### Expected behavior

Tooltip content should remain present when a user moves their cursor to the tooltip and when the cursor is hovering over the tooltip content. This should follow the WCAG 1.4.13 success criterion https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus.

### Suggested solution

Two options that I know of:
1. Create an invisible track from the triggering element to the tooltip content that won't trigger `mouseOut` event.
2. Add a timeout for when the tooltip disappears and cancel the timeout when the mouse hovers the tooltip content or the triggering element.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | @radix-ui/react-tooltip | 0.0.16 |
| React            | n/a     |         |
| Browser          | all  |         |
| Assistive tech   | n/a        |         |
| Node             | n/a     |         |
| npm/yarn         | n/a |         |
| Operating System | all |         |

benoitgrelard:
Thanks for bringing this to our attention.

I am curious to understand if this applies to `Tooltip` as it's purely for non-interactive content (ie. the same as `title` attribute, which they state is exempt from this requirement).

I can understand the requirement for other forms of components popping up on hover. For example, we have recently added a `HoverCard` component (not released yet) which does exactly this, you can move from the trigger to the content and it wont' close until you leave the content.

Are you able to share more information about this?

akornmeier:
Quoted from the 1.4.13 success criterion:

> Custom tooltips, sub-menus, and other nonmodal popups that display on hover and focus are examples of additional content covered by this criterion.

benoitgrelard:
Yeah I saw this, I'm just curious about what the difference is between title attribute and custom tooltips. ie. if they behave the same way exactly, why is one exempt and the other isn't?

akornmeier:
That is a good point. I can reach out to some subject matter experts to see if I can get more clarity around this part of the spec.

okry:
[...]

PR #1490, opened 2022-06-23
[Tooltip] Content accessible via pointer

andy-hook:
fixes https://github.com/radix-ui/primitives/issues/620

Of the options available we decided on:
> Create an invisible track from the triggering element to the tooltip content that won't trigger mouseOut event.

https://user-images.githubusercontent.com/11708259/175526607-c926c537-dd37-43b7-a029-90944356a8de.mp4

### DIFF
--- a/packages/react/tooltip/src/Tooltip.stories.tsx
+++ b/packages/react/tooltip/src/Tooltip.stories.tsx
@@ -40,40 +40,38 @@ export const Controlled = () => {
 };
 
 export const CustomDurations = () => (
-  <>
+  <Tooltip.Provider>
     <h1>Delay duration</h1>
     <h2>Default (700ms)</h2>
 
     <div style={{ display: 'flex', gap: 50 }}>
-      <Tooltip.Provider>
-        <Tooltip.Root>
-          <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
-          <Tooltip.Portal>
-            <Tooltip.Content className={contentClass()} sideOffset={5}>
-              Nicely done!
-              <Tooltip.Arrow className={arrowClass()} offset={10} />
-            </Tooltip.Content>
-          </Tooltip.Portal>
-        </Tooltip.Root>
-        <Tooltip.Root>
-          <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
-          <Tooltip.Portal>
-            <Tooltip.Content className={contentClass()} sideOffset={5}>
-              Nicely done!
-              <Tooltip.Arrow className={arrowClass()} offset={10} />
-            </Tooltip.Content>
-          </Tooltip.Portal>
-        </Tooltip.Root>
-        <Tooltip.Root>
-          <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
-          <Tooltip.Portal>
-            <Tooltip.Content className={contentClass()} sideOffset={5}>
-              Nicely done!
-              <Tooltip.Arrow className={arrowClass()} offset={10} />
-            </Tooltip.Content>
-          </Tooltip.Portal>
-        </Tooltip.Root>
-      </Tooltip.Provider>
+      <Tooltip.Root>
+        <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
+        <Tooltip.Portal>
+          <Tooltip.Content className={contentClass()} sideOffset={5}>
+            Nicely done!
+            <Tooltip.Arrow className={arrowClass()} offset={10} />
+          </Tooltip.Content>
+        </Tooltip.Portal>
+      </Tooltip.Root>
+      <Tooltip.Root>
+        <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
+        <Tooltip.Portal>
+          <Tooltip.Content className={contentClass()} sideOffset={5}>
+            Nicely done!
+            <Tooltip.Arrow className={arrowClass()} offset={10} />
+          </Tooltip.Content>
+        </Tooltip.Portal>
+      </Tooltip.Root>
+      <Tooltip.Root>
+        <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
+        <Tooltip.Portal>
+          <Tooltip.Content className={contentClass()} sideOffset={5}>
+            Nicely done!
+            <Tooltip.Arrow className={arrowClass()} offset={10} />
+          </Tooltip.Content>
+        </Tooltip.Portal>
+      </Tooltip.Root>
     </div>
 
     <h2>Custom (0ms = instant open)</h2>
@@ -145,35 +143,33 @@ export const CustomDurations = () => (
     <h1>Skip delay duration</h1>
     <h2>Default (300ms to move from one to another tooltip)</h2>
     <div style={{ display: 'flex', gap: 50 }}>
-      <Tooltip.Provider>
-        <Tooltip.Root>
-          <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
-          <Tooltip.Portal>
-            <Tooltip.Content className={contentClass()} sideOffset={5}>
-              Nicely done!
-              <Tooltip.Arrow className={arrowClass()} offset={10} />
-            </Tooltip.Content>
-          </Tooltip.Portal>
-        </Tooltip.Root>
-        <Tooltip.Root>
-          <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
-          <Tooltip.Portal>
-            <Tooltip.Content className={contentClass()} sideOffset={5}>
-              Nicely done!
-              <Tooltip.Arrow className={arrowClass()} offset={10} />
-            </Tooltip.Content>
-          </Tooltip.Portal>
-        </Tooltip.Root>
-        <Tooltip.Root>
-          <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
-          <Tooltip.Portal>
-            <Tooltip.Content className={contentClass()} sideOffset={5}>
-              Nicely done!
-              <Tooltip.Arrow className={arrowClass()} offset={10} />
-            </Tooltip.Content>
-          </Tooltip.Portal>
-        </Tooltip.Root>
-      </Tooltip.Provider>
+      <Tooltip.Root>
+        <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
+        <Tooltip.Portal>
+          <Tooltip.Content className={contentClass()} sideOffset={5}>
+            Nicely done!
+            <Tooltip.Arrow className={arrowClass()} offset={10} />
+          </Tooltip.Content>
+        </Tooltip.Portal>
+      </Tooltip.Root>
+      <Tooltip.Root>
+        <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
+        <Tooltip.Portal>
+          <Tooltip.Content className={contentClass()} sideOffset={5}>
+            Nicely done!
+            <Tooltip.Arrow className={arrowClass()} offset={10} />
+          </Tooltip.Content>
+        </Tooltip.Portal>
+      </Tooltip.Root>
+      <Tooltip.Root>
+        <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
+        <Tooltip.Portal>
+          <Tooltip.Content className={contentClass()} sideOffset={5}>
+            Nicely done!
+            <Tooltip.Arrow className={arrowClass()} offset={10} />
+          </Tooltip.Content>
+        </Tooltip.Portal>
+      </Tooltip.Root>
     </div>
 
     <h2>Custom (0ms to move from one to another tooltip = never skip)</h2>
@@ -241,7 +237,7 @@ export const CustomDurations = () => (
         </Tooltip.Root>
       </Tooltip.Provider>
     </div>
-  </>
+  </Tooltip.Provider>
 );
 
 export const CustomContent = () => (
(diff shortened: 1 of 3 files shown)


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -35,18 +35,20 @@ type TooltipProviderContextValue = {
   delayDuration: number;
   onOpen(): void;
   onClose(): void;
+  isPointerInTransitRef: React.MutableRefObject<boolean>;
+  closeOnLeave: boolean;
```
andy-hook, line 39: Ideas on a better name for this? I'd prefer it to be opt in to avoid something like `PointerAccessibleContent={false}`

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -248,13 +278,18 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
           aria-describedby={context.open ? context.contentId : undefined}
           data-state={context.stateAttribute}
           {...triggerProps}
-          ref={composedTriggerRef}
-          onPointerEnter={composeEventHandlers(props.onPointerEnter, (event) => {
+          ref={composedRefs}
+          onPointerMove={composeEventHandlers(props.onPointerMove, (event) => {
+            if (wasPointerDown.current) return;
```
andy-hook, line 283: Needed to move `onPointerEnter` -> `onPointerMove` so that a tooltip opens if you're hovering a trigger as the grace area is dismissed. Would be easier to demo the behaviour but this check stops the tooltip from re-opening on move after it's been dismissed with `pointerDown`

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -330,16 +365,132 @@ interface TooltipContentProps extends TooltipContentImplProps {
...
+      trigger.addEventListener('pointerleave', handleCreateGraceArea);
+      content.addEventListener('pointerleave', handleCreateGraceArea);
+
+      return () => {
+        trigger.removeEventListener('pointermove', handleRemoveGraceArea);
+        content.removeEventListener('pointermove', handleRemoveGraceArea);
+        trigger.removeEventListener('pointerleave', handleCreateGraceArea);
+        content.removeEventListener('pointerleave', handleCreateGraceArea);
+      };
+    }
```
andy-hook, line 467: LMK what you think about all of this, I wanted to co-locate as much of the behaviour as I could in single removable place.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -448,6 +599,85 @@ TooltipArrow.displayName = ARROW_NAME;
...
+  };
+}
+
+function isPointerInGraceArea(event: PointerEvent, area: Polygon | null) {
+  if (!area) return false;
+  const cursorPos = { x: event.clientX, y: event.clientY };
+  return isPointInPolygon(cursorPos, area);
+}
+
+function createDebugArea(area: Polygon) {
```
andy-hook, line 657: Left this in just for review, we will strip it out before changes go in.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -330,16 +366,109 @@ interface TooltipContentProps extends TooltipContentImplProps {
...
+  const context = useTooltipContext(TRIGGER_NAME, props.__scopeTooltip);
+  const providerContext = useTooltipProviderContext(TOOLTIP_NAME, props.__scopeTooltip);
+  const ref = React.useRef<TooltipContentHoverableElement>(null);
+  const composedRefs = useComposedRefs(forwardedRef, ref);
+  const [pointerGraceArea, setPointerGraceArea] = React.useState<Polygon | null>(null);
+  const pointerGraceTimerRef = React.useRef(0);
+
+  const { onClose } = context;
+  const content = ref.current;
+  const trigger = context.trigger;
```
benoitgrelard, line 403:
```suggestion
  const { trigger, onClose } = context;
  const content = ref.current;
```

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -330,16 +366,109 @@ interface TooltipContentProps extends TooltipContentImplProps {
...
+
+  const handleRemoveGraceArea = React.useCallback(() => {
+    setPointerGraceArea(null);
+    removeDebugArea();
+    window.clearTimeout(pointerGraceTimerRef.current);
+    providerContext.isPointerInTransitRef.current = false;
```
benoitgrelard, line 409: I think generally we try to avoid mutating refs directly on context, and instead provide an interface with callbacks.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -330,16 +366,109 @@ interface TooltipContentProps extends TooltipContentImplProps {
...
+      const pointerExitPosition = { x: event.clientX, y: event.clientY };
+      const targetPoints = getPointsFromRect(target.getBoundingClientRect());
+      const graceArea = getHull([pointerExitPosition, ...targetPoints]);
+      setPointerGraceArea(graceArea);
+      createDebugArea(graceArea);
+
+      pointerGraceTimerRef.current = window.setTimeout(() => {
+        onClose();
+        handleRemoveGraceArea();
+      }, 500);
```
benoitgrelard, line 423: Why does this solution still need a timer?
Shouldn't it be based purely on whether the pointer is in the grace area? Or is it too disruptive this way?

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -330,16 +366,109 @@ interface TooltipContentProps extends TooltipContentImplProps {
...
+      const handleContentLeave = (event: PointerEvent) => handleCreateGraceArea(event, trigger);
+
+      trigger.addEventListener('pointerleave', handleTriggerLeave);
+      content.addEventListener('pointerleave', handleContentLeave);
+      return () => {
+        trigger.removeEventListener('pointerleave', handleTriggerLeave);
+        content.removeEventListener('pointerleave', handleContentLeave);
+      };
+    }
+  }, [trigger, content, context.trigger, handleCreateGraceArea, handleRemoveGraceArea]);
```
benoitgrelard, line 446:
```suggestion
  }, [trigger, content, handleCreateGraceArea, handleRemoveGraceArea]);
```

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -330,16 +366,109 @@ interface TooltipContentProps extends TooltipContentImplProps {
...
+
+  React.useEffect(() => {
+    if (pointerGraceArea) {
```
benoitgrelard, line 449: Is the timer for that reason, so you can clear the area after some time, and make sure we're not tracking the movements anymore?

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -330,16 +366,109 @@ interface TooltipContentProps extends TooltipContentImplProps {
...
+      const handleTrackPointerGrace = (event: PointerEvent) => {
+        const target = event.target as HTMLElement;
+        const isTargetTrigger = trigger?.contains(target);
+        const isTargetContent = content?.contains(target);
+        const pointerPosition = { x: event.clientX, y: event.clientY };
+        const isPointerOutsideGraceArea = !isPointInPolygon(pointerPosition, pointerGraceArea);
+
+        if (isTargetTrigger || isTargetContent) {
+          handleRemoveGraceArea();
```
benoitgrelard, line 458: So does this mean, if you reach the target clear the grace area?

packages/react/tooltip/src/Tooltip.stories.tsx
```diff
@@ -762,7 +820,7 @@ export const Chromatic = () => (
       <Tooltip.Root>
         <Tooltip.Trigger className={triggerClass()}>open</Tooltip.Trigger>
         <Tooltip.Portal>
-          <Tooltip.Content className={contentClass()} sideOffset={5}>
+          <Tooltip.Content className={contentClass()} side="bottom" sideOffset={5}>
```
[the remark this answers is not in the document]
andy-hook: Good question, I changed these to keep the title visible in chromatic and maintain the passing tests, would we prefer for it to represent the default now?

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -125,44 +133,50 @@ interface TooltipProps {
...
   delayDuration?: number;
+  /**
+   * Whether to close the tooltip immediately when the pointer leaves the trigger.
+   * @defaultValue false
+   */
+  disableHoverableContent?: boolean;
 }
 
 const Tooltip: React.FC<TooltipProps> = (props: ScopedProps<TooltipProps>) => {
+  const providerContext = useTooltipProviderContext(TOOLTIP_NAME, props.__scopeTooltip);
```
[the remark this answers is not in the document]
andy-hook: I was previously using it to assign a default value in the props destructure (as we do with the portal force mount stuff), it can be moved back now though.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -57,6 +54,11 @@ interface TooltipProviderProps {
    * @defaultValue 300
    */
   skipDelayDuration?: number;
+  /**
+   * Whether to close the tooltip immediately when the pointer leaves the trigger.
```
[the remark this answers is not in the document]
andy-hook: Ah yes, good suggestion 👍

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -330,16 +366,109 @@ interface TooltipContentProps extends TooltipContentImplProps {
 const TooltipContent = React.forwardRef<TooltipContentElement, TooltipContentProps>(
   (props: ScopedProps<TooltipContentProps>, forwardedRef) => {
     const portalContext = usePortalContext(CONTENT_NAME, props.__scopeTooltip);
-    const { forceMount = portalContext.forceMount, ...contentProps } = props;
+    const { forceMount = portalContext.forceMount, side = 'top', ...contentProps } = props;
```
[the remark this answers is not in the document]
andy-hook: Yep

packages/react/tooltip/src/Tooltip.stories.tsx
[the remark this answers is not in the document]
benoitgrelard: Oh yeah I see, might make sense to reflect default yeah.

packages/react/tooltip/src/Tooltip.tsx
andy-hook: It's for the same reasons as we have one in `menu`, though perhaps subjective for `Tooltip`. I'll explain on a call.

packages/react/tooltip/src/Tooltip.tsx
andy-hook: Presence of a `pointerGraceArea` determines whether we need these document handlers or not, so it's not specifically related to the timer. It's more that when we remove the area at any time it also cleans up the event handlers without having to call `removeEventListener` in multiple places.

packages/react/tooltip/src/Tooltip.tsx
andy-hook: Yes, though you asking makes me think that they should be combined into a more descriptive name

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -232,8 +261,10 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
     const { __scopeTooltip, ...triggerProps } = props;
     const context = useTooltipContext(TRIGGER_NAME, __scopeTooltip);
     const popperScope = usePopperScope(__scopeTooltip);
-    const composedTriggerRef = useComposedRefs(forwardedRef, context.onTriggerChange);
+    const ref = React.useRef<TooltipTriggerElement>(null);
+    const composedRefs = useComposedRefs(forwardedRef, ref, context.onTriggerChange);
     const isPointerDownRef = React.useRef(false);
+    const wasPointerDown = React.useRef(false);
```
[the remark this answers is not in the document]
andy-hook: There may be a better solution for this but it'll be easier to explain why it's needed on a call.

packages/react/tooltip/src/Tooltip.tsx
andy-hook: I added a single callback that takes an arg, let me know if this appropriate or if we prefer two.

packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -231,9 +259,12 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
   (props: ScopedProps<TooltipTriggerProps>, forwardedRef) => {
     const { __scopeTooltip, ...triggerProps } = props;
     const context = useTooltipContext(TRIGGER_NAME, __scopeTooltip);
+    const providerContext = useTooltipProviderContext(TRIGGER_NAME, __scopeTooltip);
     const popperScope = usePopperScope(__scopeTooltip);
-    const composedTriggerRef = useComposedRefs(forwardedRef, context.onTriggerChange);
+    const ref = React.useRef<TooltipTriggerElement>(null);
+    const composedRefs = useComposedRefs(forwardedRef, ref, context.onTriggerChange);
     const isPointerDownRef = React.useRef(false);
+    const hasTriggerActivated = React.useRef(false);
```
benoitgrelard, line 267: What does this one mean?

When I read the name I thought it was about the trigger activation (click, keydown space/enter) but looking at where the ref is updated it seems to mean something else.

benoitgrelard (approved):
Just one question left (maybe a rename needed) but otherwise it's good to go from me.

Obviously we need to remove the debug area first though 😀

packages/react/tooltip/src/Tooltip.tsx
andy-hook: We needed to use `pointerMove` after all for the polygon bleed change. Because of this we only want to trigger an open on the very first `pointerMove` event, this flag is to describe that the tooltip "has opened" already so don't do it again until we leave and come back.

Perhaps something like `hasPointerMoveOpened`?

benoitgrelard: Yeah that makes more sense.

PR #1516, opened 2022-07-08
[Popper] Change default strategy to use position: fixed

benoitgrelard:


benoitgrelard:
This has sadly uncovered some other holes with regards to new stacking contexts created on fixed element when a transform is involved, so we likely won't be able to go ahead with this.

We will do a spike on using `floating-ui` as a replacement.

martaradziszewska:
Hey @benoitgrelard  @andy-hook,
You are doing a really great job with providing all of those components! Saves lives definitely 😃 

Currently, I'm stuck with this issue and trying to solve this with some ugly workaround that I'm not proud of. Just wondering, Is there any chance you could provide some temporary fix until `floating-ui` spike will be ready? 

Thank you for all your work!
Cheers, Marta

andy-hook:
> Currently, I'm stuck with this issue and trying to solve this with some ugly workaround that I'm not proud of. Just wondering, Is there any chance you could provide some temporary fix until `floating-ui` spike will be ready?

Glad you're finding the work useful! if you have a working fix then I suggest sticking with it for now. We're unlikely to merge a halfway solution in the meantime.

Feel free to post your problem and current solution in one of the linked issues though and we can follow up there.

Issue #1525, opened 2022-07-13
ToolbarPrimitive.ToggleItem.onClick fires twice


wallynm:
Bug report
Current Behavior
Check console in sandbox.

https://codesandbox.io/s/frosty-kapitsa-3b6hbr?file=/App.js

Expected behavior
onClick only fires once

Similar issue related with: https://github.com/radix-ui/primitives/issues/1145

bukinoshita:
I think you need to use the `onValueChange` on the `ToolbarToggleGroup` instead of using the `onClick` on the `ToolbarToggleItem`.

Eg: (and here's a [CodeSandbox](https://codesandbox.io/s/kind-snyder-o4ik44?file=/App.js))

```tsx
<ToolbarToggleGroup
  type="single"
  aria-label="Text formatting"
  onValueChange={(value) => console.log(value)}
>
  <ToolbarToggleItem value="bold" aria-label="Bold">
    <FontBoldIcon />
  </ToolbarToggleItem>
  <ToolbarToggleItem value="italic" aria-label="Italic">
    <FontItalicIcon />
  </ToolbarToggleItem>
  <ToolbarToggleItem value="strikethrough" aria-label="Strike through">
    <StrikethroughIcon />
  </ToolbarToggleItem>
</ToolbarToggleGroup>
```

benoitgrelard:
Yeah that would be the correct way to use the component most of the time, that being said, it's still a bug in case someone was to attach an `onClick`, let's say for analytics reasons for example.

bukinoshita:
@benoitgrelard I have worked on a PR that fixes that which uses a similar approach on https://github.com/radix-ui/primitives/issues/1145, but wasn't sure since it worth opening since it's we should be using the `onValueChange` prop instead. If this is something that we still want to fix, I can open that PR for review

wallynm:
I will finish the patch today, so we can merge

benoitgrelard:
Fixed by #1526

PR #1526, opened 2022-07-13
Prevent ToolbarItem click from firing twice

wallynm:
### Description
Fixes issue with Toggle Item firing Click event twice, bug reported here:
https://github.com/radix-ui/primitives/issues/1525

This PR was based into another related issue reported back in january
https://github.com/radix-ui/primitives/pull/1157

The PR follows same principle, creating an scoped preventing multiple click event to be triggered, also i've added a Unit test to ensure the Click gets called just once.

### DIFF
--- a/.yarn/versions/8da3c08f.yml
+++ b/.yarn/versions/8da3c08f.yml
@@ -1,0 +1,2 @@
+releases:
+  "@radix-ui/react-toolbar": patch
--- a/.yarn/versions/8da3c08f.yml
+++ b/.yarn/versions/8da3c08f.yml
@@ -1,2 +1,5 @@
 releases:
   "@radix-ui/react-toolbar": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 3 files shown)


benoitgrelard:
> I'm seeing that `Version / check (pull_request)` step is failing - but i don't know how to make it pass, i see that i need to generate a release version but in the Contributing Guidelines don't have any on how to achieve this. How can i generate release files?

They do have a bit about it, but it's not super clear:

> You also need to inform Yarn workspaces that a particular package has changed for proper versioning. Run `yarn version check -i` to mark the appropriate type of change for those packages.

You can make the change a patch, you might have to re-run the check too to "decline" the top-level package.

🙏

wallynm:
I don't know why but yarn version is getting stuck.
This command takes so long? Been running for minutes... Maybe something wrong because im using mac m1
Gonna try into another computer

wallynm:
I was having trouble to generate version with m1, might be some issue with the command, it was freezing the cmd
Solved at windows :P

wallynm:
@benoitgrelard could you approve the workflows please? :D

benoitgrelard (approved):

benoitgrelard:
Thanks for your contribution!

PR #1528, opened 2022-07-13
[Toast] Fix regression with `altText` announce

andy-hook:
Addresses https://github.com/radix-ui/primitives/pull/1468#issuecomment-1169243656

This change restores the original announce comp pattern but instead of duplicating children we render them to a document fragment and then extract the bare SR visible text content for the announcement. This avoids the framer motion issue while retaining original a11y intent with `altText`.


packages/react/toast/src/Toast.tsx
```diff
@@ -806,6 +860,26 @@ function useNextFrame(callback = () => {}) {
...
+    const childNodes = Array.from(walker.currentNode.childNodes);
+    const filteredTextNodes = childNodes
+      .filter((node) => node.nodeType === node.TEXT_NODE)
+      .map((node) => node.textContent);
+    textContent.push(...filteredTextNodes);
+  }
+
+  return textContent;
+}
+
```
benoitgrelard, line 882: Can you explain why `node.textContent` from the top-level isn't good enough? Why do we need to walk it? From what I understand `textContent` will already only return the text bits.

packages/react/toast/src/Toast.tsx
```diff
@@ -651,6 +651,60 @@ ToastImpl.propTypes = {
...
+              {context.label} {children}
+            </div>
+          </div>,
+          fragment as any
+        )}
+      <Portal asChild>
+        <VisuallyHidden asChild>
+          <div {...announceProps}>{renderChildren && announceTextContent}</div>
+        </VisuallyHidden>
+      </Portal>
```
benoitgrelard, line 703: I'm confused here a bit why we use `createPortal` and then also our `Portal`, why does the announcement need to be portalled into `body`?

packages/react/toast/src/Toast.tsx
```diff
@@ -651,6 +651,60 @@ ToastImpl.propTypes = {
...
+  // setting the fragment in `useLayoutEffect` as `DocumentFragment` doesn't exist on the server
+  useLayoutEffect(() => {
+    setFragment(new DocumentFragment());
+  }, []);
+
+  // render children in the next frame to ensure toast is announced in NVDA
+  useNextFrame(() => setRenderChildren(true));
+
+  React.useEffect(() => {
+    const timer = window.setTimeout(() => setIsAnnounced(true), 1000);
```
benoitgrelard, line 684: What is this timer for? Seeing hardcoded times like this makes my senses tingle 😀

packages/react/toast/src/Toast.tsx
```diff
@@ -543,10 +548,8 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
               >
                 <Primitive.li
                   role="status"
-                  aria-live={type === 'foreground' ? 'assertive' : 'polite'}
+                  aria-live="off"
```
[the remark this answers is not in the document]
andy-hook: I went back and tested this a lot in VO and NVDA, it didn't seem to make a difference without so I've removed them.

andy-hook:
@benoitgrelard Tidied this up a little, let me know what you think.

packages/react/toast/src/Toast.tsx
```diff
@@ -542,11 +547,6 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
                 })}
               >
                 <Primitive.li
-                  role="status"
-                  aria-live={type === 'foreground' ? 'assertive' : 'polite'}
-                  aria-atomic
```
jjenzz, line 547: this stuff should exist still but with `aria-live="off"` so that SRs read aloud that it is a status list or status when focused on a particular toast.  without this, it communicates it as a group list or group when focused on a particular toast.  

You can test the following in VO to see what i mean:

- current version: https://codesandbox.io/s/dawn-cherry-6iwxpm?file=/src/App.js
- previous version: https://codesandbox.io/s/sad-stallman-53xhxb?file=/src/App.js

andy-hook: Ah thanks, so basically this is to announce as "status" rather than "group"?

I guess that answers this question then! https://github.com/radix-ui/primitives/pull/1528#discussion_r920923915

jjenzz: > Ah thanks, so basically this is to announce as "status" rather than "group"?

yep 👍

PR #1530, opened 2022-07-13
[RadioGroup] Add `disabled` prop to `Root`

YarnSphere:
### Description

Allows setting `disabled` on the root `RadioGroup`, setting all `Radio` items as `disabled`.

The `RovingFocusGroup` has be altered to become unfocusable when it has no focusable items.

### DIFF
--- a/packages/react/radio-group/src/RadioGroup.stories.tsx
+++ b/packages/react/radio-group/src/RadioGroup.stories.tsx
@@ -248,7 +248,7 @@ export const Chromatic = () => {
         </RadioGroup.Item>
       </RadioGroup.Root>
 
-      <h1>Disabled</h1>
+      <h1>Disabled item</h1>
       <RadioGroup.Root className={rootClass()}>
         <RadioGroup.Item className={itemClass()} value="1">
           <RadioGroup.Indicator className={indicatorClass()} />
@@ -257,6 +257,34 @@ export const Chromatic = () => {
           <RadioGroup.Indicator className={indicatorClass()} />
         </RadioGroup.Item>
         <RadioGroup.Item className={itemClass()} value="3">
+          <RadioGroup.Indicator className={indicatorClass()} />
+        </RadioGroup.Item>
+      </RadioGroup.Root>
+
+      <h1>Disabled root</h1>
+      <RadioGroup.Root className={rootClass()} disabled>
+        <RadioGroup.Item className={itemClass()} value="1">
+          <RadioGroup.Indicator className={indicatorClass()} />
+        </RadioGroup.Item>
+        {/* Not possible to set `disabled` back to `false` since it's set on the root (this item
+            should still be disabled). */}
+        <RadioGroup.Item className={itemClass()} value="2" disabled={false}>
+          <RadioGroup.Indicator className={indicatorClass()} />
+        </RadioGroup.Item>
+        <RadioGroup.Item className={itemClass()} value="3">
+          <RadioGroup.Indicator className={indicatorClass()} />
+        </RadioGroup.Item>
+      </RadioGroup.Root>
+
+      <h1>All items disabled</h1>
+      <RadioGroup.Root className={rootClass()}>
+        <RadioGroup.Item className={itemClass()} value="1" disabled>
+          <RadioGroup.Indicator className={indicatorClass()} />
+        </RadioGroup.Item>
+        <RadioGroup.Item className={itemClass()} value="2" disabled>
+          <RadioGroup.Indicator className={indicatorClass()} />
+        </RadioGroup.Item>
+        <RadioGroup.Item className={itemClass()} value="3" disabled>
           <RadioGroup.Indicator className={indicatorClass()} />
         </RadioGroup.Item>
       </RadioGroup.Root>
@@ -330,7 +358,7 @@ export const Chromatic = () => {
         </RadioGroup.Item>
       </RadioGroup.Root>
 
-      <h2>Disabled</h2>
+      <h2>Disabled item</h2>
       <RadioGroup.Root className={rootAttrClass()} defaultValue="3">
         <RadioGroup.Item className={itemAttrClass()} value="1">
           <RadioGroup.Indicator className={indicatorAttrClass()} />
@@ -351,6 +379,32 @@ export const Chromatic = () => {
           <RadioGroup.Indicator className={indicatorAttrClass()} />
         </RadioGroup.Item>
         <RadioGroup.Item className={itemAttrClass()} value="3">
+          <RadioGroup.Indicator className={indicatorAttrClass()} />
+        </RadioGroup.Item>
+      </RadioGroup.Root>
+
+      <h2>Disabled root</h2>
+      <RadioGroup.Root className={rootAttrClass()} defaultValue="3" disabled>
+        <RadioGroup.Item className={itemAttrClass()} value="1">
+          <RadioGroup.Indicator className={indicatorAttrClass()} />
+        </RadioGroup.Item>
+        <RadioGroup.Item className={itemAttrClass()} value="2">
+          <RadioGroup.Indicator className={indicatorAttrClass()} />
+        </RadioGroup.Item>
+        <RadioGroup.Item className={itemAttrClass()} value="3">
+          <RadioGroup.Indicator className={indicatorAttrClass()} />
+        </RadioGroup.Item>
+      </RadioGroup.Root>
+
+      <h2>All items disabled</h2>
+      <RadioGroup.Root className={rootAttrClass()} defaultValue="3">
+        <RadioGroup.Item className={itemAttrClass()} value="1" disabled>
+          <RadioGroup.Indicator className={indicatorAttrClass()} />
+        </RadioGroup.Item>
+        <RadioGroup.Item className={itemAttrClass()} value="2" disabled>
+          <RadioGroup.Indicator className={indicatorAttrClass()} />
+        </RadioGroup.Item>
+        <RadioGroup.Item className={itemAttrClass()} value="3" disabled>
           <RadioGroup.Indicator className={indicatorAttrClass()} />
         </RadioGroup.Item>
       </RadioGroup.Root>
@@ -422,6 +476,8 @@ const styles = {
   backgroundColor: 'rgba(0, 0, 255, 0.3)',
   border: '2px solid blue',
   padding: 10,
+
+  '&[tabindex="0"]': { boxShadow: 'inset 0 0 0 2px yellow' },
 
   '&:disabled': { opacity: 0.5 },
   '&[data-disabled]': { borderStyle: 'dashed' },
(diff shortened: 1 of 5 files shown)


YarnSphere:
My reasoning was 2-fold:
- At least one other library I was testing (Chakra) does this, not that this means much.
- The current behaviour is simply a "shorthand" to manually setting `disabled` on each item, so it made sense to be able to override on individual items if needed.

I agree with your reasoning, though. I'll change the implementation to do that. I'll further set `data-disabled` on the root, then, and I was thinking of adding a `tabIndex={-1}` to the root as well, what are your thought on this?

benoitgrelard:
From Discord:

> @benoit Testing the disabled on the radio-group's root, I noticed something on the current radio-groups that I'm not sure is wanted: why does the group itself have tabIndex="0"? This means that the group will catch focus even if all radios within are disabled.

> In a "vanilla" radio group this doesn't seem to happen. Is this behaviour correct?
just tested a bit more and this might make sense, it seems like in a "vanilla" group if the currently selected radio is disabled, it doesn't let me tab into the whole group because it tries to tab into the selected one only 

Yeah, that's how we handle focusing the correct item when tabbing through the page, focus lands on the group, and the groups throws it back to where it knows it should be.

That does mean that in this case, when the whole group is disabled here, we need to override that `tabIndex`.
I think something like this to only override when we need and otherwise leave it to what is set by `RovingFocusGroup`:

```diff
<RovingFocusGroup.Root
  asChild
  {...rovingFocusGroupScope}
  orientation={orientation}
  dir={direction}
  loop={loop}
+  {...(disabled ? { tabIndex: -1 } : {})}
>
```

YarnSphere:
This still doesn't cover the case where all items are manually set to `disabled`, in which case the group is still focusable. Would you like me to do something about this?

We could use the context to track the number of items vs number of disabled items and set `tabIndex` to `-1` when all items are disabled.

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -31,19 +31,21 @@ const useRadioScope = createRadioScope();
...
   name?: string;
   required: boolean;
+  disabled: boolean;
   value?: string;
   onValueChange(value: string): void;
 };
 
 const [RadioGroupProvider, useRadioGroupContext] =
-  createRadioGroupContet<RadioGroupContextValue>(RADIO_GROUP_NAME);
+  createRadioGroupContext<RadioGroupContextValue>(RADIO_GROUP_NAME);
```
benoitgrelard, line 40: 👍

packages/react/radio-group/src/RadioGroup.tsx
```diff
@@ -125,6 +129,7 @@ const RadioGroupItem = React.forwardRef<RadioGroupItemElement, RadioGroupItemPro
   (props: ScopedProps<RadioGroupItemProps>, forwardedRef) => {
     const { __scopeRadioGroup, disabled, ...itemProps } = props;
     const context = useRadioGroupContext(ITEM_NAME, __scopeRadioGroup);
+    const isDisabled = disabled ?? context.disabled;
```
benoitgrelard, line 132: Yeah so I think this should be the other way around.

benoitgrelard:
> This still doesn't cover the case where all items are manually set to disabled, in which case the group is still focusable. Would you like me to do something about this?

That's a great point!
It looks like perhaps a better place to deal with this would be in `RovingFocusGroup` itself, if all items are "non-focusable" then we'd remove the `tabIndex` requirement from the top.

Then it would just work automatically for radio group here, because all the items would be made disabled/thus non-focusable from `RovingFocusGroup` POV.

YarnSphere:
> It looks like perhaps a better place to deal with this would be in RovingFocusGroup itself, if all items are "non-focusable" then we'd remove the tabIndex requirement from the top.

I've pushed the other changes for now, I'll try to have a go at implementing that.

YarnSphere:
@benoitgrelard, I had a go at it and I don't think it's an easy change.

I think we'd have to actively track the `disabled` state across items, which would likely involve changes in all components using roving focus.

I tried to use `getItems`, check if none are `focusable`, and set `tabIndex={-1}` in such case; however, this needs to run whenever there are changes to the disabled state of individual items, when new items are added to the group, etc. Either that or we'd want to run this _just before_ the user tabs into the group, which I don't think is possible.

Anyway, unless I'm missing something, I suggest we land this as-is, as it's still a strict improvement over the current behaviour (at least with this implementation setting `disabled` on the group does prevent focusing it).

YarnSphere:
> It looks like perhaps a better place to deal with this would be in RovingFocusGroup itself, if all items are "non-focusable" then we'd remove the tabIndex requirement from the top.

> Then it would just work automatically for radio group here, because all the items would be made disabled/thus non-focusable from RovingFocusGroup POV.

The above is implemented.

This PR should be ready for review. The only limitation of the current implementation is that if all `focusable` items are also hidden (`display: none`), then the group will still be focusable. I think that handling this edge case reliably would be quite difficult and probably not worth it.

andy-hook (commented):
Sorry for the delay, I had a couple of additional thoughts

> A limitation of the current implementation is that if all focusable items are also hidden (display: none), then the group will still be focusable. I think that handling this edge case might not be worth it.

  I think this is fine, I can’t imagine it being very common.

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -119,6 +124,8 @@ const RovingFocusGroupImpl = React.forwardRef<
   const handleEntryFocus = useCallbackRef(onEntryFocus);
   const getItems = useCollection(__scopeRovingFocusGroup);
   const isClickFocusRef = React.useRef(false);
+  const [focusableStates, setFocusableStates] = React.useState<FocusableStates>({});
+  const noneFocusable = Object.values(focusableStates).every((state) => !state);
```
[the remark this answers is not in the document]
YarnSphere: If I understand what you're saying correctly, how do we then update the set? Create a new instance every time an item is added/removed? Doesn't React track state by reference, meaning that we can't just do `set.add()`, `set.delete()`?

I can change the implementation to pass `onFocusableItemAdd`/`onFocusableItemRemove` via context, but I'm not sure that a set is a better data structure than an object here due to having to create a new instance every time.

I'll give this a go and see how it goes, maybe I'm misunderstanding how you'd use a set here.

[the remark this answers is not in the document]
YarnSphere: Rewriting this, unless I'm missing something, I don't think there's even a need for tracking ids, so no data structure is needed; a simple counter with the number of focusable items seems enough. Will push an update soon.

packages/react/radio-group/src/RadioGroup.stories.tsx
```diff
@@ -248,6 +248,21 @@ export const Chromatic = () => {
         </RadioGroup.Item>
       </RadioGroup.Root>
 
+      <h1>Disabled root</h1>
```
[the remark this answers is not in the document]
YarnSphere: Done.

packages/react/roving-focus/src/RovingFocusGroup.tsx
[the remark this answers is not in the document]
YarnSphere: Done.

YarnSphere:
Rebased against `main`. This should be ready for review again.

YarnSphere:
I think the release should be marked as `minor` instead of `patch` as it introduces a new API and new behaviour. I'll change the yarn versions file to reflect that.

YarnSphere:
I marked all changes as `minor`, let me know if I should mark some dependencies as `patch` instead.

.yarn/versions/7e2a8b9b.yml
```diff
@@ -0,0 +1,12 @@
+releases:
+  "@radix-ui/react-context-menu": minor
+  "@radix-ui/react-dropdown-menu": minor
+  "@radix-ui/react-menu": minor
+  "@radix-ui/react-radio-group": minor
```
andy-hook, line 5: Let's mark `react-radio-group` as `minor` and the rest can be `patch`

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -140,9 +143,11 @@ const RovingFocusGroupImpl = React.forwardRef<
         [setCurrentTabStopId]
       )}
       onItemShiftTab={React.useCallback(() => setIsTabbingBackOut(true), [])}
+      onFocusableItemAdd={React.useCallback(() => setFocusableItemsCount((n) => n + 1), [])}
+      onFocusableItemRemove={React.useCallback(() => setFocusableItemsCount((n) => n - 1), [])}
```
andy-hook, line 147:
```suggestion
      onFocusableItemAdd={React.useCallback(() => setFocusableItemsCount((prevCount) => prevCount + 1), [])}
      onFocusableItemRemove={React.useCallback(() => setFocusableItemsCount((prevCount) => prevCount - 1), [])}
```

Nit: We typically prefer to explicitly label previous values in setters

packages/react/roving-focus/src/RovingFocusGroup.tsx
```diff
@@ -198,10 +203,20 @@ const RovingFocusGroupItem = React.forwardRef<RovingFocusItemElement, RovingFocu
...
+    );
     const isCurrentTabStop = context.currentTabStopId === id;
     const getItems = useCollection(__scopeRovingFocusGroup);
 
+    React.useEffect(() => {
+      if (focusable) {
+        onFocusableItemAdd();
+        return () => onFocusableItemRemove();
+      }
+    }, [focusable, onFocusableItemAdd, onFocusableItemRemove]);
```
andy-hook, line 218:
```suggestion
    const context = useRovingFocusContext(ITEM_NAME, __scopeRovingFocusGroup);
    const isCurrentTabStop = context.currentTabStopId === id;
    const getItems = useCollection(__scopeRovingFocusGroup);
    
    const { onFocusableItemAdd, onFocusableItemRemove } = context
    
    React.useEffect(() => {
      if (focusable) {
        onFocusableItemAdd();
        return () => onFocusableItemRemove();
      }
    }, [focusable, onFocusableItemAdd, onFocusableItemRemove]);
```

Nit: Small style thing but we prefer to keep `context` whole and destructure methods next to their effects when we need to please the linter

andy-hook (approved):
Good shout on the counter over structure, even simpler still! Just a few nits but otherwise this looks great to me.

Thanks!

packages/react/roving-focus/src/RovingFocusGroup.tsx
YarnSphere: Done.

packages/react/roving-focus/src/RovingFocusGroup.tsx
YarnSphere: Done.

.yarn/versions/7e2a8b9b.yml
YarnSphere: Done.

Issue #1087, opened 2022-01-14
[DropdownMenu]  data-side on Trigger


stereobooster:
### Overview

I want to style trigger depending where dropdown menu opens. `Content` already support this feature, I want the same for `Trigger`. WDYT?

https://www.radix-ui.com/docs/primitives/components/dropdown-menu#collision-aware-animations

benoitgrelard:
Hi @stereobooster, can you show what you have in mind in terms of design?

stereobooster:
<img width="265" alt="Screenshot 2022-01-14 at 15 22 45" src="https://user-images.githubusercontent.com/179534/149530579-426dfb6a-7663-4b74-b89b-a23f4b9a06d9.png">

benoitgrelard:
I see!
It is definitely a menu with actionable items right? You're not trying to build a Select out of the dropdown menu are you?

stereobooster:
Well in this cases it is a menu. But is there a problem building a Select out of it? 

PS: I'm aware that I can style select
<img width="292" alt="Screenshot 2022-01-14 at 15 26 18" src="https://user-images.githubusercontent.com/179534/149531203-cb28621f-c4de-48f9-bf67-86ce565530f3.png">

stereobooster:
> Yes there is, they are completely different accessibility pattern so it would be wrong to build a Select out of a DropdownMenu. 

For a different case, I actually did it:

```tsx
<Content align="end">
        <RadioGroup value={value} onValueChange={onValueChange as (value: string) => void}>
          {(Object.entries(itemsPrepared) as Array<[K, AddaptiveOption]>).map(([v, i]) => (
            <RadioItem key={v} value={v}>
              {i.desktop}
            </RadioItem>
          ))}
        </RadioGroup>
        <Arrow />
      </Content>
```

> We have a Select primitive coming soon btw.

nice

benoitgrelard:
I would recommend using a native select until the primitive is out.

benoitgrelard:
> **Note**: This may be achieved by the proposal of adding an `onSideChange` as suggested in #791

Issue #945, opened 2021-10-30
[Dropdown Menu] Causes screen offset in mobile


luisorbaiceta:
### Current Behavior

When working on small screens, if there is a nested menu that doesn't fit the screen, it offsets the entire layout to fit.

![image](https://user-images.githubusercontent.com/44276180/139530594-6537ec44-2325-41cc-9d80-28702ea92a59.png)

![image](https://user-images.githubusercontent.com/44276180/139530604-57dd762e-96d7-4741-a42d-5bd8f732fbd7.png)

### Expected behavior

Maybe on mobile devices it would be helpful to hover the nested menu over the parent one.

### Your environment

Radix UI Docs website.

jjenzz:
Thank you @luisorbaiceta. 

For anyone following along, this occurs when there is not enough space on either side of the menu so the collision detection preserves its originally intended position. This is definitely something we plan to improve. 

We have another similar issue that may also be fixed by the solution we come up with here https://github.com/radix-ui/primitives/issues/557

benoitgrelard:
We're going to close this one as currently with the new version, in just gets clipped, rather than expanding the body.
More importantly, this is really more down to a product decision, as this component probably shouldn't end up being use on mobile (at least submenus).

Issue #791, opened 2021-07-21
[Popover][Tooltip][HoverCard][Menu] Access to the runtime `data-side` value for JS animation


jjenzz:
If using a JS lib for animation, the runtime value of `data-side` is not available to these libs so they cannot apply origin-aware animations.

This is a [sandbox provided in our discord](https://codesandbox.io/s/prod-snowflake-z4g2z?file=/App.js) where someone has attempted to use the `data-side` attribute as a prop by utilising our `Slot` component. Unfortunately, the attribute value will not be available when the animation component mounts so the animation never plays out.

I'm not sure how we can solve this given that we need the content to render for a frame in order to work out what side it should appear but I'm raising this ticket for us to have a think.

benoitgrelard:
Yeah, that might be quite tricky indeed! We would have to think about this quite a bit, maybe some kind of hook or render prop could help.

jjenzz:
Perhaps an `onSideChange` callback so they can control it and then our current `side` prop would become `defaultSide` which suits what it is actually doing more accurately.

Lexachoc:
Any update on this?

Also, there's `[data-side]` for [Select](https://www.radix-ui.com/primitives/docs/components/select#content). Will Select also be considered to have the callback on side changes?

I saw there's a PR to expose `data-side` and `data-align` but for Popover anchor.

3pleFly:
I ended up using a mutation observer too, but it would be really nice to have this :)

chaance:
No plans to implement at the moment, but we may revisit in the future.

PR #1531, opened 2022-07-15
[Popper] Use `floating-ui`

### LINKED ISSUE #370: [Popper] Improve calculations to account for parent positioning / scroll offsets
This is for cases when people aren't rendering in portals.

See https://github.com/radix-ui/primitives/issues/367 for example.

### LINKED ISSUE #557: [Popover] Better collision prevention
## Feature request

When using `align="start"` or `align="end"` the alignment may flip abruptly when a collision happens. Sometimes this introduces just another collision, which neither uses the available space efficiently, nor keeps the preferred alignment intact.

For example, when using `align="start" side="right"` the desired behaviour would be:
- If `align="start"` results in a collision, try `align="center"`
- If `align="center"` does not resolve the collision, try `align="end"`
- If `align="end"` does not resolve the collision, revert to `align="start"`

Current behaviour:

https://user-images.githubusercontent.com/8441036/111156630-a5ec8400-859e-11eb-906a-3fc953ce4a10.mov

### LINKED ISSUE #740: [Popover] content positioned incorrectly on Safari when combined with Framer Motion scale
## Bug report
When using the Popover component with it's Popover.Content set to `align="end"` and animating its scale with Framer Motion it's positioned incorrectly on the x-axis on Safari. 

### Current Behavior
The popover content is offset to the right.

Screenshot 2021-06-22 at 20 47 58

### Expected behavior
The popover content should be aligned to the trigger. This works in Chrome/FF.

### Reproducible example

https://codesandbox.io/s/radix-framer-bug-myc5h

### Additional context

Very specific bug for my use-case ( opening a right-aligned popover and animating it's scale with framer motion), and not sure if fixable but it works as expected in Chrome/FF

### Your environment

| Radix Package(s) |    @radix-ui/react-popover     |    0.0.18     |
| Browser          |    Safari     |     14.1    |
| Node             | n/a     |    16.0.0     |
| npm/yarn         |    yarn     |     1.22.10    |
| Operating System |    MacOS     |    11.3.1     |

### LINKED ISSUE #781: [Popper] Issue when trigger is position fixed?
### Current Behavior

CleanShot 2021-07-13 at 15 22 59

### LINKED ISSUE #800: [Popover] Off-center arrow position with scale transform in Safari
### Current Behavior

In Safari, using a scale transform to animate `Popover.Content` makes the arrow element go off-center.

Screenshot 2021-07-27 at 1 04 39 PM

P.S. Using code from Radix documentation: https://www.radix-ui.com/docs/primitives/components/popover#origin-aware-animations

### Expected behavior

Correct positioning of the arrow, and content elements.

### Reproducible example

CodeSandbox: https://codesandbox.io/s/radix-popover-animation-2w1p0?file=/src/App.js

### Your environment

| Radix Package(s) | @radix-ui/react-popover | 0.0.19 |
| Stitches             | n/a | 0.2.3 |
| React            | n/a     | 17.0.2 |
| Browser          | Safari | 15.0 (17612.1.22.1.4) |
| Assistive tech   | - | - |
| Node             | n/a     | 16.5.0 |
| npm/yarn         | npm | 7.19.1 |
| Operating System | macOS | 12.0 Beta |

### LINKED ISSUE #1445: Component `Hover Card` does not respect inner containers
### Current Behavior

At the moment component `Hover Card` is working only based on the viewport and there are some edge cases where it stops working if we're trying to use it inside inner scrollable containers.

<details>
<summary>Demo 👀</summary>
<video src="https://user-images.githubusercontent.com/9263565/171867455-365af036-c3c8-4e1a-aff1-75077301f6d5.mov" />
</details>

I found a temporary solution, however it does not solve the issue itself:

1. Add `portalled={false}` to `HoverCard.Content` component.
2. Add `z-index: 0` to `HoverCard.Content` component.
3. Add `background-color` as well as `z-index: 1` to containers above and below.

<details>
<summary>Demo 👀</summary>
<video src="https://user-images.githubusercontent.com/9263565/171867948-bf72793c-5088-47f2-b784-bf2a581bfa42.mov" />
</details>

As you can see it solves the issue, but this approach leads to another bug. We're losing `avoidCollisions` and `collisionTolerance` features. The card itself is being hidden, however it does not avoid collisions anymore and being invisible, the result is bad UX.

### Expected behavior

`HoverCard` component should respect inner container boundaries and try to avoid collisions.

### Suggested solution

I guess there are few solutions and I would be more than happy to contribute and fix it by myself, but I am not sure which solution is the best.

1. Hide `HoverCard.Content` component on scroll. I am not sure about this one, because we should make sure that we're not scrolling for example inside `HoverCard.Content` component. Also I am not sure if it should be working like that by default, maybe we should just simply add additional prop?
2. Add ability to pass the reference of parent container, in this case, `HoverCard.Content` component would respect container boundaries instead of viewport. I'm not sure if this solution can be easily implemented.
3. Your suggestions? 💡

### LINKED ISSUE #1464: [DropdownMenu] DropdownMenu doesn't locate properly.
### Current Behavior
I'm using dropdown menu for my i18n options. However, in big screen the menu doesn't located properly by default.
Screen Shot 2022-06-11 at 12 53 38

### Expected behavior
I think the menu should locate like this by default. (I'm now using tailwind class 'sm:-top-8' to fix this)
Screen Shot 2022-06-11 at 12 55 47

### Suggested solution

Maybe add some css like -top-8?

### Your environment

| Radix Package(s) |    react-dropdown-menu     |   0.1.6      |
| React            | n/a     |     18.1.0    |
| Browser          |     Microsoft Edge |   100.0.1185.29      |
| Assistive tech   |     TailwindCSS    |     3.0.24    |
| npm/yarn         |     npm    |   8.6.0      |
| Operating System |     MacOS Monterey    |     12.3.1    |

benoitgrelard:
Supersedes #1516

Fixes #370 
Fixes #557
Fixes #740
Fixes #781 
Fixes #800
Fixes #1445
Fixes #1464

Others potentially linked to investigate:
#791 — explore this
#945 — may be doable now with use of mainAxis shift, but not sure it's worth it, probably needs a diff pattern
#1087 — linked to #791

---

> **Warning**
Docs updates:
- [ ] 🔥 `Arrow` `offset` prop is gone
- [ ] 🔥 `Content` `collisionTolerance` prop is renamed to `collisionPadding` and accept `number | Padding`
- [ ] ✨ `Content` `arrowPadding` prop is new
- [ ] ✨ `Content` `sticky` prop is new
- [ ] ✨ `Content` `hideWhenDetached` prop is new
- [ ] No need for `Portal` part around submenu contents (in `DropdownMenu` and `ContextMenu`)
- [ ] Update demos accordingly
---

- [ ] Also once PR is merged, go on `main` and delete references to `core/popper` in version files (only 1 I think)


packages/react/popper/src/Popper.tsx
```diff
@@ -105,71 +126,136 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
-      align,
-      alignOffset,
-      shouldAvoidCollisions: avoidCollisions,
-      collisionBoundariesRect,
-      collisionTolerance,
-    });
-    const isPlaced = placedSide !== undefined;
+    const isPlaced = x !== null && y !== null;
+    const [placedSide, placedAlign] = getSideAndAlignFromPlacement(placement);
+    const { arrow: { x: arrowX, y: arrowY } = {} } = middlewareData;
```
andy-hook, line 178:
```suggestion
    const arrowX = middlewareData.arrow?.x;
    const arrowY = middlewareData.arrow?.y;
```

Is this equivalent? I found the nested destructure trickier to follow at first glance.

packages/react/popper/src/Popper.tsx
```diff
@@ -105,71 +126,136 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
+      if (!isRoot) {
+        positionUpdateFns.add(update);
+        return () => {
+          positionUpdateFns.delete(update);
+        };
+      }
+    }, [isRoot, positionUpdateFns, update]);
+
+    // when nested contents are rendered in portals, they are appended out of order
+    // we need to re-compute the positioning once the parent has finally been placed.
```
andy-hook, line 199:
```suggestion
    // when nested contents are rendered in portals, they are appended out of order causing
    // children to be positioned incorrectly if initially open.
    // we need to re-compute the positioning once the parent has finally been placed.
    // https://github.com/floating-ui/floating-ui/issues/1531
```

How'd you feel about being really specific on this one?

packages/react/popper/src/Popper.tsx
```diff
@@ -105,71 +126,136 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
+      <div
+        ref={floating}
+        data-radix-popper-content-wrapper=""
+        style={{
+          position: strategy,
+          left: 0,
+          top: 0,
+          transform: isPlaced
+            ? `translate3d(${Math.round(x)}px, ${Math.round(y)}px, 0)`
+            : 'translate3d(0, -200vh, 0)',
```
andy-hook, line 231: I've forgotten the purpose of the `-200vh`, do we still need this?

packages/react/use-size/src/useSize.tsx
```diff
@@ -30,10 +34,9 @@ function useSize(element: HTMLElement | SVGElement | null) {
           height = borderSize['blockSize'];
         } else {
           // for browsers that don't support `borderBoxSize`
-          // we calculate a rect ourselves to get the correct border box.
```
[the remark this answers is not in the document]
andy-hook: Ah nice 👍  I vaguely remember running into something related to this on the nav menu.

packages/react/popper/src/Popper.tsx
benoitgrelard: This is to make sure it's not overlapping other elements.
See here for more details on its origin: https://github.com/radix-ui/primitives/pull/541

andy-hook (approved):

PR #1551, opened 2022-07-21
[DropdownMenu] fix typo on DropdownMenuProps

MaeWolff:
### Description

just a little typo fix :)


andy-hook:
Thanks for your effort here but for our purposes these are equivalent and we stylistically prefer the former

PR #1137, opened 2022-02-04
[Menubar] Add Menubar primitive

luisorbaiceta:
This PR is intended for the Menubar implementation. Closes #1133

- [x] Discuss API (Following the [WAI-ARIA pattern for Menubar](https://www.w3.org/WAI/ARIA/apg/example-index/menubar/menubar-editor.html#top))
- [ ] Add integration tests
- [ ] Add stories for visual testing
- [ ] On right or left arrow press, if next focusable element is next tab it should open the tab and focus its trigger as described in WAI-ARIA (depends on #1574)

### [Suggested API](https://github.com/radix-ui/primitives/issues/1133#issue-1123075390)

```tsx
<Menubar.Root>
  <Menubar.Menu>
    <Menubar.Trigger />
    <Menubar.Content>
      <Menubar.Item />
    </Menubar.Content>
  </Menubar.Menu>
  <Menubar.Menu>
    <Menubar.Trigger />
    <Menubar.Content>
      <Menubar.Item />
      <Menubar.SubMenu>
        <Menubar.Trigger />
        <Menubar.Content>
          <Menubar.Item />
        </Menubar.Content>
      </Menubar.SubMenu>
    </Menubar.Content>
  </Menubar.Menu>
</Menubar.Root>
```

For `Menus` and `Submenus` the `Dropdown` implementation will be used. Any additional behaviour will be added via its controlled api

PR #1574, opened 2022-07-25
[Menu] Expose onEntryFocus

luisorbaiceta:
This PR exposes the cited event to allow some primitives to opt out of the default `focusFirst` when menu is opened


luisorbaiceta:
This issue is blocking #1137 as well

PR #1579, opened 2022-07-28
[NavigationMenu] Fix menu reopening on mouse move when closed with Esc

### LINKED ISSUE #1529: [NavigationMenu] Menu closed with Esc immediately reopens on mouse move
## Bug report

When you close NavigationMenu with Esc, moving the mouse opens it again, which defeats the purpose of being able to close the menu like this in the first place.

It should stay closed until the next time you hover in; similarly to how it works when you close it via a mouse click and move the mouse.

### Reproducible example

https://www.radix-ui.com/docs/primitives/components/navigation-menu

rortan134:
Fixes #1529 

Makes the menu stay closed after pressing the Esc key

### DIFF
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -400,6 +400,7 @@ const NavigationMenuTrigger = React.forwardRef<
   const triggerId = makeTriggerId(context.baseId, itemContext.value);
   const contentId = makeContentId(context.baseId, itemContext.value);
   const wasClickCloseRef = React.useRef(false);
+  const wasEscapeCloseRef = React.useRef(false);
   const open = itemContext.value === context.value;
 
   return (
@@ -417,11 +418,12 @@ const NavigationMenuTrigger = React.forwardRef<
             ref={composedRefs}
             onPointerEnter={composeEventHandlers(props.onPointerEnter, () => {
               wasClickCloseRef.current = false;
+              wasEscapeCloseRef.current = false;
             })}
             onPointerMove={composeEventHandlers(
               props.onPointerMove,
               whenMouse(() => {
-                if (disabled || wasClickCloseRef.current) return;
+                if (disabled || wasClickCloseRef.current || wasEscapeCloseRef.current) return;
                 context.onItemOver(itemContext.value);
               })
             )}
@@ -441,11 +443,16 @@ const NavigationMenuTrigger = React.forwardRef<
               const entryKey = { horizontal: 'ArrowDown', vertical: verticalEntryKey }[
                 context.orientation
               ];
-
-              if (open && event.key === entryKey) {
-                itemContext.onEntryKeyDown();
-                // Prevent FocusGroupItem from handling the event
-                event.preventDefault();
+              const isEscapeKey = event.key === 'Escape' || event.key === 'Esc';
+              if (open) {
+                if (event.key === entryKey) {
+                  itemContext.onEntryKeyDown();
+                  // Prevent FocusGroupItem from handling the event
+                  event.preventDefault();
+                }
+                if (isEscapeKey) {
+                  wasEscapeCloseRef.current = open;
+                }
               }
             })}
           />
--- a/.yarn/versions/5e841cbb.yml
+++ b/.yarn/versions/5e841cbb.yml
@@ -1,2 +1,5 @@
 releases:
   "@radix-ui/react-navigation-menu": patch
+
+declined:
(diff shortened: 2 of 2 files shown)


packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -441,11 +443,16 @@ const NavigationMenuTrigger = React.forwardRef<
...
-                // Prevent FocusGroupItem from handling the event
-                event.preventDefault();
+              const isEscapeKey = event.key === 'Escape' || event.key === 'Esc';
+              if (open) {
+                if (event.key === entryKey) {
+                  itemContext.onEntryKeyDown();
+                  // Prevent FocusGroupItem from handling the event
+                  event.preventDefault();
+                }
+                if (isEscapeKey) {
```
benoitgrelard, line 453: `if (event.key === 'Escape')` should be sufficient here.

packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -441,11 +443,15 @@ const NavigationMenuTrigger = React.forwardRef<
...
-                itemContext.onEntryKeyDown();
-                // Prevent FocusGroupItem from handling the event
-                event.preventDefault();
+              if (open) {
+                if (event.key === entryKey) {
+                  itemContext.onEntryKeyDown();
+                  // Prevent FocusGroupItem from handling the event
+                  event.preventDefault();
+                }
+                if (event.key === 'Escape') {
```
benoitgrelard, line 452:
```suggestion
                } else if (event.key === 'Escape') {
```

benoitgrelard (commented):
Sorry just one final nit before I merge!

benoitgrelard (commented):
Unfortunately after testing this, I've just realised that this approach won't work.
To clarify it works if the trigger is focused, but not if you're just hovering with your mouse.
We cannot rely on the fact that the trigger is focused.
Instead we are going to have to rely on the `onEscapeKeydown` event handler from `DismissableLayer`.

So the way to go is:

- lift up `wasEscapeCloseRef` definition into `NavigationMenuItem`
- pass it down through context (`NavigationMenuItemContextProvider`)
- set the ref to `true` in `onEscapeKeydown` on `DismissableLayer` in `NavigationMenuContentImpl` instead of in the trigger keydown event

Hopefully that's enough for you to get going, but let me know otherwise and I can take it from here otherwise! 🙏

rortan134:
I made the changes to make a shallower approach, though the root cause of the focus issue seems to be related to the document focus variable in `UseEscapeKeydown` and the fact that Storybook has two documents (One inside an iframe). Quoting from the jQuery Docs:
>[the keydown event] can be attached to any element, but the event is only sent to the element that has the focus. Focusable elements can vary between browsers

Therefore, since Storybook uses an iframe, there are two documents and only one of them can be focused, and if the external one (Sidebar + Header) is focused then the `UseEscapeKeydown` handler won't trigger and the dropdown won't close, which is likely what you saw happening.

This is issue is likely to be exclusive to Storybook though, since regular web pages usually don't have more than one document, and it won't be a problem.
An easy way to test this behavior is adding a console.log(ownerDocument.hasFocus()) inside the useEffect on the useEscapeKeydown hook, you'll see that when you focus the sidebar and proceed to hover on a menu it'll say false and Escape won't close the menu, though if you focus the inner document (iframe) then it'll say true and the menu will work fine. Tell me if you find any issues 😄

benoitgrelard (approved):
> though the root cause of the focus issue seems to be related to the document focus variable in UseEscapeKeydown and the fact that Storybook has two documents
> Therefore, since Storybook uses an iframe, there are two documents and only one of them can be focused, and if the external one (Sidebar + Header) is focused then the UseEscapeKeydown handler won't trigger and the dropdown won't close, which is likely what you saw happening.

I think you are getting your wires crossed a bit. Just to clarify, this isn't what I was referring to in my [earlier comment](https://github.com/radix-ui/primitives/pull/1579#pullrequestreview-1123566633).

To reproduce the issue I was discussing before with the naïve trigger `onKeyDown` approach:
- click somewhere on the page to make sure the document is focused (or open the frame directly itself)
- hover a menu item (don't focus it with keyboard) -> menu opens
- press escape -> menu closes
- mouse mouse (still over the item) -> item reopens

This is because we were marking the prevention in the ref only in the onKeyDown of the trigger, which in this case would never happen because the trigger isn't focus so wouldn't receive a keydown event.

Hence why I requested the changes.
This is all working great now, thanks!

PR #1597, opened 2022-08-03
[Popper] Add support for limiting height to available space

### LINKED ISSUE #1596: Support a Popover that would overflow the container
### Overview

Currently if you have a popover that would overflow the container, it get's clipped off, without a chance to provide a scroll bar instead.

image

FloatingUI provides an easy middleware to handle [this](https://floating-ui.com/docs/size) use case.

Mostly advanced users will benefit from it. Anyone building a reusable set of components (specifically selects) can leverage this functionality.

paul-sachs:
### Description

Add support for limiting height to available space. Fixes #1596

Basically introduces a prop for popper based components that can limit the height of the content based on available space:

image

### DIFF
--- a/packages/react/popper/src/Popper.stories.tsx
+++ b/packages/react/popper/src/Popper.stories.tsx
@@ -334,6 +334,54 @@ export const Chromatic = () => {
     </div>
   );
 };
+
+export const LimitHeightToAvailableSpace = () => {
+  const [open, setOpen] = React.useState(false);
+  return (
+    <Scrollable>
+      <Popper.Root>
+        <Popper.Anchor className={anchorClass()} onClick={() => setOpen(true)}>
+          open
+        </Popper.Anchor>
+
+        {open && (
+          <Popper.Content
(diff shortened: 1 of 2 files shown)


paul-sachs:
Turns out this still has bugs. Going to kick it back to draft.

andy-hook:
Hey @paul-sachs , thanks for your effort here and apologies for the delay. We’ve had quite a few requests for extra flexibility in the positioning, sizing, collision handling and have been discussing internally the best approach for extending these options while providing an intuitive public api.

For that reason we’re likely to address these via a holistic change, but I wanted to provide a follow up and let you know we've seen and are considering the options 🙏

paul-sachs:
@andy-hook any movement on this? Was hoping to close a few bugs in our design system caused by this. Moving to the updated Select will solve some of the issues but we've got some custom popovers that will still suffer from this issue.

benoitgrelard:
Superseded by #1942

PR #1624, opened 2022-08-17
[Menu, DropdownMenu] Allow CheckboxItem state to be indeterminate

### LINKED ISSUE #1581: Allow indeterminate state for DropdownMenu CheckboxItem
### Overview

The Checkbox primitive allows for three states via `[data-state]`: checked, unchecked, and indeterminate. However, the DropdownMenu's CheckboxItem appears to only allow two states: checked and unchecked. I'd love to see parity between these two, as I have a need to create a checkbox with an indeterminate state within a dropdown.

It seems I'll need to fall back to using a Checkbox instead of a CheckboxItem in this case, though I'm not clear yet on what all the ramifications of doing so may be.

N/A

This impacts any user looking for feature parity between the Checkbox and DropdownMenu.CheckboxItem primitives. :)

### Additional context

That's it, pretty simple! Thanks for the consideration!

colepeters:
### Description

Closes #1581 — allows the `Menu.CheckboxItem`/`DropdownMenu.CheckboxItem` to have an indeterminate state (and `aria-checked='mixed'` attribute). Implementation was largely referenced from the core Checkbox primitive, where this was already available.

Some notes:

- I'm extremely late to learning Typescript in my career, so please let me know if there are any cleanups/improvements that could be made in that respect (or any other respect, of course)
- I exported `isIndeterminate`, `getState`, and the `CheckboxState` type from the Checkbox package instead of recreating these for `MenuCheckboxItem` — I'm not sure if this change warrants a `minor` release, but I figured all the affected packages should get the same bump. Let me know if this is incorrect?
- ContextMenu was flagged for an update only because I needed to add type arguments to the `useState` calls in the Checkbox Item story — again, not sure if this should warrant a release or not?

I've tested my changes via updated stories for Menu and DropdownMenu ('Checkbox Items' in both).

Happy to make any changes as necessary!

### DIFF
--- a/.yarn/versions/5ca86df5.yml
+++ b/.yarn/versions/5ca86df5.yml
@@ -1,5 +1,4 @@
 releases:
-  "@radix-ui/react-checkbox": minor
   "@radix-ui/react-context-menu": minor
   "@radix-ui/react-dropdown-menu": minor
   "@radix-ui/react-menu": minor
--- a/.yarn/versions/5ca86df5.yml
+++ b/.yarn/versions/5ca86df5.yml
@@ -1,7 +1,0 @@
-releases:
-  "@radix-ui/react-context-menu": minor
-  "@radix-ui/react-dropdown-menu": minor
-  "@radix-ui/react-menu": minor
-
-declined:
-  - primitives
--- a/packages/react/context-menu/src/ContextMenu.stories.tsx
+++ b/packages/react/context-menu/src/ContextMenu.stories.tsx
@@ -3,6 +3,7 @@ import { css, keyframes } from '../../../../stitches.config';
 import * as ContextMenu from '@radix-ui/react-context-menu';
 import { foodGroups } from '../../../../test-data/foods';
 import { classes, TickIcon } from '../../menu/src/Menu.stories';
+import type { CheckedState } from '@radix-ui/react-checkbox';
 
 const { contentClass, itemClass, labelClass, separatorClass, subTriggerClass } = classes;
 
@@ -454,10 +455,10 @@ export const WithLabels = () => (
 
 export const CheckboxItems = () => {
   const checkboxItems = [
-    { label: 'Bold', state: React.useState(false) },
-    { label: 'Italic', state: React.useState(true) },
-    { label: 'Underline', state: React.useState(false) },
-    { label: 'Strikethrough', state: React.useState(false), disabled: true },
+    { label: 'Bold', state: React.useState<CheckedState>(false) },
+    { label: 'Italic', state: React.useState<CheckedState>(true) },
+    { label: 'Underline', state: React.useState<CheckedState>(false) },
+    { label: 'Strikethrough', state: React.useState<CheckedState>(false), disabled: true },
   ];
 
   return (
--- a/packages/react/context-menu/src/ContextMenu.stories.tsx
+++ b/packages/react/context-menu/src/ContextMenu.stories.tsx
@@ -453,16 +453,8 @@ export const WithLabels = () => (
 );
 
 export const CheckboxItems = () => {
-  const checkboxItems = [
-    { label: 'Bold', state: React.useState<boolean | 'indeterminate'>(false) },
-    { label: 'Italic', state: React.useState<boolean | 'indeterminate'>(true) },
-    { label: 'Underline', state: React.useState<boolean | 'indeterminate'>(false) },
-    {
-      label: 'Strikethrough',
-      state: React.useState<boolean | 'indeterminate'>(false),
-      disabled: true,
-    },
-  ];
+  const checkboxItems = ['Bold', 'Italic', 'Underline'];
+  const [selection, setSelection] = React.useState<string[]>([]);
 
   return (
     <div style={{ textAlign: 'center', padding: 50 }}>
@@ -480,20 +472,32 @@ export const CheckboxItems = () => {
               Smaller
             </ContextMenu.Item>
             <ContextMenu.Separator className={separatorClass()} />
-            {checkboxItems.map(({ label, state: [checked, setChecked], disabled }) => (
+            {checkboxItems.map((item) => (
               <ContextMenu.CheckboxItem
-                key={label}
+                key={item}
                 className={itemClass()}
-                checked={checked}
-                onCheckedChange={setChecked}
-                disabled={disabled}
+                checked={selection.includes(item)}
+                onCheckedChange={() =>
+                  setSelection((current) =>
+                    current.includes(item)
+                      ? current.filter((el) => el !== item)
+                      : current.concat(item)
+                  )
+                }
               >
-                {label}
+                {item}
                 <ContextMenu.ItemIndicator>
                   <TickIcon />
                 </ContextMenu.ItemIndicator>
               </ContextMenu.CheckboxItem>
             ))}
+            <ContextMenu.Separator />
+            <ContextMenu.CheckboxItem className={itemClass()} disabled>
+              Strikethrough
+              <ContextMenu.ItemIndicator>
+                <TickIcon />
+              </ContextMenu.ItemIndicator>
+            </ContextMenu.CheckboxItem>
           </ContextMenu.Content>
         </ContextMenu.Portal>
       </ContextMenu.Root>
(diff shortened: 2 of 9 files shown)


jjenzz:
i'm not sure it makes much sense to create a dependency between `@radix-ui/react-checkbox` and `@radix-ui/react-menu` here given that the menu never renders the checkbox component.

personally i'd copy/paste the type and utils you're re-using here, they're very small.

colepeters:
@jjenzz I wondered about that as well. I thought syncing up these up might help to prevent breaking this feature parity in the future, but if the cost of creating the dependency outweighs that risk, I'm of course happy to change it!

benoitgrelard:
I agree with Jenna, let's duplicate the bits mentioned rather than add a dependency.

colepeters:
@jjenzz @benoitgrelard Done! Please let me know if you see any other changes/improvements to be made.

benoitgrelard:
> Not quite sure how to roll that back as it doesn't appear in the release options now, any ideas?)

You can delete the version file and re-run the command.

colepeters:
Thank you, done!

colepeters:
@jjenzz @benoitgrelard Just checking in — any other changes you'd like to see, or do you reckon this can proceed to workflows & merging? Thanks!

.yarn/versions/5ca86df5.yml
```diff
@@ -0,0 +1,7 @@
+releases:
+  "@radix-ui/react-context-menu": minor
+  "@radix-ui/react-dropdown-menu": minor
+  "@radix-ui/react-menu": minor
```
benoitgrelard, line 4: I'm wondering if this will need to be `major` now because of the change in types for `defaultChecked`/`checked`/`onCheckedChange`.

@andy-hook I think technically we have to right? Because it might break people's builds otherwise (that is, if they are using `CheckboxItem` only…)

If people had built their type on top of ours it would be fine, but I highly doubt it, given it was just a boolean…

packages/react/menu/src/Menu.stories.tsx
```diff
@@ -235,32 +235,46 @@ export const Typeahead = () => (
...
-  const checkboxItems = [
-    { label: 'Bold', state: React.useState(false) },
-    { label: 'Italic', state: React.useState(true) },
-    { label: 'Underline', state: React.useState(false) },
-    { label: 'Strikethrough', state: React.useState(false), disabled: true },
+  const options = [
+    { label: 'Crows' },
+    { label: 'Ravens' },
+    { label: 'Magpies' },
+    { label: 'Jackdaws' },
```
benoitgrelard, line 242: nit: If we're only going to deal with the labels now, let's just make it an array of strings instead.

packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
```diff
@@ -668,43 +668,67 @@ export const MultipleItemsAsDialogTriggers = () => {
...
-  const checkboxItems = [
-    { label: 'Bold', state: React.useState(false) },
-    { label: 'Italic', state: React.useState(true) },
-    { label: 'Underline', state: React.useState(false) },
-    { label: 'Strikethrough', state: React.useState(false), disabled: true },
+  const corvids = [
+    { label: 'Crows' },
+    { label: 'Ravens' },
+    { label: 'Magpies' },
+    { label: 'Jackdaws' },
```
benoitgrelard, line 675: nit: If we're only going to deal with the labels now, let's just make it an array of strings instead.

packages/react/context-menu/src/ContextMenu.stories.tsx
```diff
@@ -454,10 +454,14 @@ export const WithLabels = () => (
...
-    { label: 'Underline', state: React.useState(false) },
-    { label: 'Strikethrough', state: React.useState(false), disabled: true },
+    { label: 'Bold', state: React.useState<boolean | 'indeterminate'>(false) },
+    { label: 'Italic', state: React.useState<boolean | 'indeterminate'>(true) },
+    { label: 'Underline', state: React.useState<boolean | 'indeterminate'>(false) },
+    {
+      label: 'Strikethrough',
+      state: React.useState<boolean | 'indeterminate'>(false),
+      disabled: true,
+    },
```
benoitgrelard, line 464: Can we update this story the same as `DropdownMenu` and `Menu`?

packages/react/menu/src/Menu.tsx
```diff
@@ -746,9 +746,16 @@ const MenuItemImpl = React.forwardRef<MenuItemImplElement, MenuItemImplProps>(
 const CHECKBOX_ITEM_NAME = 'MenuCheckboxItem';
 
 type MenuCheckboxItemElement = MenuItemElement;
+
+type CheckedState = boolean | 'indeterminate';
+
+function isIndeterminate(checked?: CheckedState): checked is 'indeterminate' {
+  return checked === 'indeterminate';
+}
```
benoitgrelard, line 754: nit: Could we move this at the bottom, just between `getOpenState` and `getCheckedState`?

packages/react/menu/src/Menu.tsx
```diff
@@ -758,13 +765,13 @@ const MenuCheckboxItem = React.forwardRef<MenuCheckboxItemElement, MenuCheckboxI
...
           role="menuitemcheckbox"
-          aria-checked={checked}
+          aria-checked={isIndeterminate(checked) ? 'mixed' : checked}
           {...checkboxItemProps}
           ref={forwardedRef}
           data-state={getCheckedState(checked)}
           onSelect={composeEventHandlers(
             checkboxItemProps.onSelect,
-            () => onCheckedChange?.(!checked),
+            () => onCheckedChange?.(isIndeterminate(checked) || !checked),
```
benoitgrelard, line 774: nit, let's use this form to stay consistent with `Checkbox.tsx`:

```suggestion
            () => onCheckedChange?.(isIndeterminate(checked) ? true : !checked),
```

packages/react/menu/src/Menu.tsx
```diff
@@ -849,9 +856,16 @@ MenuRadioItem.displayName = RADIO_ITEM_NAME;
...
+type CheckboxContextValue = {
+  checked: CheckedState;
+};
+
+const [ItemIndicatorProvider, useItemIndicatorContext] = createMenuContext<CheckboxContextValue>(
+  ITEM_INDICATOR_NAME,
+  {
+    checked: false,
+  }
+);
```
benoitgrelard, line 868: nit, formatting:

```suggestion
type CheckboxContextValue = { checked: CheckedState; };

const [ItemIndicatorProvider, useItemIndicatorContext] = createMenuContext<CheckboxContextValue>(
  ITEM_INDICATOR_NAME,
  { checked: false, }
);
```

benoitgrelard (approved):
@colepeters Thank you so much for spearheading this change!
This is looking great and all my comments are mostly cosmetic nits.
We should be able to get that in very soon now!

.yarn/versions/5ca86df5.yml
andy-hook: Strictly speaking yes, though it does feel like quite the leap for such a change.

colepeters: Noted! Just to be clear then, there should be a `major` for all of the affected packages, right?

packages/react/context-menu/src/ContextMenu.stories.tsx
colepeters: Can do, but that will prevent using a disabled checkbox in the story, which was something that was present before my changes. Any concerns about that?

.yarn/versions/5ca86df5.yml
benoitgrelard: @colepeters we're still debating I guess. I'll discuss this with @andy-hook tomorrow and we'll come to a choice and will report here.

andy-hook: @colepeters We caught up and agreed major is the way to go, thanks for the contribution!

packages/react/context-menu/src/ContextMenu.stories.tsx
benoitgrelard: Maybe just add an extra disabled static one below after a separator?

colepeters:
@benoitgrelard Requested changes are in, let me know if you'd like any other updates!

benoitgrelard (approved):

PR #1626, opened 2022-08-21
Remove  prettier-ignore from @radix-ui/react-primitive

CrutchTheClutch:
### Description

Removes the prettier-ignore rule from the react primitive component as the referenced pull request has been merged


jjenzz:
it's not the `prettier-ignore` that was temporary but the `PropsWithoutRef` type. so, the change here would be to remove that type or mark it as deprecated for removal later.

CrutchTheClutch:
Ahh my bad!  Going to close this for now as I don't have the time to flesh out this change

PR #1647, opened 2022-09-06
[DropdownMenu] Prevent passing keydown event to the focused MenuItem from MenuTrigger

### LINKED ISSUE #1268: [DropdownMenu] DropdownMenu.CheckboxItem asChild keyboard issue
### Current Behavior

Given a [CheckboxItem](https://www.radix-ui.com/docs/primitives/components/dropdown-menu#checkboxitem) that has `asChild={true}` and is the first item in the dropdown.

When you use `enter` to activate the trigger:
- Menu will not open
- `onCheckChange` is called on the first menu item
- If you add `onSelect={(e) => e.preventDefault()}` the menu does stay open

When you use `space` to activate the trigger:
- Menu will open
- `onCheckChange` is called on the first menu item

### Expected behavior

Menu should open and first item should not have `onCheckChange` called.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/old-shadow-j307zg)

Steps to reproduce in this codesandbox:
- Focus the trigger
- `enter`, `esc`, `enter`, `esc` 

You should see the menu open and close and the first item toggles on / off as you do this.

### Your environment

| Radix Package(s) |   dropdown-menu      |    0.1.6     |
| Browser          |    chrome     |    99.0.4844.82      |
| Operating System |    windows     |         |

### LINKED ISSUE #1397: DropDownMenu.Trigger fires the first "asChild (as button)" item's function with keyboard activation.
### Current Behavior

When setting asChild to a DropDownMenu item, with the child being a button, using keyboard enter key on the trigger will fire that child's onClick function instead of opening the menu. 

### Expected behavior

Pressing the enter key on the trigger should open the menu.

### Reproducible example

Use keyboard to navigate on the trigger, then press enter key.

[CodeSandbox Template](https://codesandbox.io/s/epic-sara-rubxj7?file=/App.js)

Sample of the code that causes the issue :

```js
          <DropdownMenuItem asChild>
            <button onClick={() => alert("this will fire on keyboard")}>
              This is a button
            </button>
          </DropdownMenuItem>
```

### Suggested solution

DropDownMenu items should be either anchors or buttons at all time, and built with those in mind. asChild seems to cause a lot of issues in this component and could be removed by having correct and accessible default semantics on the menu item itself. It is important that onClick events are never bound to a div, as users can have specific needs requiring them to style, announce and present buttons, anchors and possible interactivity in specific ways. 

### Your environment

| Radix Package(s) |         |     latest    |
| Browser          |     Chrome    |     100    |
| Assistive tech   |    keyboard navigation     |         |

### LINKED ISSUE #1622: [DropdownMenu] pressing Enter on Trigger opens anchor link on Firefox
### Current Behavior

On Firefox, if you have a dropdown where the first `DropdownMenu.Item` with `asChild` and the child is an `a` tag, when navigating with the keyboard and focusing on the `DropdownMenu.Trigger` and pressing Enter the anchor tag link is triggered as if it was clicked/selected.

### Expected behavior

When pressing Space with the `Trigger` focused it only opens the dropdown, as expected. On Chrome both keys only open the dropdown
On Firefox this should work the same way.

### Reproducible example

I forked the menu example from the docs to illustrate, the only thing I changed is adding `asChild` to the first dropdown item, and adding a `<a>` around it with an `href`
https://codesandbox.io/s/patient-http-il6epv?file=/App.js

### Suggested solution

This should work as in Chrome, both the Space and Enter keys should only open the dropdown and not trigger any interactions with any items inside

### Your environment

| Radix Package(s) |     DropdownMenu    |   v1.0.0      |
| React            | n/a     |      18   |
| Browser          |    Firefox     |     103.0.2    |
| Operating System |     Manjaro    |    Latest     |

### LINKED ISSUE #1655: [DropdownMenu] Using `<button>` for the first menu item (with `asChild`) breaks opening the menu with "Enter"
### Current Behavior

[CodeSandbox here (fork of the example in the documentation)](https://codesandbox.io/s/delicate-frost-n4007n); search for `<button` in `App.s` to see the change I made.

1. Use the keyboard to focus on the dropdown button.
2. Press the "Enter" key to open the dropdown menu.

Result: The menu doesn't open (or rather, I think it opens and closes instantly). Note that it still works with "Space".

### Expected behavior

Pressing "Enter" when the dropdown button is focused should open the menu, and not close it instantly. It works in [the example from the documentation](https://codesandbox.io/s/y9cuiu).

### Environment

I was able to reproduce the issue in all the browsers I have on my Mac (Safari, Chrome, Edge, and Firefox).

paperstick:
### Description

This fix should prevent passing `onKeyDown` to the focused `DropdownMenuItem` from `DropdownMenuTrigger` which triggered `onClick` event on `a` or `button` elements.

---
Closes #1268
Closes #1397
Closes #1622
Closes #1655

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -126,7 +126,11 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           })}
           onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
             if (disabled) return;
-            if (['Enter', ' '].includes(event.key)) context.onOpenToggle();
+            if (['Enter', ' '].includes(event.key)) {
+              context.onOpenToggle();
+              // prevent passing keydown event to the focused MenuItem element in Firefox
+              if (!context.open) event.preventDefault();
+            }
             if (event.key === 'ArrowDown') context.onOpenChange(true);
             // prevent keypresses from scrolling window
             if ([' ', 'ArrowDown'].includes(event.key)) event.preventDefault();
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -126,14 +126,11 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           })}
           onKeyDown={composeEventHandlers(props.onKeyDown, (event) => {
             if (disabled) return;
-            if (['Enter', ' '].includes(event.key)) {
-              context.onOpenToggle();
-              // prevent passing keydown event to the focused MenuItem element in Firefox
-              if (!context.open) event.preventDefault();
-            }
+            if (['Enter', ' '].includes(event.key)) context.onOpenToggle();
             if (event.key === 'ArrowDown') context.onOpenChange(true);
-            // prevent keypresses from scrolling window
-            if ([' ', 'ArrowDown'].includes(event.key)) event.preventDefault();
+            // prevent keydown from scrolling window / first focused item to execute
+            // that keydown (inadvertently closing the menu)
+            if (['Enter', ' ', 'ArrowDown'].includes(event.key)) event.preventDefault();
           })}
         />
       </MenuPrimitive.Anchor>
--- a/.yarn/versions/975c6613.yml
+++ b/.yarn/versions/975c6613.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-dropdown-menu": patch
+
+declined:
+  - primitives


packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -126,7 +126,11 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
...
             if (disabled) return;
-            if (['Enter', ' '].includes(event.key)) context.onOpenToggle();
+            if (['Enter', ' '].includes(event.key)) {
+              context.onOpenToggle();
+              // prevent passing keydown event to the focused MenuItem element in Firefox
+              if (!context.open) event.preventDefault();
+            }
             if (event.key === 'ArrowDown') context.onOpenChange(true);
             // prevent keypresses from scrolling window
             if ([' ', 'ArrowDown'].includes(event.key)) event.preventDefault();
```
benoitgrelard, line 136: I think this would be enough no?
I don't think we need the extra `if (!context.open)`

```suggestion
            if (['Enter', ' '].includes(event.key)) context.onOpenToggle();
            if (event.key === 'ArrowDown') context.onOpenChange(true);
            // prevent keydown from scrolling window / first focused item to execute
            // that keydown (inadvertently closing the menu)
            if (['Enter', ' ', 'ArrowDown'].includes(event.key)) event.preventDefault();
```

paperstick: Yes, looks like you are right, we can get rid of unnecessary check for `open` menu

andy-hook (approved):

PR #1659, opened 2022-09-12
[Popper] Add containHeight prop to menus

hipstersmoothie:
### Description

With the current menus if you have content that taller than the available space (ex: menu with many options) that menu will be cut off by the viewport

Example from docs website:

CleanShot 2022-09-12 at 12 29 45

While this might be able to be cleaned up in user land it requires a whole bunch of spaghetti code and is hard to maintain.

Luckily the underlying popper library `floating-ui` has [a middle ware that supports sizing the popper content based on available space.](https://floating-ui.com/docs/size). This PR add a prop `containHeight` that enables this behavior. When passed true the `size` middleware get called without any options. `containHeight` can also take all the options of `size` as an object as well.

Result:

https://user-images.githubusercontent.com/1192452/189739968-272c50ac-a5c5-4407-887a-bc67f68cc7e6.mp4

now the menu will be fully on screen 🎉

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
@@ -205,7 +205,11 @@ export const Submenus = () => {
         <DropdownMenu.Root dir={rtl ? 'rtl' : 'ltr'}>
           <DropdownMenu.Trigger className={triggerClass()}>Open</DropdownMenu.Trigger>
           <DropdownMenu.Portal>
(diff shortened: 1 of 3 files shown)


paul-sachs:
This looks very similar to my PR #1597 which has also been kinda just sitting there.

andy-hook:
Thanks for this @hipstersmoothie , I left [a follow up](https://github.com/radix-ui/primitives/pull/1597#issuecomment-1255952703) in the linked PR explaining our current position regarding changes to these options.

benoitgrelard:
Superseded by #1942

PR #1668, opened 2022-09-16
[Toast] Fix dom conflicts with announce

andy-hook:
While working on toast I realised that internal refs attached to children were behaving strangely, they were returning `null` after 1 second.

I tracked this back to the changes in https://github.com/radix-ui/primitives/pull/1528/files where we render `ToastAnnounce` into a `documentFragment`. As soon as that fragment is removed we lose the reference to it as well.

This change avoids rendering two sets of children entirely. Instead we query for text content from `Toast.Root` and filter out what we don't need using data attrs

### DIFF
--- a/packages/react/toast/src/Toast.tsx
+++ b/packages/react/toast/src/Toast.tsx
@@ -425,7 +425,6 @@ type SwipeEvent = { currentTarget: EventTarget & ToastElement } & Omit<
 >;
 
 const [ToastInteractiveProvider, useToastInteractiveContext] = createToastContext(TOAST_NAME, {
-  isInteractive: false,
   onClose() {},
 });
 
@@ -463,8 +462,8 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
       ...toastProps
     } = props;
     const context = useToastProviderContext(TOAST_NAME, __scopeToast);
-    const ref = React.useRef<ToastImplElement>(null);
-    const composedRefs = useComposedRefs(forwardedRef, ref);
+    const [node, setNode] = React.useState<ToastImplElement | null>(null);
+    const composedRefs = useComposedRefs(forwardedRef, (node) => setNode(node));
     const pointerStartRef = React.useRef<{ x: number; y: number } | null>(null);
     const swipeDeltaRef = React.useRef<{ x: number; y: number } | null>(null);
     const duration = durationProp || context.duration;
@@ -475,7 +474,7 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
     const handleClose = useCallbackRef(() => {
       // focus viewport if focus is within toast to read the remaining toast
       // count to SR users and ensure focus isn't lost
-      const isFocusInToast = ref.current?.contains(document.activeElement);
+      const isFocusInToast = node?.contains(document.activeElement);
       if (isFocusInToast) context.viewport?.focus();
       onClose();
     });
@@ -522,21 +521,27 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
       return () => onToastRemove();
     }, [onToastAdd, onToastRemove]);
 
+    const announceTextContent = React.useMemo(() => {
+      return node ? getAnnounceTextContent(node) : null;
+    }, [node]);
+
     if (!context.viewport) return null;
 
     return (
       <>
-        <ToastAnnounce
-          __scopeToast={__scopeToast}
-          // Toasts are always role=status to avoid stuttering issues with role=alert in SRs.
-          role="status"
-          aria-live={type === 'foreground' ? 'assertive' : 'polite'}
-          aria-atomic
-        >
-          {props.children}
-        </ToastAnnounce>
+        {announceTextContent && (
+          <ToastAnnounce
+            __scopeToast={__scopeToast}
+            // Toasts are always role=status to avoid stuttering issues with role=alert in SRs.
+            role="status"
+            aria-live={type === 'foreground' ? 'assertive' : 'polite'}
+            aria-atomic
+          >
+            {announceTextContent}
+          </ToastAnnounce>
+        )}
 
-        <ToastInteractiveProvider scope={__scopeToast} isInteractive onClose={handleClose}>
+        <ToastInteractiveProvider scope={__scopeToast} onClose={handleClose}>
           {ReactDOM.createPortal(
             <Collection.ItemSlot scope={__scopeToast}>
               <DismissableLayer.Root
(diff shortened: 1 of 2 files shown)


packages/react/toast/src/Toast.tsx
```diff
@@ -794,21 +778,47 @@ const ToastClose = React.forwardRef<ToastCloseElement, ToastCloseProps>(
...
 ToastClose.displayName = CLOSE_NAME;
 
 /* ---------------------------------------------------------------------------------------------- */
 
+type ToastAnnounceExcludeElement = React.ElementRef<typeof Primitive.div>;
+interface ToastAnnounceExcludeProps extends PrimitiveDivProps {
+  altText?: string;
+}
+
+const ToastAnnounceExclude = React.forwardRef<
```
andy-hook, line 804: I added a new comp with idea of colocating `getAnnounceTextContent` near it in the file, I haven't moved the code yet to spare the diff but let me know if you think this is overkill in general.

packages/react/toast/src/Toast.tsx
```diff
@@ -794,21 +778,47 @@ const ToastClose = React.forwardRef<ToastCloseElement, ToastCloseProps>(
...
+  ToastAnnounceExcludeElement,
+  ToastAnnounceExcludeProps
+>((props: ScopedProps<ToastAnnounceExcludeProps>, forwardedRef) => {
+  const { __scopeToast, altText, ...announceHiddenProps } = props;
+
+  return (
+    <Primitive.div
+      data-radix-toast-announce-exclude={altText ? undefined : ''}
+      data-radix-toast-announce-alt={altText || undefined}
+      {...announceHiddenProps}
```
benoitgrelard, line 814: Should this be renamed `announceExcludeProps`?

packages/react/toast/src/Toast.tsx
```diff
@@ -794,21 +778,47 @@ const ToastClose = React.forwardRef<ToastCloseElement, ToastCloseProps>(
...
-      <Primitive.button
-        type="button"
-        {...closeProps}
-        ref={forwardedRef}
-        onClick={composeEventHandlers(props.onClick, interactiveContext.onClose)}
-      />
-    ) : null;
+
+    return (
+      <ToastAnnounceExclude asChild>
```
benoitgrelard, line 783: What happens for this one as it has no `altText`?
I can imagine it will read the `textContent` (say "close") but what if it's a close icon with an `aria-label`?

packages/react/toast/src/Toast.tsx
benoitgrelard: I'm not fussed about `getAnnounceTextContent` being near it in the file really.
But seeing the data attributes here in one place, rather than spread out in both places's implementation is a good thing to keep, so I guess we get the bonus of the collocation if you want to move it before merging the PR.

packages/react/toast/src/Toast.tsx
```diff
@@ -864,8 +874,16 @@ function getAnnounceTextContent(container: HTMLElement) {
   childNodes.forEach((node) => {
     if (node.nodeType === node.TEXT_NODE && node.textContent) textContent.push(node.textContent);
     if (isHTMLElement(node)) {
-      const isHidden = node.ariaHidden || node.hidden || node.style.display === 'none';
-      if (!isHidden) textContent.push(...getAnnounceTextContent(node));
+      const isHidden =
+        node.ariaHidden ||
+        node.hidden ||
+        node.style.display === 'none' ||
+        node.dataset.radixToastAnnounceExclude === '';
```
benoitgrelard, line 881: It bothers me that this is bundled in `isHidden`.
I'd rather we have an extra `const isExcluded` and the do `if (!isHidden & !isExcluded)` later.
Thoughts?

andy-hook: Yeah I had something like this originally, sounds good.

packages/react/toast/src/Toast.tsx
```diff
@@ -805,18 +809,43 @@ const ToastAnnounceExclude = React.forwardRef<
   ToastAnnounceExcludeElement,
   ToastAnnounceExcludeProps
 >((props: ScopedProps<ToastAnnounceExcludeProps>, forwardedRef) => {
-  const { __scopeToast, altText, ...announceHiddenProps } = props;
+  const { __scopeToast, altText, ...announceExcludeProps } = props;
 
   return (
     <Primitive.div
-      data-radix-toast-announce-exclude={altText ? undefined : ''}
+      data-radix-toast-announce-exclude=""
```
benoitgrelard, line 816: nit:
```suggestion
      data-radix-toast-announce-exclude
```

packages/react/toast/src/Toast.tsx
```diff
@@ -805,18 +809,43 @@ const ToastAnnounceExclude = React.forwardRef<
...
 
+function getAnnounceTextContent(container: HTMLElement) {
+  const textContent: string[] = [];
+  const childNodes = Array.from(container.childNodes);
+
+  childNodes.forEach((node) => {
+    if (node.nodeType === node.TEXT_NODE && node.textContent) textContent.push(node.textContent);
+    if (isHTMLElement(node)) {
+      const isHidden = node.ariaHidden || node.hidden || node.style.display === 'none';
+      const isExcluded = node.dataset.radixToastAnnounceExclude === '';
```
benoitgrelard, line 832: nit (perhaps? untested): 
```suggestion
      const isExcluded = 'radixToastAnnounceExclude' in node.dataset;
```

packages/react/toast/src/Toast.tsx
```diff
@@ -805,18 +809,43 @@ const ToastAnnounceExclude = React.forwardRef<
...
+
+      if (!isHidden && isExcluded) {
+        const altText = node.dataset.radixToastAnnounceAlt;
+        if (altText) textContent.push(altText);
+        return;
+      }
+
+      if (!isHidden) textContent.push(...getAnnounceTextContent(node));
```
benoitgrelard, line 840: Thoughts on this?

```suggestion
      if (!isHidden) {
        if (isExcluded) {
          const altText = node.dataset.radixToastAnnounceAlt;
          if (altText) textContent.push(altText);
        } else {
          textContent.push(...getAnnounceTextContent(node));
        }
      }
```

packages/react/toast/src/Toast.tsx
andy-hook: This will output `data-radix-toast-announce-exclude="true"`, though I notice we tend to prefer empty string for the more compact `data-radix-toast-announce-exclude` in other data attributes.

I'm cool with the change but curious on the distinction here?

benoitgrelard: Ah yeah good point, I forgot that.

packages/react/toast/src/Toast.tsx
benoitgrelard: In this case, I guess we can keep this as is too then.

benoitgrelard (approved):

PR #1677, opened 2022-09-21
[DismissableLayer] Add DOM event handlers to ownerDocument

### LINKED ISSUE #1676: [DismissableLayer] Elements rendered in popup windows are not handled correctly
## Bug report

Dropdown menus rendered in popup windows (i.e. separate windows created via `window.open`) are not dismissed correctly when clicking outside of them. This appears to be due to `DismissableLayer` always adding event listeners to the `document` global, instead of using `node.ownerDocument`.

### Current Behavior

Clicking outside of a menu in a popup window does nothing:
dropdow-menu-popup

### Expected behavior

Clicking outside of a  menu in the main window dismisses it, the same should happen in popups:

dropdow-menu-main

### Reproducible example

https://codesandbox.io/s/agitated-stitch-txdns4 is a simple test case. The "…" button opens a menu. It works as expected in the main window, but when we render the component in a popup window (opened by the "Open Popup" button) dismissal does not work.

### Suggested solution

Event handlers such as the ones from https://github.com/radix-ui/primitives/blob/6c398d9810fc52577780b5d987d2a6f11e86276b/packages/react/dismissable-layer/src/DismissableLayer.tsx#L233-L238 should be added on `node.ownerDocument` rather than `document`. Possibly a helper ref like `ownerDocumentRef ` from https://github.com/radix-ui/primitives/blob/fbfa1e8072df7b12bcdfdf30ecc2fbead2022303/packages/react/announce/src/Announce.tsx#L99-L107 may be useful.

### Additional context

I work on a multi-window application that renders components in popup windows (to allow window management). All of the logic runs in the main window for performance reasons (so that we don't need to load a copy of the application's code in each window).

### Your environment

The environment is not really relevant since this occurs everywhere

| Radix Package(s) |  DropdownMenu,   DismissableLayer     |  1.0.0       |
| React            | n/a     | 18.2.0        |
| Browser          | Chrome        | 105.0.5195.102        |
| Operating System | macOS        | 12.6        |

mihaip:
### Description

Adding event handlers  to the `document` global is not correct when the element is rendered in a popup window. This changes the `usePointerDownOutside`, `useFocusOutside` and `useEscapeKeydown` hooks used by `DimissableLayer ` to take into account the correct document.

Fixes #1676

### DIFF
--- a/packages/react/dismissable-layer/src/DismissableLayer.stories.tsx
+++ b/packages/react/dismissable-layer/src/DismissableLayer.stories.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable jsx-a11y/accessible-emoji */
 
 import React from 'react';
+import ReactDOM from 'react-dom';
 import { FocusScope } from '@radix-ui/react-focus-scope';
 import * as Popper from '@radix-ui/react-popper';
 import { Portal } from '@radix-ui/react-portal';
@@ -643,3 +644,23 @@ function DummyPopover({
     </Popper.Root>
   );
 }
+
+export const InPopupWindow = () => {
+  const handlePopupClick = React.useCallback(() => {
+    const popupWindow = window.open(undefined, undefined, 'width=300,height=300,top=100,left=100');
+    if (!popupWindow) {
+      console.error('Failed to open popup window, check your popup blocker settings');
+      return;
+    }
+
+    const containerNode = popupWindow.document.createElement('div');
+    popupWindow.document.body.append(containerNode);
+
+    ReactDOM.render(<DismissableBox />, containerNode);
+  }, []);
+  return (
+    <div style={{ fontFamily: 'sans-serif', textAlign: 'center' }}>
+      <button onClick={handlePopupClick}>Open Popup</button>
+    </div>
+  );
+};
--- a/packages/react/dismissable-layer/src/DismissableLayer.tsx
+++ b/packages/react/dismissable-layer/src/DismissableLayer.tsx
@@ -73,6 +73,7 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
     } = props;
     const context = React.useContext(DismissableLayerContext);
     const [node, setNode] = React.useState<DismissableLayerElement | null>(null);
+    const ownerDocument = node?.ownerDocument ?? document;
     const [, force] = React.useState({});
     const composedRefs = useComposedRefs(forwardedRef, (node) => setNode(node));
     const layers = Array.from(context.layers);
@@ -89,7 +90,7 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
       onPointerDownOutside?.(event);
       onInteractOutside?.(event);
       if (!event.defaultPrevented) onDismiss?.();
-    });
+    }, ownerDocument);
 
     const focusOutside = useFocusOutside((event) => {
       const target = event.target as HTMLElement;
@@ -98,7 +99,7 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
       onFocusOutside?.(event);
       onInteractOutside?.(event);
       if (!event.defaultPrevented) onDismiss?.();
-    });
+    }, ownerDocument);
 
     useEscapeKeydown((event) => {
       const isHighestLayer = index === context.layers.size - 1;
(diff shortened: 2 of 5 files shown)


packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -73,16 +73,22 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
     } = props;
     const context = React.useContext(DismissableLayerContext);
     const [node, setNode] = React.useState<DismissableLayerElement | null>(null);
+    const [ownerDocument, setOwnerDocument] = React.useState<Document>(document);
```
benoitgrelard, line 76: Do we need this in state, given that it can be derived from `node` which is already in state?

mihaip: You're right, it's not necessary. I've changed it to get it from `node`.

mihaip:
@benoitgrelard any chance for another review?

packages/react/dismissable-layer/src/DismissableLayer.stories.tsx
```diff
@@ -643,3 +644,23 @@ function DummyPopover({
     </Popper.Root>
   );
 }
+
+export const InPopupWindow = () => {
```
benoitgrelard, line 648: nit: can we move this just above the "Chromatic" story, as we like to keep this one last. Thanks!

packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
```diff
@@ -1569,6 +1570,43 @@ export const Chromatic = () => {
 };
 Chromatic.parameters = { chromatic: { disable: false } };
 
+export const InPopupWindow = () => {
```
benoitgrelard, line 1573: nit: can we move this just above the "Chromatic" story, as we like to keep this one last. Thanks!

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -219,7 +220,10 @@ type FocusOutsideEvent = CustomEvent<{ originalEvent: FocusEvent }>;
  * to mimic layer dismissing behaviour present in OS.
  * Returns props to pass to the node we want to check for outside events.
  */
-function usePointerDownOutside(onPointerDownOutside?: (event: PointerDownOutsideEvent) => void) {
+function usePointerDownOutside(
+  ownerDocument: Document,
+  onPointerDownOutside?: (event: PointerDownOutsideEvent) => void
+) {
```
benoitgrelard, line 226: For consistency, could we follow the same API as you did in `onEscapeKeydown` where the `ownerDocument` comes second and default to `document`?

packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -293,7 +297,10 @@ function usePointerDownOutside(onPointerDownOutside?: (event: PointerDownOutside
  * Listens for when focus happens outside a react subtree.
  * Returns props to pass to the root (node) of the subtree we want to check.
  */
-function useFocusOutside(onFocusOutside?: (event: FocusOutsideEvent) => void) {
+function useFocusOutside(
+  ownerDocument: Document,
+  onFocusOutside?: (event: FocusOutsideEvent) => void
+) {
```
benoitgrelard, line 303: For consistency, could we follow the same API as you did in `onEscapeKeydown` where the `ownerDocument` comes second and default to `document`?

benoitgrelard (commented):
Thanks for making the change @mihaip, just a few final bits before we can merge this.

packages/react/dismissable-layer/src/DismissableLayer.stories.tsx
mihaip: `DismissableLayer` does not appear to have a `Chromatic` story.

packages/react/dropdown-menu/src/DropdownMenu.stories.tsx
mihaip: Done.

packages/react/dismissable-layer/src/DismissableLayer.tsx
mihaip: Done.

packages/react/dismissable-layer/src/DismissableLayer.tsx
mihaip: Done.

mihaip:
> Edit: it looks like we need to re-run the version check also since changes on `@radix-ui/use-esca-keydown` (and dependents potentially)

Done.

benoitgrelard (approved):

PR #1669, opened 2022-09-16
[Toast] Add `onPause` and `onResume`

andy-hook:
builds on https://github.com/radix-ui/primitives/pull/1668
closes https://github.com/radix-ui/primitives/issues/1453
supercedes https://github.com/radix-ui/primitives/pull/1477

PR #1682, opened 2022-09-22
[Toast] Fix timer issues

andy-hook:
Address timer issues mentioned in https://github.com/radix-ui/primitives/pull/1669#issuecomment-1252282997

The underlying issue is that the pause handling doesn't account for multiple calls in succession. Because we're now exposing `onPause` and `onResume` it made more sense to address the issue at source with the duplicate events being fired.

@benoitgrelard might be easier to hop on a call to go over this.

### DIFF
--- a/packages/react/toast/src/Toast.tsx
+++ b/packages/react/toast/src/Toast.tsx
@@ -190,40 +190,11 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
           }
         };
 
-        const elementIsFocusProxy = (element: HTMLElement | null) => {
-          const tailProxyNode = tailFocusProxyRef.current;
-          const headProxyNode = headFocusProxyRef.current;
-          return element === tailProxyNode || element === headProxyNode;
-        };
-
         const handleFocusOutResume = (event: FocusEvent) => {
-          const nextFocusedElement = event.relatedTarget as HTMLElement;
-
-          // Only attempt to resume when focus moves out of the wrapper
-          if (!wrapper.contains(nextFocusedElement)) {
-            const prevFocusedElement = event.target as HTMLElement;
-            const prevFocusedElementIsProxy = elementIsFocusProxy(prevFocusedElement);
-
-            // activeElement called within `focusout` will always return `document.body` unless focus
-            // is within the wrapper and the user clicks outside of the window,
-            // we use this to determine the intent by which `focusout` was dispatched
-            const activeElementIsInWrapper = wrapper.contains(document.activeElement);
-            const hasClickedOutOfWindow = activeElementIsInWrapper && nextFocusedElement === null;
-            const hasTabbedOutOfWindow = prevFocusedElementIsProxy && nextFocusedElement === null;
-
-            // If the intent is leaving the window then we prevent dispatching resume
-            // as window.blur will handle this event instead
-            if (hasTabbedOutOfWindow || hasClickedOutOfWindow) return;
-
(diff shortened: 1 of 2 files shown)


packages/react/toast/src/Toast.tsx
```diff
@@ -173,36 +173,70 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
     React.useEffect(() => {
       const wrapper = wrapperRef.current;
       const viewport = ref.current;
-      if (wrapper && viewport) {
+      if (hasToasts && wrapper && viewport) {
```
andy-hook, line 176: afaik these handlers shouldn't be necessary when there are no toasts.

packages/react/toast/src/Toast.tsx
```diff
@@ -173,36 +173,70 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
         const handlePause = () => {
-          const pauseEvent = new CustomEvent(VIEWPORT_PAUSE);
-          viewport.dispatchEvent(pauseEvent);
-          context.isClosePausedRef.current = true;
+          if (!context.isClosePausedRef.current) {
```
andy-hook, line 178: If we're already paused / running then we shouldn't be dispatching duplicate events.

packages/react/toast/src/Toast.tsx
```diff
@@ -173,36 +173,70 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
+          }
+        };
+
+        const isFocusProxy = (element: HTMLElement | null) => {
+          const tailProxyNode = tailFocusProxyRef.current;
+          const headProxyNode = headFocusProxyRef.current;
+          return element === tailProxyNode || element === headProxyNode;
+        };
+
+        const handleFocusOutResume = (event: FocusEvent) => {
```
andy-hook, line 199: When focusing away from `window` we were seeing duplicate events as there was an overlap between `focusout` and `window.blur`, they both fire resulting in `onPause` -> `onResume` -> `onPause`.

As a consumer I wouldn't expect to see this, if they are already paused then the handlers shouldn't fire until that perceivably changes.

packages/react/toast/src/Toast.tsx
```diff
@@ -173,36 +173,47 @@ const ToastViewport = React.forwardRef<ToastViewportElement, ToastViewportProps>
...
+          if (context.isClosePausedRef.current) {
+            const resumeEvent = new CustomEvent(VIEWPORT_RESUME);
+            viewport.dispatchEvent(resumeEvent);
+            context.isClosePausedRef.current = false;
+          }
+        };
+
+        const handleFocusOutResume = (event: FocusEvent) => {
+          // Only resume when focus moves out of the wrapper
+          if (!wrapper.contains(event.relatedTarget as HTMLElement)) {
```
andy-hook, line 195: We still need to handle for this as `focusout / focusin` will fire while tabbing within wrapper and we don't want to flip the flag until we're back outside.

benoitgrelard:
I think I spotted another issue:

https://user-images.githubusercontent.com/1539897/192280746-b25a2aa5-784b-421e-b487-3ca549fee40b.mp4

Leave should only trigger resume if focus isn't inside right?

andy-hook:
> Leave should only trigger resume if focus isn't inside right?

Good spot, tweaked a little. Let me know how that feels.

benoitgrelard (approved):

Issue #1568, opened 2022-07-25
[DropdownMenu] no way to "flip" content instead of shifting in v1.0.0


MikeYermolayev:
### Current Behavior

I'm in the process of upgrading to dropdown-menu to 1.0.0 and noticed a small change in behaviour. 
On version 0.1.6 it used to switch 'align' when collision was there but now it just moves moves content slightly along the axis. 

Is there a way to switch to previous behaviour?

On attached screenshots you can see that data-aligned was changed to 'end' when collision occurs but 1.0.0 keeps 'start' 

0.1.6
![image](https://user-images.githubusercontent.com/8783498/180780342-b1f930c9-deeb-4697-904b-c4a2bb50c307.png)
1.0.0
![image](https://user-images.githubusercontent.com/8783498/180780397-32d3b708-576c-4cff-96ad-37a545868598.png)

### Expected behavior

I'd like to have a way to disable shifting content and prefer Flip when collision occurs

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/nice-chihiro-d8hvq6)

### Suggested solution

Introduce a new prop to control it?
Maybe `avoidCollisions` can be converted to enum, e.g. `avoidCollisionsMode: 'off' | 'flip' | 'shift' | 'auto'`

### Additional context

The issue is based on discussions:
https://discord.com/channels/752614004387610674/803656530259738674/1001062835306119218
https://discord.com/channels/752614004387610674/803656530259738674/1001064659119509545

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |   @radix-ui/react-dropdown-menu      |     1.0.0    |
| React            | n/a     |     16.14    |
| Browser          |    Brave     |     [Version 1.40.113 Chromium: 103.0.5060.114 (Official Build) (x86_64)](https://brave.com/latest/)    |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |   MacOs      |     12.4    |

andy-hook:
@benoitgrelard This came up in Discord and I was re-examining our decision to always `Shift` along one axis, it makes sense for `ContextMenu` but I can see the argument for having it on `DropdownMenu` in some situations.

benoitgrelard:
It seems potentially we need to give more flexibility/access to the underlying `floating-ui` configuration.

jgoz:
[...]

PR #1683, opened 2022-09-22
[Popper] Allow more control over avoidCollisions

jgoz:
Ref #1568.

### Description

Expands the possible values for `avoidCollisions` to include the following:

- `auto` &mdash; current behavior (as of 1.0.0) &mdash; Shift along the main axis and flip over the cross axis
- `flip` &mdash; Flip along both axes, but don't shift
- `shift` &mdash; Shift along the main axis only, but don't flip
- `off` &mdash; disable collision avoidance

Boolean values are still accepted for compatibility:
- `true` &rarr; `auto`
- `false` &rarr; `off`

I looked for Stories that would be appropriate to extend, but couldn't find any that demonstrated the shift+flip/auto behavior. Happy to add one if a maintainer pointed me in the right direction.

### Open questions

- In `shift` mode, should we set `crossAxis: true`? Or should `flip` and `shift` act as strict behavioral subsets of `auto`, per the current implementation?

### DIFF
--- a/.yarn/versions/11f57b40.yml
+++ b/.yarn/versions/11f57b40.yml
@@ -1,0 +1,13 @@
+releases:
+  "@radix-ui/react-context-menu": minor
+  "@radix-ui/react-dropdown-menu": minor
+  "@radix-ui/react-hover-card": minor
+  "@radix-ui/react-menu": minor
+  "@radix-ui/react-menubar": minor
+  "@radix-ui/react-popover": minor
+  "@radix-ui/react-popper": minor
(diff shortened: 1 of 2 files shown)


andy-hook:
thanks @jgoz , I left a [follow up](https://github.com/radix-ui/primitives/pull/1597#issuecomment-1255952703) regarding similar options and wanted to share that sentiment here 🙏

jgoz:
Thanks @andy-hook! Feel free to close this PR to reduce noise. Looking forward to the solution you come up with

PR #1686, opened 2022-09-26
[Label] Use native `label`

### LINKED ISSUE #1233: [Label] hover effect should match label element
### Current Behavior

If you hover over a `<label />` associated with a Radix Checkbox (or `<input />`), the checkbox's hover state *is* triggered.

If you hover over a `<LabelPrimitive.Root />` associated with a checkbox (or `<input />`), the checkbox's hover state *is not* triggered.

### Expected behavior

The behaviour of `<LabelPrimitive.Root />` should match `<label />` .

If you follow the documentation for Checkbox:

https://www.radix-ui.com/docs/primitives/components/checkbox

It shows usage with a standard label element, including a hover effect. If you then switch to using Radix' label component, you lose the hover effect.

### Reproducible example

Hover over the labels in this example:

https://codesandbox.io/s/ecstatic-rosalind-eky4tl

### Your environment

| Radix Package(s) |   react-checkbox, react-label      |    0.1.5     |
| React            | 17.0.2   |         |
| Browser          |   Chromium      |     98.0.4758.109     |
| Operating System |    macOS     |     11.6    |

benoitgrelard:
Closes #1233


.yarn/versions/28fccc50.yml
```diff
@@ -0,0 +1,42 @@
...
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-collapsible": patch
+  "@radix-ui/react-collection": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
+  "@radix-ui/react-dismissable-layer": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-focus-scope": patch
+  "@radix-ui/react-hover-card": patch
+  "@radix-ui/react-label": major
```
benoitgrelard, line 18: `major` in a sea of `patch`es

packages/react/label/src/Label.stories.tsx
```diff
@@ -1,111 +1,31 @@
...
-
-export const WithHtmlFor = () => (
-  <>
-    <Label htmlFor="test" className={rootClass()}>
-      This should add an `aria-labelledby` to the control
-    </Label>
-    <Control id="test" />
-  </>
-);
-
```
benoitgrelard, line 30: Removed these now irrelevant stories.

packages/react/label/src/Label.stories.tsx
```diff
@@ -1,111 +1,31 @@
...
-    <h1>With htmlFor</h1>
-    <Label htmlFor="two" className={chromaticRootClass()}>
-      {' '}
-      This should add an `aria-labelledby` to the control
-    </Label>
-    <Control id="two" />
-  </>
-);
-Chromatic.parameters = { chromatic: { disable: false } };
```
benoitgrelard, line 101: Not much we now can/need to test visually.

packages/react/select/src/Select.tsx
```diff
@@ -233,14 +225,19 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
...
         data-state={context.open ? 'open' : 'closed'}
         disabled={disabled}
         data-disabled={disabled ? '' : undefined}
         data-placeholder={context.value === undefined ? '' : undefined}
         {...triggerProps}
         ref={composedRefs}
+        onClick={composeEventHandlers(triggerProps.onClick, (event) => {
+          // enable compatibility with native label or custom `Label` "click"
+          // this doesn't create any other side-effect because we are preventing default
+          // in `onPointerDown` so effectively this only runs for a label "click"
```
andy-hook, line 238: My memory is already failing but was this a Safari specific compatibility? if so should we be more explicit about that here?

andy-hook (approved):
Change looks good, just a tiny nit about a comment that may / not apply.

Do we need to do any extra as relates to `Slider` given the now native label?

packages/react/select/src/Select.tsx
andy-hook: Thanks, reads great 👍

PR #1692, opened 2022-09-28
[HoverCard] Contain user text selections

andy-hook:
closes https://github.com/radix-ui/primitives/issues/1470
supersedes https://github.com/radix-ui/primitives/pull/1479

Scope user selections to the content, improve dismiss handling to improve highlighting UX.

Builds on the work by @innub in https://github.com/radix-ui/primitives/pull/1479 but with added Safari support and some improvements to avoid dismissing while attempting to highlight in certain situations:

https://user-images.githubusercontent.com/11708259/192782489-9ae3effb-9f32-43bf-b93c-4b4a7f22118e.mp4

### DIFF
--- a/packages/react/hover-card/src/HoverCard.tsx
+++ b/packages/react/hover-card/src/HoverCard.tsx
@@ -17,6 +17,8 @@ import type { Scope } from '@radix-ui/react-context';
  * HoverCard
  * -----------------------------------------------------------------------------------------------*/
 
+let originalBodyUserSelect: string;
+
 const HOVERCARD_NAME = 'HoverCard';
 
 type ScopedProps<P> = P & { __scopeHoverCard?: Scope };
@@ -31,6 +33,7 @@ type HoverCardContextValue = {
   onOpen(): void;
   onClose(): void;
   onDismiss(): void;
+  isTextSelectingRef: React.MutableRefObject<boolean>;
 };
 
 const [HoverCardProvider, useHoverCardContext] =
@@ -58,6 +61,7 @@ const HoverCard: React.FC<HoverCardProps> = (props: ScopedProps<HoverCardProps>)
   const popperScope = usePopperScope(__scopeHoverCard);
   const openTimerRef = React.useRef(0);
   const closeTimerRef = React.useRef(0);
+  const isTextSelectingRef = React.useRef(false);
 
   const [open = false, setOpen] = useControllableState({
     prop: openProp,
@@ -72,7 +76,9 @@ const HoverCard: React.FC<HoverCardProps> = (props: ScopedProps<HoverCardProps>)
 
   const handleClose = React.useCallback(() => {
     clearTimeout(openTimerRef.current);
-    closeTimerRef.current = window.setTimeout(() => setOpen(false), closeDelay);
+    if (!isTextSelectingRef.current) {
+      closeTimerRef.current = window.setTimeout(() => setOpen(false), closeDelay);
+    }
   }, [closeDelay, setOpen]);
 
   const handleDismiss = React.useCallback(() => setOpen(false), [setOpen]);
@@ -93,6 +99,7 @@ const HoverCard: React.FC<HoverCardProps> = (props: ScopedProps<HoverCardProps>)
       onOpen={handleOpen}
       onClose={handleClose}
       onDismiss={handleDismiss}
+      isTextSelectingRef={isTextSelectingRef}
     >
       <PopperPrimitive.Root {...popperScope}>{children}</PopperPrimitive.Root>
     </HoverCardProvider>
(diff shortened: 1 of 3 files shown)


packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -256,6 +263,50 @@ const HoverCardContentImpl = React.forwardRef<
...
   const ref = React.useRef<HoverCardContentImplElement>(null);
   const composedRefs = useComposedRefs(forwardedRef, ref);
+  const [containTextSelection, setContainTextSelection] = React.useState(false);
+
+  React.useEffect(() => {
+    if (containTextSelection) {
+      const body = document.body;
+
+      // Safari requires prefix
+      originalBodyUserSelect = body.style.userSelect || body.style.webkitUserSelect;
```
andy-hook, line 273: Do we feel this restoration is necessary?

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -256,6 +263,50 @@ const HoverCardContentImpl = React.forwardRef<
...
+    }
+  }, [containTextSelection]);
+
+  const { onDismiss } = context;
+
+  React.useEffect(() => {
+    if (ref.current) {
+      const handlePointerUp = (event: PointerEvent) => {
+        // Delay a frame to ensure we always access the latest selection
+        setTimeout(() => {
```
andy-hook, line 290: I did a lot of testing between browsers and found that the selection would sometimes lag `pointerup` by a frame, I found delaying here to be the simplest of the options available as we're not handling any consumer events.

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: I mean, unlikely that someone makes it anything on body by default, but for completeness I think we should yes.

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -277,12 +328,19 @@ const HoverCardContentImpl = React.forwardRef<
       <PopperPrimitive.Content
         {...popperScope}
         {...contentProps}
+        onPointerDown={composeEventHandlers(contentProps.onPointerDown, () => {
+          setContainTextSelection(true);
+          context.isTextSelectingRef.current = true;
```
benoitgrelard, line 333: Do we need the extra ref if we have a state already that tells us we're containing a selection?
I know it's currently lower down, but we could lift it?

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -256,6 +263,50 @@ const HoverCardContentImpl = React.forwardRef<
...
+  React.useEffect(() => {
+    if (ref.current) {
+      const handlePointerUp = (event: PointerEvent) => {
+        // Delay a frame to ensure we always access the latest selection
+        setTimeout(() => {
+          const hasEmptySelection = document.getSelection()?.toString() === '';
+          setContainTextSelection(false);
+
+          if (hasEmptySelection) {
+            const isPointerUpOutsideContent = !ref.current?.contains(event.target as HTMLElement);
```
benoitgrelard, line 295: this won't respect the React tree? Do we care?

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -256,6 +263,50 @@ const HoverCardContentImpl = React.forwardRef<
...
+            context.isTextSelectingRef.current = false;
+            if (isPointerUpOutsideContent) onDismiss();
```
benoitgrelard, line 297: It seems native closes on mousemove outside?

packages/react/hover-card/src/HoverCard.tsx
andy-hook: I don't think we do, this is similar to the dismissible handling afaik which also uses `document`, is there even a way to handle that within the React tree?

This actually makes me wonder if we should be using `ownerDocument` here

packages/react/hover-card/src/HoverCard.tsx
andy-hook: Are you referring to the VSC example? or another?

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: In DismissableLayer we do use the React tree right? for `onPointerDownOutside`/`onFocusOutside`.

packages/react/hover-card/src/HoverCard.tsx
benoitgrelard: Yes VSC

packages/react/hover-card/src/HoverCard.tsx
andy-hook: I thought internally it still needs to attach on `document`:
https://github.com/radix-ui/primitives/blob/5972f26d60dd1bb3b16888be4c9e3ab50e74aaf9/packages/react/dismissable-layer/src/DismissableLayer.tsx#L222

Perhaps I misunderstand how that functions?

packages/react/hover-card/src/HoverCard.tsx
andy-hook: Gotcha, I had a version that did this and found in practical terms it added little as the card dismisses immediately if you're at all moving the pointer following `pointerUp` (which I assume would be most of the time unless you intentionally try not to)

benoitgrelard: Yeah I did wonder, because it unmounts, you wouldn't be able to copy the text if you haven't done so before moving your mouse. So probably better to leave it as is?

packages/react/hover-card/src/HoverCard.tsx
andy-hook: We need this to handle the "has a selection" case which prevents closing if there's a selection after pointerUp and pointer leaves the card.

It also informs parent layers that a child has a selection and to not close when "pointerLeave" bubbles

packages/react/hover-card/src/HoverCard.tsx
andy-hook: Thanks for clarifying this. I've implemented a similar approach but for `pointerUp` https://github.com/radix-ui/primitives/pull/1692/commits/d0bfcbd126f756e7100c5e82f7d082ab2e851b61

packages/react/hover-card/src/HoverCard.tsx
andy-hook: > So probably better to leave it as is?

Yeah I think what we have is good enough

packages/react/hover-card/src/HoverCard.tsx
```diff
@@ -271,18 +322,33 @@ const HoverCardContentImpl = React.forwardRef<
       onInteractOutside={onInteractOutside}
       onEscapeKeyDown={onEscapeKeyDown}
       onPointerDownOutside={onPointerDownOutside}
-      onFocusOutside={onFocusOutside}
+      onFocusOutside={composeEventHandlers(onFocusOutside, (event) => {
+        event.preventDefault();
+      })}
```
andy-hook, line 327: Fixes:
I don't think we need this behaviour anyway as we disable focus access in the cards and we're already handling close on trigger `blur`

https://user-images.githubusercontent.com/11708259/193234945-107aa0a1-b398-4755-a0ea-59f84e74bf23.mp4

benoitgrelard (approved):

PR #1695, opened 2022-09-29
[Slider] Allow visual inversion

### LINKED ISSUE #1635: [Slider] Prop to flip direction of values in vertical orientation
### Overview
This is feature request concerning the Slider Primitive
When the current slider is in vertical orientation, value = 0 is at the bottom of the slider, while value = max is at the top. 

It would be great to have an option of switch the direction, so that value = 0 is at the top of the slider and value = max is at the bottom. 

There is currently a prop on Slider.Root to switch the direction on the slider if it is horizontal, (dir = rtl or ltr), it would be useful to have the same but for vertical! 

I'm asking because I need it. My app has an array of cards, and a slider that focuses each card as it moves down. However, because the direction of the slider is the way it is, I need to do some Math to make the position of the thumb align to the card that is focuses. I'll link a prototype below. The relevant slider code starts on line 35
https://replit.com/join/ofcqsnrbeg-cyberspaceline

I think, however, it would be useful for anyone who would like a vertical slider. It's just a little extra confusing to the use the value output of the slider when it's backwards!

### Additional context

Thank you for all your hard work I appreciate it :D

benoitgrelard:
Closes #1635

### DIFF
--- a/.yarn/versions/f8fa5d29.yml
+++ b/.yarn/versions/f8fa5d29.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-slider": patch
+
+declined:
+  - primitives
--- a/packages/react/slider/src/Slider.stories.tsx
+++ b/packages/react/slider/src/Slider.stories.tsx
@@ -71,6 +71,60 @@ export const Vertical = () => (
   </>
 );
 
+export const Inversions = () => (
+  <div style={{ padding: 50 }}>
+    <h1>Horizontal</h1>
+    <h2>LTR (default)</h2>
+    <Slider.Root className={rootClass()} defaultValue={[20]}>
+      <Slider.Track className={trackClass()}>
+        <Slider.Range className={rangeClass()} />
+      </Slider.Track>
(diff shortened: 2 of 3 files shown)


andy-hook (commented):
Code looks great! super clear and easy to follow 🎉 

Only question would be about the stories:
- Thoughts on adding RTL for the vertical orientation on inversions? (if only for completeness when user testing, we might feel this redundant given you already omitted though)
- Can we add the new inversion case/s to the chromatic stories?

andy-hook:
> For vertical, dir shouldn't have any impact, but I guess we might want that to double check?

Yeah that was my thinking but I checked out the branch just to make sure, figured it might be useful to have around for future manual tests but no biggie.

benoitgrelard:
I'm now using the thumb with the value in everywhere, makes it easier to read I feel.
I'll let you review the Chromatic story change in Chromatic, obviously will all look quite different, but I've just inserted the "Insertions" block almost at the end.

andy-hook (approved):
Great stuff

PR #1696, opened 2022-09-30
[Slider] Add `onValueCommit` for discrete value changes

### LINKED ISSUE #903: [Slider] Distinguish onInput from onValueChange
## Feature request
Slider should have two triggers.
One to act onInput and one to act on blur that would fire onValueChange.

See discussion at the bottom of this issue:

https://github.com/radix-ui/primitives/issues/570

### Overview

Like in the demo created by @jjenzz it is typical for sliders (as well as inputs) to denounce the value so that devs can render the real-time value but only trigger an update/network request onFinalChange.

The problem with the demo is that the denounced value would render with a delay(obviously), while the use case is to render real-time values in the UI and send an update (redux/remote/etc) once dragging is finished.

I understand radix provides minimal API so I'm not asking for the value to be denounced, just the events to be separated into two distinct actions.

An example can also be found on the very popular react-range repo.
https://github.com/tajo/react-range

This is for all users as it's a very common use case.

### Additional context

Thank you!

benoitgrelard:
Closes #903

### DIFF
--- a/.yarn/versions/a263f870.yml
+++ b/.yarn/versions/a263f870.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-slider": minor
+
+declined:
+  - primitives
--- a/packages/react/slider/src/Slider.tsx
+++ b/packages/react/slider/src/Slider.tsx
@@ -124,7 +124,6 @@ const Slider = React.forwardRef<SliderElement, SliderProps>(
       const nextValue = values[valueIndexToChangeRef.current];
       const hasChanged = nextValue !== prevValue;
       if (hasChanged) onValueCommit(values);
-      valuesBeforeSlideStartRef.current = values;
     }
 
     function updateValues(value: number, atIndex: number, { commit } = { commit: false }) {
--- a/packages/react/slider/src/Slider.tsx
+++ b/packages/react/slider/src/Slider.tsx
@@ -69,6 +69,7 @@ interface SliderProps
   value?: number[];
   defaultValue?: number[];
   onValueChange?(value: number[]): void;
+  onValueCommit?(value: number[]): void;
   inverted?: boolean;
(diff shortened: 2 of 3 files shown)


.yarn/versions/a263f870.yml
```diff
@@ -0,0 +1,5 @@
+releases:
+  "@radix-ui/react-slider": minor
```
benoitgrelard, line 2: Should have been minor in the previous slider PR really with the new `inverted` prop, but this fixes it anyway.

packages/react/slider/src/Slider.tsx
```diff
@@ -150,13 +162,19 @@ const Slider = React.forwardRef<SliderElement, SliderProps>(
               data-disabled={disabled ? '' : undefined}
               {...sliderProps}
               ref={composedRefs}
+              onPointerDown={composeEventHandlers(sliderProps.onPointerDown, () => {
+                if (!disabled) valuesBeforeSlideStartRef.current = values;
```
andy-hook, line 166: Looking at this made me curious about the behaviour of `onSlideStart`, when does that fire and what's the expectation? I would have expected to see this there but seems this wouldn't be accurate?

benoitgrelard: Yeah that's what I originally tried, but the realised there are cases where it isn't called (basically when pointer down on a thumb):

CleanShot 2022-10-10 at 11 46 28@2x

I believe that is so that we don't see a jump straight as we point down, but rather as we start sliding (because of the thumb offset).

andy-hook: > I believe that is so that we don't see a jump straight as we point down, but rather as we start sliding (because of the thumb offset).

Ah yes, I can see this now 👍

PR #1699, opened 2022-10-04
[Select] Add `disabled` prop to `Root`

itsfaqih:
### Description

This will add `disabled` prop to `<Select.Root>` to disable its trigger and the BubbleSelect when it's inside a form.
It's different from disabling the trigger only because the select value will not be submitted

image

### DIFF
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -60,6 +60,7 @@ type SelectContextValue = {
   onOpenChange(open: boolean): void;
   dir: SelectProps['dir'];
   triggerPointerDownPosRef: React.MutableRefObject<{ x: number; y: number } | null>;
+  disabled?: boolean;
 };
 
 const [SelectProvider, useSelectContext] = createSelectContext<SelectContextValue>(SELECT_NAME);
@@ -84,6 +85,7 @@ interface SelectProps {
   dir?: Direction;
   name?: string;
   autoComplete?: string;
+  disabled?: boolean;
 }
 
 const Select: React.FC<SelectProps> = (props: ScopedProps<SelectProps>) => {
@@ -99,6 +101,7 @@ const Select: React.FC<SelectProps> = (props: ScopedProps<SelectProps>) => {
     dir,
     name,
     autoComplete,
+    disabled,
   } = props;
   const [trigger, setTrigger] = React.useState<SelectTriggerElement | null>(null);
   const [valueNode, setValueNode] = React.useState<SelectValueElement | null>(null);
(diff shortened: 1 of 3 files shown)


packages/react/select/src/Select.tsx
```diff
@@ -196,8 +201,9 @@ interface SelectTriggerProps extends PrimitiveButtonProps {}
 
 const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>(
   (props: ScopedProps<SelectTriggerProps>, forwardedRef) => {
-    const { __scopeSelect, disabled = false, ...triggerProps } = props;
+    const { __scopeSelect, ...triggerProps } = props;
     const context = useSelectContext(TRIGGER_NAME, __scopeSelect);
+    const disabled = props.disabled || context.disabled;
```
benoitgrelard, line 206: We should still pull out the prop otherwise the trigger prop can end up overriding the value from context which wouldn't make sense.

We generally do this like:

```tsx
const { __scopeSelect, disabled: disabledProp = false, ...triggerProps } = props;

…

const disabled = disabledProp || context.disabled;
```

benoitgrelard:
Thanks for the contribution @itsfaqih, that makes a lot of sense regarding form submitting.
Could you run the version check and pick "minor" for select, you may need to run it again to pick "decline" on the top level project too.

packages/react/select/src/Select.tsx
itsfaqih: Thanks for the review, I just did what you suggested

packages/react/select/src/Select.tsx
```diff
@@ -196,8 +201,9 @@ interface SelectTriggerProps extends PrimitiveButtonProps {}
 
 const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>(
   (props: ScopedProps<SelectTriggerProps>, forwardedRef) => {
-    const { __scopeSelect, disabled = false, ...triggerProps } = props;
+    const { __scopeSelect, disabled: disabledProp = false, ...triggerProps } = props;
     const context = useSelectContext(TRIGGER_NAME, __scopeSelect);
+    const disabled = disabledProp || context.disabled;
```
benoitgrelard, line 206: Sorry my bad, just realised we do it slightly differently in `RadioGroup` so let's keep it consistent (we want context to take over):

```suggestion
    const { __scopeSelect, disabled = false, ...triggerProps } = props;
    const context = useSelectContext(TRIGGER_NAME, __scopeSelect);
    const disabled = context.disabled || disabled;
```

itsfaqih: I'm not sure if it's possible to redeclare the `disabled`. So I look up the `RadioGroup`, and it's using `isDisabled` instead of `disabled`. Should I change it to `isDisabled` as well?

itsfaqih: Nvm, I did it anyway 😅

benoitgrelard: Yes sorry that's what I meant to write 😀

benoitgrelard (approved):

Issue #1265, opened 2022-03-23
[Accordion] Initial animation missing in Firefox and Safari


te-online:
### Current Behavior

Accordion elements only animate correctly on the 2nd open in Firefox and Safari. It's a bit subtle to see on the example page, but if you put longer text into the accordion elements it is a lot more noticeable.

https://user-images.githubusercontent.com/4127167/159681799-dbf90a1b-5527-481b-9727-e1d3ee993d3d.mov

### Expected behavior

Accordion elements animate smoothly on 1st open, same as in Chrome.

### Reproducible example

[CodeSandbox (clone of the original example with more text)](https://codesandbox.io/s/flamboyant-rain-mdil3w)

### Suggested solution

I don't really know what's causing this, unfortunately. But I think that maybe the `[data-state="open"]` attribute is applied before computation of the final height of the element is finished.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |    Accordion     |  0.1.6       |
| React            | n/a     |  17.0.2       |
| Browser          |  Firefox       |     98.0.1    |
| Browser          |  Safari       |     15.4  |
| Operating System |    macOS & Window     |   n/a      |

nemdo:
I have the exact same issue.

Issue present on:
- iOS Safari
- iOS Chrome
- iOS Firefox
- iOS Ecosia
- every other iOS browser?
- Desktop Firefox

Not present on:
- Desktop Chrome

This is the second Radix UI component I'm trying (after the Dialog) and both have UX issues that make them unusable in production. Of course, this is open-source and I'm grateful for your incredible work, but do you think that this issue will be patched soon or should I go with another library? Thank you in advance.

andy-hook:
Hey @nemdo it's on our radar to be addressed but I can’t promise when we'll be able to get a patch out.

> both have UX issues that make them unusable in production

What other problems have you encountered? feel free to raise issues you've encountered.

benjaminleonard:
I don't imagine this is a proper fix, but I can get it working in Safari with the following:

```
  const [mount, setMount] = useState(true)

  useEffect(() => {
    setMount(false)
  }, [])
```

and then `<Accordion.Content forceMount={mount}`>

benoitgrelard:
@benjaminleonard, unfortunately probably not as `forceMount` is meant to be used only for JS animation support, and can't be dynamic either.

inje-hottestlab:
[...]

Issue #1040, opened 2021-12-14
[Accordion] Closing an element causes all other elements to close momentarily


axelhzf:
### Current Behavior

https://user-images.githubusercontent.com/175627/145973966-2489181c-1f0f-43b8-9000-de8335ca67d5.mp4

The first time that an element is closed, it causes all the other elements to close momentarily. After the initial close, the rest of the open or close actions work properly.

### Expected behavior

The first time the element is closed should work like the other ones

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/lucid-noyce-59nhm?file=/App.js)

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |  @radix-ui/react-accordion       |  0.1.4        |
| React            | n/a     |  17.0.2        |
| Browser          |   Chrome      |  96.0.4664.93     |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System | macOs        |  12.0.1        |

benoitgrelard:
Hey @axelhzf, thanks or reporting this.
That's a bizarre one haha.
It seems to only happen when `value` is controlled AND all items are expanded 🤔

anudeepreddy:
Hi, 

I would like to work on this issue. I have observed that this behavior is observed because of `animation: 0s ease 0s 1 normal none running none; ` being injected in styles. The first render has these styles for AccordionContent but consecutive renders doesn't have that style. Can someone point me to how that particular style is being injected?

jjenzz:
@anudeepreddy Oh gosh I'm so sorry! We were already working on it and have just merged a fix. Apologies for jumping in there 🙈

axelhzf:
Thank you for the fix @jjenzz

PR #1710, opened 2022-10-10
[Collapsible][Accordion] Fix initial animation playback in Firefox and Safari

andy-hook:
closes https://github.com/radix-ui/primitives/issues/1265

After some exploration it dawned that perhaps these browsers need a little bump to replay the animation from scratch for the initial size retrieval.

I'll try and validate this behaviour in a smaller example but this does fix the issue.

@benoitgrelard Do you remember why we moved away from the [clearing the shorthand](https://github.com/radix-ui/primitives/pull/1044/files) in the first place? I remember that had problems but i'm questioning if removing the name property alone is all we need.

### DIFF
--- a/.yarn/versions/4f235e95.yml
+++ b/.yarn/versions/4f235e95.yml
@@ -1,0 +1,6 @@
+releases:
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-collapsible": patch
+
+declined:
+  - primitives
--- a/packages/react/collapsible/src/Collapsible.tsx
+++ b/packages/react/collapsible/src/Collapsible.tsx
@@ -176,11 +176,13 @@ const CollapsibleContentImpl = React.forwardRef<
     if (node) {
       originalStylesRef.current = originalStylesRef.current || {
         transitionDuration: node.style.transitionDuration,
+        animationName: node.style.animationName,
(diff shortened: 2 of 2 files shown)


benoitgrelard:
> @benoitgrelard Do you remember why we moved away from the [clearing the shorthand](https://github.com/radix-ui/primitives/pull/1044/files) in the first place? I remember that had problems but i'm questioning if removing the name property alone is all we need.

It seems because we had the issue in #1040 so maybe run that scenario and see if your new change creates that regression?

benoitgrelard (approved):
Approved pending check in other comment 👍

andy-hook:
> maybe run that scenario and see if your new change creates that regression

Looks good.

> i'm questioning if removing the name property alone is all we need.

Tested in all browsers + IOS and still looks good, let me know if you're good with this additional change.

benoitgrelard (approved):
I'm wondering if we could add some cypress tests or chromatic tests to test all these behaviours, although not sure how feasible that is given the tight criterias with animation.

andy-hook:
> I'm wondering if we could add some cypress tests or chromatic tests to test all these behaviours, although not sure how feasible that is given the tight criterias with animation.

This would be ideal though feasibility is definitely an issue given the real time nature. Browser differences make this even harder 😞

PR #1716, opened 2022-10-13
[NavigationMenu] Add `delayDuration` and `skipDelayDuration` to `Root`

andy-hook:
closes https://github.com/radix-ui/primitives/issues/1457
supersedes https://github.com/radix-ui/primitives/pull/1478

### DIFF
--- a/packages/react/navigation-menu/src/NavigationMenu.tsx
+++ b/packages/react/navigation-menu/src/NavigationMenu.tsx
@@ -61,8 +61,10 @@ type NavigationMenuContextValue = {
   onViewportChange(viewport: NavigationMenuViewportElement | null): void;
   onViewportContentChange(contentValue: string, contentData: ContentData): void;
   onViewportContentRemove(contentValue: string): void;
-  onItemOver(itemValue: string): void;
-  onItemLeave(): void;
+  onTriggerEnter(itemValue: string): void;
+  onTriggerLeave(): void;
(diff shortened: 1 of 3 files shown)


packages/react/navigation-menu/src/NavigationMenu.tsx
```diff
@@ -79,35 +81,119 @@ type PrimitiveNavProps = Radix.ComponentPropsWithoutRef<typeof Primitive.nav>;
...
 const NavigationMenu = React.forwardRef<NavigationMenuElement, NavigationMenuProps>(
   (props: ScopedProps<NavigationMenuProps>, forwardedRef) => {
     const {
       __scopeNavigationMenu,
-      value,
+      value: valueProp,
       onValueChange,
       defaultValue,
+      delayDuration = 0,
+      skipDelayDuration = 0,
```
andy-hook, line 109: Avoiding breaking change by not disturbing the current behaviour

benoitgrelard: I'm not sure it would be considered a breaking change, I'd rather we start with good defaults and put a note in the changelog.

andy-hook: That's true, especially if these defaults are blanket improvements.

benoitgrelard (approved):

PR #1730, opened 2022-10-19
Resolve dev console `scrollbar size calculation` warning

andy-hook:
See https://github.com/radix-ui/primitives/discussions/1594#discussioncomment-3913318 and https://github.com/theKashey/react-remove-scroll-bar/issues/35#issuecomment-1276809669

### DIFF
--- a/.yarn/versions/be773afe.yml
+++ b/.yarn/versions/be773afe.yml
@@ -1,0 +1,2 @@
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


benoitgrelard:
This won't be propagated to our users right?

andy-hook:
> This won't be propagated to our users right?

No, they need to refresh their lock afaik. It does validate the change works on our side though.

Issue #983, opened 2021-11-18
[Checkbox] Reset checkbox 'checked' state when form is reset using `form.reset()`


jjenzz:
### Discussed in https://github.com/radix-ui/primitives/discussions/982

<div type='discussions-op-text'>

<sup>Originally posted by **rkumorek** November 18, 2021</sup>
Hey,

I'm wondering if it would be possible to reset Radix Checkbox checked state on `form.reset()` action, exactly as it works with plain <input type="checkbox"/> elements?

I am not sure if it's a limitation due to how React works or it's a bug or a feature request issue so apologies again if it's in wrong place.

You can see current behaviour in below code sandbox:
https://codesandbox.io/s/heuristic-orla-cnd7c

Thanks,
Rafal</div>

PR #1733, opened 2022-10-21
[Checkbox] Reset checkbox checked state when form is reset

raulburigo:
Implements #983

### DIFF
--- a/.yarn/versions/6ad01e88.yml
+++ b/.yarn/versions/6ad01e88.yml
@@ -1,0 +1,3 @@
+releases:
+  "@radix-ui/react-checkbox": minor
+  primitives: minor
--- a/packages/react/checkbox/src/Checkbox.tsx
+++ b/packages/react/checkbox/src/Checkbox.tsx
@@ -57,11 +57,25 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
     const hasConsumerStoppedPropagationRef = React.useRef(false);
     // We set this to true by default so that events bubble to forms without JS (SSR)
     const isFormControl = button ? Boolean(button.closest('form')) : true;
+
+    const initialCheckState = React.useRef(defaultChecked ? true : checkedProp);
+
     const [checked = false, setChecked] = useControllableState({
       prop: checkedProp,
       defaultProp: defaultChecked,
       onChange: onCheckedChange,
     });
+
+    const resetCheck = React.useCallback(() => {
+      setChecked(initialCheckState.current);
+    }, [setChecked]);
+
+    React.useEffect(() => {
+      if (!button) return;
+      const form = button.closest('form');
+      form?.addEventListener('reset', resetCheck);
+      return () => form?.removeEventListener('reset', resetCheck);
+    }, [button, resetCheck]);
 
     return (
       <CheckboxProvider scope={__scopeCheckbox} state={checked} disabled={disabled}>
--- a/packages/react/checkbox/src/Checkbox.stories.tsx
+++ b/packages/react/checkbox/src/Checkbox.stories.tsx
@@ -166,6 +166,7 @@ export const WithinForm = () => {
       <br />
       <br />
 
+      <button type="reset">Reset</button>
       <button>Submit</button>
     </form>
   );


packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -57,12 +57,26 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
     const hasConsumerStoppedPropagationRef = React.useRef(false);
     // We set this to true by default so that events bubble to forms without JS (SSR)
     const isFormControl = button ? Boolean(button.closest('form')) : true;
+
+    const initialCheckState = React.useRef(defaultChecked ? true : checkedProp);
```
benoitgrelard, line 61: It might make more sense to put that below `useControllableState` and use `checked` in the ref instead as this already should take care of the `defaultChecked` vs `checkedProp` case.

Also, then we can group it with `resetCheck` below.

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -57,12 +57,26 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
...
+    const resetCheck = React.useCallback(() => {
+      setChecked(initialCheckState.current);
+    }, [setChecked]);
+
+    React.useEffect(() => {
+      if (!button) return;
+      const form = button.closest('form');
+      form?.addEventListener('reset', resetCheck);
+      return () => form?.removeEventListener('reset', resetCheck);
+    }, [button, resetCheck]);
```
benoitgrelard, line 78: We can take advantage of the auto-assigned `form` property on the control.

```suggestion
    React.useEffect(() => {
      const form = button?.form;
      if (form) {
        form.addEventListener('reset', resetCheck);
        return () => form.removeEventListener('reset', resetCheck);
      }
    }, [button, resetCheck]);
```

benoitgrelard (commented):
Thanks for your contribution @raulburigo!

packages/react/checkbox/src/Checkbox.tsx
raulburigo: totally!

.yarn/versions/6ad01e88.yml
```diff
@@ -0,0 +1,3 @@
+releases:
+  "@radix-ui/react-checkbox": minor
+  primitives: minor
```
benoitgrelard, line 3: Could you re-run the version check and `primitives` should be declined.
Also, I think it should just a be a patch, not a minor.

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -57,12 +57,26 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
     const hasConsumerStoppedPropagationRef = React.useRef(false);
     // We set this to true by default so that events bubble to forms without JS (SSR)
     const isFormControl = button ? Boolean(button.closest('form')) : true;
+
```
benoitgrelard, line 60: Can we remove this empty line?

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -57,12 +57,26 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
...
+    }, [setChecked]);
+
+    React.useEffect(() => {
+      const form = button?.form;
+      if (form) {
+        form.addEventListener('reset', resetCheck);
+        return () => form.removeEventListener('reset', resetCheck);
+      }
+    }, [button, resetCheck]);
+
```
benoitgrelard, line 79:
```suggestion
    const initialCheckedStateRef = React.useRef(checked);
    React.useEffect(() => {
      const form = button?.form;
      if (form) {
        const reset = () => setChecked(initialCheckedStateRef.current);
        form.addEventListener('reset', reset);
        return () => form.removeEventListener('reset', reset);
      }
    }, [button, setChecked]);

```

benoitgrelard: Less pieces this way and easier to move around/maintain.

benoitgrelard (commented):
This is great @raulburigo, only a few small tweaks and we can get this merged!

packages/react/checkbox/src/Checkbox.tsx
```diff
@@ -62,6 +62,16 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
       defaultProp: defaultChecked,
       onChange: onCheckedChange,
     });
+    const initialCheckedStateRef = React.useRef(checked);
+
+    React.useEffect(() => {
```
benoitgrelard, line 67: nit:

```suggestion

    const initialCheckedStateRef = React.useRef(checked);
    React.useEffect(() => {
```

benoitgrelard (approved):
🎉 Congrats on landing your first PR!

matthova:
@raulburigo with React v19, submitting a form with a defaultValue and then on reset setting to the initial value causes some weird race-condition behavior

See the related discussion here: https://github.com/facebook/react/issues/29034

I'm proposing a potential backwards-compatible next-move: Introduce an opt-in new prop, like `dontResetStateOnSubmit` to not reset the checkbox to its initial value on form reset

wdyt?

PR #1746, opened 2022-10-29
[ContextMenu] Add "disabled" prop to Context Menu trigger

### LINKED ISSUE #1745: Add `disabled` to `ContextMenu`
## Feature request

It would be nice if `ContextMenu` could be disabled. We have a situation in which we conditionally want to disable the context menu - we can do that by conditionally rendering the `ContextMenu` wrapper but that remounts the wrapped tree.

paperstick:
### Description

This PR adds `disabled` prop to `<ContextMenu />` trigger 

---
Closes #1745

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -133,22 +133,33 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
                   event.preventDefault();
                 })
           }
-          onPointerDown={composeEventHandlers(
-            props.onPointerDown,
-            whenTouchOrPen((event) => {
-              // clear the long press here in case there's multiple touch points
-              clearLongPress();
-              // if trigger is disabled, enable the native Context Menu
-              if (disabled) return;
-              longPressTimerRef.current = window.setTimeout(() => handleOpen(event), 700);
-            })
-          )}
-          onPointerMove={composeEventHandlers(props.onPointerMove, whenTouchOrPen(clearLongPress))}
-          onPointerCancel={composeEventHandlers(
-            props.onPointerCancel,
-            whenTouchOrPen(clearLongPress)
-          )}
-          onPointerUp={composeEventHandlers(props.onPointerUp, whenTouchOrPen(clearLongPress))}
+          onPointerDown={
+            disabled
+              ? props.onPointerDown
+              : composeEventHandlers(
+                  props.onPointerDown,
+                  whenTouchOrPen((event) => {
+                    // clear the long press here in case there's multiple touch points
+                    clearLongPress();
+                    longPressTimerRef.current = window.setTimeout(() => handleOpen(event), 700);
+                  })
+                )
+          }
+          onPointerMove={
+            disabled
+              ? props.onPointerMove
+              : composeEventHandlers(props.onPointerMove, whenTouchOrPen(clearLongPress))
+          }
+          onPointerCancel={
+            disabled
+              ? props.onPointerCancel
+              : composeEventHandlers(props.onPointerCancel, whenTouchOrPen(clearLongPress))
+          }
+          onPointerUp={
+            disabled
+              ? props.onPointerUp
+              : composeEventHandlers(props.onPointerUp, whenTouchOrPen(clearLongPress))
+          }
         />
       </>
     );
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -104,7 +104,6 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
       []
     );
     const handleOpen = (event: React.MouseEvent | React.PointerEvent) => {
-      if (disabled) return;
       pointRef.current = { x: event.clientX, y: event.clientY };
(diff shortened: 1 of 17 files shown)


packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -102,6 +104,7 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
       []
     );
     const handleOpen = (event: React.MouseEvent | React.PointerEvent) => {
+      if (disabled) return;
```
Andarist, line 107: I think that this should be moved outside of this handler to event handlers. The problem with the current implementation is that it still prevents the default in the `onContextMenu` handler and thus it prevents the native context menu to open.

I've prepared a PR with this suggestion implemented: https://github.com/paperstick/primitives/pull/1

paperstick: Thanks, @Andarist, a good suggestion! I have just cleaned up the code a bit and merged the PR, thanks for the contibution!

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -113,11 +115,14 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
         <MenuPrimitive.Anchor {...menuScope} virtualRef={virtualRef} />
         <Primitive.span
           data-state={context.open ? 'open' : 'closed'}
+          data-disabled={disabled ? '' : undefined}
```
benoitgrelard, line 118: 👍

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -113,11 +115,14 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
...
           {...triggerProps}
           ref={forwardedRef}
           // prevent iOS context menu from appearing
           style={{ WebkitTouchCallout: 'none', ...props.style }}
           onContextMenu={composeEventHandlers(props.onContextMenu, (event) => {
+            // if trigger is disabled, enable the native Context Menu
+            if (disabled) return;
```
benoitgrelard, line 125: Instead of those internal checks, could we conditionally pass the handlers like so:

```tsx
onContextMenu={disabled ? undefined : composeEventHandlers(props.onContextMenu, (event) => {
  // clearing the long press here because some platforms already support
  // long press to trigger a `contextmenu` event
  clearLongPress();
  handleOpen(event);
  event.preventDefault();
})}
```

We can do that for the 5 handlers to ensure there are no side-effects of disabling.

benoitgrelard (commented):
Thanks for the contribution @paperstick!

packages/react/context-menu/src/ContextMenu.tsx
Andarist: This version would be incorrect - you'd have to only disable the "own" handler - without disabling the `props.onContextMenu`.

benoitgrelard: oops, yep you're right, it needs to be:

```tsx
onContextMenu={disabled ? props.onContextMenu : …}
```

Andarist: On top of that - I would add this:
```js
useEffect(() => {
  if (disabled) clearLongPress()
}, [disabled])
```

benoitgrelard: Yep, that's good for completeness too, like we do on unmount 👍

Andarist: https://github.com/paperstick/primitives/pull/2/files

gunnartorfis:
Would it be possible to add disabled conditions based on PointerDown's type/button?
In our case, we want to disable the context menu entirely on touch devices and only allow it on right click.

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -107,28 +109,37 @@ const ContextMenuTrigger = React.forwardRef<ContextMenuTriggerElement, ContextMe
...
+            disabled
+              ? props.onContextMenu
+              : composeEventHandlers(props.onContextMenu, (event) => {
+                  // clearing the long press here because some platforms already support
+                  // long press to trigger a `contextmenu` event
+                  clearLongPress();
+                  handleOpen(event);
+                  event.preventDefault();
+                })
+          }
```
benoitgrelard, line 135: We'll need this `disabled` check for all subsequent handlers too.

Andarist:
@benoitgrelard is there anything that we could do to move this forward? :)

benoitgrelard:
Hey @Andarist, it seems the original contributor hasn't followed up, so if you want to address the [remaining feedback](https://github.com/radix-ui/primitives/pull/1746#discussion_r1010369370) we can move it forward.

paperstick:
Hi, @Andarist  and @benoitgrelard! I have already made needed changes in the following [commit](https://github.com/radix-ui/primitives/pull/1746/commits/a8b26b6d4e8447e58cdc3acab1d5b28e1f3e5e53). Let me know if we need to add anything else, I will check

benoitgrelard:
Oh sorry my bad!

benoitgrelard (approved):

andy-hook (approved):

Issue #1751, opened 2022-11-02
[DropdownMenu] Within Submenu, escape key exits fullscreen on Firefox & Safari


cpmsmith:
### Current Behavior

* Enter fullscreen mode in Firefox or Safari
* Open dropdown menu
* Open sub-menu of that dropdown menu
* Press escape
* Browser exits fullscreen

### Expected behavior

The menu should close, and the browser should not exit fullscreen.

### Reproducible example

Reproducible on the [basic demo](https://www.radix-ui.com/docs/primitives/components/dropdown-menu) on Radix's site

### Suggested solution

Well, I said in #1423 I wasn't sure about the semantics of a `DismissableLayer` not having an `onDismiss`, and here we are! I made the intentional decision not to call `preventDefault` on the keydown event in that case, and we now seem to be running into that very situation—[`MenuSubContent` provides an `onEscapeKeyDown` prop, but not an `onDismiss`](https://github.com/radix-ui/primitives/blob/e861c16f20fe8e2e440983355edf2ec78cc2a51c/packages/react/menu/src/Menu.tsx#L1167-L1203).

There are a number of places another `preventDefault` could be put, it's just a question of what's most correct. (I'm leaning toward adding it to the `onEscapeKeyDown` prop [here](https://github.com/radix-ui/primitives/blob/e861c16f20fe8e2e440983355edf2ec78cc2a51c/packages/react/menu/src/Menu.tsx#L1190))

### Additional context

This is a partial regression of #1420 introduced, I think, when `Submenu`s became their own component.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | `react-dropdown-menu` | 2.0.1 |
| React            | n/a     | 18.3 |
| Browser          | Firefox | 106.0.3 |
| Operating System | macOS | 12.3.1 |

PR #1752, opened 2022-11-02
[DropdownMenu][ContextMenu] Don't exit fullscreen when esc-ing out of a submenu

cpmsmith:
Fixes radix-ui/primitives#1751

### Description

I kept looking at it, and I think this is probably the right place for this `preventDefault`? It even seems like another option for `composeEventHandlers` that would automatically `preventDefault` might make sense. I don't wanna tell you how to design your library, though.

### DIFF
--- a/.yarn/versions/c21771b2.yml
+++ b/.yarn/versions/c21771b2.yml
@@ -1,0 +1,7 @@
+releases:
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+
+declined:
+  - primitives
--- a/.yarn/versions/c21771b2.yml
+++ b/.yarn/versions/c21771b2.yml
@@ -2,6 +2,7 @@ releases:
   "@radix-ui/react-context-menu": patch
   "@radix-ui/react-dropdown-menu": patch
   "@radix-ui/react-menu": patch
+  "@radix-ui/react-menubar": patch
 
 declined:
   - primitives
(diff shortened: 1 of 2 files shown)


packages/react/menu/src/Menu.tsx
```diff
@@ -1187,7 +1187,10 @@ const MenuSubContent = React.forwardRef<MenuSubContentElement, MenuSubContentPro
                 // on pointer interaction.
                 if (event.target !== subContext.trigger) context.onOpenChange(false);
               })}
-              onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, rootContext.onClose)}
+              onEscapeKeyDown={composeEventHandlers(props.onEscapeKeyDown, (event) => {
+                event.preventDefault();
+                rootContext.onClose();
```
benoitgrelard, line 1192:
```suggestion
                rootContext.onClose();
                // ensure pressing escape in submenu doesn't escape full screen mode
                event.preventDefault();
```

.yarn/versions/c21771b2.yml
```diff
@@ -0,0 +1,7 @@
+releases:
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+
+declined:
```
benoitgrelard, line 6:
```suggestion
  "@radix-ui/react-dropdown-menu": patch
  "@radix-ui/react-menu": patch
  "@radix-ui/react-menubar": patch

declined:
```

benoitgrelard (approved):

PR #1753, opened 2022-11-03
[Select] fix invalid pointerId in Firefox Cypress test

myrjola:
### Description

Fixes https://github.com/cypress-io/cypress/issues/24459

> Looks like the invalid pointer ID is thrown If the pointerId does not match any the active pointers and so returns void and throws a DOMException with the name InvalidPointerId . I believe this is happening because the captured pointer is being automatically released by firefox a fraction earlier, so by the time the pointerdown listener is called, the component no longer has a valid pointerID as it has implicitly been released thereby throwing an error.
>
> According to https://www.w3.org/TR/pointerevents3/#implicit-pointer-capture "the [hasPointerCapture](https://www.w3.org/TR/pointerevents3/#dom-element-haspointercapture) API may be used (eg. within any [pointerdown](https://www.w3.org/TR/pointerevents3/#dfn-pointerdown) listener) to determine if `setPointerCapture was called on the target element"

To verify fix, run `npx cypress run --headed --browser firefox --spec cypress/integration/Select.spec.ts`

image

### Known issues

Headless Firefox does not work. See https://github.com/cypress-io/cypress/issues/24459#issuecomment-1301673757

### DIFF
--- a/.yarn/versions/82e6e50f.yml
+++ b/.yarn/versions/82e6e50f.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-select": patch
+
+declined:
+  - primitives
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -257,7 +257,10 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
         onPointerDown={composeEventHandlers(triggerProps.onPointerDown, (event) => {
           // prevent implicit pointer capture
           // https://www.w3.org/TR/pointerevents3/#implicit-pointer-capture
-          (event.target as HTMLElement).releasePointerCapture(event.pointerId);
+          const target = event.target as HTMLElement;
+          if (target.hasPointerCapture(event.pointerId)) {
+            target.releasePointerCapture(event.pointerId);
+          }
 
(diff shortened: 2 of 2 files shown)


benoitgrelard:
Hey @myrjola,

Is this a Firefox issue only?
How did you stumble on the symptom?
Why are we not seeing this error when using Firefox manually?

myrjola:
To my understanding, this only affects Radix UI Select inside Cypress tests running on Firefox. It's likely caused by differences in how Cypress' `.click()` does not work as a real browser click. Essentially, the fix is defensive programming by adding `hasPointerCapture` check before the `releasePointerCapture` that can fail in some circumstances. I'd like this fix merged since my team has Firefox tests in our CI pipeline that won't pass if we replace select component with Radix UI Select.

I didn't encounter any regressions by introducing this fix, but do you see any risks in merging it?

benoitgrelard:
No adding it should be fine, it's rather that I wanted to understand why we didn't see it in Firefox itself as a user but I guess that makes sense. I'm not a fan of having to add defensive code to please tools, but in this case I think that's a fair ask.

benoitgrelard (approved):

andy-hook:
Should we be standardising on this across our pointer capture code?

benoitgrelard:
> Should we be standardising on this across our pointer capture code?

Yes, I'll update other places.

PR #1754, opened 2022-11-03
Standardize checking for pointer capture before releasing

benoitgrelard:
As discussed in https://github.com/radix-ui/primitives/pull/1753#issuecomment-1302047373


andy-hook (approved):
Interesting that `Slider` was already covered, I wonder if that was for a specific reason at that time.

benoitgrelard:
haha yeah not sure!

PR #1771, opened 2022-11-08
[Select] add scroll button onClick for better touch experience

myrjola:
### Description

These changes introduce an onClick event handler for the Select scroll buttons for better user experience on touch devices. This issue was highlighted by our QA engineer that scrolling the Select content with swipe gestures might not be obvious to some users. If they tried to click on the scroll buttons, nothing happened. After this change, clicking on the scroll buttons scrolls the viewport.

This change only enhances the touch experience. To test, 

1. Navigate to the Select Within Dialog story http://localhost:9009/?path=/story/components-select--within-dialog
2. Open browser devtools and enable mobile mode
3. Open the select and click on the scroll buttons

https://user-images.githubusercontent.com/36122/200658548-28ee7a1a-cb15-42ce-96cc-9eacdc11aaae.mov

### DIFF
--- a/.yarn/versions/03e4cd61.yml
+++ b/.yarn/versions/03e4cd61.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-select": minor
+
+declined:
+  - primitives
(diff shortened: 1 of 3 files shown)


benoitgrelard (approved):
Thanks for getting this going @myrjola!

You are right, on touch screen if a user wasn't doing the interaction in one whole touch (touch down, touchmove over button) then it would be a bit fiddly to kickstart the scroll.
It was possible but akward (ie, touch down and then move over the button).

I have tweaked the implementation a little to reflect the new functionality.

🙏

PR #1796, opened 2022-11-18
[Dialog] Allow first focus on anchors

paul-sachs:
### Description

Introducing a flag to allow first focus on anchors. This can be extremely useful for dialogs that have navigation elements. Think a drawer with nav elements.

### DIFF
--- a/.yarn/versions/694d49d6.yml
+++ b/.yarn/versions/694d49d6.yml
@@ -1,0 +1,10 @@
+releases:
+  "@radix-ui/react-dialog": minor
+  "@radix-ui/react-focus-scope": minor
+  primitives: minor
+
+declined:
+  - "@radix-ui/react-alert-dialog"
+  - "@radix-ui/react-menu"
+  - "@radix-ui/react-popover"
+  - "@radix-ui/react-select"
--- a/packages/react/dialog/src/Dialog.stories.tsx
+++ b/packages/react/dialog/src/Dialog.stories.tsx
@@ -92,6 +92,25 @@ export const FocusTrap = () => (
   </>
 );
 
+export const AutoFocusOnAnchor = () => (
+  <>
+    <Dialog.Root>
+      <Dialog.Trigger>open</Dialog.Trigger>
+      <Dialog.Portal>
+        <Dialog.Overlay className={overlayClass()} />
+        <Dialog.Content className={contentDefaultClass()}>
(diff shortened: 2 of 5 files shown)


benoitgrelard:
Interesting, this is something we intentionally added in #925 although I did express some doubts at the time.
@paul-sachs Could you please show us what your use-case currently looks like?

@andy-hook, either way we go, I don't think I'm in favour of adding to the API surface with a new boolean prop, so to me the choice is between:

- we don't do anything and keep things as is, if a use-case arises where a user would like to focus a link on open, they do it manually with `onOpenAutoFocus`
- we revert the change from #925 and links are part of the focusable elements on mount, if it isn't the behaviour the developer wants in their specific case, they manually change that with `onOpenAutoFocus` (what I originally suggested)

paul-sachs:
Hi @benoitgrelard, I feel like option 2 makes more sense to me since i was very surprised when this happened originally.

My use case is actually pretty normal, I have a mobile menu rendered as a dialog with navigation elements as a list of options (and they are all links - naturally). Currently, the dialog opens and all it can focus on is the close button which is really not ideal.

andy-hook:
@paul-sachs is this your primary use case or did you have a wider need in your project? only this looks easily achievable by manually focusing your element via `onOpenAutoFocus `

> i was very surprised when this happened originally.

That the first anchor wasn't focused? or that in your case the close button was focused?

Thanks for raising this though! just want to understand the needs and expectations better.

@benoitgrelard Currently I feel reluctant to make this the default. AFAIK we don't get many requests for this and behaviour wise it's a breaking change right? agree with you on the API point btw.

paul-sachs:
This is my only use case at this moment, however we are trying to use the router more and more as the source of truth for state so I expect more links to drive state in the future. Manually focusing each of these cases feels clunky at best, especially when it is unclear as to why we're skipping over anchor tags at all.

paul-sachs:
> That the first anchor wasn't focused? or that in your case the close button was focused?

The fact that the first anchor wasn't focused. If i replaced that with a button, it would be focused but in this case it's a nav link so it doesn't make sense to use a button.

benoitgrelard:
Thanks again for raising this @paul-sachs, we're always happy to reconsider past decisions.

However in this case, we are going to keep things as they are currently as this is very rare feedback and there is a simple way to achieve what you want regardless.

PR #1810, opened 2022-12-02
[Select] Add aria-invalid when BubbleSelect is invalid

itsfaqih:
### Description

This will allow us to test whether the select component is invalid or not.
For example:
```js
expect(dialog.getByRole("combobox", { name: "Select Fruit" })).toBeInvalid()
```

### DIFF
--- a/.yarn/versions/2bb20bec.yml
+++ b/.yarn/versions/2bb20bec.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-select": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


chaance:
I think validity state is best left to consumers. Different users may have different events that trigger validity errors which could conflict with `onInvalid`, and this approach is simple enough to add in userland (just lift the validity state to your component and pass `aria-invalid` directly).

We have useful abstractions for dealing with validity in our [Form components](https://www.radix-ui.com/primitives/docs/components/form), which you should be able to compose over Select if you'd like.

Issue #1020, opened 2021-12-09
[Presence] `onAnimationEnd` doesn't fire on close


dmwin72015:
## Question
I want to execute some functions at the end of the animation.  How do i know the animation is over!

Is there something like `afterClose`  props
thanks!!!

jjenzz:
Hi @dmwin72015 👋

You should be able to use [any of React's supported events](https://reactjs.org/docs/events.html#supported-events) on our parts but it looks like `onAnimationEnd` isn't firing on close 🤔

I'll convert this to a bug for us to investigate. Thanks for raising!

dmwin72015:
> You should be able to use [any of React's supported events](https://reactjs.org/docs/events.html#supported-events) on our parts but it looks like `onAnimationEnd` [isn't firing on close](https://codesandbox.io/s/staging-smoke-ssxiq?file=/App.js) 🤔
> 
> Seems to be a bug in `Presence` somewhere. Thanks for raising! I'll rename the ticket and we'll take a look when we can.

thanks!

gadishimwe:
For someone looking for a workaround while we're waiting for https://github.com/radix-ui/primitives/pull/1849 to be merged, you can use `onAnimationEndCapture` instead. It's working fine for me.

PR #1849, opened 2022-12-19
[Presence] Ensure `onAnimationEnd` when closing

YarnSphere:
Make sure that `onAnimationEnd` is fired for consumers, after the closing animation of a component.

Fixes: #1020

_I hit #1020 when implementing an imperative API over toasts, where I want to remove a toast from the list of open toasts after it finishes its "close" animation. I expect wanting to know when an element finishes closing to be a relatively common need._

### DIFF
--- a/.yarn/versions/8a3bd6ee.yml
+++ b/.yarn/versions/8a3bd6ee.yml
@@ -1,0 +1,22 @@
+releases:
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-collapsible": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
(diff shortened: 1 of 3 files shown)


YarnSphere:
A completely different approach to solving #1020 would change the Presence component to receive an extra property `onAnimationEnd` (the same event handler that would otherwise have been passed to the component we are animating) and manually calling `onAnimationEnd` before we call `flushSync`.

This would require going through every usage of `Presence` in Radix and pass it the `onAnimationEnd` handler that was being passed to the animating component.

The biggest downside of this approach (besides the fact that we're complicating `Presence`) is that we'd have mismatched signatures for `onAnimationEnd`, since we're dealing with "real" events here, rather than synthetic ones, so we'd likely have to deal with that discrepancy somehow.

What are your thoughts? Do you have any other ideas on how to solve this? I'm leaving this PR here with this possible solution, but am happy to change the approach if you find it inadequate.

chaance (approved):

Issue #881, opened 2021-09-28
[Accordion] Add `orientation` prop


jjenzz:
The `Accordion` currently assumes vertical orientation only (e.g. keyboard).

PR #1850, opened 2022-12-19
[Accordion] Add `orientation` prop to support horizontal orientation

dliu-workos:
### Description

Resolves https://github.com/radix-ui/primitives/issues/881

### DIFF
--- a/.yarn/versions/fa513961.yml
+++ b/.yarn/versions/fa513961.yml
@@ -1,0 +1,3 @@
+releases:
+  "@radix-ui/react-accordion": minor
+  primitives: patch
--- a/.yarn/versions/fa513961.yml
+++ b/.yarn/versions/fa513961.yml
@@ -1,3 +1,6 @@
 releases:
   "@radix-ui/react-accordion": minor
-  primitives: patch
+
+declined:
+  - primitives
+
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -237,20 +237,14 @@ interface AccordionImplProps extends PrimitiveDivProps {
    */
   orientation?: React.AriaAttributes['aria-orientation'];
   /**
-   * The language read direction. Defaults to RTL.
+   * The language read direction.
    */
   dir?: Direction;
 }
 
 const AccordionImpl = React.forwardRef<AccordionImplElement, AccordionImplProps>(
   (props: ScopedProps<AccordionImplProps>, forwardedRef) => {
-    const {
-      __scopeAccordion,
-      disabled,
-      dir = 'ltr',
-      orientation = 'vertical',
-      ...accordionProps
-    } = props;
+    const { __scopeAccordion, disabled, dir, orientation = 'vertical', ...accordionProps } = props;
     const accordionRef = React.useRef<AccordionImplElement>(null);
     const composedRefs = useComposedRefs(accordionRef, forwardedRef);
     const getItems = useCollection(__scopeAccordion);
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -17,7 +17,7 @@ import type { Scope } from '@radix-ui/react-context';
  * -----------------------------------------------------------------------------------------------*/
 
 const ACCORDION_NAME = 'Accordion';
-const ACCORDION_KEYS = ['Home', 'End', 'ArrowDown', 'ArrowUp'];
+const ACCORDION_KEYS = ['Home', 'End', 'ArrowDown', 'ArrowUp', 'ArrowLeft', 'ArrowRight'];
 
 const [Collection, useCollection, createCollectionScope] =
   createCollection<AccordionTriggerElement>(ACCORDION_NAME);
@@ -212,6 +212,7 @@ const AccordionImplMultiple = React.forwardRef<
 
 type AccordionImplContextValue = {
   disabled?: boolean;
+  orientation: AccordionImplProps['orientation'];
 };
 
 const [AccordionImplProvider, useAccordionContext] =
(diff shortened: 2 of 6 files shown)


packages/react/accordion/src/Accordion.tsx
```diff
@@ -255,13 +261,30 @@ const AccordionImpl = React.forwardRef<AccordionImplElement, AccordionImplProps>
...
+          if (orientation === 'vertical') {
+            nextIndex = triggerIndex + 1;
+          }
+          break;
+        case 'ArrowLeft':
+          if (orientation === 'horizontal') {
+            nextIndex = triggerIndex - 1;
+            if (nextIndex < 0) {
+              nextIndex = triggerCount - 1;
+            }
```
benoitgrelard, line 279: Now that there's a couple places with this adjustement, it might make more sense to let the index run out of bounds negatively (as we do positively already) and cycle it afterwards where we are already clamping on line 292.

benoitgrelard (commented):
This is looking great @dliu-workos!
We may need to add `aria-orientation` somewhere too (root?)

packages/react/accordion/src/Accordion.tsx
andy-hook: Worth noting that the original `index` based implementation breaks when hiding triggers with css.

Not sure if it's worth doing now but should we eventually be using a [`focusFirst` approach](https://github.com/radix-ui/primitives/blob/81b25f4b40c54f72aeb106ca0e64e1e09655153e/packages/react/navigation-menu/src/NavigationMenu.tsx#L1125-L1137)?

Do we also need to support RTL with this change?

benoitgrelard: Yeah, focusFirst probably in the end, although I think we can skip for now here (let us know later if you want to learn a bit more about what we're talking about here.)

RTL though yes that would make sense given we're now talking about horizontal orientation.

@dliu-workos, you'll need to use `useDirecttion` to get access to the normalized `dir` passed from context and adjust keyboard handling accordingly. We do this already in a bunch of components.

packages/react/accordion/src/Accordion.tsx
```diff
@@ -11,6 +11,9 @@ import { useId } from '@radix-ui/react-id';
 
 import type * as Radix from '@radix-ui/react-primitive';
 import type { Scope } from '@radix-ui/react-context';
+import { useDirection } from '@radix-ui/react-direction';
```
benoitgrelard, line 14: We'll need to add this package as a dependency in accordion now and run `yarn` so it gets added to the lockfile.

packages/react/accordion/src/Accordion.tsx
```diff
@@ -228,18 +232,30 @@ interface AccordionImplProps extends PrimitiveDivProps {
...
+  dir?: Direction;
 }
 
 const AccordionImpl = React.forwardRef<AccordionImplElement, AccordionImplProps>(
   (props: ScopedProps<AccordionImplProps>, forwardedRef) => {
-    const { __scopeAccordion, disabled, orientation = 'vertical', ...accordionProps } = props;
+    const {
+      __scopeAccordion,
+      disabled,
+      dir = 'ltr',
```
benoitgrelard, line 250: I think we normally don't set a default for `dir` and rely in implicit default.

packages/react/accordion/src/Accordion.tsx
```diff
@@ -228,18 +232,30 @@ interface AccordionImplProps extends PrimitiveDivProps {
    */
   disabled?: boolean;
   /**
-   * The functional direction in which the Accordion operates.
+   * The layout in which the Accordion operates.
    * @default vertical
    */
   orientation?: React.AriaAttributes['aria-orientation'];
+  /**
+   * The language read direction. Defaults to RTL.
```
benoitgrelard, line 240: We don't usually set a default I think.

packages/react/accordion/src/Accordion.tsx
```diff
@@ -254,37 +270,56 @@ const AccordionImpl = React.forwardRef<AccordionImplElement, AccordionImplProps>
...
+          nextIndex = homeIndex;
+        }
+      };
+
+      const movePrev = () => {
+        nextIndex = triggerIndex - 1;
+        if (nextIndex < homeIndex) {
+          nextIndex = endIndex;
+        }
+      };
```
benoitgrelard, line 288: It might be nice for consistency to get closer to how we generally handle direction aware arrow keys. Maybe closer to what we do in `Slider` potentially?

dliu-workos: I'm not following this 100%. Are you suggesting we add an OrientationProvider and pass down separate implementations for each key based on the orientation?

The functions you highlighted here are just helpers to avoid having duplicate code and are agnostic of direction. The direction logic only applies to `orientation="horizontal"` and `isDirectionLTR`

.yarn/versions/fa513961.yml
```diff
@@ -0,0 +1,3 @@
+releases:
+  "@radix-ui/react-accordion": minor
+  primitives: patch
```
benoitgrelard, line 3:
```suggestion
releases:
  "@radix-ui/react-accordion": minor

declined:
  - primitives

```

packages/react/accordion/src/Accordion.tsx
```diff
@@ -271,10 +323,17 @@ const AccordionImpl = React.forwardRef<AccordionImplElement, AccordionImplProps>
...
+        scope={__scopeAccordion}
+        disabled={disabled}
+        direction={dir}
+        orientation={orientation}
+      >
         <Collection.Slot scope={__scopeAccordion}>
           <Primitive.div
             {...accordionProps}
+            // aria-orientation={orientation}
+            data-orientation={orientation}
```
benoitgrelard, line 336:
```suggestion
            data-orientation={orientation}
```

packages/react/accordion/src/Accordion.tsx
benoitgrelard: On second thoughts, I think this is fine for now as it does the job and we will potentially rework this using `focusFirst` approach as suggested by @andy-hook here: https://github.com/radix-ui/primitives/pull/1850#discussion_r1053136773

andy-hook (approved):

Issue #393, opened 2021-01-11
[DropdownMenu] Ensure menu height is constrained by the view height


chaance:
See https://github.com/radix-ui/primitives/discussions/375#discussioncomment-275343

> Yeah I believe that's to do with the fact that the menu can overflow the view now and we'd like to constrain its height (at least in some situations). I wanted to do that before xmas but didn't have time to get to it.

andy-hook:
@benoitgrelard happy to look at this. I think I get the gist but could you provide some more context on the issue?

benoitgrelard:
Had a call with @andy-hook about this and we decided for now to wait and see what new requirements are introduced by adding submenu support to `Menu` and adding `MenuBar` to our portfolio to see what use cases emerges.

andy-hook:
For posterity – the issue is with tall menus exceeding the viewport height. It’s possible to get into situations where popper has nowhere else to place the menu, causing layout shifting and other overflow problems, for example:

![image](https://user-images.githubusercontent.com/11708259/115211313-59ddb380-a0f7-11eb-8571-9ae2e685773b.png)

The proposed solution is to cap the max height of the menu relative to the viewport and provide an overflow scroll (native would be fine initially)

Position logic is calculated via [`getPlacementData`](https://github.com/radix-ui/primitives/blob/7144c4a3df7b6864a84cf627be27675d38bd0e29/packages/core/popper/src/popper.ts#L44-L67) and some thought would be needed on the best way to supply additional context for these calculations.

PR #1868, opened 2023-01-04
Always constrain popper to collison boundary

timothympace:
### Fixes
https://github.com/radix-ui/primitives/issues/393

### Description
Adds [`size`](https://floating-ui.com/docs/size) to the middleware pipeline to ensure the resulting placement doesn't overflow the collision boundary. Instead, the content will be given a max height and overflows will be handled with scrolling.

### Goal / Motivation
Allow Dropdown menus that go outside the collision boundary to be scrollable instead of just going off screen where users cannot use them.

### Approach / Alternatives
I'm not married to the approach I currently have in the PR. What I dislike about it is that it makes the assumption that consumers of the library will want their content to scroll when the popper goes outside the collision boundary. I do think that it's a good default however, so it's the approach I went with. One of the primary downsides is that having the scroll behavior on the content is gets coupled with other styles that get passed in for the content itself). For example, a menu with a border-radius & border will cause clipping with the scrollbar:
Screen Shot 2023-01-04 at 2 11 16 PM

Alternative:
- Store the `availableWidth` and `availableHeight` in the `middlewareData` (also passed into the `apply` fn used with `size`) and then store them on the content wrapper as CSS variables (e.g. `--radix-popper-available-height` & `--radix-popper-available-width`). This would avoid introducing new behavior, but also allow consumers to apply their own styling using the CSS vars (example below). Additionally, it would allow people to setup their own scroll behavior, which would allow them to use a [parent/child div to separate the content styles from the border styles, which would allow better looking scrollbars](https://stackoverflow.com/a/60163005).
```
<DropdownMenu.Content
  style={{
    maxHeight: 'var(--radix-popper-available-height)',
    overflowY: 'auto',
  }}
>
  {children}
</DropdownMenu.Content>
```

### DIFF
--- a/packages/react/popper/src/Popper.tsx
+++ b/packages/react/popper/src/Popper.tsx
@@ -8,6 +8,7 @@ import {
   hide,
   arrow as floatingUIarrow,
   flip,
+  size,
 } from '@floating-ui/react-dom';
 import * as ArrowPrimitive from '@radix-ui/react-arrow';
 import { useComposedRefs } from '@radix-ui/react-compose-refs';
@@ -188,6 +189,14 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
           : undefined,
         arrow ? floatingUIarrow({ element: arrow, padding: arrowPadding }) : undefined,
         avoidCollisions ? flip({ ...detectOverflowOptions }) : undefined,
+        size({
+          apply({ availableWidth, availableHeight, elements }) {
+            Object.assign(elements.floating.style, {
+              maxWidth: `${availableWidth}px`,
+              maxHeight: `${availableHeight}px`,
+            });
+          },
+        }),
         transformOrigin({ arrowWidth, arrowHeight }),
         hideWhenDetached ? hide({ strategy: 'referenceHidden' }) : undefined,
       ].filter(isDefined),
@@ -246,6 +255,8 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
         animation: !isPlaced ? 'none' : undefined,
         // hide the content if using the hide middleware and should be hidden
         opacity: middlewareData.hide?.referenceHidden ? 0 : undefined,
+        maxHeight: 'inherit',
+        overflow: 'auto',
       },
     };
 


packages/react/popper/src/Popper.tsx
```diff
@@ -188,6 +189,14 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
           : undefined,
         arrow ? floatingUIarrow({ element: arrow, padding: arrowPadding }) : undefined,
         avoidCollisions ? flip({ ...detectOverflowOptions }) : undefined,
+        size({
+          apply({ availableWidth, availableHeight, elements }) {
```
timothympace, line 193: `apply` also receives `middlewareData` which could be used to support the alternative approach I describe in the PR description.

packages/react/popper/src/Popper.tsx
```diff
@@ -188,6 +189,14 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
+            Object.assign(elements.floating.style, {
+              maxWidth: `${availableWidth}px`,
+              maxHeight: `${availableHeight}px`,
+            });
```
timothympace, line 197: The "content" element is not available here, so it's applied to the root and then inherited on the "content" later on.

packages/react/popper/src/Popper.tsx
```diff
@@ -246,6 +255,8 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
         animation: !isPlaced ? 'none' : undefined,
         // hide the content if using the hide middleware and should be hidden
         opacity: middlewareData.hide?.referenceHidden ? 0 : undefined,
+        maxHeight: 'inherit',
```
timothympace, line 258: Actually constrains the content height by inheriting the value from the root.

packages/react/popper/src/Popper.tsx
```diff
@@ -246,6 +255,8 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
+        overflow: 'auto',
```
timothympace, line 259: Should consumers of the library be responsible for setting this? It's being added as an inline style with highest specificity.

benoitgrelard:
Hey @timothympace,

Interestingly I have already solved this in this PR #1853.
And rather satisfyingly I went for the approach you described in your alternative.

I will therefore close your PR as it's addressed there.

PR #1880, opened 2023-01-10
Adds composePreventableEvent

### LINKED ISSUE #905: [core/primitive][composeEventHandlers] Replace `checkForDefaultPrevented` api
Internally, the `checkForDefaultPrevented` api makes event composition difficult to follow. The suggested path is to split this out into two utils:

`composeEvent`
`composePreventableEvent`
(naming tbd)

In rare occurrences where we have a mixture of preventable and non-preventable handlers, multiple composition is preferred to better communicate the intent:

```jsx
onBlur={composePreventableEvent(
  composeEvent(props.onBlur, context.onNotPreventableEvent), 
  context.onPreventableEvent
)}
```

CrutchTheClutch:
### Description

I noticed #905 and I had a similar need for a utility like this.  

- Aims to improve the `checkForDefaultPrevented` api by splitting `composeEventHandlers` into two separate functions (`composeEventHandlers` & `composePreventableEventHandlers`).
- Replaces a majority of the calls for `composeEventHandlers` with the new `composePreventableEventHandlers`.
- Replaces calls where `checkForDefaultPrevented` was false with the reworked `composeEventHandlers` function.  

closes #905

### DIFF
--- a/.yarn/versions/17c4f847.yml
+++ b/.yarn/versions/17c4f847.yml
@@ -1,0 +1,30 @@
+releases:
+  "@radix-ui/primitive": major
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-collapsible": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
+  "@radix-ui/react-dismissable-layer": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-hover-card": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-navigation-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-radio-group": patch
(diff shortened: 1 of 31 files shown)


chaance:
@crutchtheclutch I'm sorry no one got back to you on this. Are you still interested in seeing this one across the finish line? If so we'd need to resolve conflicts. Totally understand if not, just let me know and I'll put this on my todo list.

CrutchTheClutch:
@chaance For sure! No worries at all, I should be able to take a look at this again early next week

CrutchTheClutch:
@chaance I believe this gets us back up to date with the current changes.  I also went ahead and updated Form and Menubar since they were not components when I originally made this pull request.  I'm a little unfamiliar with the codebase as it's been a while since I last looked at this, but it the changes should be good to go now.  I have also gone ahead and confirmed I can build locally

chaance:
I actually think I'm going to pass on this. IMO `checkForDefaultPrevented` can be avoided if we just don't use the utility in cases where we don't need the check. Something I'll probably change down the road, should time allow.

PR #1913, opened 2023-01-29
Cleanup: remove custom ComponentPropsWIthoutRef type

MiroslavPetrik:
### Description

Removing temporary helper type, as the upstream fix was merged:
https://github.com/radix-ui/primitives/blob/main/packages/react/primitive/src/Primitive.tsx#L23

### DIFF
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -9,7 +9,6 @@ import * as CollapsiblePrimitive from '@radix-ui/react-collapsible';
 import { createCollapsibleScope } from '@radix-ui/react-collapsible';
 import { useId } from '@radix-ui/react-id';
 
-import type * as Radix from '@radix-ui/react-primitive';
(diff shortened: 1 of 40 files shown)


benoitgrelard:
Hey @MiroslavPetrik thanks for the contribution!

@andy-hook I am wondering what this means for the support question, given that we support older versions of React, so theoretically, someone on that would install older versions of the types too?

andy-hook:
> I am wondering what this means for the support question, given that we support older versions of React, so theoretically, someone on that would install older versions of the types too?

Do you mean this might only apply to v18 onwards? do we know which version of the types this shipped in?

benoitgrelard:
I tracked it down and it was released in **v17.0.21** of the types: https://unpkg.com/browse/@types/react@17.0.21/index.d.ts

andy-hook:
> But we support React 16.8+ right, so that does mean for these people it could still be wrong?

Potentially, should we revise our stance on 16?

benoitgrelard:
You mean drop support?

PR #1931, opened 2023-02-05
Remove use of deprecated set-output

penx:
### Description

Remove use of set-output from GitHub actions, which has been deprecated

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,13 +11,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+
       - name: Read Node version
+        id: node_version
         run: echo "NVMRC=$(cat .nvmrc)" >> $GITHUB_OUTPUT
-        id: nvm
       - name: Set Node version
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
         with:
-          node-version: '${{ steps.nvm.outputs.NVMRC }}'
+          node-version: '${{ steps.node_version.outputs.NVMRC }}'
+
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
         run: echo "dir=$(yarn config get cacheFolder)" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -12,15 +12,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Read Node version
-        run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
```
penx, line 15: not sure why this was a hash before...

penx: and not `::set-output`

andy-hook: I'm not sure either, perhaps we should also replace the separate read and set with the approach from `publish-rc`?

penx: I think publish-rc is doing the same thing, it has 2 steps (read the nvmrc followed by use actions/setup-node) just the second step in publish-rc is not named.

andy-hook: Should have been clearer, I meant align them on the naming as they perform the same task. The latter also uses the `v2` action.

penx: Sure, will take a look later

Have you considered adding your actions to dependabot? Though I think has security implications if you use 3rd party actions.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

andy-hook: Thanks mate.

> Have you considered adding your actions to dependabot?

  To be honest we haven’t invested much time in our CI config as it has been doing the job we've needed so far, these are good suggestions though.

penx: @andy-hook I've brought them in line with each other, hopefully what you were wanting :)

andy-hook (approved):
Perfect, thanks @penx !

PR #1942, opened 2023-02-08
Re-namespace new custom properties from `Popper`

### LINKED ISSUE #393: [DropdownMenu] Ensure menu height is constrained by the view height

### LINKED ISSUE #1596: Support a Popover that would overflow the container

benoitgrelard:
Fixes #1596
Closes #393

### Description

This PR re-namespaces the newly available custom properties from `Popper` into dependent components (`ContextMenu`, `DropdownMenu`, `HoverCard`, `Popover`, `Tooltip`).

These were introduced in `Popper` and used in `Select` here: #1853

---

Related PRs (now redudant): #1597, #1659

### DIFF
--- a/packages/react/context-menu/src/ContextMenu.tsx
+++ b/packages/react/context-menu/src/ContextMenu.tsx
@@ -229,9 +229,14 @@ const ContextMenuContent = React.forwardRef<ContextMenuContentElement, ContextMe
         }}
         style={{
           ...props.style,
-          // re-namespace exposed content custom property
-          ['--radix-context-menu-content-transform-origin' as any]:
-            'var(--radix-popper-transform-origin)',
+          // re-namespace exposed content custom properties
+          ...{
+            '--radix-context-menu-content-transform-origin': 'var(--radix-popper-transform-origin)',
+            '--radix-context-menu-content-available-width': 'var(--radix-popper-available-width)',
+            '--radix-context-menu-content-available-height': 'var(--radix-popper-available-height)',
+            '--radix-context-menu-trigger-width': 'var(--radix-popper-anchor-width)',
+            '--radix-context-menu-trigger-height': 'var(--radix-popper-anchor-height)',
+          },
         }}
       />
     );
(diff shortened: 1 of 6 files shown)


packages/react/dropdown-menu/src/DropdownMenu.tsx
```diff
@@ -467,9 +474,14 @@ const DropdownMenuSubContent = React.forwardRef<
       ref={forwardedRef}
       style={{
         ...props.style,
-        // re-namespace exposed content custom property
-        ['--radix-dropdown-menu-content-transform-origin' as any]:
```
benoitgrelard, line 471: This was exposed as `…-dropdown-menu-content-…`, even though it's the sub content. I've done the same for now, but I'm wondering whether it was an omission or on purpose, I can see a reason for both…

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -501,9 +506,14 @@ const ContextMenuSubContent = React.forwardRef<
       ref={forwardedRef}
       style={{
         ...props.style,
-        // re-namespace exposed content custom property
-        ['--radix-context-menu-content-transform-origin' as any]:
```
benoitgrelard, line 505: This was exposed as `…-context-menu-content-…`, even though it's the sub content. I've done the same for now, but I'm wondering whether it was an omission or on purpose, I can see a reason for both…

packages/react/context-menu/src/ContextMenu.tsx
```diff
@@ -501,9 +506,14 @@ const ContextMenuSubContent = React.forwardRef<
...
-          'var(--radix-popper-transform-origin)',
+        // re-namespace exposed content custom properties
+        ...{
+          '--radix-context-menu-content-transform-origin': 'var(--radix-popper-transform-origin)',
+          '--radix-context-menu-content-available-width': 'var(--radix-popper-available-width)',
+          '--radix-context-menu-content-available-height': 'var(--radix-popper-available-height)',
+          '--radix-context-menu-trigger-width': 'var(--radix-popper-anchor-width)',
```
benoitgrelard, line 514: Similarly to the point above, should this be trigger or sub-trigger? Again I can see an argument for both.

packages/react/context-menu/src/ContextMenu.tsx
andy-hook: I remember discussing this and name spacing with `sub` made it harder to share a common block of content styles. I think we ended up leaving it this way for the ergonomics of that.

benoitgrelard: Yeah exactly that's what I was thinking too.
So we're happy with the same thinking for all these new variables representing content/trigger (vs subcontent/subtrigger)?

andy-hook: Sub triggers are a bit different as they wouldn't have many common styles but I think this is fine tbh

benoitgrelard: Yeah I think it's ok too because it's more about the relative aspect of styling the content according to "its" trigger.

andy-hook (approved):

PR #1960, opened 2023-02-15
Fix hovering on tooltip by increasing grace area

### LINKED ISSUE #1855: Hovering on Tooltip is hard in some situations
### Overview

Currently, it's hard to hover on some tooltips because of the graceArea building algorithm.
Example:
When I move the mouse out of the text-box, a grace area is built using the mouse leaving point (ypos - 5)
And it works fine in the following example.

image

but when I try to move the mouse from the corners of text-box this doesn't work, impossible to hold the mouse inside  graceArea

image

Probably would be better to add a small rect around leaving point here, instead of the start point calculation
 
https://github.com/radix-ui/primitives/blob/main/packages/react/tooltip/src/Tooltip.tsx#L423

```js
getHull([
    {x: event.clientX+5, y: event.clientY+5},
    {x: event.clientX+5, y: event.clientY-5},
    {x: event.clientX-5, y: event.clientY+5},
    {x: event.clientX-5, y: event.clientY-5},
    ...hoverTargetPoints
]);
```

That gives the following graceArea and solves the issue above

image

lokeswaran-aj:
### Description

closes #1855 

Updated the grace area algorithm to allow an extra 5 px around the button or text to show the tooltip.

https://user-images.githubusercontent.com/74011196/219046497-54086d16-de44-47c1-99b4-90e1409770b4.mov

Please leave a comment if you feel something has to be changed✨

### DIFF
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -409,18 +409,14 @@ const TooltipContentHoverable = React.forwardRef<
 
   const handleCreateGraceArea = React.useCallback(
     (event: PointerEvent, hoverTarget: HTMLElement) => {
-      const currentTarget = event.currentTarget as HTMLElement;
-      const exitPoint = { x: event.clientX, y: event.clientY };
-      const exitSide = getExitSideFromRect(exitPoint, currentTarget.getBoundingClientRect());
-
-      const bleed = exitSide === 'right' || exitSide === 'bottom' ? -5 : 5;
-      const isXAxis = exitSide === 'right' || exitSide === 'left';
-      const startPoint = isXAxis
-        ? { x: event.clientX + bleed, y: event.clientY }
-        : { x: event.clientX, y: event.clientY + bleed };
-
       const hoverTargetPoints = getPointsFromRect(hoverTarget.getBoundingClientRect());
-      const graceArea = getHull([startPoint, ...hoverTargetPoints]);
+      const graceArea = getHull([
+        { x: event.clientX + 5, y: event.clientY + 5 },
+        { x: event.clientX + 5, y: event.clientY - 5 },
+        { x: event.clientX - 5, y: event.clientY + 5 },
+        { x: event.clientX - 5, y: event.clientY - 5 },
+        ...hoverTargetPoints,
+      ]);
       setPointerGraceArea(graceArea);
       onPointerInTransitChange(true);
     },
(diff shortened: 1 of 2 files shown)


packages/react/tooltip/src/Tooltip.tsx
```diff
@@ -409,18 +409,14 @@ const TooltipContentHoverable = React.forwardRef<
...
-        ? { x: event.clientX + bleed, y: event.clientY }
-        : { x: event.clientX, y: event.clientY + bleed };
-
       const hoverTargetPoints = getPointsFromRect(hoverTarget.getBoundingClientRect());
-      const graceArea = getHull([startPoint, ...hoverTargetPoints]);
+      const graceArea = getHull([
+        { x: event.clientX + 5, y: event.clientY + 5 },
+        { x: event.clientX + 5, y: event.clientY - 5 },
+        { x: event.clientX - 5, y: event.clientY + 5 },
+        { x: event.clientX - 5, y: event.clientY - 5 },
```
benoitgrelard, line 417: This is smart as the extra points will be swallowed by the hull so it can be lazy and not side-dependent. 👍

benoitgrelard (approved):
Hey @lokeswaran-aj, thanks for your contribution!

I like this solution, it's pragmatic and takes less code overall.

@andy-hook I'm happy to go with this, essentially because we're adding a small square buffer (with the 2 inner points being swallowed by the hull) around the start coordinate it gives an area like this:

CleanShot 2023-02-16 at 12 18 41@2x

packages/react/tooltip/src/Tooltip.tsx
benoitgrelard: The only caveat I see where we might still want to bring back some side-dependent checks is this:

CleanShot 2023-02-16 at 12 24 18@2x

as admittedly, exiting from this side should be able to dismiss instantly and not have to suffer the extra 5px wait as we know it's the wrong direction anyway.

benoitgrelard: Does that make sense @lokeswaran-aj?

benoitgrelard (changes requested):
Yeah on second thoughts, I think we should bring back some side-dependency for the case mentioned above.

packages/react/tooltip/src/Tooltip.tsx
lokeswaran-aj: Hi, @benoitgrelard
yes, I get it. Right we have an extra 5px on the side and down also. Do you want to instantly exit only on the sides or down also?

benoitgrelard: Yeah essentially we should only add extra points relative to the exit side.

I think basically it's almost the same as what we currently have, but instead of just one point which we pull back towards the opposite side, we'd add 2 points.

To give a concrete example:

- if we exist on the "top" side
- we'd add 2 points:
  - `{ x: event.clientX - 5, y: event.clientY - 5 }`
  - `{ x: event.clientX + 5, y: event.clientY - 5 }`

Does that make sense?

benoitgrelard: Essentially given a side:
- the 2 points should extend along that side in both directions
- the 2 points should be pushed towards the the opposite side

lokeswaran-aj:
Hi, I m closing this PR as I can't contribute to this project

benoitgrelard:
> Hi, I m closing this PR as I can't contribute to this project

I'm confused, you did contribute. Did you mean you no longer have the time to continue making changes?

lokeswaran-aj:
Yes, you are right

Issue #1933, opened 2023-02-06
[DropdownMenu] Escape does not comply with ARIA authoring practices


dkeele:
### Current Behavior

When focusing a menu item within a submenu, `escape` key closes all menus and submenus.

### Expected behavior

When focusing a menu item within a submenu, `escape` key closes only the containing submenu.

### Reproducible example

[Radix Docs](https://www.radix-ui.com/docs/primitives/components/dropdown-menu)

### Additional context

According to [ARIA authoring practices](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/):
> Escape: Close the menu that contains focus and return focus to the element or context, e.g., menu button or parent menuitem, from which the menu was opened.

Also, here's an [example demonstrating correct behavior](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/)

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |    @radix-ui/react-dropdown-menu     |     Latest, see docs example.    |
| Browser          | Edge, Safari, Firefox, Chrome    |         |
| Operating System | osx Ventura      |       13.1     |

PR #1964, opened 2023-02-17
[Menu] Only close one submenu per escape keypress, per WAI-ARIA

cpmsmith:
Fixes radix-ui/primitives#1933

### Description

Pretty simple: [the WAI-ARIA pattern says](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/) when you press escape in a submenu, instead of closing the submenu and all its ancestors, just close the submenu and return focus to the triggering menuitem, as if you had pressed the left arrow.

> Escape: Close the menu that contains focus and return focus to the element or context, e.g., menu button or parent menuitem, from which the menu was opened.

I didn't realize this was in the pattern, and it's not what macOS does (it matches the existing behaviour of closing the entire menu), but it's not unreasonable. And I suppose matching the WAI-ARIA pattern is the stated design goal.

### DIFF
--- a/.yarn/versions/a03fb8cd.yml
+++ b/.yarn/versions/a03fb8cd.yml
@@ -1,0 +1,8 @@
+releases:
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-menubar": patch
+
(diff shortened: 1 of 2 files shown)


benoitgrelard:
See explanation for closure on related issue: https://github.com/radix-ui/primitives/issues/1933#issuecomment-1472079614

PR #1977, opened 2023-02-22
RFC: `Form` primitive

benoitgrelard:
This RFC proposes adding a new primitive to Radix UI primitives: `Form`. It will provide an easier way to create forms in React. The goal is to provide a simple, declarative, uncontrolled (but controllable) API (à la Radix) to create forms that includes client-side validation in an accessible manner, as well as handling of server errors accessibly too.

### [View rendered RFC](https://github.com/radix-ui/primitives/blob/form-rfc/rfcs/2023-radix-form-primitive.md)

We would love your input on this 🙏

### DIFF
--- a/rfcs/2023-radix-form-primitive.md
+++ b/rfcs/2023-radix-form-primitive.md
@@ -1,6 +1,13 @@
 - Start Date: 2023-02-17
 - RFC PR: (leave this empty, to be filled in later)
 - Authors: Benoît Grélard
+- Updates:
+  - 2023-03-03:
+    - Reworked message API to remove distinction between client/server validation
+      - Now a single `Form.Message` component
+      - A single `match` prop now supports all use-cases
+    - Reworked server error handling to be less prescriptive (now relying more on messages, rather than the prescriptive `serverErrors` prop)
+    - Validation messages / state available outside of `Field` by passing explicit `name` prop
 
 # Radix Form primitive
 
@@ -43,23 +50,23 @@ function Page() {
       <Form.Field name="name">
         <Form.Label>Full name</Form.Label>
         <Form.Control /> {/* renders a `input[type="text"]` by default */}
-        <Form.ClientMessage type="missingValue">Please enter your name.</Form.ClientMessage>
+        <Form.Message match="missingValue">Please enter your name.</Form.Message>
       </Form.Field>
 
       <Form.Field name="email">
         <Form.Label>Email address</Form.Label>
         <Form.Control type="email" />
-        <Form.ClientMessage type="missingValue">Please enter your email.</Form.ClientMessage>
-        <Form.ClientMessage type="typeMismatch">Please provide a valid email.</Form.ClientMessage>
+        <Form.Message match="missingValue">Please enter your email.</Form.Message>
+        <Form.Message match="typeMismatch">Please provide a valid email.</Form.Message>
       </Form.Field>
 
       <Form.Field name="age">
         <Form.Label>Age</Form.Label>
         <Form.Control type="number" min="18" max="57" step="1" />
-        <Form.ClientMessage type="missingValue">Please enter your age.</Form.ClientMessage>
-        <Form.ClientMessage type="rangeOverflow">You must be 18 or older.</Form.ClientMessage>
-        <Form.ClientMessage type="rangeUnderflow">You must be 57 or younger.</Form.ClientMessage>
-        <Form.ClientMessage type="stepMismatch">Please enter a whole number.</Form.ClientMessage>
+        <Form.Message match="missingValue">Please enter your age.</Form.Message>
+        <Form.Message match="rangeOverflow">You must be 18 or older.</Form.Message>
+        <Form.Message match="rangeUnderflow">You must be 57 or younger.</Form.Message>
+        <Form.Message match="stepMismatch">Please enter a whole number.</Form.Message>
       </Form.Field>
 
       <Form.Submit>Submit</Form.Submit>
@@ -70,8 +77,9 @@ function Page() {
 
 Note that there is not state present about the client, everything is uncontrolled by default (yet can be controlled). Also note that there are no attributes necessary to handle accessibility, this is all handled by the primitive:
 
-- label and controls are associated using the `name` provided on `Form.Field`.
-- when one or more client-side error message display, they are automatically associated to their matching control and focus is moved to the first invalid control.
+- label and controls are associated using the `name` provided on `Form.Field`
+- when one or more client-side error message display, they are automatically associated to their matching control
+- focus is moved to the first invalid control
 
 ### Styling
 
@@ -107,53 +115,58 @@ It can also be used to compose other types of controls, such as a `select`:
 
 #### Providing your own validation messages
 
-When no `children` are provided, `Form.ClientMessage` will render a default error message for the given `type`.
+When no `children` are provided, `Form.Message` will render a default error message for the given `match`.
 
 ```jsx
-<ClientMessage type="missingValue" /> // will yield "This value is missing."
+<Form.Message match="missingValue" /> // will yield "This value is missing"
 ```
 
 You can provide a more meaningful message by providing your own `children`. This also allows for internationalization.
 
 ```jsx
-<ClientMessage type="missingValue">Please provide a name.</ClientMessage> // will yield "Please provide a name."
+<Form.Message match="missingValue">Please provide a name</Form.Message> // will yield "Please provide a name"
 ```
 
-#### Client-side validation types
+#### Client-side validation matching
 
-`Form.ClientMessage` accepts a required `type` prop which is used to determine when the message should show. It matches the native HTML validity state (`ValidityState` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState)) which result from using attributes such as `required`, `min`, `max`. The message will show if the given `type` is `true` on the control’s validity state.
+`Form.Message` accepts a `match` prop which is used to determine when the message should show. It matches the native HTML validity state (`ValidityState` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState)) which result from using attributes such as `required`, `min`, `max`. The message will show if the given `match` is `true` on the control’s validity state.
 
 ```ts
-// `type` is one of the following keys from the DOM ValidityState interface:
+// `match`'s prop accepts a `ValidityMatcher` (see below):
 
-interface ValidityState {
-  readonly badInput: boolean;
-  readonly customError: boolean;
-  readonly patternMismatch: boolean;
-  readonly rangeOverflow: boolean;
-  readonly rangeUnderflow: boolean;
-  readonly stepMismatch: boolean;
-  readonly tooLong: boolean;
-  readonly tooShort: boolean;
-  readonly typeMismatch: boolean;
-  readonly valid: boolean;
-  readonly valueMissing: boolean;
-}
+type ValidityMatcher =
+  | 'badInput'
+  | 'patternMismatch'
+  | 'rangeOverflow'
+  | 'rangeUnderflow'
+  | 'stepMismatch'
+  | 'tooLong'
+  | 'tooShort'
+  | 'typeMismatch'
+  | 'valid'
+  | 'valueMissing';
 ```
 
 This means you can even show something whe the field is valid:
 
 ```jsx
-<Form.ClientMessage type="valid">✅ Valid!</Form.ClientMessage>
+<Form.Message match="valid">✅ Valid!</Form.Message>
 ```
 
 #### Custom validation
 
-On top of all the built-in client-side validation types described above you can also provide your own custom validation whislt still making use of the platform's validation abilities. It uses the `customError` type present in the constraint validity API.
+On top of all the built-in client-side validation matches described above you can also provide your own custom validation whilst still making use of the platform's validation abilities. It uses the `customError` type present in the constraint validity API.
 
-You can pass `type="customError"` on `Form.ClientMessage` and provide an `isValid` prop with your custom validation logic.
+You can pass your own validation function into the `match` prop on `Form.Message`.
 
-> `isValid` is called with the current value of the control and the current values of all other controls in the form.
+```ts
+// The `match` prop also accepts a `CustomMatcher` (see below):
+type MatchProp = ValidityMatcher | CustomMatcher;
+type SyncCustomMatcher = (value: string, formData: FormData) => boolean;
+type AsyncCustomMatcher = (value: string, formData: FormData) => Promise<boolean>;
+```
+
+> `match` will be called with the current value of the control as first argument and the entire `FormData` as second argument.
 
 Here's a contrived example:
 
@@ -161,20 +174,16 @@ Here's a contrived example:
 <Form.Field name="name">
   <Form.Label>Full name</Form.Label>
   <Form.Control />
-  <Form.ClientMessage type="customError" isValid={(value, fields) => value === 'John'}>
-    Only John is allowed.
-  </Form.ClientMessage>
+  <Form.Message match={(value, formData) => value !== 'John'}>Only John is allowed.</Form.Message>
 </Form.Field>
 ```
 
-> `isValid` can also be an `async` function (or return a promise) to perform async validation (see type below):
+> `match` can also be an `async` function (or return a promise) to perform async validation (see type below):
 
-```ts
-type CustomValidatorFn = ValidatorSyncFn | ValidatorAsyncFn;
-type ValidatorSyncFn = (value: string, fields: FormFields) => boolean;
-type ValidatorAsyncFn = (value: string, fields: FormFields) => Promise<boolean>;
-type FormFields = { [index in string]?: FormDataEntryValue };
-```
+#### Messages outside of a `Form.Field`
+
+When nested inside a `Form.Field` part, `Form.Message` will automatically be wired to the control's validity which exists in the field.
+Alternatively you can render a message outside of a `Form.Field` part and pass it a `name` prop explicitly.
 
 #### Accessing the validity state for even more control
 
@@ -193,18 +202,19 @@ Sometimes, you may need to access the raw validity state of a field in order to 
 </Form.Field>
 ```
 
-> Note: `Form.ValidityState` should be nested inside a `Form.Field` part and gives you access to the validity of that field.
+When nested inside a `Form.Field` part, `Form.ValidityState` will automatically be wired to the control's validity which exists in the field.
+Alternatively you can render it outside of a `Form.Field` part and pass it a `name` prop explicitly.
 
 #### Server-side validation
 
-The component also supports server-side validation via `Form.ServerMessage`.
+The component also supports server-side validation using the same `Form.Message` component.
+You can re-use the same messages you defined for client-side errors by passing a `forceMatch` prop which will for the message to show regardless of the client-side matching logic. If the message doesn't exist on the client-side, you can render a `Form.Message` without a `match` too.
+The field is marked as invalid by passing a `serverInvalid` boolean prop to the `Form.Field` part.
 
-Given that the server logic is completely outside of the scope of this component, the errors are provided to it using a controlled API: `serverErrors` and `onServerErrorsChange`: It will display accordingly to the errors passed into `serverErrors` on `Form.Root` (typically mapped from the actual errors returned by your server call). When inside a `Field` part, it will display the errors matching that field (`serverErrors[fieldName]`) When outside a `Field` part, it will display global errors that aren't tied to a specific field (`serverErrors.global`).
+The accessibility relating to server errors is handled by the primitive in a similar fashion:
 
-Similary to `Form.ClientMessage`, all accessibility relating to server errors is handled by the primitive:
-
-- when one or more server-side error message display, they are automatically associated to their matching control and focus is moved to the first invalid control.
-- when there are only global server errors, focus is moved to the submit button and the message is associated with it so screen readers announce it.
+- when one or more server-side error message display, they are automatically associated to their matching control
+- focus is moved to the first invalid control
 
 Let's see the same example as above but with added server-side error handling:
 
@@ -213,7 +223,7 @@ import * as React from 'react';
 import * as Form from '@radix-ui/react-form';
 
 function Page() {
-  const [serverErrors, setServerErrors] = React.useState({});
+  const [serverErrors, setServerErrors] = React.useState({ name: false, email: false, age: false });
 
   return (
     <Form.Root
@@ -224,68 +234,69 @@ function Page() {
 
         // maybe do an async server call and gather some errors from the server to display
         submitForm(data)
-        .then(() => …)
-        /**
-         * Map errors from your server response into the expected format of `serverErrors` (see type below)
-         * For example:
-         * {
-         *   email: [ { code: 'email_invalid', message: 'This is not a valid email.' } ],
-         *   global: [ { code: 'server_error', message: 'Something went wrong.' }]
-         * }
-         */
-        .catch((errors) => setServerErrors(mapServerErrors(errors)));
+          .then(() => {})
+          /**
+           * Map errors from your server response into a structure you'd like to work with.
+           * In this case resulting in this object: `{ name: false, email: true, age: true }`
+           */
+          .catch((errors) => setServerErrors(mapServerErrors(errors)));
       }}
-      serverErrors={serverErrors}
-      onServerErrorsChange={setServerErrors}
+      onClearServerErrors={() => setServerErrors({ name: false, email: false, age: false })}
     >
-      <Form.Field name="name">
+      <Form.Field name="name" serverInvalid={serverErrors.name}>
         <Form.Label>Full name</Form.Label>
         <Form.Control />
-        <Form.ClientMessage type="missingValue">Please enter your name.</Form.ClientMessage>
-        <Form.ServerMessage />
+        <Form.Message match="missingValue" forceMatch={serverErrors.name}>
+          Please enter your name.
+        </Form.Message>
       </Form.Field>
 
-      <Form.Field name="email">
+      <Form.Field name="email" serverInvalid={serverErrors.email}>
         <Form.Label>Email address</Form.Label>
         <Form.Control type="email" />
-        <Form.ClientMessage type="missingValue">Please enter your email.</Form.ClientMessage>
-        <Form.ClientMessage type="typeMismatch">Please provide a valid email.</Form.ClientMessage>
-        <Form.ServerMessage /> {/* will yield "This is not a valid email." */}
+        <Form.Message match="missingValue">Please enter your email.</Form.Message>
+        <Form.Message match="typeMismatch" forceMatch={serverErrors.email}>
+          Please provide a valid email.
+        </Form.Message>
       </Form.Field>
 
-      <Form.Field name="age">
+      <Form.Field name="age" serverInvalid={serverErrors.age}>
         <Form.Label>Age</Form.Label>
         <Form.Control type="number" min="18" max="57" step="1" />
-        <Form.ClientMessage type="missingValue">Please enter your age.</Form.ClientMessage>
-        <Form.ClientMessage type="rangeOverflow">You must be 18 or older.</Form.ClientMessage>
-        <Form.ClientMessage type="rangeUnderflow">You must be 57 or younger.</Form.ClientMessage>
-        <Form.ClientMessage type="stepMismatch">Please enter a whole number.</Form.ClientMessage>
-        <Form.ServerMessage />
+        <Form.Message match="missingValue">Please enter your age.</Form.Message>
+        <Form.Message match="rangeOverflow">You must be 18 or older.</Form.Message>
+        <Form.Message match="rangeUnderflow">You must be 57 or younger.</Form.Message>
+        <Form.Message match="stepMismatch">Please enter a whole number.</Form.Message>
+        {serverErrors.age ? (
+          <Form.Message>Registration not open to people aged 40 exactly.</Form.Message>
+        ) : null}
       </Form.Field>
 
       <Form.Submit>Submit</Form.Submit>
-
-      <Form.ServerMessage /> {/* will yield "Something went wrong." */}
     </Form.Root>
   );
 }
 ```
 
-If no `children` are provided to `Form.ServerError`, it will display a concatenation of error messages from the server. If passing a React node, it will render this node instead. For finer control, you can pass a render function instead which will be called with the server errors.
+You should also clear the server errors using the `onClearServerErrors` callback prop on the `Form.Root` part.
 
-The server messages will be routed to the correct fields, based on the `name` attribute on `Form.Field` matching in the `serverErrors` object (see type below). There is also a `global` key which will be used for global errors (displayed in the `Form.ServerMessage` sitting outside a `Form.Field`).
+> This will clear the server errors before the form is re-submitted, and when the form is reset.
 
-```ts
-interface ServerError {
-  code: string;
-  message: React.ReactNode;
-}
+On top of this, it means you have control over when to reset single server errors.
+For example you could reset the email server error as soon as the user edits it:
 
-type ServerErrors = {
-  [fieldName in string]?: ServerError[];
-} & {
-  global?: ServerError[];
-};
+```jsx
+<Form.Field name="email" serverInvalid={serverErrors.email}>
+  <Form.Label>Email address</Form.Label>
+  <Form.Control
+    type="email"
+    onChange={() => setServerErrors((prev) => ({ ...prev, email: false }))}
+  />
+  <Form.Message match="missingValue">Please enter your email.</Form.Message>
+  <Form.Message match="typeMismatch" forceMatch={serverErrors.email}>
+    Please provide a valid email.
+  </Form.Message>
+</Form.Field>
 ```
 
 ## Open questions
@@ -293,9 +304,5 @@ type ServerErrors = {
 - Does this API make sense?
   - Is it easy to use?
   - Is everything named as you'd expect?
-  - What about the distinction between `ClientMessage` and `ServerMessage`?
-    - Would you call this something else?
-  - Does the server side stuff make sense?
-    - Would you call `serverErrors` something else as it's not directly server errors, but a structure wanted by the UI
 - Does this API cover most form use-cases? Have we missed some use-cases?
 - Is the API flexible enough to allow composition with other components?
--- a/rfcs/2023-radix-form-primitive.md
+++ b/rfcs/2023-radix-form-primitive.md
@@ -14,7 +14,7 @@ Forms are a huge part of building for the web, and whilst we do offer building b
 
 Additionally, similarly to how Radix introduced a great out of the box experience for specific components (not having to wire state or refs for example) we see a big opportunity to do the same for forms as this is a space where we see a lot of people struggle or overcomplicate things.
 
-At [WorkOS](https://workos.com), we have also been working on components specifically geared towards building great authentication experiences (see [Radix Auth RFC](2023-radix-form-primitive.md)). As these rely heavily on forms, we thought it would make sense to provide some fundations in Radix to help with this.
+At [WorkOS](https://workos.com), we have also been working on components specifically geared towards building great authentication experiences (see [Radix Auth RFC](https://github.com/radix-ui/primitives/pull/1978)). As these rely heavily on forms, we thought it would make sense to provide some fundations in Radix to help with this.
 
 ## Detailed design
 
--- a/rfcs/2023-radix-form-primitive.md
+++ b/rfcs/2023-radix-form-primitive.md
@@ -229,8 +229,8 @@ function Page() {
          * Map errors from your server response into the expected format of `serverErrors` (see type below)
          * For example:
          * {
-         *   email: [ { code: 'email_invalid', 'This is not a valid email.' } ],
-         *   global: [ { code: 'server_error', 'Something went wrong.' }]
+         *   email: [ { code: 'email_invalid', message: 'This is not a valid email.' } ],
+         *   global: [ { code: 'server_error', message: 'Something went wrong.' }]
          * }
          */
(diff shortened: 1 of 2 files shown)


peduarte:
This is great 👏

benoitgrelard:
Thanks for the feedback @Murkrage!

> I have one question regarding the error type and messages. Perhaps instead of a component for each desired error type you could provide <Form.ClientMessage /> with an object that matches the following: <type, message>. If I wanted to use a wide range of messages, this would eliminate multiple components form having to be created. Many of which will be sharing styling.

That's interesting. Although the issue with that is that it potentially moves us away from the 1:1 mapping between part and elements. As there may be situations where multiple messages are visible at once, you'd generally want access to the element for styling for example.

> Since valid is a type, and usually this would involve some form of green styling, a data attribute could be added to indicate whether the input is valid or invalid: data-state='valid' which can then be used with styling.

Yeah that's included in the RFC already in the styling section: https://github.com/radix-ui/primitives/blob/form-rfc/rfcs/2023-radix-form-primitive.md#styling

Murkrage:
> As there may be situations where multiple messages are visible at once, you'd generally want access to the element for styling for example.

That's not something I had considered, to be honest. I think there's merit for both implementations and ultimately it's subjective whether someone "likes" the extra components for messaging or not. I suppose someone could always create a wrapping component around all of the messages to achieve the same result.

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+
+> `isValid` is called with the current value of the control and the current values of all other controls in the form.
+
+Here's a contrived example:
+
+```jsx
+<Form.Field name="name">
+  <Form.Label>Full name</Form.Label>
+  <Form.Control />
+  <Form.ClientMessage type="customError" isValid={(value, fields) => value === 'John'}>
```
peduarte, line 164: isValid` seems like a boolean prop at first glance, instead of a function. How about `validator`?

ttsirkia:
If using this, how it would integrate with React Hook Form or similar libraries? Or is the idea that you wouldn't need those anymore?

bdsqqq:
Great work on the RFC! 

> Does this API make sense?

Yes! Combining the browser constraint validation API makes the logic part intuitive, while the Radix pattern of splitting a component into multiple parts + the asChild approach makes composability simple and advanced things like dynamically generated fields frictionless.

As with every other primitive, the data- attributes provide an easy way to style different states, and I especially like the control given by ClientMessage being a component for each error type. For this bit how would someone achieve "I want to add a comma between each of my errors", would checking the `validityState` after each ClientMessage and conditionally adding a comma be the best/only approach?

The server Errors made sense and having them based on the name of the Form.Field is nice, but I'd having something like `<Form.ServerMessage name="email" />` outside a Form.Field would be great.

jamesopstad:
This is exciting! I'm interested in how this would integrate with the React Router `Form` component (https://reactrouter.com/en/main/components/form)? Could this be done directly by using `asChild` on `Form.Root` or would you have to instead use the `useSubmit` hook (https://reactrouter.com/en/main/hooks/use-submit) somehow?

benoitgrelard:
> This is exciting! I'm interested in how this would integrate with the React Router `Form` component (https://reactrouter.com/en/main/components/form)? Could this be done directly by using `asChild` on `Form.Root` or would you have to instead use the `useSubmit` hook (https://reactrouter.com/en/main/hooks/use-submit) somehow?

@jamesopstad I am not familiar with React Router `Form` component, but looking at their docs at a glance, it seems like it could be a match made in heaven.

They specifically mention:

> The Form component is a wrapper around a plain HTML [form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) that emulates the browser for client side routing and data mutations. **It is not a form validation/state management library like you might be used to in the React ecosystem (for that, we recommend the browser's built in [HTML Form Validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation) and data validation on your backend server).**

Which is exactly what ours is.

I think as long as their `Form` component forwards all props and ref and generally behaves like a normal native `form`, using `asChild` should be all one need to compose the 2.

```tsx
import * as Form from '@radix-ui/react-form';
import { Form as RouterForm } from "react-router-dom";

<Form.Root asChild>
  <RouterForm>
    …
  </RouterForm>
</Form.Root>
```

One to play around with to check compatibility but I think it's promising.

jamesopstad:
@benoitgrelard This sounds very promising!

> I think as long as their `Form` component forwards all props and ref and generally behaves like a normal native `form`, using `asChild` should be all one need to compose the 2.

I've just had a quick look at the source code and all of the above are true. The nice thing about the `asChild` approach is that people can just create their own `Form` component that composes the two. FYI the the React Router `Form` component is also used in Remix.

ttsirkia:
> @ttsirkia Yes that's the idea, it should tackle yours needs around form validation so it wouldn't make much sense to use both. Are there things you can think of that wouldn't be possible with our API but would be with React Hook Form or similar libs?

I could see the benefits of using Radix to provide the accessibility part and React Hook Form the state management and validation (maybe with Zod) etc.

benoitgrelard:
> I've just had a quick look at the source code and all of the above are true. The nice thing about the `asChild` approach is that people can just create their own `Form` component that composes the two. FYI the the React Router `Form` component is also used in Remix.

@jamesopstad, that's awesome. I'll play around with it but yeah sounds like things should slot in nicely 🙂 And yes that is the nice thing about `asChild`!

rothsandro:
Did you consider supporting schema-based validation with a library like Zod?

jjenzz:
okay, i am so fricken' excited that this is finally happening 🎉 yay!

a few bits i would love to see that i didn't see mentioned (or may have missed, 'pologies). it was common in ye olden' days to be able to name fields like `pets[]` and PHP would understand that this is an array and give you an array on the server. it would be great if the `onSubmit` callback was a little smarter and gave you the form data object with array values in a similar fashion if you were to use this naming convention.

here's a version of my own attempts at this declarative Form idea from many moons ago (so ignore how scrappy it is) that also uses named form groups (rendered as `fieldset`s). these give you nesting in your form data object shape which might also be a nice addition: https://codesandbox.io/s/form-components-6gdck

another thing worth pointing out from the above example is that validation is not restricted to being rendered alongside the field itself because it associates itself using a `name`/`field` prop relationship much like `for`/`id`. so, i can render all my errors at the top of my form if that is a preferred design choice. i'm not sure if that is okay in terms of a11y tho, especially if Radix will be moving focus to the first invalid field. is the focus an a11y requirement?

also, i agree with the `isValid` concerns. definitely sounds like a boolean. something like `validator` would be preferred 👍

jamiebuilds:
It would be nice to have `<Form.Description>` for adding `aria-describbedby` to your field.

jjenzz:
> It would be nice to have <Form.Description> for adding `aria-describbedby` to your field.

i presume Radix does this internally already because that is needed to associate `Form.ClientMessage` with the input i believe. that will all be part of the a11y work that Radix handles for you.

jamiebuilds:
`<Form.Field>` should optionally accept an `id` prop so you could pass in your own in cases where you want to use the `id` outside of its children (ex. `<output>`, various aria attributes, etc.)

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+        // maybe do an async server call and gather some errors from the server to display
+        submitForm(data)
+        .then(() => …)
+        /**
+         * Map errors from your server response into the expected format of `serverErrors` (see type below)
+         * For example:
+         * {
+         *   email: [ { code: 'email_invalid', 'This is not a valid email.' } ],
+         *   global: [ { code: 'server_error', 'Something went wrong.' }]
+         * }
```
jamiebuilds, line 234: Missing a `message` property here? `{ ..., message: 'This is not a valid email' }`

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+
+function Page() {
+  return (
+    <Form.Root
+      onSubmit={(event) => {
+        // `onSubmit` only triggered if it passes client-side validation
+        const data = Object.fromEntries(new FormData(event.currentTarget));
+        console.log(data);
+        event.preventDefault();
+      }}
```
jamiebuilds, line 41: It's not called out explicitly in this RFC, but I assume without the `event.preventDefault()` this will submit the form natively?

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+        <Form.ClientMessage type="missingValue">Please enter your age.</Form.ClientMessage>
+        <Form.ClientMessage type="rangeOverflow">You must be 18 or older.</Form.ClientMessage>
+        <Form.ClientMessage type="rangeUnderflow">You must be 57 or younger.</Form.ClientMessage>
+        <Form.ClientMessage type="stepMismatch">Please enter a whole number.</Form.ClientMessage>
+        <Form.ServerMessage />
+      </Form.Field>
+
+      <Form.Submit>Submit</Form.Submit>
+
+      <Form.ServerMessage /> {/* will yield "Something went wrong." */}
```
jamiebuilds, line 268: I think I'd rather operate on codes returned from the server so that I can manage the translations in one place.

```tsx
<Form.ServerMessage type="email_taken">
  {/* react-intl */}
  <FormattedMessage defaultMessage="Account with email already exists" ... />
</Form.ServerMessage>
```

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+interface ServerError {
+  code: string;
+  message: React.ReactNode;
+}
+
+type ServerErrors = {
+  [fieldName in string]?: ServerError[];
+} & {
+  global?: ServerError[];
+};
```
jamiebuilds, line 288: These types are sometimes annoying, the other option here is:

```ts
type ServerErrors = {
  form: ServerError[],
  fields: Record<string, ServerError[]>,
}

// You could make it optional in the prop so that it's easier to pass in
type ServerErrorsProp = Partial<ServerErrors>
```

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+Note that there is not state present about the client, everything is uncontrolled by default (yet can be controlled). Also note that there are no attributes necessary to handle accessibility, this is all handled by the primitive:
+
+- label and controls are associated using the `name` provided on `Form.Field`.
+- when one or more client-side error message display, they are automatically associated to their matching control and focus is moved to the first invalid control.
+
+### Styling
+
+Similar to other Radix primtives, the `Form` primitive is unstyled by default. This means that you can style it using any CSS solution of your choice. Each part is a node you can style. Like other primitives, data attributes are present to aid styling states such as `data-invalid` and `data-valid`.
```
jacobrask, line 78: Would it also apply an `aria-invalid` styling hook? I know Radix components so far seem to exclusively expose data attributes for state styling, any particular reason for that if matching aria-states already exists?

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+</Form.Field>
+```
+
+> `isValid` can also be an `async` function (or return a promise) to perform async validation (see type below):
+
+```ts
+type CustomValidatorFn = ValidatorSyncFn | ValidatorAsyncFn;
+type ValidatorSyncFn = (value: string, fields: FormFields) => boolean;
+type ValidatorAsyncFn = (value: string, fields: FormFields) => Promise<boolean>;
+type FormFields = { [index in string]?: FormDataEntryValue };
```
jacobrask, line 176: I’d recommend also looking at https://github.com/brophdawg11/remix-validity-state by @brophdawg11, passing FormData directly is very webby and nice

bdsqqq:
@benoitgrelard 
>  The easiest way to achieve this might actually to wrap all of the messages in an element like a span and do it with css pseudo content...

This makes a lot of sense. I'll raise this as another pro to exposing each message as a component instead of a single object or something similar.

> I haven't played much with ZOD, if it gives you a function back to run validation then you should be able to use it in `isValid`. Are there specific things you can think of here in terms of integration?

The biggest reasons why someone would use ZOD are to share a validation schema between the back and front end and to be able to infer the type of the form data. I think you could validate each field individually using zod, but the big win would come if you were able to pass the whole schema to the form as a validator. 

as an example, react-hook-form would do:

```tsx
import { useForm } from 'react-hook-form';
import { zodResolver } from '@hookform/resolvers/zod';
import * as z from 'zod';

const schema = z.object({
  name: z.string().min(1, { message: 'Required' }),
  age: z.number().min(10),
});

const App = () => {
  const {
    register,
    handleSubmit,
    formState: { errors },
  } = useForm({
    resolver: zodResolver(schema),
  });

return (
    <form onSubmit={handleSubmit((d) => console.log(d))}>
      <input {...register('name')} />
      {errors.name?.message && <p>{errors.name?.message}</p>}
      <input type="number" {...register('age', { valueAsNumber: true })} />
      {errors.age?.message && <p>{errors.age?.message}</p>}
      <input type="submit" />
    </form>
  );
};
```

I see why this might not belong in the primitive itself, but I think it would be useful to expose a way for the consumer(or more realistically, library authors) to define custom resolvers that apply to the whole form instead of single fields.

---

Additionally, this is probably outside of 99% of use cases, but would come up if custom resolvers were a thing. 
Is it possible to tap into the default ValidityState values with custom validation?

eg:

```tsx
<Form.Field name="something">
  <Form.Control type="number"/>
  <Form.ClientMessage type="rangeUnderflow" isValid={(value, fields) => value <= 10}>
    Too big.
  </Form.ClientMessage>
</Form.Field>
```

instead of 

```tsx
<Form.Field name="something">
  <Form.Control type="number" max="10"/>
  <Form.ClientMessage type="rangeUnderflow">
    Too big.
  </Form.ClientMessage>
</Form.Field>
```

Again, amazing work on the RFC, excited to see it develop!

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+        <Form.ClientMessage type="typeMismatch">Please provide a valid email.</Form.ClientMessage>
+      </Form.Field>
+
+      <Form.Field name="age">
+        <Form.Label>Age</Form.Label>
+        <Form.Control type="number" min="18" max="57" step="1" />
+        <Form.ClientMessage type="missingValue">Please enter your age.</Form.ClientMessage>
+        <Form.ClientMessage type="rangeOverflow">You must be 18 or older.</Form.ClientMessage>
+        <Form.ClientMessage type="rangeUnderflow">You must be 57 or younger.</Form.ClientMessage>
+        <Form.ClientMessage type="stepMismatch">Please enter a whole number.</Form.ClientMessage>
```
joaopcm, line 62: This can get a little confusing when there are several possible types of errors for an input. It may not be the best way to create a `Form.ClientMessage` for each `type`.

What about rendering just one `<Form.ClientMessage />` and giving it choices of types and error messages to render? Here is an example of what I believe may be a better approach (or not):

```jsx
const ageErrorsMessages = {
   missingValue: 'Please enter your age.',
   rangeOverflow: 'You must be 18 or older.',
   rangeUnderflow: 'You must be 57 or younger.',
   stepMismatch: 'Please enter a whole number.',
   customErrors: [
      {
         message: 'This is my first custom error message.',
         validate: (value, fields) => value === 'John'
      },
      // could do it forever
   ]
}

// ...

<Form.ClientMessage messages={ageErrorsMessages} />
```

If you like this approach, we could rename the `Form.ClientMessage` to something like `Form.ClientMessages` to handle multiple options.

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+        <TextField variant="primary" state={getTextFieldState(validity)} />
+      </Form.Control>
+    )}
+  </Form.ValidityState>
+</Form.Field>
+```
+
+> Note: `Form.ValidityState` should be nested inside a `Form.Field` part and gives you access to the validity of that field.
+
+#### Server-side validation
```
SoYoung210, line 198: First of all, this is a component I really needed and I can't wait to try it out!!! :heart:

I have a comment regarding the Server/Client Error.

1. location-independent error message representation

> When inside a `Field` part, it will display the errors matching that field (`serverErrors[fieldName]`) When outside a `Field` part, it will display global errors that aren't tied to a specific field (`serverErrors.global`).

Based on the above, my current understanding is that a `<Form.ServerErrorMessage />` component must be located below the targeted `<Form.Field />` component in order to display an error for a specific field. However, as I've worked with Forms, there have been times when I've wanted to display an error regardless of location.

What if we treated `name` as an optional prop, and allowed the error to be expressed if there was a field associated with it?

```jsx
{ name: 'email', type: 'invalid', message: 'This is not a valid email.' }

// ok
<Form.Field name='email'>
  <Form.ServerErrorMessage />  
</Form.Field>

// also ok
<Form.Field name='email'>
  {/* ... */}
</Form.Field> // also ok

// outside of Form.Field
<div > show all errors </div>
<Form.ServerErrorMessage name='email' />
```

2. Should we separate the ServerErrorMessage component from the ClientErrorMessage component?

There are two kinds of errors that can occur in a form, client and server, but they don't seem to be different in that they should eventually be 'passed as an error' to the user.

It would make the code a bit more concise for the user if static and dynamic validation were handled by a single Error component, but I'm curious as to the reason for separating them into two.

```jsx
// for example, 
<Form.Root
  onSubmit={() => { serverside_validation }}
>
  <Form.Field name='email' isValid={(value, fields) => value === 'John'}>
    <Form.Errors>
      {/* errors: array containing both client and server errors */}
      {errors => errors.map.join}
    </Form.Errors>
  </Form.Field>
</Form.Root>
```

rfcs/2023-radix-form-primitive.md
benoitgrelard: Interesting. What specifically is annoying with this type?

rfcs/2023-radix-form-primitive.md
benoitgrelard: There should be no need for `aria-invalid` as `Form.Control` renders an `input` so the native validity already works. If you are composing a custom control, we would expect the responsibility of making sure that component is accessible belongs in its implementation (similar to how we do with some of our primitives). This way, as long as the custom control relies on some native control internally, it should all work fine.

rfcs/2023-radix-form-primitive.md
jjenzz: @benoitgrelard yeah i also have concerns about the `Form.ClientMessage` and `Form.ServerMessage` being two separate components. from a progressive enhancement mindset, it feels strange to separate them.

i think i would rather have `Form.ValidationMessage` or similar that renders client-side and server-side, with a `defaultValid` prop or similar for specifying uncontrolled default state from server:

```jsx
<Form.ValidationMessage type="missingValue" defaultValid={!serverData.errors.name}>Name is required</Form.ValidationMessage>
```

rfcs/2023-radix-form-primitive.md
jacobrask: Styling using `:invalid` is problematic since it will apply immediately when a form control is rendered, even before the user has interacted with it. An empty input with `required` will be invalid from the time it's rendered.

`:user-invalid` is designed to fix this and only apply after a user has interacted with the form, and has been selected as a priority for [Browser Interop 2023](https://github.com/web-platform-tests/interop/issues/178) but is still not supported in Chrome. `aria-invalid` is [specced](https://github.com/whatwg/html/issues/7848) slightly differently to `:invalid`/`element.validity.valid`, and more closely matches how `:user-invalid` works.

This is something you'll have to decide for your validation logic as well - should a message be shown immediately when the form is invalid (a required input will report `input.validity.valid === false` directly when it's rendered), or should it only happen after "significant user interaction" as with the `:user-invalid` selector.

rfcs/2023-radix-form-primitive.md
joaopcm: That makes sense! Thank you for clarifying all this stuff! Now, I agree that the current implementation is a more flexible and better approach.

rfcs/2023-radix-form-primitive.md
benoitgrelard: Yeah the separation between `ClientMessage` and `ServerMessage` is the part I'm the last happy with hence why it was one of the open questions. Glad folks are suggesting ideas for this too. Will iterate on them.

Thanks! 🙏

rfcs/2023-radix-form-primitive.md
benoitgrelard: Yes, that's why we expose data attributes, as it's less reliant on implementation details (and also for this case you are mentioning in particular relating to `:invalid`).

rfcs/2023-radix-form-primitive.md
benoitgrelard: Yeah I think we're unlikely to change the 1 part = 1 node requirement, seeing as that's how Radix primitives work in general. Also it's kinda trivial to build abstractions around it.

rfcs/2023-radix-form-primitive.md
benoitgrelard: Thanks for the feedback!

> 1. location-independent error message representation

Yes having an optional `name` on them to be available outside is something we have considered and may implement.
The reason we went for `Field` collocation at the moment is that it works automatically, and it's unlikely/bad ux that you'd want your errors far away from the controls. That said flexibility is always something we consider so will think about it.

> 2. Should we separate the ServerErrorMessage component from the ClientErrorMessage component?

Yeah others (https://github.com/radix-ui/primitives/pull/1977#discussion_r1115548609) have raised similar point. It's also the part I'm the last happy about so we'll probably revise that.

Thanks!

benoitgrelard:
Thanks for you feedback @jjenzz, a lot of this contains some

> okay, i am so fricken' excited that this is finally happening 🎉 yay!

🎉 🎉 🎉 

> 
> a few bits i would love to see that i didn't see mentioned (or may have missed, 'pologies). it was common in ye olden' days to be able to name fields like `pets[]` and PHP would understand that this is an array and give you an array on the server. it would be great if the `onSubmit` callback was a little smarter (`CustomEvent` detail or even a separate callback) and gave you the form data object with array values in a similar fashion. for example something like the following:

Yeah, this could be a nice enhancement, especially for SPA scenarios where submit gets `event.preventDefault()`'ed.

> 
> another thing perhaps worth pointing out from the above example is that validation is not restricted to being rendered alongside the field itself because it associates itself using a `name`/`field` prop relationship much like `for`/`id`. so, i can render all my errors at the top of my form/fieldset if that is a preferred design choice. i'm not sure if that is okay in terms of a11y tho, especially if Radix will be moving focus to the first invalid field. i'm assuming the control focus is an a11y requirement? this might make more sense for the server errors rather than the client ones but i like the flexibility it provides.

Yeah I explained some of the reasoning here: https://github.com/radix-ui/primitives/pull/1977#discussion_r1115855539
But yeah I think UX-wise having the errors far from the field isn't great.

> also, i agree with the `isValid` concerns. definitely sounds like a boolean. something like `validator` would be preferred 👍

Yep, agree too, will make some changes there.

edzis:
If the input is failing on multiple validations I need flexibility to control which of those I show.

A popular variant of that would be to show just one message (the most important/urgent on), which is frequently done by showing it not next to the input but somewhere else (e.g. toast, header/footer, tooltip on submit button).

Most likely this is too naive, but currently I am imagining:
- a mirror of ClientMessage to avoid render anything but still allow to declaratively gather error data to be used further - e.g ErrorBuilder
- a `getErrorData` function on both ClientMessage and ErrorBuilder that returns custom data to be used by any error renderer
- a `renderChildren` function on ClientMessage that receives errorData generated by it's validate or getErrorData function and generates content instead of `children`
- a Form.Errors that is like Form.Validity but scoped to the Form instead of a single Field, also with a `renderChildren` function that receives errorData from all fields (from CustomError, ClientMessage and maybe even standard validation)

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,301 @@
...
+
+```jsx
+<Form.Field name="name">
+  <Form.Label>Full name</Form.Label>
+  <Form.ValidityState>
+    {(validity) => (
+      <Form.Control asChild>
+        <TextField variant="primary" state={getTextFieldState(validity)} />
+      </Form.Control>
+    )}
```
pacocoursey, line 191: Radix got so far without having to resort to render props — why add this pattern now?

Couldn't it be a component instead?

```tsx
<Form.ValidityState value="success">
  ... success state ...
</Form.ValidityState>
```

benoitgrelard: Hey Paco, sure a component-based API could be created to render based on certain state matching (pretty much a simpler version of `ClientMessage`) but sometimes you do need more access.

ie. the case you highlighed above where your DS expects a prop for some variant or state.

Worth nothing also that although we we haven't historically used render props, we are not completely opposed to it. It depends what the use-case is and if there are other technical solution with less API surface. For most of the states, we do believe a controlled prop fulfills the need.

With validity though, that isn't something that can be "controlled". However, we could follow the same logic and expose a callback for when the validity of a field changes so you have access to it potentially?

I do feel like a render prop is a good fit for this particular case though.

Do you have specific concerns about it?

dwayneholmberg:
I was really excited to see this come across my twitter feed. I'm a fan of Radix, and have not found the happy place when it comes to accessible form handling.

Two thoughts:

1) I sometimes use a pattern where form errors are summarised at the top of the form, with links to the invalid fields, rather than moving focus directly to the invalid field on submit. (This is most useful for complex forms where it might be difficult for the user to understand the context if they are just immediately moved to field 37 without warning.) If I'm understanding the RFC correctly, I don't think this would be possible? It's entirely understandable if this is outside the project goals. (One thing I found myself wishing for with react-hook-forms was something like `onBeforeSubmit` that could access the form state. Having only `handleSubmit`, and it only being called if validation passed led to some ugly workarounds. Among other things, this would enable the above scenario.)

2) Under **Custom validation**, I wasn't quite clear what kind of information is available in the FormFields type - does this include information about other field states? I'm thinking about cases where one field's validation depends on the value provided in another field.

benoitgrelard:
Thanks for your feedback @dwayneholmberg!

> 1. I sometimes use a pattern where form errors are summarised at the top of the form, with links to the invalid fields, rather than moving focus directly to the invalid field on submit. (This is most useful for complex forms where it might be difficult for the user to understand the context if they are just immediately moved to field 37 without warning.) If I'm understanding the RFC correctly, I don't think this would be possible? It's entirely understandable if this is outside the project goals.

Yeah at the moment that wouldn't be part of this proposal. This is interesting though, but might require more work as ideally we wouldn't want to leave it up to the users to have to do the necessary wiring for accessibility.

> (One thing I found myself wishing for with react-hook-forms was something like `onBeforeSubmit` that could access the form state. **Having only `handleSubmit`, and it only being called if validation passed led to some ugly workarounds.** Among other things, this would enable the above scenario.)

Could you expand on this?

> 2. Under **Custom validation**, I wasn't quite clear what kind of information is available in the FormFields type - does this include information about other field states? I'm thinking about cases where one field's validation depends on the value provided in another field.

Yes, the validation function gives you the value of that field as first argument, and an object keyed by name with the values of the other fields if there are dependencies for example.

ArekBartnik:
Where would you place `Form.Label` for a radio or checkbox?
Could there be multiple `Form.Control`s in one `Form.Field`?

benoitgrelard:
> Where would you place Form.Label for a radio or checkbox?
Could there be multiple Form.Controls in one Form.Field?

@ArekBartnik these are great questions! In the original proposal `Field` would only really take care of labelling a single control so that would be an issue. You could use a single `Field` (thus defining a `name`) and then pairs of `Label`/`Control` but re-associate them manually  with `id` and `htmlFor` as we always make it possible to override everything.

I think we can do better though, I will look specifically around supporting groups of controls like radio/checkbox groups that share a `name`.

ArekBartnik:
> I think we can do better though, I will look specifically around supporting groups of controls like radio/checkbox groups that share a `name`.

@jjenzz in her example used `Form.Group` to render `fieldset`. Maybe it is the solution 🤔

jrestall:
I have a use case which requires disabling the form submit button until all mandatory fields pass validation. Is that something this primitive could easily support?

benoitgrelard:
> I have a use case which requires disabling the form submit button until all mandatory fields pass validation. Is that something this primitive could easily support?

Yeah we could easily add support for this now that the validity state was lifted to the root.

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,308 @@
...
+</Form.Field>
+```
+
+When nested inside a `Form.Field` part, `Form.ValidityState` will automatically be wired to the control's validity which exists in the field.
+Alternatively you can render it outside of a `Form.Field` part and pass it a `name` prop explicitly.
+
+#### Server-side validation
+
+The component also supports server-side validation using the same `Form.Message` component.
+You can re-use the same messages you defined for client-side errors by passing a `forceMatch` prop which will for the message to show regardless of the client-side matching logic. If the message doesn't exist on the client-side, you can render a `Form.Message` without a `match` too.
```
SoYoung210, line 211: Maybe it's just me, but I feel like the validation representation is a bit cluttered.

There are three main props involved in validation: `math`, `forceMath`, and `serverInvalid`, and I think it can be difficult to understand the relationship between them when first using this component.

- As described in the documentation, the `forceMath` prop can display messages regardless of client-side matching. While this is a useful feature, it can also be thought of as a sort of hack prop in that it "forces" something to be shown.
- If `serverInvalid` is false (not undefined) and `forceMatch` on `Form.Message` is true (if server-side or client-side validation results in an error), it's confusing whether the field evaluates as valid or invalid.

Why not limit the places where server-side validation can be handled to either `Form.Field` or `Form.Message`?

```jsx
const serverErrors = { missingValue: true, customError: true }

<Form.Field name="name" errors={serverErrors}>
  <Form.Label>Full name</Form.Label>
  <Form.Control />
    <Form.Message match="missingValue">
      Please enter your name.
    </Form.Message>
    <Form.Message match="customError">
      My Custom Error Message.
    </Form.Message>
</Form.Field>
```

rfcs/2023-radix-form-primitive.md
```diff
@@ -0,0 +1,308 @@
...
+  </Form.Message>
+</Form.Field>
+```
+
+## Open questions
+
+- Does this API make sense?
+  - Is it easy to use?
+  - Is everything named as you'd expect?
+- Does this API cover most form use-cases? Have we missed some use-cases?
```
SoYoung210, line 307: Q. How should I express different validations when one field is dependent on another?

For example, if I have a `country` field and an `age` field, as shown below, and I need the max/min value of age to be different depending on the value of the country field, how can I use it?

```jsx
<Form.Field name="country">
  <Form.Label>Country</Form.Label>
  <Form.Control asChild>
    <select>
      <option value="uk">United Kingdom</option>
      <option value="us">United States</option>
    </select>
  </Form.Control>
</Form.Field>

<Form.Field name="age">
  <Form.Label>Age</Form.Label>
  <Form.Control type="number" min="18" max="57" step="1" />
</Form.Field>
```

jjenzz: you can use react state as you usually would:

```jsx
const [country, setCountry] = React.useState('');

<Form.Field name="country">
  <Form.Label>Country</Form.Label>
  <Form.Control asChild>
    <select value={country} onChange={event => setCountry(event.currentTarget.value)}>
      <option value="uk">United Kingdom</option>
      <option value="us">United States</option>
    </select>
  </Form.Control>
</Form.Field>

<Form.Field name="age">
  <Form.Label>Age</Form.Label>
  <Form.Control type="number" min={country === 'uk' ? 18 : 21} max="57" step="1" />
</Form.Field>
```

rfcs/2023-radix-form-primitive.md
benoitgrelard: Yeah, @jjenzz has expressed similar concerns (allbeit maybe for different reasons) and we have chatted about it quite a bit. It might be nice to align closer to usual Radix components using uncontrolled/controlled props but it's not immediately clear how we may achieve that.

In the meantime, I think there's value in putting a release candidate in people's hands so you get to play with it and we can get more pointed feedback earlier.

rfcs/2023-radix-form-primitive.md
benoitgrelard: Or you could also do it using a custom validation function as it has access to the entire form data so it can check other fields values.

franciscohanna92:
Really excited about this! 🤯 

My two cents:
_Field Arrays_, available in both [react-hook-form#useFieldArray](https://react-hook-form.com/api/usefieldarray/) and [remix-validated-form#FIeldArray](https://www.remix-validated-form.io/reference/field-array), is one of those features you don't know how much you need until you do. It would be great to have it as a Form primitive. 

Not having it would mean some people coming up with their own implementation, whic adds a lot of maintenance to users, a would sometimes be a reason to use another lib that supports it (personal experience)

benoitgrelard:
> Really excited about this! 🤯
> 
> My two cents: _Field Arrays_, available in both [react-hook-form#useFieldArray](https://react-hook-form.com/api/usefieldarray/) and [remix-validated-form#FIeldArray](https://www.remix-validated-form.io/reference/field-array), is one of those features you don't know how much you need until you do. It would be great to have it as a Form primitive.
> 
> Not having it would mean some people coming up with their own implementation, whic adds a lot of maintenance to users, a would sometimes be a reason to use another lib that supports it (personal experience)

Hey @franciscohanna92 yeah that is something we have been thinking about. We are likely to include an API to handle such cases.

benoitgrelard:
> Great work on this guys! Really liked it!
> 
> Have you thought about supporting multiple validation strategies? I think that currently it is only possible to validate when submit? it would be nice to have something like:
> 
> ```tsx
> <Form.Field validationStrategy="onSubmit | onBlur | onChange"
>   ...
> </Form.Field>
> ```
> 
> as the Constraint validation API is used this can be done on the user-land with `reportValidity()` but it would be nice if it was builtin

Yes definitely something on the list, or at least something we thought people would ask for having more control over. For now we went with what we believe is a good default UX-wise but I think there's definitely room for that in the future.

rostero1:
Does anyone have an example of how this would work with react-select?  An example would be (1) requires a selection and (2) requires a number greater than 5.

kentcdodds:
This appears to already be implemented: https://www.radix-ui.com/docs/primitives/components/form

Is there a reason this hasn't been merged?

benoitgrelard:
We kept it open for now as the main place for feedback since the component is in "preview" for now.

slavanga:
I really like how simple and composable the `Form` primitive is.

One thing that does not work yet is checking if a form is valid without triggering all validation messages. Natively there are `checkValidity` and `reportValidity`. Calling `checkValidity` on the form should just report the status without showing any messages. Are there other ways to achieve this or could this be improved?

slavanga:
@benoitgrelard yes, calling `checkValidity` works and reports the status but it also triggers all validation messages. showing the messages is unwanted in this case and should only occur when calling `reportValidity`.

benoitgrelard:
> @benoitgrelard yes, calling `checkValidity` works and reports the status but it also triggers all radix validation messages. showing the messages is unwanted in this case and should only occur when calling `reportValidity`.
> 
> Edit: Here's a codepen showing the difference with native behavior: https://codepen.io/chriscoyier/pen/mOzgBZ

I understand the difference, I just assumed it would work the same given we just build on top.
But reading MDN a bit more:

>**checkValidity()**
Returns true if the element's child controls are subject to [constraint validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation) and satisfy those constraints; returns false if some controls do not satisfy their constraints. **Fires an event named [invalid](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/invalid_event) at any control that does not satisfy its constraints; such controls are considered invalid if the event is not canceled.** It is up to the programmer to decide how to respond to false.

So, because we hook onto `onInvalid` the message shows. It looks like the events are cancellable, but we'd have to find a way to decide when to cancel them. I'm not sure that's trivial considering we don't know what caused them… 🤔

slavanga:
@benoitgrelard ok, seems hard to make the native way work while using `onInvalid`.
what about a radix API that reports the validation status? maybe something like `ValidityState` but for the whole form.

benoitgrelard:
> @benoitgrelard ok, seems hard to make the native way work while using `onInvalid`. what about a radix API that checks the validation status? maybe something like `ValidityState` but for the whole form.

Yeah will need to look into it, the way the API call fire events seems counter-intuitive, so not sure if I'm missing something or what. What is the use-case btw?

jjenzz:
@slavanga something [like this](https://codesandbox.io/s/friendly-framework-9312uq?file=/src/App.tsx) would work for now if that's what you need?

slavanga:
@jjenzz the link does not work for me

slavanga:
@jjenzz thank you for putting this together.
it works great! although accessing the form validity state via an API from the Form primitive would be much more convenient.

jjenzz:
agreed, an `onValidityChange` prop that internally does something like my sandbox might be nice :)

snelsi:
Have been using the Form component for a while now, and it's been working fine.

There are two very critical points, which I hope will be fixed before the release:

1) No possibility to use `Radix.Checkbox` and `Radix.Select` primitives together with `Form.Control`
2) No built-in solutions/docs for working with arrays

@benoitgrelard Are there any plans/estimates for these two issues?

benoitgrelard:
@snelsi, definitely planning to work on it yet. No estimates at the moment I'm afraid.

mattandryc:
@benoitgrelard I noticed that error messages are linked to the `input` via `aria-describedby` at this point. Is there a reason for using `aria-describedby` over `aria-errormessage` here? 

Screen Shot 2023-06-20 at 9 15 55 PM

kentcdodds:
Sadly, support for that aria attribute is lacking, so it's best to use describedby

jkohlin:
> 

Radix.Select need to be able to get an id attribute from Form.Control, or at least be able to set an id attribute to Radix.Select

PR #1978, opened 2023-02-22
RFC: Radix Auth project

benoitgrelard:
This RFC proposes adding a new project under the Radix UI umbrella (at the same level as "Radix Primitives", "Radix Colors", "Radix Icons"). This new product would be called "Radix Auth" and would provide a set of components specifically tailored to building great authentication experiences.

It would build on top of the new `Form` primitive discussed in the [Radix Form primitive RFC](https://github.com/radix-ui/primitives/pull/1977) but provide higher level components geared towards specific scenarios such as sign-up, sign-in, password reset flows, etc.

### [View rendered RFC](https://github.com/radix-ui/primitives/blob/radix-auth-rfc/rfcs/2023-radix-auth.md)

We would love your input on this 🙏

### DIFF
--- a/rfcs/2023-radix-auth.md
+++ b/rfcs/2023-radix-auth.md
@@ -1,0 +1,135 @@
+- Start Date: 2023-02-21
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: Benoît Grélard
+
+# Radix Auth
+
+## Summary
+
+This RFC proposes adding a new project under the Radix UI umbrella (at the same level as "Radix Primitives", "Radix Colors", "Radix Icons"). This new product would be called "Radix Auth" and would provide a set of components specifically tailored to building great authentication experiences.
+
+It would build on top of the new `Form` primitive discussed in the [Radix Form primitive RFC](2023-radix-form-primitive.md) but provide higher level components geared towards specific scenarios such as sign-up, sign-in, password reset flows, etc.
+
+## Motivation
+
+At [WorkOS](https://workos.com), we build enterprise-ready features such as Single Sign-On, Directory-Sync and Audit Logs. A lot of these features revolve around great authentication experiences.
+
+We also build and maintain [Radix UI](https://radix-ui.com). We believe we could bring a similar level of quality to authentication experiences as people are used to with Radix primitives. Think everything you love about Radix: similar multi-parts API, great accessibility, great developer experience, etc.
+
+## Design
+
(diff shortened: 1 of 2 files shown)


jonathangiardino:
First of all, this is great and would be very useful. 

> What would you like to see in this project?
> What use cases should it cover?

A great use case could be an error boundary/component that automatically returns the native HTML form validation errors, and the possibility of having them localized using HTML Language declaration passing the current locale or the browser preferred language e.g `<Error lang="en" />`.

blechatellier:
Would be good to have some social auth logos (ie: Google) in [radix-icons](https://icons.radix-ui.com/).

saurabhan:
This is great, we can also include OTP flow/Input as it is used for both sign in and sign up 2fa flows.

rfcs/2023-radix-auth.md
```diff
@@ -0,0 +1,135 @@
...
+
+https://user-images.githubusercontent.com/1539897/220597235-316c0251-0173-4c37-a522-f0b19cec447d.mp4
+
+---
+
+#### Other primitives to visually represent password strength
+
+https://user-images.githubusercontent.com/1539897/220597276-8896cf0f-7e8b-4d91-8c23-2a6a2ea866af.mp4
+
+Over time, we would keep providing value by incorporating learnings from our own products around single sign-on, multi-factor authentication, etc.
```
joaopcm, line 47: What if we allow adding a custom function to calculate the strength of a password? Some products may force special characters; some may not. Others may require at least one capital letter, and some may not 🤔

We should have a standard strength calculation, but allowing overwriting of this calculation seems interesting to reach different requirements for each project.

Maybe something like:
## Donut
```jsx
<Donut.Root strengthEstimator={myCustomPasswordStrengthEstimator}>
   <Donut.Track />
   <Donut.Value />
</Donut.Root>
```

Here, my `myCustomPasswordStrengthEstimator` function may return an integer as the percentage where 100% is strong enough, and 0% is too weak.

## Bars
```jsx
<Bars.Root strengthEstimator={myCustomPasswordStrengthEstimator}>
   <Bars.Bar />
   <Bars.Bar />
   <Bars.Bar />
   <Bars.Bar />
   <Bars.Bar />
</Bars.Root>
```

Here, my `myCustomPasswordStrengthEstimator` function may return an integer as the percentage where 100% is strong enough, and 0% is too weak. We could get the result and divide it by the number of `<Bars.Bar />`'s to calculate which bars should be filled and which should not (maybe using `Math.floor`).

What do you think?

benoitgrelard: At the moment the prototypes for those are more low-level so you could wire them to whatever rules you want/have.
However, I can absolutely see some built-in way to provide rules too. We have already explored password strength validation which could check solid strength using https://github.com/dropbox/zxcvbn for example, and also check password breach databases.

joaopcm: Got it! Sounds good!

PR #1988, opened 2023-02-26
[Slider] Clamp thumbs position within range

kombucha:
### Description
It is possible to pass values that are outside of the defined `min` and `max`, in which case the positioning of the ui elements currently behaves unexpectedly. This change should bring the behaviour closer to what the native range input does.

- Clamp thumb and range position percentages between 0 and 100 to avoid stretching UI outside of its natural boundaries
- Added 2 cases in the Slider's Chromatic story to prevent regressions when dealing with out of bound values 🧪 

In the before, the slider is stretched beyond its natural width (200px) 👀 
|Before|After|
|-|-|
|Screenshot 2023-02-26 at 21 51 58|Screenshot 2023-02-26 at 21 53 28|


packages/react/slider/src/Slider.tsx
```diff
@@ -663,10 +663,20 @@ function getNextSortedValues(prevValues: number[] = [], nextValue: number, atInd
...
+ * @example
+ * // returns 50
+ * getPositionPercentage(2, 0, 4)
+ *
+ * @example
+ * // returns 100
+ * getPositionPercentage(2000, 0, 4)
+ */
+function getPositionPercentage(value: number, min: number, max: number) {
+  return clamp(100 * ((value - min) / (max - min)), [0, 100]);
```
benoitgrelard, line 679:
```suggestion
function convertValueToPercentage(value: number, min: number, max: number) {
  const maxSteps = max - min;
  const percentPerStep = 100 / maxSteps;
  const percentage = percentPerStep * (value - min);
  return clamp(percentage, [0, 100]);
```

I'd rather we keep the original name and update the code like this ☝️ as it's clearer IMO.

Also don't think we need the docs above as the code is then self describing.

packages/react/slider/src/Slider.test.tsx
benoitgrelard, line 1: Thanks for adding tests for this.
However, I am really not a fan of how it's testing implementation details.
ie. if we ever changed the method and sized it with an explicit `width` these would all break unnecessarily.
Instead of these unit tests which don't bring much value, I think we should add 2 additional cases in the Chromatic story for out of bounds values (negative/positive) so we'll know instantly if we break it.

benoitgrelard (changes requested):
Thanks for your contribution @kombucha, I've left a few comments.

packages/react/slider/src/Slider.tsx
kombucha: I applied your suggestions as is 👍 
The change in the implementation was unnecessary and readability is subjective.

I do think the jsdoc helped clarify the behaviour of the function: it's not obvious from the name `convertValueToPercentage` that the output will be clamped (130% is a valid percentage after all ^^)

packages/react/slider/src/Slider.test.tsx
kombucha: 100% agree it did feel icky, I didn't realise you guys had visual regression testing in place 😍 
I added the two test cases to the Chromatic story 👍

packages/react/slider/src/Slider.tsx
kombucha: Alright np

PR #1995, opened 2023-03-04
Fix position when use anchor css properties

### LINKED ISSUE #1968: [Select] wrong position of content
As you can see in the video the content has wrong position (because of wrong transform) only when i open the select. If i scroll the position changes and it get fixed.
https://user-images.githubusercontent.com/63153191/220315409-35fc6c46-acb8-4545-8b44-6c998f80d1a8.mov

"radix-ui": "^1.0.1",
"react": "18.2.0",
yarn

### LINKED ISSUE #2017: Popover "Constrain the content size" example doesn't align properly
## Bug report

As per this example, it should be possible to constrain the width to e.g. trigger with with CSS variables: https://www.radix-ui.com/docs/primitives/components/popover#constrain-the-content-size

This works, however, messes up the alignment if alignment is set to "center" (until. If alignment is set to "start", it works fine, but the arrow is still misaligned until window is resized.

### Reproducible example

[CodeSandbox example](https://codesandbox.io/s/radix-ui-popover-bug-hrub04?file=/src/App.js)

### Your environment

Tested on Safari and Chrome, popover v1.0.5

joaom00:
### Description

[Reset the middleware lifecycle](https://floating-ui.com/docs/middleware#resetting-the-lifecycle) to calculate the right position when using `--radix-popper-anchor-width` or `--radix-popper-anchor-height`.

See the behavior in some primitives although works with `Select`

https://codesandbox.io/p/sandbox/cocky-wood-wgqnuy

Fixes #1968
Fixes #2017

### DIFF
--- a/packages/react/popper/src/Popper.tsx
+++ b/packages/react/popper/src/Popper.tsx
@@ -193,11 +193,11 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
         avoidCollisions ? flip({ ...detectOverflowOptions }) : undefined,
         size({
           ...detectOverflowOptions,
-          apply: ({ elements, rects, availableWidth: width, availableHeight: height }) => {
+          apply: ({ elements, rects, availableWidth, availableHeight }) => {
             const { width: anchorWidth, height: anchorHeight } = rects.reference;
             const contentStyle = elements.floating.style;
-            contentStyle.setProperty('--radix-popper-available-width', `${width}px`);
-            contentStyle.setProperty('--radix-popper-available-height', `${height}px`);
+            contentStyle.setProperty('--radix-popper-available-width', `${availableWidth}px`);
+            contentStyle.setProperty('--radix-popper-available-height', `${availableHeight}px`);
             contentStyle.setProperty('--radix-popper-anchor-width', `${anchorWidth}px`);
             contentStyle.setProperty('--radix-popper-anchor-height', `${anchorHeight}px`);
           },
--- a/packages/react/popper/src/Popper.tsx
+++ b/packages/react/popper/src/Popper.tsx
@@ -181,7 +181,6 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
       placement: desiredPlacement,
       whileElementsMounted: autoUpdate,
       middleware: [
-        anchorCssProperties(),
         offset({ mainAxis: sideOffset + arrowHeight, alignmentAxis: alignOffset }),
         avoidCollisions
           ? shift({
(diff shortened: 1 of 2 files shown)


benoitgrelard:
I can see that fixes the issue, but I'm a little unclear as to why it needs to be in this middleware, or why it happens.
Do you have a bit more understanding here @joaom00?

benoitgrelard:
I see that makes sense.

> In fact, this code is very similar to the one internally in the size middleware, if you want we can set the anchor CSS variables there and remove the anchorCssProperties middleware.

If we did this, would this mean we don't need all this extra logic as it would already be invalidated by the internals of the size middleware?

joaom00:
> If we did this, would this mean we don't need all this extra logic as it would already be invalidated by the internals of the size middleware?

Yes, what I understood is that `size()` was made to perform any mutation in the size of the floating element

benoitgrelard:
Let's try that then as I think it might make more sense to group anything size related within that middleware and not have the extra invalidation code as its the same concern.

joaom00:
I had to change the position of the `arrow` middleware to after the `size` because it was not able to position correctly

packages/react/popper/src/Popper.tsx
```diff
@@ -191,15 +190,22 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
+          apply: ({ elements, rects, availableWidth: width, availableHeight: height }) => {
+            const { width: popperWidth, height: popperHeight } = rects.reference;
+
             elements.floating.style.setProperty('--radix-popper-available-width', `${width}px`);
             elements.floating.style.setProperty('--radix-popper-available-height', `${height}px`);
+            elements.floating.style.setProperty('--radix-popper-anchor-width', `${popperWidth}px`);
+            elements.floating.style.setProperty(
+              '--radix-popper-anchor-height',
+              `${popperHeight}px`
+            );
```
benoitgrelard, line 205:
```suggestion
          apply: ({ elements, rects, availableWidth, availableHeight }) => {
            const { width: anchorWidth, height: anchorHeight } = rects.reference;
            const contentStyle = elements.floating.style;
            contentStyle.setProperty('--radix-popper-available-width', `${availableWidth}px`);
            contentStyle.setProperty('--radix-popper-available-height', `${availableHeight}px`);
            contentStyle.setProperty('--radix-popper-anchor-width', `${anchorWidth}px`);
            contentStyle.setProperty('--radix-popper-anchor-height', `${anchorHeight}px`);
```

benoitgrelard: Just in case you didn't know, you can accept suggestions and commit them straight away from GitHub!

benoitgrelard: No worries! I think you missed one last bit from my suggestions, let's not rename the available width/height stuff now as it's all in the same context, it's clearer named correctly.

joaom00: Done

benoitgrelard (approved):

UTkzhang:
+ 1 still seeing this issue in react-dropdown-menu

PR #2003, opened 2023-03-07
[Small Feat] Use inertOthers instead of hideOthers

frykten:
### Description

This is not a big feature; while trying to open a PR on another issue (the forced "modal mode" the Select component), I found that aria-hidden had been updated and had implemented the "inertOthers".
Inert is the new [W3C's](https://html.spec.whatwg.org/multipage/interaction.html#modal-dialogs-and-inert-subtrees) way to hide elements, instead of relying on an aria attribute.

### DIFF
--- a/.yarn/versions/0f4adb39.yml
+++ b/.yarn/versions/0f4adb39.yml
@@ -1,0 +1,12 @@
+releases:
+  "@radix-ui/react-dialog": minor
+  "@radix-ui/react-menu": minor
+  "@radix-ui/react-popover": minor
+  "@radix-ui/react-select": minor
+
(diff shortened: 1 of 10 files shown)


BrianHung:
`suppressOthers` might be better here if `inert` is not supported.

https://github.com/theKashey/aria-hidden/blob/9220c8f4a4fd35f63bee5510a9f41a37264382d4/src/index.ts#L207C14-L207C29

chaance:
Yeah I'd definitely prefer `suppressOthers` in this case for back compat. Closing for now, but ping me if you want to update the PR and I can re-open.

Issue #1193, opened 2022-02-24
[Select] Remaining tasks


benoitgrelard:
- Add chromatic stories
- Add tests
- Improve mobile experience
  - pointer handling works well when using slowly, but it has the tendancy to close from under you when going quickly, need to check what's going on here
  - on mobile, there's an issue with positioning when the page is pinch-zoomed. I think it might be due to visual viewport vs. layout viewport and the fact we position: fixed.
- Investigate to see if we can add support for animation (opening/closing)
- see if we can improve this scenario: if I press down on the `ScrollUpButton` or `ScrollDownButton` and hold while it scrolls, when I release, it will select the first or last item because the scroll button disappears and mouse is now released on an item

Kinbaum:
@benoitgrelard Can you also look at the option of supplying a max-height to the content / viewport? Currently when setting the max-height for the content and attempting to scroll, the container sticks to the bottom of the window.

Kinbaum:
Also, when using the select with groups of items, the separator is being announced as a group. 
In the example you have with food, the screen reader says there's 5 groups when there are only 3 (plus 2 separators)

benoitgrelard:
Hey @Kinbaum, someone is keen! haha
How have been playing with it given it's not even out yet 😀
Guessing you're following the PRs/npm, etc? 😀

In any case:

> Also, when using the select with groups of items, the separator is being announced as a group.
In the example you have with food, the screen reader says there's 5 groups when there are only 3 (plus 2 separators)

I've handled that one and made the separators visual only for now.

> Can you also look at the option of supplying a max-height to the content / viewport? Currently when setting the max-height for the content and attempting to scroll, the container sticks to the bottom of the window.

I don't think I understand what you mean here.

jjenzz:
Ah I see the issue here...

![CleanShot 2022-02-25 at 10 03 14](https://user-images.githubusercontent.com/175330/155696466-46459994-c4bd-4331-90eb-5392ade9614d.gif)

@Kinbaum it would be good to understand your use-case? why do you need to set a `max-height`?

benoitgrelard:
This version of `Select` is designed to function like the MacOS select in terms of UX. So the heuristics are basically to present as many options as possible on the screen whilst maintaining alignment with the trigger as close as possible.

So setting a `maxHeight` wouldn't be desirable here.

[...]

PR #2004, opened 2023-03-07
[Feat] Add a modal/non-modal mode to the Select component

### LINKED ISSUE #1927: [Feature] [Select | Menu] Make the hideOthers feature optional
### Overview

We started using Radix recently and we found something that is quite an issue on our side: when opening a simple Select or Menu "content", the page scrollbar disappears and there is a small lag.

Both stem from the same line of code (a feature):
```
    React.useEffect(() => {
      if (content) return hideOthers(content);
    }, [content]);
```

That "hideOthers" method/feature, though it has good intentions, breaks the UX for our mouse users as it hides the page scrollbar, and creates a lag in certain pages as it adds to all nodes the aria-hidden property (and we do not have a small website, but a rather big application with some heavy pages/documents and modifying the whole dom is just... not really something we want to do).

We understand that the intention was good and to trap the focus inside the element/popover and only that element/popover. Still, while we see the use for modals/dialogs, for Menus and Selects, we do not see the real plus-value---especially since it hinders our users' experience (page scroll difference, plus performance), and we would like to request a parameter to make that behavior optional via a param.

Personally, I'm inclined to make a PR, if you wish.

It is mostly that other libraries do not use such behavior for elements like Menus/Selects. Just look at GitHub, the Google suite, Spotify, or any tool/app/website you use every day; this is way too brutal for such components that are used in forms, in navigation, and which do not break the flow (contrary to a modal, where it is advised in every specs/docs/tutorial to add such a behavior).
In aria-hidden's documentation, they advise/enforce the use of the addon for modals/dialogs, not for smaller block items.

### Whom does this impact? Who is this for?

This impacts Users---at least, ours.
Making the property optional, with true as default, would not break the implementation of all those who use those components and are happy about it, but it would permit those who are not to fix it.

### Additional context

Additionally, now, the repo aria-hidden has a "suppressOthers" method that turns everything into the new and better `inert`. Maybe you should turn to this method for the Modal/Dialog components?

Another thing/question: while implementing the component, we had to suppress an axe rule "Ensures aria-hidden elements are not focusable nor contain focusable elements"; when axe runs while the popover is opened (whether it is the selection list or the menu dropdown), the component's trigger is aria-hidden as well, even though it is tabbable/interactive and axe doesn't like that very much. Maybe it isn't supposed to be an issue, since the focus is trapped inside the popover, but it felt weird to our Product/QA teams and during User tests.

frykten:
### Description

As described in https://github.com/radix-ui/primitives/issues/1927, when using the Select component, the lack of a modal/non-modal mode may cause issues for some users (at least, our Product team isn't very keen on it).
The removal of the scrollbar, the performance hindrance with the "hideOthers" on some pages, and the event.preventDefault when clicking on another form element are some features that may be great sometimes but not always.

This modal mode was developed in the DropdownMenu and in the Popover; I tried to imitate their code and implementation. I know there are lots of impacted lines, but the component seems old as everything was done in the same method and it resulted in a very cumbersome functional component with very different and imbricated bits, difficult to cleanly refactor. I tried to stay small, so as to not mess with the codebase.

I also added a Story, but I added no tests: I found only one and was unsure of how you treat these.

I have some time that has been dedicated to resolving this, as our designers and owners really dislike that "modal mode"; I am willing to listen to any comments you would add to the PR, limited by my understanding of the codebase, understanding which is but humble.

Closes #1927

### DIFF
--- a/.yarn/versions/c1cdd5c6.yml
+++ b/.yarn/versions/c1cdd5c6.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-select": minor
+
+declined:
+  - primitives
--- a/packages/react/select/src/Select.stories.tsx
+++ b/packages/react/select/src/Select.stories.tsx
@@ -56,6 +56,113 @@ export const Styled = () => (
   </div>
 );
 
+export const Modality = () => (
+  <div
+    style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', minHeight: '110vh' }}
+  >
+    <div style={{ display: 'grid', gap: 50 }}>
+      <div style={{ display: 'inline-flex', alignItems: 'center', flexDirection: 'column' }}>
+        <h1>Modal (default)</h1>
+        {POSITIONS.map((position) => (
+          <Label key={'modal' + position}>
+            Choose a number:
+            <Select.Root defaultValue="two">
+              <Select.Trigger className={triggerClass()}>
+                <Select.Value />
+                <Select.Icon />
+              </Select.Trigger>
+              <Select.Portal>
+                <Select.Content className={contentClass()} position={position} sideOffset={5}>
+                  <Select.Viewport className={viewportClass()}>
+                    <Select.Item className={itemClass()} value="one">
+                      <Select.ItemText>
+                        One<span aria-hidden> 👍</span>
+                      </Select.ItemText>
+                      <Select.ItemIndicator className={indicatorClass()}>
+                        <TickIcon />
+                      </Select.ItemIndicator>
+                    </Select.Item>
+                    <Select.Item className={itemClass()} value="two">
+                      <Select.ItemText>
+                        Two<span aria-hidden> 👌</span>
+                      </Select.ItemText>
+                      <Select.ItemIndicator className={indicatorClass()}>
+                        <TickIcon />
+                      </Select.ItemIndicator>
+                    </Select.Item>
+                    <Select.Item className={itemClass()} value="three">
+                      <Select.ItemText>
+                        Three<span aria-hidden> 🤘</span>
+                      </Select.ItemText>
+                      <Select.ItemIndicator className={indicatorClass()}>
+                        <TickIcon />
+                      </Select.ItemIndicator>
+                    </Select.Item>
+                  </Select.Viewport>
+                  <Select.Arrow />
+                </Select.Content>
+              </Select.Portal>
+            </Select.Root>
+          </Label>
+        ))}
+        <textarea
+          style={{ width: 500, height: 100, marginTop: 10 }}
+          defaultValue="Lorem ipsum dolor sit amet consectetur adipisicing elit. Quaerat nobis at ipsa, nihil tempora debitis maxime dignissimos non amet."
+        />
+      </div>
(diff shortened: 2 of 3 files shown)


frykten:
I'm still around if ya ever wanna discuss this PR/code; though my planning is full at my company, I'm willing to make those kinda subjects go on. Maybe it's easier for you if I join your discord to chitchat about it?

frykten:
@benoitgrelard @andy-hook Heya team! Still willing to do lotsa work on this PR to make it work :) I know it might not be ideal (I do not know much about how you usually work around your own codebase and I merely tried to copycat what I saw on the component and the other components)

The need is real for my team, and it seems to exist for others (cf. [the issue](https://github.com/radix-ui/primitives/issues/1927#issuecomment-1536154220) where I first referenced it, though I hadn't fully understood the issue), and, I guess, for the Select component the need must exist since that feature (modal/non-modal mode) has been released for the other components (the Dropdown menu and the Popover).

benoitgrelard:
Hi @frykten,

Thanks for the contribution.

> but the component seems old as everything was done in the same method and it resulted in a very cumbersome functional component with very different and imbricated bits, difficult to cleanly refactor.

This component is pretty complex, so I understand it's pretty daunting to make changes.
It isn't old though, it just happens that the initial version was a very specific layout (what is now called `item-aligned`) and we then added later support for the usual `popper` positioning (same found in other components).
Thus why see a split with 2 content implementations: `SelectItemAlignedPosition` and  `SelectPopperPosition`.

As explained here: https://github.com/radix-ui/primitives/issues/1193#issuecomment-1124836293, getting the same kind of non-modal for `SelectItemAlignedPosition` will be more difficult.

The current approach in this PR is too dissimilar for us to take it on.
What I suggest and what I'd like to see (if you are still keen to work on this) would be the following:

- First, lets concentrate on a PR that adds non-modal / modal split only for `SelectPopperPosition`
  - This mean there would be an implementation split in this component (like in other primitives like `Dialog` for example):
    - `SelectPopperPositionModal`
    - `SelectPopperPositionNonModal`
- We can then see next how best to bring the behaviour to `SelectItemAlignedPosition`, however it will have to be different (ie. close on scroll for example).

---

I have done these kinds of changes enough to know that it's a lot of juggling around, and refactoring takes a while.
Let me know if you still feel like you want to get on with it, or if you'd rather we take it on.

✌️

frykten:
Hey @benoitgrelard !

Thank you for the feedback and the care for the subject.

Personally, I am willing to work on it; I am currently in a rush at my company but I may be able to work on this subject next week.

I'll keep you tuned on whether I am to do all this "juggling around" or not (company schedules do not always give love to open-source unfortunately), and I hope to push some commits by the end of next week.

joaom00:
Hey @frykten,

you said on discord that you wouldn't have time for that PR, unfortunately. Would you mind if I take care of it?

Thanks!

frykten:
Heya @joaom00 !

Yes, of course! As I said, I am currently unable to make this subject move on, so, if you wish to make it work, I'll only be glad that it is taken care of.

You have my thanks :slightly_smiling_face:

frykten:
@joaom00 @benoitgrelard Hello!

Any news on this PR? This creates annoying behaviours in inline forms for users and as I continue to see thumbs up in the different issues/discords, I guess it still would be appreciated by lots.

PR #2014, opened 2023-03-08
[Slider] better extend Thumb styles

greg-schrammel:
### Description

Hey created a pr but maybe there is a something I'm missing

It removes the `Thumb` wrapping `span` tag and add the positioning styles straight to the `Primitive.span`
so that I can for example use a transform transition and have the thumb animate nicely to it's new value

here's a small before and after of applying a tailwind `transitions-all` class to the `Thumb` component

https://user-images.githubusercontent.com/6232729/223790318-40cbdc76-2988-4ef0-80be-cbaef8b9a58f.mov

https://user-images.githubusercontent.com/6232729/223790332-5ba0178c-839d-4061-b5cc-ca107576e7c8.mov

### DIFF
--- a/packages/react/slider/src/Slider.tsx
+++ b/packages/react/slider/src/Slider.tsx
@@ -586,13 +586,6 @@ const SliderThumbImpl = React.forwardRef<SliderThumbImplElement, SliderThumbImpl
     }, [thumb, context.thumbs]);
 
     return (
-      <span
-        style={{
-          transform: 'var(--radix-slider-thumb-transform)',
-          position: 'absolute',
-          [orientation.startEdge]: `calc(${percent}% + ${thumbInBoundsOffset}px)`,
-        }}
-      >
         <Collection.ItemSlot scope={props.__scopeSlider}>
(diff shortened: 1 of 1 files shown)


benoitgrelard:
Hi @greg-schrammel,

Unfortunately I don't think this is a change we should make for a couple of reasons:
- The wrapper is important because of the built-in transform. If we remove it, it would mean it's impossible to have other transforms on it without breaking the whole thing. For example you might have a slight scale transform on hover
- Although it might look fine when controlling with keyboard, it's probably a bad idea in general UX-wise as it will break the direct manipulation feeling of dragging the thumb:

https://user-images.githubusercontent.com/1539897/225619903-21f91ee4-94cc-4b90-ba85-0755550de2f6.mp4

Based on those reasons, I will close the PR but please feel free to discuss still and if I've missed something we can always re-open it.

greg-schrammel:
yes I ended up having to remove the transition animation on focus, do you have a better idea on how to implement that transition with the current design?

PR #2015, opened 2023-03-09
[Form] Export FormLabel correctly

joaom00:
### Description

`FormLabel` is not exported correctly as `Label`.


benoitgrelard:
Thanks @joaom00 this is actually due to a parcel bug that we usually go around writing our exports differently and I completely forgot on this one. I'll close yours in favour of #2016

PR #2027, opened 2023-03-16
[RovingFocusGroup] Make roving focus group tabbable by default until we know if we have items or not

### LINKED ISSUE #1910: First item in radio group does not automatically receive focus when dialog is opened
### Current Behavior

When a radio group is nested inside a Radix UI dialog, the first item in the radio group does not receive focus when the dialog is opened even if it is the first focusable element.

### Expected behavior

The first item in the radio group should automatically receive focus when a dialog is opened and it is the first focusable element. This would mimic the behavior of native radio inputs, as well as the behavior of the Headless UI radio group component.

### Reproducible example

[StackBlitz Reproduction](https://stackblitz.com/edit/vitejs-vite-17yzf7?file=src/App.tsx) (Note: the preview needs to be opened in a separate window in order for the styles indicating the focused element to be visible.)

This reproduction shows how the Radix UI radio group implementation compares to both native radio inputs and the radio group component from Headless UI.

### Your environment

| Radix Package(s) |    @radix-ui/react-radio-group     |    1.0.2     |
| React            | n/a     |    18.2.0     |
| Browser          |    Edge     |    109     |
| Operating System |    macOS     |    13.1     |

benoitgrelard:
Fixes #1910

### DIFF
--- a/.yarn/versions/acf16116.yml
+++ b/.yarn/versions/acf16116.yml
@@ -1,0 +1,13 @@
+releases:
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-menubar": patch
+  "@radix-ui/react-radio-group": patch
+  "@radix-ui/react-roving-focus": patch
+  "@radix-ui/react-tabs": patch
+  "@radix-ui/react-toggle-group": patch
+  "@radix-ui/react-toolbar": patch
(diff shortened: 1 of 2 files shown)


andy-hook:
Should we add a test case to cover this?

benoitgrelard:
hmmm, unfortunately that broke something because now if all items of a radiogroup are disabled or the top-level is disabled, the group is still tabbable… :(

Not sure how I can fix this now then…

Issue #2018, opened 2023-03-11
Tooltip doesn't position self properly if div has a drop shadow


Zavien1:
### Current Behavior

Adding a drop shadow to a parent div makes tooltip position itself incorrectly 
<img width="665" alt="Screen Shot 2023-03-11 at 9 14 00 AM" src="https://user-images.githubusercontent.com/29804724/224502310-aadf34c5-80c1-412a-98fa-90d804e86647.png">

<img width="670" alt="Screen Shot 2023-03-11 at 9 14 30 AM" src="https://user-images.githubusercontent.com/29804724/224502332-e846a0a5-7eb2-4fce-b81d-a6784e8da566.png">

### Expected behavior

Dropshadow should not affect tool tip position

### Suggested solution

No clue

### Your environment

React + Tailwind 

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |         |         |
| React            | n/a     |         |
| Browser          |         |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

joaom00:
Do you have a reproducible example?

atomiks:
Radix is using `v0` of Floating UI and is pulling in a version of `@floating-ui/dom` from June 2022, while there have been several fixes since v1 got released in July.

https://github.com/radix-ui/primitives/blob/72018163e1fdb79b51d322d471c8fc7d14df2b59/packages/react/popper/package.json#L20

There aren't any breaking changes between the v0 version used and the latest version, so they can upgrade to `@floating-ui/react-dom^1.0.0` with no effort.  cc: @benoitgrelard 

If you override `@floating-ui/react-dom` + `dom` and `core` to the latest versions manually in your project, is the issue fixed? What browser? This should be the relevant line:

https://github.com/floating-ui/floating-ui/blob/02c127ad9d10800e0ef29e5df9bb458612e737bf/packages/dom/src/utils/is.ts#L60

benoitgrelard:
Yes, I was going to ask too: @Zavien1 are you using the `Tooltip.Portal` part?

dcyoung:
Encountering this issue as well, any resolution?

benoitgrelard:
Hey @atomiks,

[...]

PR #2032, opened 2023-03-19
[Popper] Upgrade floating-ui

### LINKED ISSUE #2128: Popover - invalid value for --radix-popper-available-width
### Current Behavior

Hey :) 

I am trying to create a design system Popover using radix Popover primitives.
We were trying to use `collisionBoundary` to force the Popover to stay in a given container. 

From what I understand, Radix is using [Floating UI](https://floating-ui.com/) to get the positionning data, and set them in css custom properties that we should use to apply the styles.

**For some reason `--radix-popper-available-height` calculation is correct, but `--radix-popper-available-width:-495px; ` it not.**. For it we have huge negative values (ex: `-495px`).

Capture d’écran 2023-05-05 à 14 40 30

### Expected behavior

I need the popover to stay in its `collisionBoundary` element.

Capture d’écran 2023-05-05 à 14 45 33

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/react-tailwind-starter-forked-3uyqqv?file=/src/PopoverDemo.js:602-619)

### Your environment

| Radix Package(s) |     @radix-ui/react-popover    |     1.0.5    |
| React            | n/a     |    18.2.0     |
| Browser          |   Chrome      |    112.0.5615.137     |

lesha1201:
### Description

This PR upgrades major version of `@floating-ui/react-dom` in `Popper` component.

Noticeble release notes from Floating UI:

- https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Fcore%401.0.0
- https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Fdom%401.2.5
- https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Freact-dom%401.1.0
- https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Freact-dom%401.0.1
- https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Freact-dom%402.0.0

The PR does the following:

- Upgrades `@floating-ui/react-dom` to `2.0.0`.
- Changes  `@floating-ui/react-dom` from fixed version to minor version. Since Floating UI uses semantic versioning, it should be safe to use minor version and it will reduce a chance of duplicate packages for Floating UI for the end users.
- Removes unnecessary logic/workarounds after the upgrade.
- Removed filtering falsy values in `middleware` because now it's done inside `@floating-ui/react-dom`.

Affected issues:
- Fixes #2128

### DIFF
--- a/packages/react/popper/package.json
+++ b/packages/react/popper/package.json
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.13.10",
-    "@floating-ui/react-dom": "^1.3.0",
+    "@floating-ui/react-dom": "^2.0.0",
     "@radix-ui/react-arrow": "workspace:*",
     "@radix-ui/react-compose-refs": "workspace:*",
     "@radix-ui/react-context": "workspace:*",
--- a/packages/react/popper/package.json
+++ b/packages/react/popper/package.json
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.13.10",
-    "@floating-ui/react-dom": "0.7.2",
+    "@floating-ui/react-dom": "^1.3.0",
     "@radix-ui/react-arrow": "workspace:*",
     "@radix-ui/react-compose-refs": "workspace:*",
     "@radix-ui/react-context": "workspace:*",
--- a/packages/react/popper/src/Popper.tsx
+++ b/packages/react/popper/src/Popper.tsx
@@ -110,7 +110,6 @@ const [PopperContentProvider, useContentContext] =
 
 const [PositionContextProvider, usePositionContext] = createPopperContext(CONTENT_NAME, {
   hasParent: false,
-  positionUpdateFns: new Set<() => void>(),
 });
 
 type Boundary = Element | null;
@@ -175,43 +174,40 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
       altBoundary: hasExplicitBoundaries,
     };
 
-    const { reference, floating, strategy, x, y, placement, isPositioned, middlewareData, update } =
-      useFloating({
-        // default to `fixed` strategy so users don't have to pick and we also avoid focus scroll issues
-        strategy: 'fixed',
-        placement: desiredPlacement,
-        whileElementsMounted: autoUpdate,
-        middleware: [
-          offset({ mainAxis: sideOffset + arrowHeight, alignmentAxis: alignOffset }),
-          avoidCollisions &&
-            shift({
-              mainAxis: true,
-              crossAxis: false,
-              limiter: sticky === 'partial' ? limitShift() : undefined,
-              ...detectOverflowOptions,
-            }),
-          avoidCollisions && flip({ ...detectOverflowOptions }),
-          size({
+    const { refs, floatingStyles, placement, isPositioned, middlewareData } = useFloating({
+      // default to `fixed` strategy so users don't have to pick and we also avoid focus scroll issues
+      strategy: 'fixed',
+      placement: desiredPlacement,
+      whileElementsMounted: autoUpdate,
+      elements: {
+        reference: context.anchor,
+      },
+      middleware: [
+        offset({ mainAxis: sideOffset + arrowHeight, alignmentAxis: alignOffset }),
+        avoidCollisions &&
+          shift({
+            mainAxis: true,
+            crossAxis: false,
+            limiter: sticky === 'partial' ? limitShift() : undefined,
             ...detectOverflowOptions,
-            apply: ({ elements, rects, availableWidth, availableHeight }) => {
-              const { width: anchorWidth, height: anchorHeight } = rects.reference;
-              const contentStyle = elements.floating.style;
-              contentStyle.setProperty('--radix-popper-available-width', `${availableWidth}px`);
-              contentStyle.setProperty('--radix-popper-available-height', `${availableHeight}px`);
-              contentStyle.setProperty('--radix-popper-anchor-width', `${anchorWidth}px`);
-              contentStyle.setProperty('--radix-popper-anchor-height', `${anchorHeight}px`);
-            },
           }),
-          arrow && floatingUIarrow({ element: arrow, padding: arrowPadding }),
-          transformOrigin({ arrowWidth, arrowHeight }),
-          hideWhenDetached && hide({ strategy: 'referenceHidden' }),
-        ],
-      });
-
-    // assign the reference dynamically once `Content` has mounted so we can collocate the logic
-    useLayoutEffect(() => {
-      reference(context.anchor);
-    }, [reference, context.anchor]);
+        avoidCollisions && flip({ ...detectOverflowOptions }),
+        size({
+          ...detectOverflowOptions,
+          apply: ({ elements, rects, availableWidth, availableHeight }) => {
+            const { width: anchorWidth, height: anchorHeight } = rects.reference;
+            const contentStyle = elements.floating.style;
+            contentStyle.setProperty('--radix-popper-available-width', `${availableWidth}px`);
+            contentStyle.setProperty('--radix-popper-available-height', `${availableHeight}px`);
+            contentStyle.setProperty('--radix-popper-anchor-width', `${anchorWidth}px`);
+            contentStyle.setProperty('--radix-popper-anchor-height', `${anchorHeight}px`);
+          },
+        }),
+        arrow && floatingUIarrow({ element: arrow, padding: arrowPadding }),
+        transformOrigin({ arrowWidth, arrowHeight }),
+        hideWhenDetached && hide({ strategy: 'referenceHidden' }),
+      ],
+    });
 
     const [placedSide, placedAlign] = getSideAndAlignFromPlacement(placement);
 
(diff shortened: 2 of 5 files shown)


packages/react/popper/src/Popper.tsx
```diff
@@ -279,8 +279,8 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
           position: strategy,
           left: 0,
           top: 0,
-          transform: isPlaced
-            ? `translate3d(${Math.round(x)}px, ${Math.round(y)}px, 0)`
+          transform: isPositioned
+            ? `translate3d(${Math.round(x as number)}px, ${Math.round(y as number)}px, 0)`
```
lesha1201, line 283: Checking `isPositioned` making sure that `x` and `y` are not `null`.

packages/react/popper/package.json
lesha1201, line 1: I didn't upgrade version for the package because I'm not sure what version it should be. My suggestion is to bump minor version since this component heavily relies on Floating UI but doesn't bring any breaking changes to the end users.

lesha1201:
@benoitgrelard @andy-hook Sorry for pinging. Can you please review? Does this PR is blocked by something? It would be nice to have this PR released because some of my projects are already using a new Floating UI version and installing radix (Popper, Tooltip) results in duplicate packages.

atomiks:
@lesha1201 in the latest version of `@floating-ui/react-dom`, there are quite a few changes that make a lot of this even simpler. I'll try to review them

packages/react/popper/src/Popper.tsx
```diff
@@ -249,12 +248,12 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
     // https://github.com/floating-ui/floating-ui/issues/1531
     useLayoutEffect(() => {
-      if (isRoot && isPlaced) {
+      if (isRoot && isPositioned) {
         Array.from(positionUpdateFns)
           .reverse()
           .forEach((fn) => requestAnimationFrame(fn));
       }
-    }, [isRoot, isPlaced, positionUpdateFns]);
+    }, [isRoot, isPositioned, positionUpdateFns]);
```
atomiks, line 256: This should no longer be necessary since `@floating-ui/dom@1.0.5`

packages/react/popper/src/Popper.tsx
atomiks: You can use the new `floatingStyles` object returned from the hook `v2` instead of computing the styles here. It performs rounding based on DPR and should be more accurate.

This point: `// keep off the page when measuring` shouldn't matter because the browser doesn't paint until it's positioned anyway.

packages/react/popper/src/Popper.tsx
```diff
@@ -175,53 +175,52 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
-      placement: desiredPlacement,
-      whileElementsMounted: autoUpdate,
-      middleware: [
-        offset({ mainAxis: sideOffset + arrowHeight, alignmentAxis: alignOffset }),
-        avoidCollisions
-          ? shift({
+    const { reference, floating, strategy, x, y, placement, isPositioned, middlewareData, update } =
+      useFloating({
+        // default to `fixed` strategy so users don't have to pick and we also avoid focus scroll issues
+        strategy: 'fixed',
```
atomiks, line 181: `isPositioned` lets you wait until the position is ready now and stick to `absolute`, but that's probably a more intensive change for just upgrading

packages/react/popper/src/Popper.tsx
```diff
@@ -175,53 +175,52 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
+          avoidCollisions && flip({ ...detectOverflowOptions }),
+          size({
+            ...detectOverflowOptions,
+            apply: ({ elements, rects, availableWidth, availableHeight }) => {
+              const { width: anchorWidth, height: anchorHeight } = rects.reference;
+              const contentStyle = elements.floating.style;
+              contentStyle.setProperty('--radix-popper-available-width', `${availableWidth}px`);
+              contentStyle.setProperty('--radix-popper-available-height', `${availableHeight}px`);
+              contentStyle.setProperty('--radix-popper-anchor-width', `${anchorWidth}px`);
+              contentStyle.setProperty('--radix-popper-anchor-height', `${anchorHeight}px`);
```
atomiks, line 202: Bugs with this that people have been filing are fixed in the latest version of `@floating-ui/dom`

lesha1201:
@atomiks Thank you for the review! I will it handle it when I can (I hope this weekend).

packages/react/popper/src/Popper.tsx
lesha1201: @atomiks I agree, I would better leave it as it is for now.

packages/react/popper/src/Popper.tsx
lesha1201: > the browser doesn't paint until it's positioned anyway

@atomiks Why not? I tried to remove this and in heavily loaded story, the difference is noticable.


https://github.com/radix-ui/primitives/assets/10157660/beb1d296-fe6a-4598-9444-3e9da5d0d1cf

packages/react/popper/src/Popper.tsx
lesha1201: @atomiks Can you please give me the context? What bugs? I think removing this is a breaking change since users might use these variables.

packages/react/popper/src/Popper.tsx
atomiks: I see! Yeah, for SSR etc, then keeping it out of view (or using `visibility: isPositioned ? undefined : 'hidden'`) is the correct approach.

packages/react/popper/src/Popper.tsx
lesha1201: Oh, I found one issue. Now I get it.

- https://github.com/radix-ui/primitives/issues/2128

lesha1201: Added a story for Popover from the issue.

packages/react/popper/src/Popper.tsx
lesha1201: Done

packages/react/popper/src/Popper.tsx
```diff
@@ -175,53 +175,52 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
...
+          arrow && floatingUIarrow({ element: arrow, padding: arrowPadding }),
+          transformOrigin({ arrowWidth, arrowHeight }),
+          hideWhenDetached && hide({ strategy: 'referenceHidden' }),
+        ],
+      });
 
     // assign the reference dynamically once `Content` has mounted so we can collocate the logic
     useLayoutEffect(() => {
       reference(context.anchor);
     }, [reference, context.anchor]);
```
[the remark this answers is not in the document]
lesha1201: Done

packages/react/popper/src/Popper.tsx
lesha1201: I just noticed that there is `translate` in `floatingStyles` and `translate3d` in Radix. I will change it to `translate` to make it consistent.

atomiks: Yeah, `willChange: 'transform` makes `translate` act as `translate3d`. It's only applied to high PPI displays to reduce blurring issue on Windows

benoitgrelard (commented):
Thanks for doing this both! I'll make a few more changes and test out our side.

packages/react/popper/src/Popper.tsx
benoitgrelard: CleanShot 2023-05-17 at 13 42 08@2x

> It performs rounding based on DPR and should be more accurate.

@atomiks I'm guessing this is what we're seeing here caught up by Chromatic?

atomiks: I'd assume so yes

benoitgrelard:
hmm, weird, some more differences in the latest chromatic run, only in some of the tooltip stories:

CleanShot 2023-05-17 at 19 45 19@2x

Timing?
That's worrying though… 🤔

atomiks:
@benoitgrelard Timing is my guess yes. 

Looking locally, all the tooltips paint at the same time (including with 6x slowdown), so I'm not sure how that screenshot is happening.

My guess is that multiple screenshots are being taken, then being stitched together, and some are being taken too early, before the JS measurement logic has run when loading the route. This only coincidentally didn't appear before as maybe it was initializing more quickly.

lesha1201:
@benoitgrelard I also think that this is timing because the performance is about the same on my machine as well. Maybe we should set delay for that Story to make it more reliable https://www.chromatic.com/docs/delay?

benoitgrelard:
Yeah we might have to introduce some delay to some of those stories. We do this in places already.
Otherwise, I fear we will keep getting some false negatives like this.

Btw, @atomiks, have you seen this my comment here? https://github.com/radix-ui/primitives/issues/2018#issuecomment-1551243712

atomiks:
Yeah I forgot, feel free to open an issue!

benoitgrelard:
> Yeah I forgot, feel free to open an issue!

https://github.com/floating-ui/floating-ui/issues/2333

packages/react/popper/src/Popper.tsx
benoitgrelard: > isPositioned lets you wait until the position is ready now and stick to absolute, but that's probably a more intensive change for just upgrading

@atomiks, can you tell me more about this, specifically around the `absolute` strategy? Are you saying `absolute` should work for everything now? If so, why is there still a strategy option? This is something that we know we want to work out at some point because although `fixed` works consistently for us, we know it isn't the most performant when scrolling, resizing, etc.

atomiks: It means that you can prevent scroll jumps using `absolute`, by doing this:

```js
useLayoutEffect(() => {
  isPositioned && element.focus();
}, [isPositioned]);
```

`fixed` is useful in two cases:

- The reference element is in a fixed container (or is fixed itself) and the floating element isn't inside the fixed container. This prevents jitters when scrolling.
- "Breaking" the floating element out of a clipping container in many cases (when a containing block isn't being formed) to keep DOM context in tact. Not relevant if you're portalling.

benoitgrelard:
I haven't see anymore false positive like [this one](https://github.com/radix-ui/primitives/pull/2032#issuecomment-1551890793), so won't introduce delays yet. If we notice them, we can add them then.

benoitgrelard (approved):
Thanks @lesha1201 and @atomiks for this, I've tested my side, and it all looks great. Merging now 🎉

Issue #2034, opened 2023-03-20
[Tabs] Unable to switch from tabpanel to another with @testing-library/react


silversonicaxel:
### Current Behavior

The library cannot test the switch of tabpanel when clicking tabitem while writing unit tests with @testing-library/react against the Tabs component.
The expected tabpanel to be displayed results still inactive.

### Expected behavior

Expected tab is active

### Reproducible example

https://github.com/radix-ui/primitives/compare/main...silversonicaxel:primitives:tabs-tests

this is the error I'm facing with the associated previously linked code
```
 FAIL  packages/react/tabs/src/Tabs.test.tsx
  given a Tab List
    with default behaviour in an uncontrolled scenario
      ✓ should have no accessibility violations in default state (227 ms)
      ✓ should check proper presence of all tablist entities (17 ms)
      ✕ should show the proper tabpanel accordingly to click navigation (20 ms)
    with default behaviour in a controlled scenario
      ✓ should have no accessibility violations in default state (42 ms)
      ✓ should check proper presence of all tablist entities (11 ms)
      ✕ should show the proper tabpanel accordingly to click navigation (12 ms)

  ● given a Tab List › with default behaviour in an uncontrolled scenario › should show the proper tabpanel accordingly to click navigation

    expect(element).toHaveAttribute("data-state", "inactive") // element.getAttribute("data-state") === "inactive"

    Expected the element to have attribute:
      data-state="inactive"
    Received:
      data-state="active"

      54 |
      55 |       expect(panel1).toHaveAttribute('data-state', 'inactive');
    > 56 |       expect(panel2).toHaveAttribute('data-state', 'inactive');
         |                      ^
      57 |       expect(panel3).toHaveAttribute('data-state', 'active');
      58 |       expect(panel4).toHaveAttribute('data-state', 'inactive');
      59 |     });

      at Object.<anonymous> (packages/react/tabs/src/Tabs.test.tsx:56:22)
      at TestScheduler.scheduleTests (node_modules/@jest/core/build/TestScheduler.js:333:13)
      at runJest (node_modules/@jest/core/build/runJest.js:404:19)

  ● given a Tab List › with default behaviour in a controlled scenario › should show the proper tabpanel accordingly to click navigation

    expect(element).toHaveAttribute("data-state", "inactive") // element.getAttribute("data-state") === "inactive"

    Expected the element to have attribute:
      data-state="inactive"
    Received:
      data-state="active"

      109 |
[...]

PR #2035, opened 2023-03-20
[Tabs] Add partial unit tests

silversonicaxel:
### Description

Added partial unit tests reagrding the Tabs component, related to issue https://github.com/radix-ui/primitives/issues/2034


chaance:
Sorry no one got back to you on this! I'm going to close this for now. Philosophically I don't think we should test implementation details, so I think this is better handled in our UI tests.

silversonicaxel:
ah, ok, fine by me, but, why not?

chaance:
Because implementation details may change and are not relevant to our end users. Whether or not we fire a click event or a series of other events under the hood is an implementation detail. Our UI tests can better simulate what an actual user does, which is limited in a virtual DOM environment unfortunately.

Issue #2049, opened 2023-03-27
Select: prevent item horizontal offset to be lower than 0


simrobin:
## Bug report

The issue occurs when a Select item has more content than the screen viewport width.

### Current Behavior

When the content of an item in a select, it goes out of the screen:
![image](https://user-images.githubusercontent.com/10052336/227983861-17816601-dcc3-4ada-9291-c9f0761959dd.png)

### Expected behavior

The Select should be in `left: 0` position and not get out of the screen.
![image](https://user-images.githubusercontent.com/10052336/228004520-2d208ea2-747e-4aa2-9c36-e46c8bd2bc2e.png)

### Reproducible example

https://codesandbox.io/p/sandbox/peaceful-moore-57lrc4

### Suggested solution

(I'm not quite sure about the `rtl` solution)
```diff
diff --git c/packages/react/select/src/Select.tsx w/packages/react/select/src/Select.tsx
index 11d0ba5b..a04f5fe2 100644
--- c/packages/react/select/src/Select.tsx
+++ w/packages/react/select/src/Select.tsx
@@ -802,17 +802,25 @@ const SelectItemAlignedPosition = React.forwardRef<
       const valueNodeRect = context.valueNode.getBoundingClientRect();
       const itemTextRect = selectedItemText.getBoundingClientRect();
 
+      console.log('selectedItemText', selectedItemText);
+
       if (context.dir !== 'rtl') {
         const itemTextOffset = itemTextRect.left - contentRect.left;
         const left = valueNodeRect.left - itemTextOffset;
         const leftDelta = triggerRect.left - left;
         const minContentWidth = triggerRect.width + leftDelta;
         const contentWidth = Math.max(minContentWidth, contentRect.width);
-        const rightEdge = window.innerWidth - CONTENT_MARGIN;
-        const clampedLeft = clamp(left, [CONTENT_MARGIN, rightEdge - contentWidth]);
 
-        contentWrapper.style.minWidth = minContentWidth + 'px';
-        contentWrapper.style.left = clampedLeft + 'px';
+        if (contentWidth >= window.innerWidth) {
+          contentWrapper.style.left = CONTENT_MARGIN + 'px';
+          contentWrapper.style.maxWidth = window.innerWidth - CONTENT_MARGIN * 2 + 'px';
+        } else {
+          const rightEdge =
+            contentWidth <= window.innerWidth ? window.innerWidth - CONTENT_MARGIN : contentWidth;
[...]

PR #2053, opened 2023-03-29
[Select] Prevent items longer than the window width to be positioned outside of the viewport

simrobin:
fixes [2049](https://github.com/radix-ui/primitives/issues/2049)

### Description
As described in issue [2049](https://github.com/radix-ui/primitives/issues/2049), if you have an item that is longer than your screen width (on a small screen for example), you will have a negative offset, pushing the content out of the viewport.

This PR proposes to compare the `contentWidth` with the `window.innerWidth` before affecting `left`/`right` position and `minWidth` to the element and fallback to a default value (default margin for the position and screenWidth - default margins for the `maxWidth`). 

This is the `simplest` solution I've found, but if you have suggestions to improve it, I'll be happy to discuss about it 🙂.

### DIFF
--- a/.yarn/versions/ed1540a2.yml
+++ b/.yarn/versions/ed1540a2.yml
@@ -1,0 +1,2 @@
+releases:
+  "@radix-ui/react-select": patch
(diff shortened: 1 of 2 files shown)


chaance:
Thank you! Opted for a slightly different approach [here](https://github.com/radix-ui/primitives/pull/3149). This will be in the next release.

simrobin:
Thanks for the feedback!

PR #2063, opened 2023-04-02
[DropdownMenu] Allow click events that are not only triggered by pointers

### LINKED ISSUE #1986: Vimium can not trigger links
## Bug report

Not really sure it wether to open this as a bug but personally for me it is very annoying. 

### Current Behavior

Radix menu triggers are not being able to trigger from vimium 

https://user-images.githubusercontent.com/16926049/221255386-9f0f1223-9fce-40a3-b09d-1ef8d8693a80.mov

### Expected behavior

HeadlessUI menus are openable by vimium 

https://user-images.githubusercontent.com/16926049/221255610-b82b9b28-293f-4113-ba98-ce901d6ed2d0.mov

### Reproducible example

Open your latest docs https://www.radix-ui.com/docs/primitives/components/dropdown-menu#dropdown-menu

### Suggested solution

This may be related to handler onTouch instead of clicks 

### Your environment

| Radix Package(s) |     latest   |         |
| React            | n/a     |    any     |
| Browser          |         |      chrome   |
| Operating System |     macos    |     13.2    |

ricci2511:
Fixes #1986.

Pretty simple change, but it makes sure that Dropdown button click events that are triggered by anything other than pointers  are also listened to. One case for this is the Vimium browser extension as mentioned in the linked issue, which doesn't trigger neither a pointer or keydown event.

### DIFF
--- a/.yarn/versions/c89fd021.yml
+++ b/.yarn/versions/c89fd021.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-dropdown-menu": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


andy-hook:
Thanks for the effort here, I agree with https://github.com/radix-ui/primitives/issues/1986#issuecomment-1493777880 though and feel that the extension should account for these event types rather than client code changing implementation to support.

PR #2070, opened 2023-04-04
Update useControllableState.tsx

lifeiscontent:
Fix types for useControllableState


benoitgrelard:
Hi @lifeiscontent, could you explain what your reasoning behind this PR? What is it intending to fix exactly?

lifeiscontent:
@benoitgrelard the types are lying `T | undefined` should just be `T` and whatever the caller is should be a concrete represenation of the possible states

lifeiscontent:
@benoitgrelard I guess from my perspective, the generic should be `T` undefined should be a part of the definition that is calling the function instead of masking it

benoitgrelard:
I see, thanks for the explanation, we'll take a look if it makes sense to incorporate this at some point.

PR #2094, opened 2023-04-20
[Dialog] fix accessibility issue - remove focusable elements from ARIA hidden element

sagar-ganatra:
### Description

DialogTrigger button should not be focusable when the Dialog box is open. Please see https://dequeuniversity.com/rules/axe/4.5/aria-hidden-focus?application=axeAPI. This PR addresses the issue by setting the value for `tabIndex` attribute to `-1` when the Dialog is open and to `0` when the Dialog is closed

Testing - 

Dialog_TabIndex

### DIFF
--- a/.yarn/versions/385e36b4.yml
+++ b/.yarn/versions/385e36b4.yml
@@ -1,0 +1,6 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-dialog": patch
+
+declined:
+  - primitives
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -107,6 +107,7 @@ const DialogTrigger = React.forwardRef<DialogTriggerElement, DialogTriggerProps>
         aria-expanded={context.open}
         aria-controls={context.contentId}
         data-state={getState(context.open)}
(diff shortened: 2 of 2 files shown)


benoitgrelard:
Hey @sagar-ganatra, I don't think this is necessary as when a modal dialog is open, not only the outside content is "hidden" from screen readers, but focus is also trapped.

sagar-ganatra:
@benoitgrelard though the modal has a focus trap, when you test for accessibility, the issue - https://dequeuniversity.com/rules/axe/4.5/aria-hidden-focus?application=axeAPI does get reported. This is because `#root` has `aria-hidden` set to `true` and element in it should have an attribute (tabIndex=-1) which would let the browser and testing tools happy

benoitgrelard:
> @benoitgrelard though the modal has a focus trap, when you test for accessibility, the issue - https://dequeuniversity.com/rules/axe/4.5/aria-hidden-focus?application=axeAPI does get reported. This is because `#root` has `aria-hidden` set to `true` and element in it should have an attribute (tabIndex=-1) which would let the browser and testing tools not report an error

I think that's one of those cases though where in practice we are accessible and it's the tool reporting that's being rather overzealous.

I will close the PR as I don't believe this is necessary (other libs don't do this either).

PR #2095, opened 2023-04-20
[Accordion] Refactor AccordionItem data-state to use getState helper

brattonross:
### Description

Makes use of the existing `getState` function in AccordionItem


andy-hook:
Thanks @brattonross , could you run `yarn version check --interactive`, decline all packages and push the change?

andy-hook (approved):

PR #2110, opened 2023-04-25
[Popover] Fix close/re-open bug on Safari

### LINKED ISSUE #2105: Popover inside element with tabIndex="0" cannot be closed by pressing its trigger on safari
### Current Behavior

Kapture 2023-04-24 at 16 25 32

In most browsers, if you create a popover inside an element with `tabindex="0"`, pressing the trigger opens the popover & pressing the trigger again closes it.

On safari (mac & ios), pressing the trigger opens the popover as expected. When pressing the trigger for a second time, the popover closes on pointer down, but re-appears again on pointer up, making it impossible to completely close a popover by pressing its trigger on safari.

### Expected behavior

Popovers in safari should close when the trigger is pressed again.

### Reproducible example

https://stackblitz.com/edit/react-ts-zhrg64?file=App.tsx

### Additional context

I spent a little while investigating this to see if I could fix it in user-land but couldn't figure something out. The issue here is that safari focuses the container with `tabindex="0"` when the button is pressed down, so radix-ui treats that as a focus-outside. Then, when the click is released, radix sees that as the end of a press, but we've already closed the popover, so it gets opened again.

### Your environment

| Radix Package(s) | @radix-ui/react-popover | 1.0.5        |
| React            | n/a     |   18.2.0      |
| Browser          |  Safari       |   16.2      |
| Assistive tech   |    n/a     |  n/a       |
| Node             | n/a     |    n/a     |
| npm/yarn         |   n/a      |   n/a      |
| Operating System | macOS, iPadOS, iOS        |  macOS 13, iOS + iPadOS 16        |

benoitgrelard:
Fixes #2105

### DIFF
--- a/.yarn/versions/ca8a29d8.yml
+++ b/.yarn/versions/ca8a29d8.yml
@@ -1,0 +1,7 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-dialog": patch
+  "@radix-ui/react-popover": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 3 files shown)


andy-hook:
Interesting, does this bug affect menu primitives at all?

Issue #2011, opened 2023-03-08
`Slot` in server component gives error


kuus:
## Bug report
Not sure if this is to be considered a bug actually, but trying to use `Slot` from `import { Slot } from "@radix-ui/react-slot";` in a server component (within next@13.2 framework) you get an error.

### Current Behavior

You get an error: `Error: Refs cannot be used in Server Components, nor passed to Client Components.`

### Expected behavior

Can `Slot` work in a server component?

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |         |   1      |
| React            | n/a     |    18.2     |

benoitgrelard:
Hey @kuus, I don't believe it could given the error message.

From my understanding with server components, you are restricted in terms of what features of React can be used, for example `useState` cannot be used, etc.

It sounds like refs can't be used either so that means `Slot` cannot be used in a server component because it needs to attach a ref to the cloned element.

Hope that makes sense.

✌️

benoitgrelard:
Pretty much all Radix components are client components though, what is your use case for `asChild` in server components?

kuus:
I think is pretty common, in the footer I have some social media links, they are styled as buttons but they are links,
my `Button` component implementation uses `asChild` prop with `Slot` component so that it can be used to style both HTML links `<a>` and `<button>` or even `<span>` if needed, etc.

Pseudo code:
```js
<Button asChild>
  <a href="https://example.com>An icon</a>
</Button>
```
This is defnitely a good candidate for a server component as it is just a plain `<a>` element with some style...
Hope you see what I mean, to sum up: `Slot`table components are often good candidates to be server components

kuus:
hi @benoitgrelard , I just tried to remove the refs usage and that is enough to make the `Slot` component play nice as RSC.
Regarding the general attempt of making the Slot component usable as RSC I don't know yet if there is an "illegal" way to conditionally use `useRef` (maybe in a `try catch..`?), and I am also very unsure that is a viable path...but I will post here if I find anything useful.

Meanwhile here is a playground:
https://codesandbox.io/p/sandbox/next-13-slot-imd6ft

cheers

benoitgrelard:
That makes sense, removing the illegal bit. But now you wouldn't be able to attach a ref to it, so I'm kinda confused how this would be useful.

btw, I don't have much knowledge of RSC yet so I'm probably talking rubbish.

PR #2116, opened 2023-04-27
Fix `Slot` when used in an RSC

jjenzz:
### Description

in RSCs you cannot pass a `ref` prop (makes sense, given it's a client only thing) so React will throw a nasty error to warn about this if the `ref` is not `undefined`. This becomes problematic with `Slot` because it passes the return type of `composeRefs` (a function) to anything it slots onto which means the `ref` is not `undefined` regardless of whether you are attaching one and the error occurs.

This PR ensures the ref is only composed by `Slot` if there is a defined `forwardRef` and allows consumers to build their own slottable comps to use in RSCs e.g.:

```jsx
// span by default
<Text asChild size={'large'}>
  <p>foo</p>
</Text>
```


packages/react/slot/src/Slot.tsx
```diff
@@ -63,7 +63,7 @@ const SlotClone = React.forwardRef<any, SlotCloneProps>((props, forwardedRef) =>
   if (React.isValidElement(children)) {
     return React.cloneElement(children, {
       ...mergeProps(slotProps, children.props),
-      ref: composeRefs(forwardedRef, (children as any).ref),
+      ref: forwardedRef ? composeRefs(forwardedRef, (children as any).ref) : (children as any).ref,
```
jjenzz, line 66: could also just do this tbh (so it isn't overriding `ref` if it doesn't need to) but lmk which you'd prefer :)

Screenshot 2023-04-27 at 16 21 20

benoitgrelard: Wouldn't that one be missing `children.ref` then?

benoitgrelard: I think I prefer the explicit way you have written though either way.

benoitgrelard (approved):

kuus:
https://github.com/radix-ui/primitives/issues/2011

Issue #2122, opened 2023-05-01
[Dialog] `DialogContent` disables pointer events for the whole page which might conflict with 3rd party modals


lesha1201:
### Current Behavior

When `DialogContent` is modal, it has `disableOutsidePointerEvents={true}` which sets `pointer-events: none`. And because of that modals from 3rd parties might not work properly when they're open from the dialog. For example, I need to implement a dialog for adding a payment method in Stripe. Stripe.js can open their dialog in some cases, for example, in case of 3D secure.

https://user-images.githubusercontent.com/10157660/235491448-5fc13399-e384-45c0-a1c0-1187523d7c91.mp4

### Suggested solution

I think we should allow to specify `disableOutsidePointerEvents` for `DialogContent`. There are several reasons for that:
- There is no sense for having it when you have `DialogOverlay` which covers the whole page.
- To have more flexibility, when you have some special cases.

It should be easy to allow to specify that prop.

ccxdev:
I also faced the same problem when I integrated the recaptcha from google

recs182:
I having this issue too! Also can confirm that version 1.0.2 the problem does not occurs, so people may consider downgrade until it is fixed.

benoitgrelard:
Hey all,

> I think we should allow to specify disableOutsidePointerEvents for DialogContent. There are several reasons for that:
> - There is no sense for having it when you have DialogOverlay which covers the whole page.

> Is there any reason for the pointer-events: none to be set on the body element when the Dialog is open? Isn’t the Overlay component supposed to stand on top of the content behind the dialog, and capture all clicks outside the Dialog?

The issue is you are thinking about modality only from the POV of `Dialog`. Although somewhat visually related sometimes, the concept of modality however has nothing to do with it. `Dialog` does happen to have an overlay, but there are plenty of other components that don't and still need to support modality (`Popover`, `DropdownMenu`, etc).

The easiest solution to the 3rd party component being open from one of our modal components would be to set `pointer-events: auto` on their root container.

benoitgrelard:
> I don't get it how it's related to Popover, DropdownMenu, etc. because the proposal is to allow setting disableOutsidePointerEvents on DialogContent which isn't used in those components.

Because modality is a general concept which applies in more than dialogs. The stripe dialog could be opened from a `Popover` just as well as a `Dialog`. And that doesn't have an overlay.

> It's not a common case to see pointer-events: none on body in case of modal.

[...]

Issue #1159, opened 2022-02-15
Change Dialog.Overlay to wrap Content in examples


hipstersmoothie:
### Relevant Radix Component(s)

Here in the docs it say to make `Overlay` a sibling to `Content`.

![CleanShot 2022-02-15 at 11 02 22](https://user-images.githubusercontent.com/1192452/154130777-fcb5420c-943a-4ba4-b89b-20afb0add8aa.png)

This setup doesn't work if you have a scrollable popover in the dialog content. On [this line](https://github.com/radix-ui/primitives/blob/main/packages/react/dialog/src/Dialog.tsx#L202) a ref is passed to `RemoveScroll` but I _think_ that by using just a ref that `RemoveScroll` doesn't consider the portaged element a part of the react tree.

In my own code I fixed this by putting my `Content` inside of my `Overlay`. After this change my scrollable popover could scroll again since `RemoveScroll` now has access to the react tree.

benoitgrelard:
Both compositions are valid, for different use cases.
In fact we have an example of a scrollable overlay here: https://www.radix-ui.com/docs/primitives/components/dialog#scrollable-overlay

hipstersmoothie:
It's not overlay that I need scrollable though. I'll come up with a code sandbox to show what I'm talking about

hipstersmoothie:
Here you can see if you follow all of the docs for both popover and dialog you can scroll a popover in a dialog. I had to read through radix's source code and dependencies to figure out that I needed to wrap my content in the overlay. While both are valid I think the docs should steer people towards code+structure that will work out of the box without having to understand how the internals are actually working.

https://codesandbox.io/s/twilight-http-veu4gw?file=/App.js

hipstersmoothie:
And if you're unwilling to change that I think the docs should at least detail why both composition are valid and what the use cases actually are.

benoitgrelard:
Oh sorry I misunderstood what you meant!
That seems like a bug potentially, I'll take a look.

jjenzz:
Hmm, looks like it could be a `RemoveScroll` shards thing? 🙈

dcastil:
I'm running into the same issue. I'd like to adopt Radix for the design system I'm working on but we have some custom selects with a scrollable area (implemented as popover) which are rendered inside modals. Would it be possible to disable the scroll-blocking in the dialog components? Or just to enable scroll-blocking of the body since the rest is not scrollable because of the dialog overlay anyway.

[...]

Issue #998, opened 2021-11-23
[Dialog] force mounting dialog content makes the dialog open by default 


tk-olari:
## Bug report

Force mounting `DialogOverlay` and `DialogContent` renders in the DOM the dialog and it is possible to control the visibility of the dialog with CSS but it also locks the scroll and traps the focus which makes the page unusable.

### Current Behavior

Force mounting `DialogOverlay` and `DialogContent` in modal mode blocks the scrolling and traps focus.

### Expected behavior

Force mounting `DialogOverlay` and `DialogContent` should work with scroll lock and focus trap the same way as `forceMount=false`. So basically the dialog should control focus trap and scroll lock regardless of `forceMount` option.

### Reproducible example

https://codesandbox.io/s/2oryx?file=/App.js

In this example I made the height of the body `200vh` for scrolling purpose. By default the `forceMount` option is true and you can observe that it is impossible to scroll or click the button. The dialog is hidden with CSS using `[data-state='closed']` selector. If `forceMount` is set to false, everything is working as expected.

### Suggested solution

I'm not sure if this is a bug or an intended behaviour. If it is intended, how the scroll lock and focus trap can be released programmatically?

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | @radix-ui/react-dialog | 0.1.0 |
| React            | n/a     | 17.0.2 |
| Browser          | any | any |
| Assistive tech   | - | - |
| Node             | n/a     | 14.16.0 |
| npm/yarn         | any | any |
| Operating System | any | any |

jjenzz:
> I'm not sure if this is a bug or an intended behaviour. If it is intended, how the scroll lock and focus trap can be released programmatically?

We had a similar issue reported for this here https://github.com/radix-ui/primitives/issues/628
It might help shed some light on the thinking behind the behaviour of this prop.

@benoitgrelard I wonder if we should rename this prop to something like `deferUnmount` so that people aren't tempted to use it for cases other than postponing unmount for animation?

benoitgrelard:
Yeah I don't think the use case is valid considering what the prop is for.
I'm also unsure changing the name of the prop would change anything 🤔

jjenzz:
> I'm also unsure changing the name of the prop would change anything 🤔

The name `forceMount` implies it can be used to keep it mounted "for any reason". Hence, I assume, @tk-olari thought he could use it to keep things mounted and hide with CSS instead.

[...]

PR #2123, opened 2023-05-01
[Dialog] Allow to set `disableOutsidePointerEvents` on DialogContent

lesha1201:
### Description

Handles https://github.com/radix-ui/primitives/issues/2122 (all the details there)

Let me know if I should add any tests or stories for that PR.

### DIFF
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -247,8 +247,7 @@ DialogContent.displayName = CONTENT_NAME;
 /* -----------------------------------------------------------------------------------------------*/
 
 type DialogContentTypeElement = DialogContentImplElement;
-interface DialogContentTypeProps
-  extends Omit<DialogContentImplProps, 'trapFocus' | 'disableOutsidePointerEvents'> {}
+interface DialogContentTypeProps extends Omit<DialogContentImplProps, 'trapFocus'> {}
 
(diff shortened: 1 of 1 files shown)


yukulelix (approved):

benoitgrelard:
Hi @lesha1201,

I wouldn't make sense to turn this feature off whilst in `modal` mode thus why we preset a bunch of things based on modality.

Also see: https://github.com/radix-ui/primitives/issues/2122#issuecomment-1557016755

Lexachoc:
When I have `modal={true}` (the default) and turned on forceMount, I can't click on the page because of this prop is set `true`, any idea how to overcome this? I saw the solution of using setTimeout to remove it later in [here](https://github.com/radix-ui/primitives/issues/2122#issuecomment-1531430926) but I wonder if it's possible to allow the expose of this prop that we can turn it off when forceMount? 

Issue with forceMount: https://github.com/radix-ui/primitives/issues/998
Related issue about scrolling: https://github.com/radix-ui/primitives/issues/1159 https://github.com/radix-ui/primitives/issues/2122

After altering the code, I was able to use the Dialog `forceMount` as expected, like the `forceMount` used for [Tabs] (https://www.radix-ui.com/primitives/docs/components/tabs#content), which I can use the class name (Tailwind) to change the visibility of the tabs instead of unmounting them.

But I stuck at the pointer event, which was forced to be added to the div when using forceMount.

Issue #1979, opened 2023-02-22
Components used with "asChild" should only accept a single child


equal-matt:
### Overview

Current state of affairs, using a Dialog as an example:
```jsx
<Dialog.Root>
    <Dialog.Trigger />
    <Dialog.Portal>
      <Dialog.Overlay />
      <Dialog.Content asChild>
        {/* runtime error here */}
        <Dialog.Title />
        <Dialog.Description />
        <Dialog.Close />
      </Dialog.Content>
    </Dialog.Portal>
  </Dialog.Root>
```

When setting asChild on a component, an error is thrown at runtime because there must be a single child.

The type system should be stricter on this and raise before the runtime error is experienced.

### Who does this impact? Who is this for?

This impacts all users. The current error feedback looks something like this (excuse the next-related error stack, I haven't made a minimal reproduction):

```
```

kjhuang-db:
I honestly do have a clear mindset what "asChild" means

benoitgrelard:
Hey @equal-matt, yeah we may be able to get slightly better errors (as at the moment they refer to a component deep in the tree that users aren't using directly (`Slot`).

With regards to types, I don't think that is possible due to TS limitations.

cubecleveland:
Guys, i have the same issue , i  was using an old version of Collapsible.
Wheni upgrade i get the following message

./node_modules/@radix-ui/react-primitive/node_modules/@radix-ui/react-slot/dist/index.mjs
Can't import the named export 'Children' from non EcmaScript module (only default export is available

I tried installing primitive but it didnt fix,
When we go back in the collpasible version (where using 0.1.6 and upgradeded to 1.0.3 -- everything works.
Any help , would be appreictaqed

ahmetkca:
I was just questioning my entire life. if anybody is having problem with this particular error. Try to search for components that has the `asChild` prop and make sure there is only one child. Even `{' '}` text counts as a separate child. My auto formatter inserted a text `{' '}` instead of removing the new line.

bhatvikrant:
> I was just questioning my entire life. if anybody is having problem with this particular error. Try to search for components that has the `asChild` prop and make sure there is only one child. Even `{' '}` text counts as a separate child. My auto formatter inserted a text `{' '}` instead of removing the new line.

I'm facing the same issue, and there are literally no `{' '}` in my `DialogContent`, idk what is going on, asChild just does not work for me in Dialogs

raffidahmad:
[...]

PR #2141, opened 2023-05-15
[Slot] Update error message for asChild with multiple children.

yuta-ike:
fix https://github.com/radix-ui/primitives/issues/1979

### Description
I changed the error message when multiple ReactNodes are specified as children with the asChild option enabled. 

It might be helpful to include the URL for the official documentation, but since there is no other implementation, I don't include. If we provide document url, the page created in [this pr](https://github.com/radix-ui/website/pull/628) would be a good choice.

Pease let me know if there are any mistakes, as I am not experienced in library implementation.


packages/react/slot/src/Slot.tsx
```diff
@@ -22,7 +22,12 @@ const Slot = React.forwardRef<HTMLElement, SlotProps>((props, forwardedRef) => {
       if (child === slottable) {
         // because the new element will be the one rendered, we are only interested
         // in grabbing its children (`newElement.props.children`)
-        if (React.Children.count(newElement) > 1) return React.Children.only(null);
+        if (1 < React.Children.count(newElement)) {
+          throw new Error(
+            `\`asChild\` cannot be set to true when passing multiple ReactNodes as children.`
```
benoitgrelard, line 27: This messaging here in `Slot` bothers me because this isn't where `asChild` is declared (that's in primitive).

Slot is a public utility and people who use it may not have an `asChild` or have a different prop.

I'll looking for a different approach.

PR #2142, opened 2023-05-15
Group stories as per website

benoitgrelard:


packages/react/announce/src/Announce.stories.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { VisuallyHidden } from '@radix-ui/react-visually-hidden';
 import { Announce } from '@radix-ui/react-announce';
 
-export default { title: 'Components/Announce' };
+export default { title: 'Utilities/Announce' };
```
andy-hook, line 5: We can probably remove this bad boy at some point.

andy-hook (approved):

PR #2148, opened 2023-05-21
[Tooltip] Add "closeDelayDuration" prop

mattandryc:
### Description

This PR adds a `closeDelayDuration` prop to the Tooltip primitive to allow the consumer to set a custom duration from when the pointer leaves the tooltip's content or trigger area until the tooltip closes. Like the existing `delayDuration` prop, this value can be set on the `TooltipProvider` or on an individual `Tooltip`. For context, in usability testing, we found the tooltip closed unexpectedly when the user accidentally moved the pointer out of the content or trigger area. 

https://github.com/radix-ui/primitives/assets/6387301/a929e2dd-be16-4be2-9f48-55504d9ee168

### DIFF
--- a/packages/react/tooltip/src/Tooltip.stories.tsx
+++ b/packages/react/tooltip/src/Tooltip.stories.tsx
@@ -237,6 +237,67 @@ export const CustomDurations = () => (
         </Tooltip.Root>
       </Tooltip.Provider>
     </div>
+    <h1>Close delay duration</h1>
+    <h2>Default (0ms after pointer leaves trigger or content area)</h2>
+    <div style={{ display: 'flex', gap: 50 }}>
+      <Tooltip.Root>
+        <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
+        <Tooltip.Portal>
+          <Tooltip.Content className={contentClass()} sideOffset={5}>
+            Nicely done!
+            <Tooltip.Arrow className={arrowClass()} offset={10} />
+          </Tooltip.Content>
+        </Tooltip.Portal>
+      </Tooltip.Root>
+      <Tooltip.Root>
+        <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
+        <Tooltip.Portal>
+          <Tooltip.Content className={contentClass()} sideOffset={5}>
+            Nicely done!
+            <Tooltip.Arrow className={arrowClass()} offset={10} />
+          </Tooltip.Content>
+        </Tooltip.Portal>
+      </Tooltip.Root>
+      <Tooltip.Root>
+        <Tooltip.Trigger className={triggerClass()}>Hover me</Tooltip.Trigger>
+        <Tooltip.Portal>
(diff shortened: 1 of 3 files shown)


mattandryc:
Sure — the first unexpected piece was that the tooltip would sometimes close before the cursor could travel from the trigger to the content area. This was more pronounced on tooltips with lengthier content (clip below).

https://github.com/radix-ui/primitives/assets/6387301/02a82680-17bc-49fe-9511-8d9b29a2e89b

The second observation was that when users attempted to move their mouse out of the way to the tooltip's edge to read its content, they'd overshoot slightly and the tooltip would immediately close (i.e. it's a pretty small target and there's no error tolerance). While increasing the padding would work here, it throws off the balance between the text and space in the design, and would occlude more of the UI behind it. 

https://github.com/radix-ui/primitives/assets/6387301/75e4046c-1b6f-4317-aa0f-de780497fa1e

To your point, I could see exposing some control over how the tooltip's grace area is calculated as a better solution conceptually, but it also seems less intuitive for developers than a "delay close" feature, which, in looking at alternate tooltip implementations (e.g. Chakra, Material, Mantine, etc.), seems to be more common.

benoitgrelard:
Thanks for getting back to me.

> the first unexpected piece was that the tooltip would sometimes close before the pointer could travel from the trigger to the content area. This was more pronounced on tooltips with lengthier content (clip below).

Yes that's a known issue. There was an attempt at fixing it here https://github.com/radix-ui/primitives/pull/1960 but the contributor didn't have enough time to continue on it. Again, this is a grace area issue, not a delay issue.

> The second observation was that when users attempted to move their pointer out of the way to the tooltip's edge to read its content, they'd overshoot slightly and the tooltip would immediately close (i.e. it's a pretty small target and there's no error tolerance). While increasing the padding would work here, it throws off the balance between the text and space in the design, and would occlude more of the UI behind it.

Ok so yeah my inkling was correct then, this is also not a delay issue, but a grace area issue. I understand you may not want it bigger visually, but it is trivial to add some tolerance using a pseudo element on `Tooltip.Content` to add some "invisible" padding. Something like:

```css
.SelectContent {
  …
  position: relative;
}

.SelectContent::before {
  content: "";
  position: absolute;
  inset: -5px;
}
```

So I don't think we need to add anything to the library for that use-case as it's a decision based on design and pretty trivial implementation user-side.

> To your point, I could see exposing some control over how the tooltip's grace area is calculated as a better solution conceptually, but it also seems less intuitive than a "delay close" feature, which, in looking at alternate tooltip implementations (e.g. Chakra, Material, Mantine, etc.), seems to be more common.

Yes, that's because most libraries deal with this "temporarily" rather than "spatially". Whilst the former can work, the latter is better because it doesn't depend on any timing, so ends up being a better solution for accessibility (no need for the user to have to move fast).

---

If you are still interested in contributing, I think taking over what #1960 was trying to achieve would be a better placed effort. There are already pointers in the PR comments about what needs to be done too.

Let me know!

mattandryc:
Sounds good! — I'll pick up #1960

mattandryc:
See https://github.com/radix-ui/primitives/pull/2155

PR #2155, opened 2023-05-24
[Tooltip] Improve grace area calculation

### LINKED ISSUE #1855: Hovering on Tooltip is hard in some situations

mattandryc:
### Description

https://github.com/radix-ui/primitives/assets/6387301/e866a6e2-dec5-4d35-b69b-c03cdac432bb

Following up on discussion in [#2148](https://github.com/radix-ui/primitives/pull/2148) and prior work in #1960, this PR improves the grace area calculation for the Tooltip component (closing #1855). The approach here approach follows the logic outlined by @benoitgrelard in #1960: Instead of a single start point, the logic now pre-appends two start points which are determined as follows relative to the exit point:

Given a side ("top" | "bottom" | "left" | "right"):
- the two points should extend along that side in both directions
- the two points should be pushed towards the the opposite side

Closes #1855

### DIFF
--- a/.yarn/versions/2590f83c.yml
+++ b/.yarn/versions/2590f83c.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -353,6 +353,7 @@ TooltipPortal.displayName = PORTAL_NAME;
  * -----------------------------------------------------------------------------------------------*/
 
 const CONTENT_NAME = 'TooltipContent';
+const GRACE_AREA_BLEED_PX = 5;
 
 type TooltipContentElement = TooltipContentImplElement;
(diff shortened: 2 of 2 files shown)


benoitgrelard (approved):
Thanks @mattandryc, great work!
I've extracted it into a function but otherwise it works great.

❤️

PR #2161, opened 2023-05-25
[Toggle]: Add unit tests

sandrina-p:
### Description

- chore(toggle): Add unit tests

Why did I add tests? Toggle is a simple component, yet a great real-world example for devs starting to learn A11Y and how to use unit tests to cover A11Y attributes in their components. I'll use this as an example for an A11Y workshop I'm conducting.


packages/react/toggle/src/Toggle.test.tsx
```diff
@@ -0,0 +1,133 @@
+import React from 'react';
```
sandrina-p, line 1: P.S. I'm not sure if the PR title follows the conventions. I looked into the commit history for reference. :)

benoitgrelard (approved):

PR #2174, opened 2023-05-30
[Select] Use `""` `value` to reset to placeholder

### LINKED ISSUE #1569: [Select] Unable to clear value and return to `placeholder`
I'm playing with the new `Select` `placeholder` prop, and am wondering if the following is possible?

As an example, let's say you have a series of 3 selects for an automobile and the placeholder values for each are "make", "model", and "year".  When you select a new "make" (after having already selected both a "make" and a "model") and the underlying list for "model" is updated, the component still technically has a `value` based on the previous list's selection, therefore the placeholder doesn't render. 

Is there a way to reset `value` when updating the list?

benoitgrelard:
Fixes #1569

### DIFF
--- a/.yarn/versions/96d1204b.yml
+++ b/.yarn/versions/96d1204b.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-select": major
+
+declined:
+  - primitives
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -251,7 +251,7 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
           data-state={context.open ? 'open' : 'closed'}
           disabled={isDisabled}
           data-disabled={isDisabled ? '' : undefined}
-          data-placeholder={context.value === undefined ? '' : undefined}
+          data-placeholder={shouldShowPlaceholder(context.value) ? '' : undefined}
           {...triggerProps}
           ref={composedRefs}
           // Enable compatibility with native label or custom `Label` "click" for Safari:
@@ -316,7 +316,7 @@ interface SelectValueProps extends Omit<PrimitiveSpanProps, 'placeholder'> {
 const SelectValue = React.forwardRef<SelectValueElement, SelectValueProps>(
   (props: ScopedProps<SelectValueProps>, forwardedRef) => {
     // We ignore `className` and `style` as this part shouldn't be styled.
-    const { __scopeSelect, className, style, children, placeholder, ...valueProps } = props;
+    const { __scopeSelect, className, style, children, placeholder = '', ...valueProps } = props;
     const context = useSelectContext(VALUE_NAME, __scopeSelect);
     const { onValueNodeHasChildrenChange } = context;
     const hasChildren = children !== undefined;
@@ -334,7 +334,7 @@ const SelectValue = React.forwardRef<SelectValueElement, SelectValueProps>(
         // through the item they came from
         style={{ pointerEvents: 'none' }}
       >
-        {context.value === undefined && placeholder !== undefined ? placeholder : children}
+        {shouldShowPlaceholder(context.value) ? <>{placeholder}</> : children}
       </Primitive.span>
     );
   }
@@ -1198,6 +1198,12 @@ const SelectItem = React.forwardRef<SelectItemElement, SelectItemProps>(
       }
     };
 
+    if (value === '') {
+      throw new Error(
+        'A <Select.Item /> must have a value prop that is not an empty string. This is because the Select value can be set to an empty string to clear the selection and show the placeholder.'
+      );
+    }
+
     return (
       <SelectItemContextProvider
         scope={__scopeSelect}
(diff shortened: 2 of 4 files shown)


.yarn/versions/96d1204b.yml
```diff
@@ -0,0 +1,5 @@
+releases:
+  "@radix-ui/react-select": major
```
benoitgrelard, line 2: Making this a major as this is potentially a breaking change for people who were relying on using an option with `value=""` instead of the `placeholder`.

packages/react/select/src/Select.tsx
```diff
@@ -251,7 +251,7 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
           data-state={context.open ? 'open' : 'closed'}
           disabled={isDisabled}
           data-disabled={isDisabled ? '' : undefined}
-          data-placeholder={context.value === undefined ? '' : undefined}
+          data-placeholder={shouldShowPlaceholder(context.value) ? '' : undefined}
```
benoitgrelard, line 254: This is combination with the next change below (`placeholder = ''`) means users will always be able to style differently, even if they don't want any visible placeholder text.

packages/react/select/src/Select.tsx
```diff
@@ -334,7 +334,7 @@ const SelectValue = React.forwardRef<SelectValueElement, SelectValueProps>(
         // through the item they came from
         style={{ pointerEvents: 'none' }}
       >
-        {context.value === undefined && placeholder !== undefined ? placeholder : children}
+        {shouldShowPlaceholder(context.value) ? <>{placeholder}</> : children}
```
benoitgrelard, line 337: The key thing here is the fragment, which enables React to find its feet and reconcile properly when we reset the state to "". This is because when a value is set, we portal into that node, and without that React seems to lose it and isn't able to render the placeholder again.

andy-hook (approved):
lgtm 👍

benoitgrelard:
You can already use the release candidates as soon as the PR merge.
https://www.npmjs.com/package/@radix-ui/react-select?activeTab=versions

aboveyunhai:
> You can already use the release candidates as soon as the PR merge. https://www.npmjs.com/package/@radix-ui/react-select?activeTab=versions

Thx a lot, didn't see the rc.

nowlena:
I may be missing something however It'd be really nice if we could just use `<Select.Item value="">Reset</Select.Item>` to reset the Select rather than having to interact with the `<Select.Root />`

aboveyunhai:
@nowlena  For this rc-release only , since I didn't install the later version, you can do it via.
```ts
    <Select.Item value={undefined as unknown as string}>
         Reset
     </Select.Item>
```
It's hacky but make sense. and you also cannot make `Select.Item` value to be `string | undefined` imo in order to satisfy our implementation goals.

Shaker-Hamdi:
So, how to use this exactly?
I'm a bit confused, I installed the latest `rc` version, and I tired to have an item with an empty string value, but I got this error ...

`Uncaught Error: A <Select.Item /> must have a value prop that is not an empty string. This is because the Select value can be set to an empty string to clear the selection and show the placeholder.`

So, how exactly should I reset the select to the placeholder?

andy-hook:
> So, how exactly should I reset the select to the placeholder?

1. Ensure all items have a unique `value`, none should use empty string
2. Pass a `placeholder` to `<Value />`
3. Use the controlled props to set the `value` of `<Root />` to an empty string

Shaker-Hamdi:
@andy-hook  Thanks a bunch, that worked.

Shaker-Hamdi:
@andy-hook  Do I have to use the `<Select>` inside a `<form>` tag in order for the this feature to work?

The reason I'm asking is because I have this case where I'm using "React Hook Form", and so that I can set the value to `null` in case the user chose the option that reset to the placeholder I have to do something like ...

``` typescript
onValueChange={(value: string) => {
    if (value === "empty_field") {
        field.onChange(null);
    } else {
        field.onChange(value);
    }
}}
value={field.value === "empty_field" ? "" : field.value}
```

But when I do this (the onValueChange) the <Select> doesn't reset to the placeholder, unless I wrap the `<Select>` inside a `<form>` tag.

I created a quick Codesandbox demo ...
https://codesandbox.io/s/radix-ui-select-yv9gsl?file=/src/App.tsx

andy-hook:
@Shaker-Hamdi I haven't looked closely at your code but `Select` renders a native element to bubble form values only when it is wrapped inside a `form`, that might help explain the different behaviour here depending on how `useForm` is attempting to interface with form elements.

Changing `field.onChange(null)` to `field.onChange('')` appears to work in both cases though.

Shaker-Hamdi:
@andy-hook You're correct, setting the value to an empty string instead of null works in both cases.
Thank you.

lauri865:
Still struggling to add a placeholder as a valid item to "reset" the select though. Setting the value to undefined works, but breaks the ItemIndicator.

I don't understand what was wrong with the previous implementation. It worked to reset the field and was closer to the native implementation of <select>.

ErKeLost:
want to insert an AI-related problem. Using AI to generate an application, if I use the select component, the problem will never be solved by AI. AI will never understand the problem and will fall into an infinite loop. In the case of vibe coding,  Thanks for this PR

PR #2175, opened 2023-05-30
[ScrollArea] `scroll-behavior: smooth` compatibility

### LINKED ISSUE #2064: ScrollArea event-based scrolling does not work with `scroll-behavior: smooth`
## Bug report

I'm setting `scroll-behavior: smooth` so that my custom `.scrollTo` calls are smoothly scrolled into view. However, it breaks the ScrollArea dragging and trackpad (`wheel` event) behavior.

### Current Behavior

https://user-images.githubusercontent.com/34928425/229560733-d6d9d109-c148-43ae-a793-38000fa6d270.mp4

### Expected behavior

https://user-images.githubusercontent.com/34928425/229560654-163cce91-3953-4db4-bf62-3f5e8c761691.mp4

### Reproducible example

https://codesandbox.io/p/sandbox/sparkling-waterfall-6pibz1

### Suggested solution

ScrollArea should manually set `viewport.style.scrollBehavior = 'auto'` when the scroll position (`.scrollTop` and `.scrollLeft`) is manually set from JS events.

### Your environment

| Radix Package(s) | ScrollArea        |   1.0.3      |
| React            | n/a     |   18.2.0      |
| Browser          | Chrome, Brave, Safari        |         |

benoitgrelard:
Fixes #2064

### DIFF
--- a/packages/react/scroll-area/src/ScrollArea.tsx
+++ b/packages/react/scroll-area/src/ScrollArea.tsx
@@ -715,6 +715,7 @@ const ScrollAreaScrollbarImpl = React.forwardRef<
             // so we remove text selection manually when scrolling
             prevWebkitUserSelectRef.current = document.body.style.webkitUserSelect;
             document.body.style.webkitUserSelect = 'none';
+            if (context.viewport) context.viewport.style.scrollBehavior = 'auto';
             handleDragScroll(event);
           }
         })}
@@ -725,6 +726,7 @@ const ScrollAreaScrollbarImpl = React.forwardRef<
             element.releasePointerCapture(event.pointerId);
           }
           document.body.style.webkitUserSelect = prevWebkitUserSelectRef.current;
+          if (context.viewport) context.viewport.style.scrollBehavior = '';
           rectRef.current = null;
         })}
       />
--- a/.yarn/versions/8e903555.yml
+++ b/.yarn/versions/8e903555.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-scroll-area": patch
+
+declined:
+  - primitives


packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -715,6 +715,7 @@ const ScrollAreaScrollbarImpl = React.forwardRef<
             // so we remove text selection manually when scrolling
             prevWebkitUserSelectRef.current = document.body.style.webkitUserSelect;
             document.body.style.webkitUserSelect = 'none';
+            if (context.viewport) context.viewport.style.scrollBehavior = 'auto';
```
andy-hook, line 718: Do we also need to do something to address the wheel interaction mentioned in the issue?

benoitgrelard: No, I tested using my mouse wheel, and using my trackpad and there was no issue with those.
Similarly, I tested with touch drag on my iPhone and it was fine too.

This makes sense to me as it makes sense that `scroll-behavior: smooth` takes that into account already.

packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -725,6 +726,7 @@ const ScrollAreaScrollbarImpl = React.forwardRef<
             element.releasePointerCapture(event.pointerId);
           }
           document.body.style.webkitUserSelect = prevWebkitUserSelectRef.current;
+          if (context.viewport) context.viewport.style.scrollBehavior = '';
```
andy-hook, line 729: Is there any risk that this event would fire? I'm curious as the capture release code is above this.

benoitgrelard: I don't think so, what are you thinking?

andy-hook: No, this looks good. Just raising the question just in case 😄

andy-hook (approved):

Issue #2171, opened 2023-05-29
Bug: Popover component closing after content dragging event on mobile when clicked after.


Mario857:
## Bug report
The popover component is not working properly when controlled. The issue occurs on mobile firstly but should be examined on desktop too. Tested on both chrome browser and mobile. Same issue.

### Current Behavior
When Popover is opened within the scrolling area and content outside is dragged/scrolled when clicking back to Popover.Content It closes Popover. But it should not as it was not clicked outside of Popover.

https://github.com/radix-ui/primitives/assets/26458264/3df18667-b763-469b-9045-7a03e49bc87c

### Expected behavior
It should not close Popover when clicking on it after scrolling.

### Reproducible example

[CodeSandbox Example](https://codesandbox.io/s/boring-tdd-kwoeb9?file=/src/App.tsx)

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |     @radix-ui/react-popover    |      1.0.5   |
| React            | n/a     |         |
| Browser          |     chrome, safari   |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |  chrome, safari   |         |

PR #2177, opened 2023-05-30
Fix dispatch onPointerDownOutside on touch when click inside

joaom00:
The click event is not properly cleared when canceled on touch devices, causing content to close when clicked inside.

Fixes https://github.com/radix-ui/primitives/issues/2171

### DIFF
--- a/packages/react/dismissable-layer/src/DismissableLayer.tsx
+++ b/packages/react/dismissable-layer/src/DismissableLayer.tsx
@@ -262,8 +262,7 @@ function usePointerDownOutside(
           handleAndDispatchPointerDownOutsideEvent();
         }
       } else {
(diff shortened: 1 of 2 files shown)


packages/react/dismissable-layer/src/DismissableLayer.tsx
```diff
@@ -261,6 +261,10 @@ function usePointerDownOutside(
         } else {
           handleAndDispatchPointerDownOutsideEvent();
         }
+      } else {
+        // We need to remove the event listener in case the outside click has been canceled
```
benoitgrelard, line 265:
```suggestion
        // We need to remove the event listener in case the outside click has been canceled.
```

benoitgrelard (approved):

PR #2178, opened 2023-05-30
Remove primitive element types from `Portal` part

### LINKED ISSUE #1974: [Portal] Props interface should not include primitive element types
Typescript is being informed of `div` prop types yet they are not forwarded in the `Portal` calls. We should omit the `primitive.div` types from the resulting props interface to avoid a discrepancy.

See https://github.com/radix-ui/primitives/discussions/1956#discussioncomment-5076435

joaom00:
Removed primitive element types from `Portal` part in all primitives and added JSDoc to the `container` prop in the `Portal` utility

Fixes #1974

### DIFF
--- a/packages/react/portal/src/Portal.tsx
+++ b/packages/react/portal/src/Portal.tsx
@@ -13,6 +13,9 @@ const PORTAL_NAME = 'Portal';
 type PortalElement = React.ElementRef<typeof Primitive.div>;
 type PrimitiveDivProps = Radix.ComponentPropsWithoutRef<typeof Primitive.div>;
 interface PortalProps extends PrimitiveDivProps {
+  /**
+   * An optional different container where the portaled content should be appended.
+   */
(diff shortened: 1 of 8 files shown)


packages/react/portal/src/Portal.tsx
```diff
@@ -13,6 +13,9 @@ const PORTAL_NAME = 'Portal';
 type PortalElement = React.ElementRef<typeof Primitive.div>;
 type PrimitiveDivProps = Radix.ComponentPropsWithoutRef<typeof Primitive.div>;
 interface PortalProps extends PrimitiveDivProps {
+  /**
+   * An optional different container where the portaled content should be appended.
```
benoitgrelard, line 17:
```suggestion
   * An optional container where the portaled content should be appended.
```

benoitgrelard (approved):

Issue #2180, opened 2023-06-01
Popover nested in Dialog does not open


kradical:
### Current Behavior
Slightly complicated, but.. 
- A Popover who's trigger renders differently based on open state
- and uses a Portal
- and is rendered in a Dialog

Does not open when the trigger is clicked. In my minimal reproduction case it only doesn't open on the first click, but in our production application it doesn't open at all. This happens because onOpenChange is being called twice in quick succession.

### Expected behavior

The popover should open

### Reproducible example

https://codesandbox.io/s/sad-darwin-rvpy1o?file=/src/App.tsx

### Suggested solution

Not sure, but it looks possibly introduced by https://github.com/radix-ui/primitives/commit/568fdc5927d91b50acb106a4931d8e865639de53.

### Additional context

We recently saw this behaviour when bumping radix-ui versions, specifically popover from `^1.0.5` to `^1.0.6`.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |     popover    |  1.0.6       |
| React            | n/a     |    18.2.0     |
| Browser          |     chrome    |  latest       |
| Assistive tech   |     n/a    |     n/a    |
| Node             | n/a     |    16.15.1     |
| npm/yarn         |    yarn     |   3.5.1      |
| Operating System |     macos    |    13.4     |

robbtraister:
The issue in our app seems to have been caused by the mutationObserver [here](https://github.com/radix-ui/primitives/commit/8a604454befce38b550456c7de6b9738c6fec85e#diff-578150720655fbca8e09f3dd24de414227526370525379e9ddee23f061a46037). I was able to achieve the expected behavior by reverting it locally.

PR #2182, opened 2023-06-01
Limit focus retention to when the document body is the activeElement

robbtraister:
### Description

DOM nodes may be removed from the tree without having focus. In this case, the mutationObserver is greedy and steals focus that it should not be managing. As described in the adjacent comment, when the focused node is removed, browser focus is assigned to the body. By ensuring the body is the activeElement, we avoid stealing focus when the removed node did not have focus to begin with.

Addresses https://github.com/radix-ui/primitives/issues/2180

### DIFF
--- a/.yarn/versions/310b65b7.yml
+++ b/.yarn/versions/310b65b7.yml
@@ -1,0 +1,13 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-context-menu": patch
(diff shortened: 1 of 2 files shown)


benoitgrelard (approved):
Hey @robbtraister, thanks for your contribution.
You are right, it was definitely a bit too eager. This is more explicit now and inline with the explanation above.

I have used an early return and removed the now unnecessary conditional too.

PR #2188, opened 2023-06-04
Add `use client` directive to support Next.js 13

zignis:
### Description

Adds `use client` directive to components so that they are treated as client components when using Next.js 13 with React server components.


jjenzz:
just a heads up that this is [not going to work with radix's current parcel set up](https://github.com/parcel-bundler/parcel/issues/9050).

jjenzz:
heads up, i've solved the parcel isue in my slot pr https://github.com/radix-ui/primitives/pull/2205

A7med3bdulBaset:
Any updates?

PR #2205, opened 2023-06-13
[Slot] fix `React.memo` when using `asChild`

jjenzz:
### Description

prevents `asChild` from breaking `React.memo`. here is a [sandbox for my original PoC](https://codesandbox.io/s/purple-sky-54qtrs?file=/src/App.js) where you can see it working.

worth noting that i have had to make `Slot` a client component (`"use client"`) to ensure it doesn't scream when used in an RSC because i've introduced a `useRef` and `useLayoutEffect`. unfortunately, this is [not going to work with parcel out of the box](https://github.com/parcel-bundler/parcel/issues/9050). to fix, i've made the following changes:

- upgraded parcel so i can add a custom optimizer
- created a custom optimizer to move `"use client";` to top of bundled output
- fixes the PURE display name babel plugin for `Slot` comp because the previous AST code assumed display name value would be a variable and not a literal string (which `Slot` has)

i built the `ssr-testing` app to make sure things still look the same (bundle output looked like the below both times) and tested in browser where things appear to be working well. let me know if there is anything else you'd like me to test here.

image


.parcelrc
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": [
+    "@parcel/config-default"
+  ],
+  "optimizers": {
+    "*.{mjs,js}": [
+      "./scripts/fix-use-client.js"
```
jjenzz, line 7: this is an optimizer instead of a babel plugin because it needs to run **after** parcel has bundled the code.

babel.config.js
```diff
@@ -8,10 +8,10 @@ const pureDisplayNames = () => ({
       if (
         path.node.left.type === 'MemberExpression' &&
         path.node.left.property.name === 'displayName' &&
-        path.node.right.name
+        (path.node.right.name || path.node.right.extra?.raw)
       ) {
         const COMPONENT = path.node.left.object.name;
-        const DISPLAY_NAME = path.node.right.name;
+        const DISPLAY_NAME = path.node.right.name || path.node.right.extra?.raw;
```
jjenzz, line 14: fixes this issue:
image

PR #2234, opened 2023-06-29
[Slot] Allow slot to override child's handlers instead of composing

bdsqqq:
First off, thanks for the amazing work. Radix raised the bar for accessibility and set an example for composable components I strive to reach in my codebases.

### Description

This PR aims to enable more complex behavior composition through the use of the `Slot` component by allowing it to overwrite the child's handlers instead of merging. 

To be conservative, I implemented a new prop `dangerouslyoverridehandlersinsteadofcomposing` that requires the consumer to explicitly set which handlers it wants to be overwritten. eg:
```
<Slot
    {...props}
    dangerouslyoverridehandlersinsteadofcomposing={['onClick']}
    onClick={(event) => {
        // ... 
    }}
 />
``` 

The patterns that emerged from this remind me of React's HoC but through JSX instead of function wrappers.

A concrete use case for this would be providing useful feedback for "disabled" buttons; For example: when a user reaches the maximum amount of teams allowed by their license, don't disable the "new team" button, instead, show a modal with useful feedback and actions on click. This behavior is achievable without the changes to Slot, but they allow for creating simple abstractions that can be easily distributed. See: https://slot-overrides-examples.vercel.app/ ([source](https://github.com/bdsqqq/slot-overrides-examples)). 

Some immediate questions I have are:
- Do you see value in this feature?
- Would it be better exposed in another way?

Any feedback is welcome, and I hope to get this PR merged to enable a rich ecosystem of composable behavior components.

### DIFF
--- a/.yarn/versions/4ff09dfe.yml
+++ b/.yarn/versions/4ff09dfe.yml
@@ -1,0 +1,12 @@
+releases:
+  "@radix-ui/react-slot": patch
+
+declined:
+  - "@radix-ui/react-alert-dialog"
+  - "@radix-ui/react-collection"
+  - "@radix-ui/react-dialog"
+  - "@radix-ui/react-menu"
+  - "@radix-ui/react-popover"
+  - "@radix-ui/react-primitive"
+  - "@radix-ui/react-select"
+  - "@radix-ui/react-tooltip"
--- a/.yarn/versions/4ff09dfe.yml
+++ b/.yarn/versions/4ff09dfe.yml
@@ -1,7 +1,8 @@
 releases:
-  "@radix-ui/react-slot": patch
+  "@radix-ui/react-slot": minor
 
 declined:
+  - primitives
   - "@radix-ui/react-alert-dialog"
   - "@radix-ui/react-collection"
   - "@radix-ui/react-dialog"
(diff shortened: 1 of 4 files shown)


jjenzz:
i'm not sure this would be the job for `Slot` given its entire purpose is composing stuff but @benoitgrelard or @andy-hook  can confirm. 

your use-case might be simpler to reason about [like this tho](https://codesandbox.io/s/wonderful-shannon-wv3n3f?file=/src/App.js) tbh:

```jsx
const Lock = React.forwardRef(({ children, locked, lockedFeedback, ...rest }, ref) => {
  const child = React.Children.only(children);
  if (!locked) return <>{child}</>;

  return (
    <Popover.Root>
      <Popover.Trigger asChild {...rest} ref={ref}>
        {React.cloneElement(child, { onClick: () => {} })}
      </Popover.Trigger>
      <Popover.Content>{lockedFeedback}</Popover.Content>
    </Popover.Root>
  );
});
```

bdsqqq:
@jjenzz Thanks for the feedback! 🙌 

I think you're right, my thought process was coming from the perspective that there are examples of preventing the parent's handler, so it would make sense to have the opposite too. The approach you shared does look better, taking `<Slot>` out as the middleman and consuming react's primitives directly is way simpler.

Waiting on input from others but am inclined to close this since the change really doesn't make sense if the role of `<Slot />` is to always merge behavior and never override.
 
_ah, and Codesandbox strikes again! I think sandboxes are private by default. But I got the snippet you shared running locally, ty again!_

bdsqqq:
Closing considering the above comments, thanks for the feedback and clarification on the purpose of the component!

Issue #2233, opened 2023-06-28
Toast doesn't dismiss because of pause/resume issue


jvzaniolo:
# Pause/Resume issue on Toast when changing durations

## Context

I created a way to handle multiple toasts inspired by the Shadcn UI Toast component and react-hot-toast library.

One possible toast is the `toast.promise` which displays a loading, success, or error toast depending on the Promise status.
It uses an update function to merge the success and error props with the loading toast (example in the CodeSandbox).

I'm using `Infinity` as the loading Toast duration so it doesn't close until the Promise finishes.

## Issue

The issue happens when the duration is paused (either by focusing on another window, focusing on the Toast itself, or hovering over it)

When I do that, the Toast doesn't close automatically.</p>

I can still swipe and close it or click the `<Toast.Close />` button.

## Example

Here's a CodeSandbox with the example:

https://codesandbox.io/s/radix-toast-pause-resume-issue-q6ghfn?file=/src/App.tsx

## Related issues

- #2221

jvzaniolo:
I forked the project and ran the Storybook. This issue also happens in your Promise Toast.

The toast automatically closes if I don't trigger the pause/resume functions, but it doesn't work correctly when I do.

![CleanShot 2023-06-29 at 18 39 12](https://github.com/radix-ui/primitives/assets/54036572/7fbce992-853a-45d1-992f-d371cf3c2ee3)

jvzaniolo:
I think the issue is here:

https://github.com/radix-ui/primitives/blob/eca6babd188df465f64f23f3584738b85dba610e/packages/react/toast/src/Toast.tsx#L491

The value of `closeTimerRemainingTimeRef.current` doesn't change from `Infinity` to the new `duration` when the component rerenders.

And on this line

https://github.com/radix-ui/primitives/blob/eca6babd188df465f64f23f3584738b85dba610e/packages/react/toast/src/Toast.tsx#L521

the Math is actually `closeTimerRemainingTimeRef.current = Infinity - elapsedTime;` 

Changing it from a ref to a normal variable fixes the issue.

I'll submit a PR

paul-sachs:
I second this, I recently ran into this when trying to create a notification whose duration is dynamic (based on when a promise resolves).

joepetrillo:
Any updates?

gdfreitas:
We're also facing this issue. Would anyone kindly give us an status update?

richard-hunter-hpe:
[...]

PR #2236, opened 2023-06-29
Fix/toast pause resume issue

jvzaniolo:
### Description

Change closeTimerRemainingTime from ref to variable

This enables the duration to update from Infinity to a number when the component rerenders

Closes https://github.com/radix-ui/primitives/issues/2233

### DIFF
--- a/packages/react/toast/src/Toast.tsx
+++ b/packages/react/toast/src/Toast.tsx
@@ -488,7 +488,7 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
     const swipeDeltaRef = React.useRef<{ x: number; y: number } | null>(null);
     const duration = durationProp || context.duration;
     const closeTimerStartTimeRef = React.useRef(0);
-    const closeTimerRemainingTimeRef = React.useRef(duration);
+    let closeTimerRemainingTime = duration;
     const closeTimerRef = React.useRef(0);
     const { onToastAdd, onToastRemove } = context;
     const handleClose = useCallbackRef(() => {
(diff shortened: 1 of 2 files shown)


packages/react/toast/src/Toast.tsx
```diff
@@ -93,7 +93,13 @@ const ToastProvider: React.FC<ToastProviderProps> = (props: ScopedProps<ToastPro
         toastCount={toastCount}
         viewport={viewport}
         onViewportChange={setViewport}
-        onToastAdd={React.useCallback(() => setToastCount((prevCount) => prevCount + 1), [])}
+        onToastAdd={React.useCallback(() => {
+          // Reset the isClosePausedRef when a new toast is added
+          isClosePausedRef.current = false;
+          // I assume this one has the same issue as the one above
+          // isFocusedToastEscapeKeyDownRef.current = false;
```
jvzaniolo, line 100: I assume the `isFocusedToastEscapeKeyDownRef` has the same issue of not initializing with `false` between rerenders, so we might need to reset that too

packages/react/toast/src/Toast.tsx
```diff
@@ -535,6 +541,10 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
     // we include `open` in deps because closed !== unmounted when animating
     // so it could reopen before being completely unmounted
     React.useEffect(() => {
+      // Needs to reset the timer when the toast is rerendered with the new duration
+      // in case the toast is initially rendered with `duration={Infinity}`.
+      closeTimerRemainingTimeRef.current = duration;
```
jvzaniolo, line 546: BTW, this solution fixes the issue of updating durations in general, not just for `Infinity`

jvzaniolo:
Another way to quickly fix both issues is to add a key to `<ToastImpl />`.

But I'm not sure if this solution is acceptable. Feel free to choose 😋

CleanShot 2023-06-30 at 14 52 01

paul-sachs:
I second this, I recently ran into this when trying to create a notification whose duration is dynamic (based on when a promise resolves).

HugeLetters:
I would also suggest moving `clearTimeout` above the guard clause - because we always want to cancel the timeout when trying to start a new one.
https://github.com/radix-ui/primitives/blob/be208ef8ce3f24b6616d916688ee1e4812d5e0e3/packages/react/toast/src/Toast.tsx#L510-L512

Say I had a duration of 5 seconds but then changed it to Infinity - if left as is it would still close after 5 second. With this fix it will correctly update itself to never close.

My personal use-case is I have my notifications stacks(like in Sonnar in shadcn) - and I need duration to be dynamic so that only the top notification has a timer on it and all the notifications below are on hold until they are on top.

thexpand:
Any reason why this is still open after over a year?

gdfreitas:
We are still interested in this fix, it would be nice to hear from a project maintainer

jvzaniolo:
Fix is to use https://ark-ui.com/react/docs/components/toast or https://sonner.emilkowal.ski/

zhaoyu1999:
Does anyone know why this is still open?

Issue #2241, opened 2023-07-03
Dropdown -  Dropdown is not expanding in mobile device's mozilla firefox browser.


hasan-deriv:
### Current Behavior

[Screen Record](https://www.awesomescreenshot.com/video/18813673?key=d66c4459acd6f4e9f1e8c1b41504d5ab)

### Expected behavior

It should be behave as dropdown.

### Reproducible example

It is happening when `DropdownMenu.Root`'s prop `modal` is `false`

[CodeSandbox Template](https://codesandbox.io/p/sandbox/zealous-fog-mzw4rz?file=%2FApp.jsx%3A17%2C6-17%2C23)

frstua:
I have the same behavior. Also you can try to reproduce it here:

https://www.radix-ui.com/
https://0x0.st/H1qh.mp4

But here it works as expected:
https://www.radix-ui.com/docs/primitives/components/dropdown-menu
https://0x0.st/H1bs.mp4

SaveliiLukash:
+1 Experiencing the same behavior in both Firefox on Android and desktop Firefox responsive design mode.

Adding `onFocusOutside={ (event) => event.preventDefault() }` logic specifically to Firefox works, but is a really hacky solution.

Would be nice to have this case handled by the library itself.

benoitgrelard:
Duplicate of #2565

Issue #1634, opened 2022-08-29
[Violation] Forced reflow while executing JavaScript took <N>ms


dan5py:
Some elements like (Dialog, Dropdown menu, Select, Popover, ...) cause a Forced reflow and this can lower the performance of the component and the entire page.
<img width="450" alt="image" src="https://user-images.githubusercontent.com/59441403/187240270-a23e7fa9-2f67-449a-ba20-14e3df331917.png">

andy-hook:
@Dan5py These types of performance warnings depend on many variables including complexity of page, performance of client and other factors, there might be more we can do to optimise but without a better understanding of your project and usage it would be hard to assess in a meaningful way.

Are you able to provide a "reduced" example showing these violations?

christiankaindl:
I am having the same issue, in my case with Overlay. Clicking the Overlay trigger takes up to 1 second, depending on the size of the page. The larger the page the bigger the lag. E.g. [this page](https://dev.gesetzefinden.at/bundesrecht/verordnungen/20009315) seems to have no delay, while [this page](https://dev.gesetzefinden.at/bundesrecht/bundesgesetze/stpo) suffers quite a bit. The issue seems a bit more severe on Firefox compared to Chrome (at least on my Intel MacBook Air), but it is very noticeable nonetheless. The following recording was taken on this page: https://dev.gesetzefinden.at/bundesrecht/bundesgesetze/abgb

<img width="842" alt="Screen Shot 2022-10-01 at 14 00 44" src="https://user-images.githubusercontent.com/15364860/193408710-5fb1f3b0-68b9-44cf-9603-900a64afb554.png">

Here's a video showing the issue on more detail:

https://user-images.githubusercontent.com/15364860/193408962-9b3d177b-7388-4306-8e82-91e5dc3fcdda.mov

However, the issue only persists when `modal` is set to `true`, which leads me to believe that the issue is caused by the "aria-hidden" package on this line, as that is the only noticeable difference I could find between modal and non-modal implementations:

https://github.com/radix-ui/primitives/blob/1f66c88538da13f5ec6b4e97f215268df1a3baa2/packages/react/dialog/src/Dialog.tsx#L262

Sam-Afshari:
Same here with select component

christiankaindl:
> It's more likely to be an issue with re-layout because of react-remove-scroll which restyles the body.

[...]

PR #2253, opened 2023-07-05
[Select] Add non modal mode in `popper` position

### LINKED ISSUE #1927: [Feature] [Select | Menu] Make the hideOthers feature optional

joaom00:
### Description

As suggested here https://github.com/radix-ui/primitives/pull/2004#issuecomment-1556985859, this PR adds non-modal mode only for the `popper` position since that mode for `item-aligned` position needs to be thought a little more.

Closes #1927

### DIFF
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -66,6 +66,7 @@ type SelectContextValue = {
   dir: SelectProps['dir'];
   triggerPointerDownPosRef: React.MutableRefObject<{ x: number; y: number } | null>;
   disabled?: boolean;
+  modal: boolean;
 };
 
 const [SelectProvider, useSelectContext] = createSelectContext<SelectContextValue>(SELECT_NAME);
@@ -92,6 +93,7 @@ interface SelectProps {
   autoComplete?: string;
   disabled?: boolean;
   required?: boolean;
+  modal?: boolean;
 }
 
 const Select: React.FC<SelectProps> = (props: ScopedProps<SelectProps>) => {
@@ -109,6 +111,7 @@ const Select: React.FC<SelectProps> = (props: ScopedProps<SelectProps>) => {
     autoComplete,
     disabled,
     required,
+    modal = true,
   } = props;
   const popperScope = usePopperScope(__scopeSelect);
   const [trigger, setTrigger] = React.useState<SelectTriggerElement | null>(null);
@@ -159,6 +162,7 @@ const Select: React.FC<SelectProps> = (props: ScopedProps<SelectProps>) => {
         dir={direction}
         triggerPointerDownPosRef={triggerPointerDownPosRef}
         disabled={disabled}
+        modal={modal}
       >
         <Collection.Provider scope={__scopeSelect}>
           <SelectNativeOptionsProvider
@@ -462,23 +466,8 @@ type SelectPopperPrivateProps = { onPlaced?: PopperContentProps['onPlaced'] };
 
 interface SelectContentImplProps
   extends Omit<SelectPopperPositionProps, keyof SelectPopperPrivateProps>,
-    Omit<SelectItemAlignedPositionProps, keyof SelectPopperPrivateProps> {
-  /**
-   * Event handler called when auto-focusing on close.
-   * Can be prevented.
-   */
-  onCloseAutoFocus?: FocusScopeProps['onUnmountAutoFocus'];
-  /**
-   * Event handler called when the escape key is down.
-   * Can be prevented.
-   */
-  onEscapeKeyDown?: DismissableLayerProps['onEscapeKeyDown'];
-  /**
-   * Event handler called when the a `pointerdown` event happens outside of the `DismissableLayer`.
-   * Can be prevented.
-   */
-  onPointerDownOutside?: DismissableLayerProps['onPointerDownOutside'];
-
+    Omit<SelectItemAlignedPositionProps, keyof SelectPopperPrivateProps>,
+    SelectContentModalityImplPublicProps {
   position?: 'item-aligned' | 'popper';
 }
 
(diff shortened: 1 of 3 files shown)


packages/react/select/src/Select.tsx
```diff
@@ -990,36 +971,196 @@ const SelectPopperPosition = React.forwardRef<
...
+      ref={contentRef}
+      // we make sure we're not trapping once it's been closed
+      // (closed !== unmounted when animating out)
+      trapFocus={context.open}
+      // make sure to only disable pointer events when open
+      // this avoids blocking interactions while animating out
+      disableOutsidePointerEvents={context.open}
+      disableOutsideScroll
+      {...modalityProps}
+      onFocusOutside={(event) => event.preventDefault()}
```
joaom00, line 1138: I have seen a problem on mobile in non-modal mode where the `Trigger` gets focus after being opened. Maybe I need to put this line in both modal and non-modal mode? 

I noticed that this same issue also occurs in the dropdown menu, I believe it is because of react 17? In the documentation it doesn't seem to have any issues :thinking:

packages/react/select/src/Select.tsx
```diff
@@ -66,6 +66,7 @@ type SelectContextValue = {
   dir: SelectProps['dir'];
   triggerPointerDownPosRef: React.MutableRefObject<{ x: number; y: number } | null>;
   disabled?: boolean;
+  modal: boolean;
```
kachkaev, line 69:
```suggestion
  modal?: boolean;
```

kachkaev:
👋 @benoitgrelard! Any chance this PR could be looked into? It potentially fixes a rather serious performance issue – see https://github.com/radix-ui/primitives/issues/1634#issuecomment-1264347705 and below for details. I use Radix Select in a project via https://ui.shadcn.com/docs/components/select. When a page has heavy markup like a large table, clicking on select makes the app freeze even on a pretty recent MacBookPro with M1. The lag is especially visible in Safari, but it’s also noticeable on Chrome and FF (especially with x4 throttling).

frykten:
@joaom00 Olá! This is just a ping :slightly_smiling_face:  is this PR/subject still alive?

joaom00:
@frykten Yes definitely. Unfortunately, I think the team didn't have time to review the PR...

SaveliiLukash:
Came here from #2241 

Thanks for the hint on `onFocusOutside={ (event) => event.preventDefault() }`. Might be a temporary, yet hacky solution, which is far from ideal.

Looking forward for this issue to be handled by the library itself.

maschad96:
beep boop bump, this is a nice contribution – any chance it could get a review soon @benoitgrelard @andy-hook?

currently resorting to using `DropdownMenu` for the non-modal support when semantically I should be using `Select`

andreareginato:
Any chance for this to be merged?

CodingAleCR:
I would also prefer it if this fix was checked and worked/merged. Is there anyway we can help to push it over the line?

bmutahhar:
I am facing the same issue. Any updates on when this will be merged?

andreareginato:
@joaom00 Thanks for you work, it's a good one. I've bee applying on my current project, and here is what I wanted to share with you. 

I've been using the "new" `Select` component you created with `modal={false}`. 

The strange thing is that it works perfectly on Safari and Firefox, but it does not work on Chrome (v.125.0.6422.61 - arm64). I'm not able to understand what is not working. Maybe you have some suggestions on the topic. 

Thanks a lot for your work 🥇

atancredi:
Hi all! This is just a ping, any updates on a possible merge?

donghoon-song:
Any..

ttpss930141011:
Will be appreciated if the PR is merged

Taaqif:
Any updates on when this might be merged?

chaance:
This one is on my radar, though I'd prefer to resolve it without additional wrapper components. One of our goals over the coming months is to reduce React tree bloat to improve the React dev tools experience.

kochai:
Bump, please fix this it's really annoying!

u3ih:
is that still alive....?

PR #2433, opened 2023-10-06
fix: Popover and Dialog trigger issue with shadow dom

artchen-db:
### Description

Popover and Dialog currently rely on event.target to determine if a dismiss click was on the trigger element. However, when rendered inside shadow root, the dismiss event listener bound on document will always get the shadow host as event.target ([reference](https://javascript.info/shadow-dom-events)). This causes clicking on the target to close then immediately re-open the popover/dialog.

This commit fixes the issue by applying 2 changes:
1. Use event.composedPath() instead of event.target, this gives the full path of elements the event propagated through as long as the shadow mode is open.
2. Check not only the custom event but also the original event. This is because the custom event's composedPath starts from the shadow host.

### Testing

Check out this sandbox: https://codesandbox.io/s/radix-ui-popover-fixing-bug-with-shadow-dom-sqxyjs?file=/src/App.js

The local compiled copy of popover.js has this fix commit, and it works in shadow root.

The popover imported from npm has issue with shadow root.

Outside of shadow DOM both behave as expected.

### DIFF
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -336,8 +336,11 @@ const DialogContentNonModal = React.forwardRef<DialogContentTypeElement, DialogC
           // Prevent dismissing when clicking the trigger.
           // As the trigger is already setup to close, without doing so would
           // cause it to close and immediately open.
-          const target = event.target as HTMLElement;
-          const targetIsTrigger = context.triggerRef.current?.contains(target);
+          const triggerEl = context.triggerRef.current;
+          const targetIsTrigger =
+            triggerEl &&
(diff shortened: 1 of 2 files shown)


molvqingtai:
It seems that it is not completely solved. ComposedPath().includes(triggerEl) is only valid for mode:'open', because when mode:'closed', path will not contain shadow's child nodes.

Check out this sandbox: 
https://codesandbox.io/s/radix-ui-popover-fixing-bug-with-shadow-dom-forked-j6n5sj?file=/src/index.js

Reference: https://javascript.info/shadow-dom-events#:~:text=trees%20are%20completely-,hidden,-.

PR #2484, opened 2023-10-25
[Checkbox] Support form attribute

Manduro:
### Description

This lets you associate the underlying native `<input>` element to a `<form>` anywhere in the document instead of requiring an ancestor `<form>` element.

Related change for Select: https://github.com/radix-ui/primitives/pull/2447

### DIFF
--- a/packages/react/checkbox/src/Checkbox.tsx
+++ b/packages/react/checkbox/src/Checkbox.tsx
@@ -83,6 +83,7 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
           data-state={getState(checked)}
           data-disabled={disabled ? '' : undefined}
           disabled={disabled}
+          form={form}
           value={value}
           {...checkboxProps}
           ref={composedRefs}
--- a/packages/react/checkbox/src/Checkbox.tsx
+++ b/packages/react/checkbox/src/Checkbox.tsx
@@ -48,6 +48,7 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
       defaultChecked,
       required,
       disabled,
+      form,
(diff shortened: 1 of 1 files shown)


chaance:
I definitely want to support the form prop, but we still need to do the DOM check if the prop isn't passed.

Manduro:
@chaance That happens automatically:

> If this attribute is not set, the <button> is associated with its ancestor <form> element, if any.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#form

chaance:
> @chaance That happens automatically:
> 
> > If this attribute is not set, the `<button>` is associated with its ancestor `<form>` element, if any.
> 
> [developer.mozilla.org/en-US/docs/Web/HTML/Element/button#form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#form)

Yes, but we use `isFormControl` to determine whether or not to render the bubble input in the first place, so my point is we still want to check for an ancestor form if there is no form attribute.

Manduro:
@chaance That is what the code does. `button.form` references the ancestor form automatically as per the mdn docs.

chaance:
Ohhh that's my bad, I misread your code. Yeah this makes sense, thanks 👍

chaance:
Actually sorry, gotta push back once more: if the user renders the checkbox as anything other than a button via `asChild` this would still break. I think we need to account for that.

chaance:
I'm actually going to close this for now. Per https://github.com/radix-ui/primitives/issues/2530#issuecomment-2380105452 I'd like to introduce a more wholistic solution to cover all affected primitives, and I still need to do some manual testing for edge cases.

Tracking in https://github.com/radix-ui/primitives/issues/2530

PR #2489, opened 2023-10-27
Bump react-remove-scroll to 2.5.7 (see changelog) to fix behavior in shadow DOM

glenntws:
### Description

react-remove-scroll fixes the problem of all children of RemoveScroll not being scrollable via mouse wheel or touch when created inside a shadow DOM. (See [react-remove-scroll #98](https://github.com/theKashey/react-remove-scroll/pull/98))

Tested in my local project and scroll now works properly in Dialogs etc. in shadow DOMs without seeing any negative change to other use cases.

### DIFF
--- a/.yarn/versions/96889917.yml
+++ b/.yarn/versions/96889917.yml
@@ -1,0 +1,17 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-context-menu": patch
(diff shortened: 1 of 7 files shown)


jwm0:
Can we get this merged in please 🙏  @benoitgrelard

chaance:
Looks like this has already been updated. Sorry it took a while to get a reply!

PR #2518, opened 2023-11-11
Remove unnecessary useEffect in React Primitive component

Meemaw:
### Description

This PR removes what seems to be an unnecessary `useEffect` in React Primitive component. Usage of `useEffect` breaks compatibility with RSC (React Server Components). Because this is used in such a low lever primitive/component, all `radix-ui` components become unusable in RSC.

### DIFF
--- a/packages/react/primitive/src/Primitive.tsx
+++ b/packages/react/primitive/src/Primitive.tsx
@@ -46,10 +46,6 @@ const Primitive = NODES.reduce((primitive, node) => {
     const { asChild, ...primitiveProps } = props;
     const Comp: any = asChild ? Slot : node;
 
-    React.useEffect(() => {
-      (window as any)[Symbol.for('radix-ui')] = true;
-    }, []);
-
     return <Comp {...primitiveProps} ref={forwardedRef} />;
   });
 


packages/react/primitive/src/Primitive.tsx
```diff
@@ -46,10 +46,6 @@ const Primitive = NODES.reduce((primitive, node) => {
     const { asChild, ...primitiveProps } = props;
     const Comp: any = asChild ? Slot : node;
 
-    React.useEffect(() => {
-      (window as any)[Symbol.for('radix-ui')] = true;
```
Meemaw, line 50: If this is for some reason needed, we can achieve it without `useEffect` as well.

sunwrobert:
@benoitgrelard @jjenzz please take a look 🙏🏼, would unblock a lot of people on NextJS

BradMcGonigle:
Any update on why or why not this wouldn't be merged?

chaance:
This is as-intended. Writing to the window is a side-effect. Instead we need to expose the primitive as a client component. Feel free to open another PR for that.

In the mean time, you can work around this in your own app by re-exporting whatever component is causing issues from your own module marked as a client component, then import from there instead of Radix.

```tsx
'use client';

export { whatever } from '@radix-ui/whatever';
```

chaance:
Actually it looks like this was done anyway in some other commit. I'd be a bit more comfortable if we didn't do this in render but I assume Benoit decided it was better to not make primitive a client component.

Issue #1912, opened 2023-01-28
[DropdownMenu] `DropdownMenu.Trigger` reacts to scrolling on mobile devices


TmLev:
### Current Behavior

Scrolling on mobile devices may trigger `DropdownMenu.Trigger` even when not intended.
In my case, I have a table where rows have `DropdownMenu.Trigger` as a last column. Scrolling with right thumb almost always triggers some row's dropdown menu.

https://user-images.githubusercontent.com/22966523/215288503-eb96c5c3-1b14-4ad3-b877-3956c3c092c1.mov

### Expected behavior

Scrolling shouldn't act as a regular tap.

### Reproducible example

https://www.radix-ui.com/docs/primitives/components/dropdown-menu#trigger

Toggle responsive mode.

### Suggested solution

This may be hard, I don't know. Track movement and open menu only when touch cursor doesn't move far enough?..

### Additional context

Nothing.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | @radix-ui/react-dropdown-menu | 2.0.2 (latest) |
| React            | n/a     | 18.2 |
| Browser          | Chrome | Any recent one |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System | Android and macOS | 13 / 12 |

Fensterbank:
I'm facing the same issue in my app.
The behavior seems to be caused by using `onPointerDown` event.
https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/dropdown-menu/src/DropdownMenu.tsx#L117

This event fires always when the trigger is touched. Wouldn't it better to use `onClick`, which also fires on mobile devices, but wouldn't fire when the user is only scrolling?

Fensterbank:
@TmLev In case you're still blocked by this issue, I created a local patch which solves the issue in our product.

_patches/@radix-ui+react-dropdown-menu+2.0.6.patch_
```
diff --git a/node_modules/@radix-ui/react-dropdown-menu/dist/index.js b/node_modules/@radix-ui/react-dropdown-menu/dist/index.js
index c05924d..205468b 100644
--- a/node_modules/@radix-ui/react-dropdown-menu/dist/index.js
+++ b/node_modules/@radix-ui/react-dropdown-menu/dist/index.js
[...]

PR #2540, opened 2023-11-21
[DropdownMenu] replace onPointerDown by onClick to avoid scroll prevention on mobile

Fensterbank:
### Description

Scrolling on mobile devices trigger DropdownMenu.Trigger even when not intended.
This is caused by using `onPointerDown`.

The problem is reported in issue https://github.com/radix-ui/primitives/issues/1912

Scrolling is prevented and the DropdownMenu opens, which is not intended by users in that case.
By using `onClick` everything seems to work normal while users can still scroll on mobile devices.

Since this is my first PR and I don't know the codebase of Radix UI well, I'm not sure, what implications or side-effects this change could have. In my tests, everything worked fine by just using `onClick`.

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -114,7 +114,7 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           disabled={disabled}
           {...triggerProps}
           ref={composeRefs(forwardedRef, context.triggerRef)}
-          onPointerDown={composeEventHandlers(props.onPointerDown, (event) => {
+          onClick={composeEventHandlers(props.onClick, (event) => {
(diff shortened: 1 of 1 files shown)


joaom00:
Hey @Fensterbank, replace onPointerDown by onClick will break the ability to select an item in one click lifecycle (down on trigger, move, up on item)

Fensterbank:
Thanks for your answer. I see. I was afraid there is a reason for using `onPointerDown`.

Not sure, what could be another way to fix the mentioned issue then... maybe discovering if we are in touch mode and use `onPointerDown` only if we are having an actual mouse? :thinking: 
The one click lifecycle seems to be desktop/mouse only, in mobile view with touch events, it doesn't work (and would be unexpected UX anyways).

joaom00:
> maybe discovering if we are in touch mode and use onPointerDown only if we are having an actual mouse?

Yes, I believe that would be a way

Issue #2542, opened 2023-11-22
[Form] Custom matcher will not be triggered when you copy text or autofill


ChocoLZS:
### Current Behavior

- Custom matcher will not be triggered when you paste text into input element.
- Autofill also has this problem.

> This video only demonstrates the problem when pasting.

https://github.com/radix-ui/primitives/assets/61224208/b4cceb98-efb1-4eb1-a55f-9d3d86cca831

### Expected behavior

- The custom form validation should be excuted when submitting, regardless of whether the validation passes or fails.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/p/devbox/hardcore-chatterjee-7gvd8y)

### Your environment

Radix-ui form official Codesandbox

```
{
  "dependencies": {
    "@radix-ui/colors": "latest",
    "@radix-ui/react-form": "latest",
    "@radix-ui/react-icons": "latest",
    "classnames": "latest",
    "react": "latest",
    "react-dom": "latest"
  },
  "devDependencies": {
    "@vitejs/plugin-react": "latest",
    "autoprefixer": "latest",
    "postcss": "latest",
    "tailwindcss": "latest",
    "vite": "latest"
  },
  "scripts": {
    "start": "vite"
  }
}
```

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |  react-form   |  0.0.3  |
| React            | n/a     |  latest  |
| Browser          |  codesandbox  |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System | windows  |   11   |

jessefischer-workco:
Funny, I came here from an almost identical problem on the same day you posted this. I believe the same issue is happening with the following steps:
1. Type "equal" into the input box
2. Blur the input
3. Observe the "Custom Matcher error" message
4. Refocus on the input box
5. Delete any part of the word "equal" and then re-type it
6. Blur the input
7. Observe that there is no "Custom Matcher error" message anymore and that you are allowed to submit the form.

rwbrockhoff:
I was able to get the form to validate a custom matcher for copy and paste: 

`const handleMatcher: SyncCustomMatcher = (value) => value === 'equal';`

PR #2560, opened 2023-12-02
Bug fix matcher not triggering

mrtnpar:
### Description

This PR intends to fix issue https://github.com/radix-ui/primitives/issues/2542

Why this issue occurs?

Since form input uses the native `change` event, programmatically updates doesn't emits.
Also if a validation error has already triggers and `input.value` gets updated with [same value](https://github.com/radix-ui/primitives/issues/2542) validation disappear, because input event doesn't detect a change hence event `change` not emits.
Currently this is happening for custom and non-custom matchers, i.e. `type="email"` inputs.

Before
https://github.com/radix-ui/primitives/assets/74157/d2be0df4-403c-491d-bf66-a1bec00e33d7

After
https://github.com/radix-ui/primitives/assets/74157/968d2091-fdd3-492d-b900-986a55707ae1

### DIFF
--- a/packages/react/form/src/Form.tsx
+++ b/packages/react/form/src/Form.tsx
@@ -371,6 +371,24 @@ const FormControl = React.forwardRef<FormControlElement, FormControlProps>(
       }
     }, [name, onFieldValiditionClear]);
 
+    React.useEffect(() => {
+      const control = ref.current;
+      if (control) {
+        // It emits a 'change' event on blur since this is an uncontrolled input.
+        // For cases when programatically `control.value` or same previous value is set on input,
+        // `change` event is not emitted, so the following manually emits on `blur`.
+        // It skips when value is "" so focus then blur on input doesn't trigger validation.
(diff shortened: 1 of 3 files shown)


packages/react/form/src/Form.tsx
```diff
@@ -371,6 +371,24 @@ const FormControl = React.forwardRef<FormControlElement, FormControlProps>(
...
+        // `change` event is not emitted, so the following manually emits on `blur`.
+        // It skips when value is "" so focus then blur on input doesn't trigger validation.
+        const handleBlur = () => {
+          if (control.value === '') {
+            return;
+          }
+          control.dispatchEvent(new Event('change'));
+        };
+        control.addEventListener('blur', handleBlur);
+        return () => control.removeEventListener('change', handleBlur);
```
karlguillotte, line 388: Looking at the code, are you sure you want to remove the 'change' event ?

PR #2587, opened 2023-12-13
Fix checkbox missing name prop when is not inside a form

thefalked:
### Description

This `primitive` uses a `button` and when inside a form adds a `input`. The name was destructured from the component and passed only to the input.

Fix:
- It was added the `name` prop to the button if the checkbox is `not` inside a form.


chaance:
I think this is probably OK but I'm reconsidering how we handle our bubble fields which would negate the need here. Need to think through this a bit before I merge.

PR #2591, opened 2023-12-14
Add forward ref to react/select:Select

### LINKED ISSUE #2590: Support ref in Select Component
### Overview

The Select component does not allow a ref to be passed to the underlying html `select` element. This breaks [react-hook-form](https://react-hook-form.com/docs/useform/register)'s `register()` function; We are using [remix-forms](https://remix-forms.seasoned.cc/) which uses `register()` to allow [any custom component](https://remix-forms.seasoned.cc/examples/forms/custom-input) to be used in a form, as long as it can pass that ref to the underlying `select`. @jaimiz expressed the same problem in https://github.com/radix-ui/primitives/issues/1851#issuecomment-1653294400

The needed change is minimal; The `Select` component must accept a ref via `React.forwardRef` and [pass the ref](https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/select/src/Select.tsx#L182) to `BubbleSelect`. `BubbleSelect` even supports the ref param, but it isn't used. Perhaps `Select` was intended to have this behavior originally?

We have already tried the solution in our project with `patch-package` and saw it work. I will create a PR with the change shortly. Is there any reason this wouldn't be accepted? Is there an alternative we haven't considered?

N/A

- Users of react-hook-form and remix-form

Roanmh:
### Description

Closes #2590 (More detail in issue description)

`Select` component needs to forward a ref to `BubbleSelect` to be compatible with `react-hook-form`'s `register()` function.

### DIFF
--- a/.yarn/versions/fd8c5361.yml
+++ b/.yarn/versions/fd8c5361.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-select": patch
+
+declined:
+  - primitives
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -94,111 +94,114 @@ interface SelectProps {
   required?: boolean;
 }
 
-const Select: React.FC<SelectProps> = (props: ScopedProps<SelectProps>) => {
-  const {
-    __scopeSelect,
-    children,
-    open: openProp,
-    defaultOpen,
-    onOpenChange,
-    value: valueProp,
-    defaultValue,
-    onValueChange,
-    dir,
-    name,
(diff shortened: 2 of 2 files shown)


phsantiago:
Hello @benoitgrelard, @andy-hook any updates here?

chaance:
I think we need to fix this in other components that do this as well, but I'm not sure this is exactly the solution we want. I need to think through this a bit more to land on a more wholistic solution. Not sure we'll keep the bubble elements in the end.

michaelcpuckett:
What was the end result here?

PR #2595, opened 2023-12-15
fix toast a11y issues

### LINKED ISSUE #1750: Toast accessibility issues
### Axe devTools extension has identified 2 accessibility issues on Toast code example

`<ol>` is not allowed children of `role="status"`.

image

`role="status"` is not valid on `<li>` elements.

image

### Expected behavior

No accessibility issues. Toasts should not be list items.

### Reproducible example
- Install Axe devTools
- Open Radix Toast code example https://codesandbox.io/s/z688kv?module=App.js&file=/App.js
- Ensure toast is open by default
- Scan page using Axe devTools

### Suggested solution

Replace `<ol>` and `<li>`'s with alternative elements.

### Your environment

Production

| Radix Package(s) |    Toast     |    2.0.1     |
| Browser          |         |     Chrome    |

42tte:
### Description

- Add axe-core tests to Toast
- Fix accessibility issues
  - Remove aria-hidden from focusable element, these elements are already empty and won't be read by screen readers
  - Remove not allowed role=status from list item element, see documentation for list item under https://www.w3.org/TR/html-aria/#docconformance
  - (Remove unneeded default aria-atomic from role=status)

Closes #1750

Tag @jjenzz since she seems to have been involved in this before


adrianaferrugento:
We are having the same accessibility issue with the aria-hidden on the span. Is there a way to move this PR forward? I'm happy to pair on it or take it on!
Screenshot 2024-04-15 at 17 48 16

packages/react/toast/src/Toast.tsx
```diff
@@ -574,10 +572,6 @@ const ToastImpl = React.forwardRef<ToastImplElement, ToastImplProps>(
                 })}
               >
                 <Primitive.li
-                  // Ensure toasts are announced as status list or status when focused
-                  role="status"
```
jjenzz, line 578: my memory on this is flaky but the comment above seems to imply i intentionally added this to ensure "status list" was read aloud. 

i recall purposefully using `<li>`s for this to ensure it communicated the number of toasts in the toasts list too so @benoitgrelard @andy-hook it would be good to double-check none of that is broken with these changes.

lfantom:
Hi, is there any update on this? In particular I'm still seeing the accessibility fail as mentioned above for the tabindex on hidden spans. Would be good to get this fixed, as it causes a fail when running axe tests. Thanks

42tte:
Updated PR with latest main.
Test by first checking out 812804f and run Cypress tests to see it fail on said accessibility issues.

chaance (approved):

Issue #2598, opened 2023-12-18
aria-describedby is not assigned for tooltip if an aria-describedby is already present


wbolduc:
### Current Behavior

If the child of Tooltip.Trigger already has the aria-describedby property the tooltip is not correctly linked to the child

### Expected behavior
According to [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) aria-describedby is a list of ids. It follows that radix tooltip should correctly add it's own id to the trigger even if aria-describedby is already present

### Reproducible example

[...]

PR #2600, opened 2023-12-19
Correctly concatenate aria-describedby in a slot

wbolduc:
### Description

aria-describedby is technically a list of ids and should be concatenated kind of like className.

This pr fixes [my own issue](https://github.com/radix-ui/primitives/issues/2598) where this is described

Tests are included


stefanofa (approved):

wbolduc:
Is there something wrong with my pr?

chaance:
I'm not entirely sure we want to do this. Yes, `aria-describedby` can be more than one string but I think the component should be explicit about which element(s) are used as descriptions. With this solution there's no way for a consumer to explicitly override what Radix does internally, which may be valid in some cases.

PR #2616, opened 2023-12-29
Fixed Dropdown Menu trigger to work only onClick

unknown503:
### Description

Changed `onPointerDown` event to `onClick` event which triggers the dropdown only when you click and release the left mouse button. 

This was done because when you clicked without releasing the button, the dropdown opened, on pc this was not really an issue, but it was an annoying issue on mobile, where you could be scrolling down and unintentionally pressed any dropdown trigger, causing the menu to show even though it was not intended.

Further explanation can be found on this issue [#173](https://github.com/radix-ui/primitives/issues/2626) which is closed by this PR. Tests were performed to verify the new changes.

### DIFF
--- a/packages/react/dropdown-menu/src/DropdownMenu.tsx
+++ b/packages/react/dropdown-menu/src/DropdownMenu.tsx
@@ -114,7 +114,7 @@ const DropdownMenuTrigger = React.forwardRef<DropdownMenuTriggerElement, Dropdow
           disabled={disabled}
           {...triggerProps}
           ref={composeRefs(forwardedRef, context.triggerRef)}
-          onPointerDown={composeEventHandlers(props.onPointerDown, (event) => {
+          onClick={composeEventHandlers(props.onClick, (event) => {
             // only call handler if it's the left button (mousedown gets triggered by all mouse buttons)
             // but not when the control key is pressed (avoiding MacOS right click)
             if (!disabled && event.button === 0 && event.ctrlKey === false) {


joaom00:
Hi @unknown503, this change will break a feature where the user can select an item in one lifecycle click (down on trigger, move, up on item)

joaom00:
> > Hi @unknown503, this change will break a feature where the user can select an item in one lifecycle click (down on trigger, move, up on item)
> 
> on what story is that feature?

I don't think there is one

mech:
Can we have an option to use onClick because when I click I can "hold" without release then move to other screen area to reject any menu opening if I change my mind.

xvvvyz:
Is there a workaround for this? I would like to have the menu appear directly above the trigger point and because of `onPointerDown` it is not possible since whichever menu item appears below your cursor is automatically opened when you release.

chaance:
I think we need to stick with pointer events and distinguish between mouse and non-mouse devices. With a mouse, pointerdown is what we want, per @joaom00's comment. I agree the current behavior is broken for touch interactions.

I dealt with this exact issue in Reach UI, in case you want to reference and update your fix: https://github.com/reach/reach-ui/blob/main/packages/listbox/src/reach-listbox.tsx#L1005-L1019

chaance:
Fixed in https://github.com/radix-ui/primitives/pull/2939

chaance:
Fix was merged in `Select` [here](https://github.com/radix-ui/primitives/pull/2939), for reference

iamJoeTaylor:
@chaance the change to Select in the PR you reference doesn't fix the issue in DropDown right?

chaance:
This PR changes the behavior entirely, which is why it's closed. The actual solution would be to replicate the fix in https://github.com/radix-ui/primitives/pull/2939. This is on our radar, but PRs that address this without changing the core behavior for mouse users are welcome.

PR #2623, opened 2024-01-03
fix: Select should not crash if getItems returns empty array

### LINKED ISSUE #1795: [Select]: sometimes, brief flash of no items (resulting in undefined ref on initial render at top of viewport)
### Current Behavior

After rendering a select near the top of the screen, on first open I'm getting `TypeError: Cannot read properties of undefined (reading 'ref')` on Chrome, `items[0] is undefined` on Firefox. This is coming from:

https://github.com/radix-ui/primitives/blob/f0a7495bcd00faa14164b2f80721149238c85d9d/packages/react/select/src/Select.tsx#L594

And no surprise, but only when I use the `ScrollUpButton`: if I remove that, no more errors.

If I just refresh the page, now everything works again, and I can use the select.

Note that this only happens when I generate a component that holds the select menu using a button.

### Expected behavior

There's no behaviour lacking (just the same behaviour as when I hit refresh), just want to avoid the error.

### Reproducible example

Sadly I actually could not reproduce this. I copied as much as I could, and it looks something like https://stackblitz.com/edit/test-radix-select?file=pages/index.js (but this example works...).

Here's a demo of this example of the action I'm taking. In our site, at the point where I open the menu, the undefined ref error appears:

https://user-images.githubusercontent.com/6945891/202635810-fe19e796-e0de-42d2-8809-19154d6cb647.mp4

You'll see in the example code I have a bunch of forwarded refs, which I thought might be the solution, but it doesn't seem to make a difference. I can try and recreate our env more closely, but I'm not sure of the issue basically.

### Suggested solution

One possibility is maybe to make the comparison at https://github.com/radix-ui/primitives/blob/f0a7495bcd00faa14164b2f80721149238c85d9d/packages/react/select/src/Select.tsx#L594 check if the ref exists before reading current.

### Additional context

I realise this is a frustrating one to help with without a reproducible example. I would however be glad to try any debugging you can recommend instead too.

### Your environment

| Radix Package(s) |  Nearly all of them!       |  All at latest        |
| React            | NextJS    |    12.1.6     |
| Browser          |    Chrome, Firefox     |  C: 107 and 108, F: 107       |
| Node             | n/a    |   16      |
| npm/yarn         |      yarn   |  1.22.19     |
| Operating System |   MacOS      |    13.0     |

### LINKED ISSUE #2535: [Select] `ref` to collection is undefined
## Bug report

I've tried to replicate behaviour in our app as much as possible. We've select built using `@radix-ui/react-select` which allows to select an option that makes an input control visible which allows user to enter custom value. When user leaves the input, we run function which adjust entered text value. It works perfectly fine until `ScrollUp/DownButton` is starting to be visible because of smaller viewport or list being rendered near the top or bottom of the viewport.

### Current Behavior

If you focus the input and right after you try to open the select and the `ScrollUp/DownButton` can be rendered it's going to throw because the ref to collection's container is `undefined`.

### Expected behavior

I expect the component to not throw.

### Reproducible example

https://codesandbox.io/s/crazy-dream-mlrwc2

1. Select option between 40-50 to make sure the scroll up button will be rendered.
2. Focus the input
3. Click on the select's trigger.

### Suggested solution

I didn't dig into your source code so can't tell right now.

### Additional context

[N/A]

### Your environment

| Radix Package(s) |    `@radix-ui/react-select`     |  `2.0.0`   |
| React            | n/a     |     `18.3.0-canary-9377e1010-20230712` (Next.js app)    |
| Browser          |    Brave     |     `v1.60.118`    |
| Node             | n/a     |     `v16.20.0`    |
| npm/yarn         |    `pnpm`     |    `8.8.0`    |
| Operating System |    MacOS    |    Sonoma  (14.1.1 (23B81))   |

benj-dobs:
### Description

- Fixes #1795 (confirmed to fix reproduction at https://github.com/radix-ui/primitives/issues/1795#issuecomment-1705308325)
- Fixes #2535

I have not added tests because I couldn't make the simple reproduction case from #2535 work in the test environment. I could reproduce it in Storybook, but this seemed like an odd thing to add to Storybook, so I didn't.

It's possible this solution is incomplete, but it fixes both of the above reproductions, so it's probably good enough for now.


maxired:
Hi. I am facing a similar crash on a use case similar to the one of https://github.com/radix-ui/primitives/issues/2535 
I think that this fix make sens

izhan:
Our customers hit this error every single day. Is it possible to get this merged? Also happy to help out with this PR if it needs changes

benj-dobs:
In case anyone else is wondering about how maintainer activity seems to have just died without a word, there are some signs of life on the Discord:

https://discord.com/channels/752614004387610674/803656831704629298/1199950373650190336

No idea when this issue will get fixed but for now using patch-package or yarn resolutions is probably your best workaround.

tomtomprince:
We're also facing this issue. Going the patch-package route for now but it'd be great to merge this in

chaance:
Hey all! I'm back on the team and will be reviewing some old PRs, cleaning up the backlog in the coming weeks. This one's first on the list. Thanks for hanging with us 🙏

chaance (approved):

tatwater:
Hey @chaance for some reason this fix got undone in a more recent version 😢 Facing this issue again now that we're on v2.1.1

tatwater:
@blackbird-chrisf Oop I meant to amend my comment last month; I reached out on Discord and heard basically exactly what you discovered from Chance himself! Sorry for leaving it so a month later you still had to dig for that. Looking forward to 2.1.2!

chaance:
For all those following: this has shipped in v2.1.2

PR #2635, opened 2024-01-06
[Select] Closes on resize when the window size has changed

### LINKED ISSUE #2634: [Select] Select closes when on-screen keyboard is shown
### Current Behavior

When using a mobile device, showing the on-screen keyboard causes the `Select` close every time.

[Video showing the issue](https://github.com/radix-ui/primitives/assets/23644327/7d567d03-3257-471d-b5c2-ba68dc524485)

### Expected behavior

It should not close the `Select` when opening the on-screen keyboard.

### Reproducible example

1. On a mobile device go to the sandbox https://codesandbox.io/p/devbox/beautiful-night-sss6gx
2. Open the `Select`
3. Click on the "Search items" `input`
4. The `Select` will instantly close.

### Additional context

After a bit of debugging I've found that the resize event is triggering whenever the on-screen keyboard is shown.
https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/select/src/Select.tsx#L601-L609

### Your environment

| Radix Package(s) | react-select        | 2.0.0        |
| Browser          | Google Chrome        | 120.0.6099.145        |
| npm/yarn         | Yarn        | 1.22.19        |
| Operating System | Android        | 13        |

georgekaran:
Closes #2634


chaance:
Hey @georgekaran, sorry it took awhile to get to this! Two quick changes I'll request:
- Could you resolve the merge conflicts? I think it's probably just a filename issue, but if it's more involved let me know and I can help
- Can we not attach the event listener when the popover is already closed?

georgekaran:
Hey @chaance, just circling back on this one. I addressed both points from your review a while ago, fixed the filename/merge conflict and made sure the resize listener isn't attached when the popover is already closed. It's been sitting for a bit, so wanted to bump it in case it slipped off the radar. Happy to rebase or add a changeset if that helps get it across the line. Thanks!

chaance:
I'm actually thinking this might be a naive solution to your problem. The check would fail on iOS and even on Android for pages that opt into `interactive-widget=resizes-content`. A more robust solution would probably check the visualViewport rather than window, but even then you can't be sure that the resize occurred because of the virtual keyboard ([React Aria infers this by coordinating between the viewport's scroll and resize events](https://github.com/adobe/react-spectrum/blob/57a1c82c87100e80ebefe8f79626bea5d8499302/packages/react-aria/src/overlays/useOverlayPosition.ts#L345-L375)).

georgekaran:
Makes sense @chaance, thanks for explaining a bit more. I see two ways forward:

**A)** Keep closing on a genuine resize, but use the `visualViewport` resize→scroll signal to ignore the keyboard. The keyboard shows up as a `visualViewport` resize immediately followed by a scroll, whereas a real window resize / orientation change fires no such scroll, same heuristic React Aria uses, just applied to our "close" intent instead of "reposition".

**B)** Follow React Aria's lead more directly and reposition the content on resize instead of closing it. This sidesteps keyboard detection entirely (we never have to classify the resize), but it changes Select's current close-on-resize behavior, so it's the bigger call.

Both are still inference rather than a guarantee, as you noted. Do you have a preference before I rework it? I'll test on real iOS/Android either way.

chaance:
I'd hold off for now, I'd need to do more research before committing to a change here. I'd suggest patching the package if you need this behavior.

Important to note that this isn't the only issue you'll have trying to implement a filterable select since we don't currently have first-class combobox support. That's a much bigger lift and not a high priority right now.

PR #2651, opened 2024-01-14
feat(react-select): create number type support for value

milad-a-kareem:
### Description

1. **Type Definition Updates**: The `ItemData` type, `SelectContextValue`, `SelectProps`, `SelectContentContextValue`, `SelectItemContextValue`, and `SelectItemProps` have all been updated so that the `value` property can accept both `string` and `number` types.

2. **Function and Callback Adjustments**: Functions and callbacks that previously accepted `value` as a string now accept it as `string | number`. This includes `itemRefCallback`, `itemTextRefCallback`, and `shouldShowPlaceholder`.

3. **Component Property Changes**: Various component properties that dealt with `value` have been updated to align with the new type definition. This includes changes in `SelectContentImpl`, `SelectLabel`, `SelectItem`, and other components where `value` is used.

4. **Consistency and Robustness**: These changes ensure that the usage of `value` is consistent throughout the codebase and that the components can handle both strings and numbers, increasing the robustness of the code.

These modifications are significant as they improve the flexibility and type safety of the components involved, allowing for a more versatile and error-resistant implementation. If you're looking for a specific analysis or have a specific question about these changes, feel free to ask!

### DIFF
--- a/.yarn/versions/c0096325.yml
+++ b/.yarn/versions/c0096325.yml
@@ -1,0 +1,2 @@
+releases:
+  "@radix-ui/react-select": patch
--- a/.yarn/versions/c0096325.yml
+++ b/.yarn/versions/c0096325.yml
@@ -1,2 +1,3 @@
 releases:
   "@radix-ui/react-select": patch
+  primitives: patch
--- a/packages/react/select/src/Select.stories.tsx
+++ b/packages/react/select/src/Select.stories.tsx
@@ -57,7 +57,7 @@ export const Styled = () => (
 );
 
 export const Controlled = () => {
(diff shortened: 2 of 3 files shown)


chaance:
This is a bit more difficult in practice. Making this change would result in type errors, which is definitely not worth doing for a patch (and may not be worth it at all, IMO). 

For example, most folks are only ever going to use a string for values. It makes no sense for them to have to type-check the value in their `onChange` callback just because we've decided on looser types. I think it's preferable to keep them as-is, even if that means it's a little more cumbersome if you are passing numbers.

Closing for now, though we may consider other ways to make this more flexible in the future.

PR #2667, opened 2024-01-26
[Select] Fix select portal by using Presence component

phsantiago:
### Description
When Select with portal was used inside a component like Dialog, Popover, etc., it was being rendered before the portal of the parent element, causing it to be shown behind it because of stacking context HTML tag order priority

With this change, it will be possible to control the mount of Select.Portal, solving the stacking problem and also allowing to be controlled by js animation libraries

### DIFF
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -370,6 +370,11 @@ SelectIcon.displayName = ICON_NAME;
 
 const PORTAL_NAME = 'SelectPortal';
 
+type PortalContextValue = { forceMount?: true };
+const [PortalProvider, usePortalContext] = createSelectContext<PortalContextValue>(PORTAL_NAME, {
+  forceMount: undefined,
+});
+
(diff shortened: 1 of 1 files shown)


phsantiago:
Hello @benoitgrelard, @andy-hook any updates here?

phsantiago:
Hi @StephenHaney @vladmoroz @lucasmotta @hadihallak @chaance, can I get a review here, please? 🙂

mxdvl:
Just a note to say that having faced this issue, I was able to solve it by finding the closest parent `dialog` and set it as the Root’s container:

```tsx
import * as Select from '@radix-ui/react-select';

function Select(props: SelectProps) {
  const [container, setContainer] = useState<HTMLDialogElement>()
  return (
    <div ref={(ref) => setContainer(ref?.closest('dialog') ?? undefined)}>
      <Select.Root container={container}>
        {/* abridged for clarity */}
        <Select.Trigger />
        <Select.Portal />
	</Select.Root>
    </div>
  );
}
``` 

This may not be suitable for all and being properly supported would be better, but it may help others in the meantime or provide insights into the correct implementation.

PR #2668, opened 2024-01-26
Fix isFormControl to check if form attribute is defined (checkbox, radio-group, select, slider, switch)

### LINKED ISSUE #2530: Checkbox and other components do not render `<input>` element with `form` attribute
### Current Behavior

In https://github.com/radix-ui/primitives/discussions/874, @benoitgrelard says...

> That is correct, if it's outside of a form, there isn't any need for any native component anyway because bubbling only makes sense within forms.

That is not correct. See a simplified example of my case:

```jsx
<div>
  <AbstractedCheckbox label="Accept terms and conditions" name="consented" form="terms-form" />
  <form id="terms-form" onSubmit={(event) => console.log(Object.fromEntries(new FormData(event.target)))}/>
</div>

// result -> {}
// result when form wraps checkbox -> { consented: "on" }
```

Forms are not always the parent of form fields. In the above case, Radix doesn't realize it's part of a form, and so doesn't render an `<input>` element, and the checkbox state doesn't end up in the `FormData`.

### Expected behavior

An `<input>`/native element is rendered when the `form` attribute is present.

### Suggested solution

This line in `Checkbox.tsx` (and in Switch and elsewhere)...

```
```
should be changed to something like...

```
const isFormControl = button ? Boolean(button.closest('form')) || !!checkboxProps.form : true;
```

Or... just always render an `<input>`?  Not sure why it was decided to avoid it, every other library, [headless](https://react-spectrum.adobe.com/react-aria/Checkbox.html) or [not](https://mui.com/material-ui/react-checkbox/), seems to always render an `<input>` and just visually and  accessibly hide it as appropriate. 

### Additional context

This is a perfectly valid feature of the HTML spec...

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#form
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fae-form

... so I don't like having to justify why I would need it, but here is some context.

I tend to do it this because that way the form does not affect layout or styling. Yes I'm aware of `display: contents` but that does not prevent it from completely changing the CSS selectors required. As such, I tend to have an abstract form component that gets portaled to the body.

I'm sure there are many other cases why someone might need the input element there.

## Workaround

As a very unpleasant workaround, to my `<AbstractCheckbox>` component I've had to track and sync a local copy of the checked state and add an invisible (visually and accessibly) `<input>` so the form can pick it up, like this (irrelevant stuff removed):

```tsx
type Props = {
  label: ReactNode;
  value?: boolean;
  onChange?: (value: boolean) => void;
} & CheckboxProps;

const Checkbox = ({ label, tooltip, value, onChange, ...props }: Props) => {
  const id = useId();

  useEffect(() => {
    if (value !== undefined) setChecked(value);
  }, [value]);

  return (
    <div>
      <Root
        {...props}
        id={id}
        checked={checked}
        onCheckedChange={(checked) => {
          if (onChange) onChange(!!checked);
          setChecked(!!checked);
        }}
      >
        <Indicator>
          <FaCheck />
        </Indicator>
      </Root>

      <label htmlFor={id}>
        {label}
      </label>

      {/* https://github.com/radix-ui/primitives/issues/2530 */}
      <input
        type="checkbox"
        form={props.form}
        name={props.name}
        style={{ display: "none" }}
        checked={checked}
        readOnly
      />
    </div>
  );
};

export default Checkbox;
```

seanlennaerts:
### Description

In some situations it is useful to render form inputs outside of a form. Inputs have a [`form`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#form) attribute where you can specify which form the input is associated with. This PR fixes the `isFormControl` check to not only check if the input is contained in a form but to also check if the form attribute has been specified. It also passes the form attribute to the hidden input so that it will be submitted correctly.

Fixes #2530

### DIFF
--- a/packages/react/checkbox/src/Checkbox.tsx
+++ b/packages/react/checkbox/src/Checkbox.tsx
@@ -50,13 +50,14 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
       disabled,
       value = 'on',
       onCheckedChange,
+      form,
       ...checkboxProps
     } = props;
     const [button, setButton] = React.useState<HTMLButtonElement | null>(null);
     const composedRefs = useComposedRefs(forwardedRef, (node) => setButton(node));
     const hasConsumerStoppedPropagationRef = React.useRef(false);
     // We set this to true by default so that events bubble to forms without JS (SSR)
-    const isFormControl = button ? Boolean(button.closest('form')) || Boolean(button.form) : true;
+    const isFormControl = button ? Boolean(button.closest('form')) || form !== undefined : true;
     const [checked = false, setChecked] = useControllableState({
       prop: checkedProp,
       defaultProp: defaultChecked,
@@ -109,6 +110,7 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
             checked={checked}
             required={required}
             disabled={disabled}
+            form={form}
             // We transform because the input is absolutely positioned but we have
             // rendered it **after** the button. This pulls it back to sit on top
             // of the button.
--- a/packages/react/checkbox/src/Checkbox.tsx
+++ b/packages/react/checkbox/src/Checkbox.tsx
@@ -56,7 +56,7 @@ const Checkbox = React.forwardRef<CheckboxElement, CheckboxProps>(
     const composedRefs = useComposedRefs(forwardedRef, (node) => setButton(node));
     const hasConsumerStoppedPropagationRef = React.useRef(false);
     // We set this to true by default so that events bubble to forms without JS (SSR)
-    const isFormControl = button ? Boolean(button.closest('form')) : true;
+    const isFormControl = button ? Boolean(button.closest('form')) || Boolean(button.form) : true;
     const [checked = false, setChecked] = useControllableState({
       prop: checkedProp,
       defaultProp: defaultChecked,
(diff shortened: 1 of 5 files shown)


seanlennaerts:
Hey @benoitgrelard / @andy-hook any updates here? Let me know what I can do to get this PR over the line. If there are plans to merge this then I can fix the conflicts.

nicolastoulemont:
Facing the same issue in my remix apps.
Indeed, Remix heavily encourages uncontrolled components and form-based mutation/query, and using form IDs is a quite common way to solve the layout challenges that complex UI raises.

So far, this lack of support for the form prop and the way the isFormControl is resolved is preventing me from using these components in anything other than basic forms.

Happy to do whatever is needed to help this PR get merged & release

DarkPurple141:
Hi folks thought I'd chime in on this one as this would be a welcome change - let us know if more discussion desired to make this work or help wanted 🙏

evadecker:
This problem is currently biting me too. I'm using Remix and React Testing Library, and currently even though I've defined `name` props for my `Select` component, they aren't showing up in tests because no native component is being rendered.

chaance:
Tracking in https://github.com/radix-ui/primitives/issues/2530. I still need to do some manual testing for edge cases, but I'm planning to address this soon. Thanks for your patience!

seanlennaerts:
Looks like this got closed by #3161. Will close this as well to clean up after myself.

Issue #2606, opened 2023-12-20
Select, DropdownMenu - keyboard interactions are not working as expected in Shadow DOM 


alexandravassova:
### Current Behavior

When Select or DropdownMenu component is used in Shadow DOM and its content is open, arrow keys are not working correctly. After pressing the ArrowDown key, the focus is moved to the last item of the list and vice versa, if ArrowUp key is pressed, the focus is moved to first item. 

### Expected behavior

When focus is on an item, ArrowDown key should move focus to the next item, ArrowUp should move focus to the previous item also in Shadow DOM.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | select, dropdown-menu |  1.2.2, 2.0.5  |
| React            | n/a     |    18.2.0     |
| Browser          |    all browsers     |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

MartinDavi:
I see the same with toggle group as well.

abrehamgezahegn:
Also experiencing the same issue.

abrehamgezahegn:
I have opened a PR for this here https://github.com/radix-ui/primitives/pull/2712

QuocThuanTruong:
I got the same issue :((( Please fix it

michael-langley:
Any updates on this one? Thanks

yc42-Sebastian:
Hi, Is there any progress on this issue? 

Or any information about how to work around it?

jHolcnerTatum:
I have the same problem with Tabs component https://www.radix-ui.com/primitives/docs/components/tabs. I would appreciate any help with how to resolve this issue

RobinLbt:
Any updates?

kachkaev:
@RobinLbt there is https://github.com/radix-ui/primitives/pull/3674 with a likely fix, but the PR has been up for three months without review.

PR #2712, opened 2024-02-18
[Select] add new prop to SelectContent to account for shadow dom rendering

abrehamgezahegn:
### Description

Fix for https://github.com/radix-ui/primitives/issues/2606 .
Select arrow navigation depends on `document.activeElement`. But if the select is rendered in shadow dom, `document.activeElement` returns top level element in the shadow dom. This breaks the keyboard navigation. 

I edited the keyboard handler to check if select is in shadow dom then get the active element from it.
B


chaance:
Instead of a new prop, can't we just get this from the element stored in state?

```
const activeElement = (content?.shadowRoot ?? document).activeElement
```

PR #2715, opened 2024-02-20
feature: use onBlur instead of native change event on FormControl change (Form package)

elliason:
### Description
I think using React's onBlur event is a better choice here for a few reasons:
1) Imagine a hypothetical situation when user types incorrect input, he then clicks outside and input loose focus. Now the native onChange event is fired and the field is correctly marked as invalid. So the user goes back, fixes the input, field is marked as valid because of the `resetControlValidity();` call on line 402. So far so good, but the user types again and unfortunately makes the same mistake as before and ends up with the same incorrect value in the input. If he now clicks outside and the input loses focus, the native onChange event is not fired at all because the value is the same as before. So the user is left with an incorrectly filled input without even realising it.
2) It's less code, and you don't have to escape from the React event system to native events.

### DIFF
--- a/packages/react/form/src/Form.tsx
+++ b/packages/react/form/src/Form.tsx
@@ -352,17 +352,6 @@ const FormControl = React.forwardRef<FormControlElement, FormControlProps>(
       [customMatcherEntries, name, onFieldCustomErrorsChange, onFieldValidityChange]
     );
 
-    React.useEffect(() => {
-      const control = ref.current;
-      if (control) {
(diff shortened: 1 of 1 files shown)


jjenzz:
in terms of potential breaking changes here... some elements fire native change event when they are still focused (so not the equivalent of blur). for example, `<select>`s fire change when you choose an option in the dropdown. file inputs trigger change when you select a file.

perhaps the issue here is that radix couples _validity_ and _error visibility_, since the validity hasn't been resolved until `change` can fire and revalidate again.

PR #2716, opened 2024-02-21
feat: make `data-state` more explicit

### LINKED ISSUE #602: Figure out why `as={Component}` doesn't use `data-state` from `Component`
Stemmed from this discussion https://github.com/radix-ui/primitives/discussions/560#discussioncomment-588638

When I did `<Dialog.Trigger as={Tooltip.Trigger}>` the `data-state` attribute was maintained from `Dialog.Trigger` instead of being overridden with the `Tooltip.Trigger` one. I'm not sure if this is a misunderstanding of `as` prop on my part (they're a confusing beast) but thought it worth looking into as a DX improvement if possible.

chaejunlee:
### Description

- fixes https://github.com/radix-ui/primitives/issues/602
- ref https://github.com/radix-ui/primitives/discussions/560#discussioncomment-3115606

Reading the comments from the discussion, I made the `date-state` more explicit by adding the name of the radix component. This feature will help the namespace of `data-state` be more clean.

I think leaving the `data-state` as a legacy and adding `data-[component]-state` will be good for now. For next major update, we can drop the support for the `data-state`.

If the maintainers agree on this approach, I can update all the storybook examples to new `data-state` representation as well. Please, any feedback is appreciated!!!

### DIFF
--- a/packages/react/accordion/src/Accordion.tsx
+++ b/packages/react/accordion/src/Accordion.tsx
@@ -391,6 +391,7 @@ const AccordionItem = React.forwardRef<AccordionItemElement, AccordionItemProps>
         <CollapsiblePrimitive.Root
           data-orientation={accordionContext.orientation}
           data-state={getState(open)}
+          data-accordion-state={getState(open)}
           {...collapsibleScope}
           {...accordionItemProps}
           ref={forwardedRef}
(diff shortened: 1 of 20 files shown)


Obiwarn:
This looks like the best solution for the problem. Keeping the simple DX from data-state plus olving the problem.

chaance:
Something similar will be implemented ahead of the next major release, but it will look slightly different. Update forthcoming soon.

PR #2720, opened 2024-02-23
fix: [@radix-ui/react-tooltip] tooltip performance issue

tangye1234:
### Description

The main purpose of this pr is to resolve the performance issue of tooltip provider.

When there are lots of tooltips inside tooltip provider, everytime tooltip open state is changing, all of the tooltips will re-render regardless of its relation to the iteraction.

I found that provider only provides some context settings and also help to maintain some lifecycle states, while tooltip itself is enough to control its own open state. So I think it is unneccessary  for the provider's states to be broadcasted to all context consumers.


divmgl:
Can we get this in? @benoitgrelard @andy-hook 

@tangye1234 is there a way I can prevent this from happening at the caller? Seems to be internal inaccessible state, but figured I'd ask.

ZackMitkin:
@divmgl, you can try grouping components that have many tooltips in their own `<TooltipProvider>`.

Would be nice for this optimization to be merged. I have over 500 tooltips on a page which cause a noticeable freeze whenever one of them is hovered. For now, I just wrap components that use many tooltips in their own provider which does fix it

chaance:
Dealt with the conflicts and merged locally. Thanks!

PR #2741, opened 2024-02-29
[NavigationMenu] Remove unsuported `disableOutsidePointerEvents` prop

### LINKED ISSUE #2731: navigation-menu disableOutsidePointerEvents does not work
### Current Behavior

Adding `disableOutsidePointerEvents` cause open/close flickering

### Expected behavior

As according to the docs:

> When true, hover/focus/click interactions will be disabled on elements outside the bounds of the component. Users will need to click twice on outside elements to interact with them: Once to close the navigation menu, and again to activate the element.

The Content should stay open when the mouse pointer leaves the Content area

### Reproducible example

Open the [CodeSandbox linked](https://codesandbox.io/p/devbox/objective-sun-8vvktr) [from the docs](https://www.radix-ui.com/primitives/docs/components/navigation-menu) and add the `disableOutsidePointerEvents` prop on the `<NavigationMenu.Content>`  components

### Additional context

Maybe I misinterpreted the prop? Or should it be used in tandem with other props?

### Your environment

| Radix Package(s) | `@radix-ui/react-navigation-menu`| 1.1.3 |
| React            | n/a     |  18.2.0 |

benoitgrelard:
Fixes #2731


.yarn/versions/e97b6ff3.yml
```diff
@@ -0,0 +1,5 @@
+releases:
+  "@radix-ui/react-navigation-menu": patch
```
andy-hook, line 2: I guess this is technically breaking? not sure how much it matters?

benoitgrelard: Yeah, though I would think probably rare, and we can always call it out in the changelog.

andy-hook (approved):

Issue #2743, opened 2024-02-29
Change visibility to hidden instead of opacity to 0


nicholaschiang:
### Current Behavior

When using `hideWhenDetached`, the opacity is changed to `0` instead of the `visibility`. This lets users click items that they cannot see. This is a horrid UX.

You can see in the video below, that the popover is properly hidden when the trigger is occluded. However, the items can still be interacted with. When I hover to inspect, you'll see that I'm inspecting the popover items, and not the stuff that you see on-screen. On clicking, the popover item is clicked.

https://github.com/radix-ui/primitives/assets/20798889/b7c9bfa5-556e-49b8-b3d4-55a5391692c1

### Expected behavior

You should, as per the `@floating-ui/react-dom` [documentation](https://floating-ui.com/docs/hide#usage), set the `visibility` to `hidden` instead of the `opacity` to `0`.

https://github.com/radix-ui/primitives/blob/main/packages/react/popper/src/Popper.tsx#L271

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/2r30e)

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |         |         |
| React            | n/a     |         |
| Browser          |         |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

benoitgrelard:
You are 100% correct! That is an oversight from our part, thanks for raising an issue!

benoitgrelard:
And fixing it too! ♥️

nicholaschiang:
One thing I want to flag here: It seems like `hover:` interactions with elements covered by a popover with `visibility: hidden` are still not triggered until the popover is closed. Clicking where the popover is registers as a click outside and closes the popover, allowing interaction with the elements underneath. I suspect `pointer-events: none` is needed to allow interactions with elements underneath the hidden popover.

https://github.com/radix-ui/primitives/assets/20798889/86e7406c-4195-4732-8a5d-6666720c1b3e

benoitgrelard:
Yes that's a great point, if the intention is for it to behave as if it was already closed, then adding `pointer-events: none` might make sense too. Would you mind opening another PR with that change?

PR #2744, opened 2024-02-29
fix(Popper): use `visibility` to hide instead of `opacity`

nicholaschiang:
When using `hideWhenDetached`, the opacity is changed to 0 instead of the visibility. This lets users click items that they cannot see. This is generally a bad user experience.

This patch updates the `Popper` to set the `visibility` to `hidden` when `hideWhenDetached` is used. This ensures the user cannot interact with the hidden element.

This aligns with what is in the `@floating-ui/react-dom` documentation.

Ref: https://floating-ui.com/docs/hide#usage

Closes: https://github.com/radix-ui/primitives/issues/2743

### DIFF
--- a/packages/react/popper/src/Popper.tsx
+++ b/packages/react/popper/src/Popper.tsx
@@ -268,7 +268,7 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
               // we prevent animations so that users's animation don't kick in too early referring wrong sides
               animation: !isPositioned ? 'none' : undefined,
               // hide the content if using the hide middleware and should be hidden
-              opacity: middlewareData.hide?.referenceHidden ? 0 : undefined,
+              visibility: middlewareData.hide?.referenceHidden ? 'hidden' : 'visible',
             }}
           />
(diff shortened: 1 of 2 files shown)


packages/react/popper/src/Popper.tsx
```diff
@@ -268,7 +268,7 @@ const PopperContent = React.forwardRef<PopperContentElement, PopperContentProps>
               // we prevent animations so that users's animation don't kick in too early referring wrong sides
               animation: !isPositioned ? 'none' : undefined,
               // hide the content if using the hide middleware and should be hidden
-              opacity: middlewareData.hide?.referenceHidden ? 0 : undefined,
+              visibility: middlewareData.hide?.referenceHidden ? 'hidden' : 'visible',
```
benoitgrelard, line 271: nit: We shouldn't need to explicitly set it as `visible` when visible as it is the default value for `visibility`

```suggestion
              visibility: middlewareData.hide?.referenceHidden ? 'hidden' : undefined,
```

benoitgrelard (approved):

PR #2746, opened 2024-02-29
fix: (@radix-ui/react-tooltip) only allow content to be dismissed by trigger if visible

upatel32:
### Description

When the tooltip trigger is clicked before the content appears, it prevents the content from ever appearing. The change in PR allows the content to be dismissed only after when the content is visible. Subsequent clicks will dismiss the content and it will not appear again. This way the content will be visible to a user before they are able to dismiss it. 

Behavior before change:
https://github.com/radix-ui/primitives/assets/17918553/0ce84391-36c5-441e-872b-db839509966b

Behavior after change:
https://github.com/radix-ui/primitives/assets/17918553/5affe843-9b18-4362-a54d-fe6abc26e654

Additionally, I have added unit tests for the `Tooltip` component. The last text case demonstrates the fix this PR addresses.

### DIFF
--- a/packages/react/tooltip/src/Tooltip.test.tsx
+++ b/packages/react/tooltip/src/Tooltip.test.tsx
@@ -1,0 +1,121 @@
+import React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
+import * as Tooltip from '@radix-ui/react-tooltip';
+import userEvent from '@testing-library/user-event';
+
+describe('Tooltip', () => {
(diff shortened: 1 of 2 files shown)


benoitgrelard:
Hi @upatel32 this is actually by design and the intended behaviour. If a user is unfamiliar with your UI, they might hover over buttons to see if there are more information in tooltips. However if a user is familiar with it, and goes to click a button quickly, we don't think showing the tooltip even if their cursor remains over the button is good UX.

PR #2761, opened 2024-03-06
Use capture phase in `useEscapeKeydown`

### LINKED ISSUE #2653: [DismissableLayer] since Escape key is handled on document level it breaks if parent element also listens for Esc
Hello, 

Recently I've found out that Escape key is handled seprately for components like DropdownMenu and event handler for it is attached to document instead of a Content element.

Imagine a scenario where DropdownMenu is rendered inside a Parent component which also listens to escape and closes itself. How would I properly stop event propagation and close only DropdownMenu since doing something like `<DropdownMenu.Content onKeyDown={(e)=>e.stopPropagation()}` breaks it completely ?

Thanks in advance for any tips

_Originally posted by @MikeYermolayev in https://github.com/radix-ui/primitives/discussions/2644_

benoitgrelard:
Fixes #2653

### DIFF
--- a/.yarn/versions/75dcd823.yml
+++ b/.yarn/versions/75dcd823.yml
@@ -1,0 +1,18 @@
+releases:
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-context-menu": patch
+  "@radix-ui/react-dialog": patch
+  "@radix-ui/react-dismissable-layer": patch
+  "@radix-ui/react-dropdown-menu": patch
+  "@radix-ui/react-hover-card": patch
+  "@radix-ui/react-menu": patch
+  "@radix-ui/react-menubar": patch
+  "@radix-ui/react-navigation-menu": patch
+  "@radix-ui/react-popover": patch
+  "@radix-ui/react-select": patch
+  "@radix-ui/react-toast": patch
+  "@radix-ui/react-tooltip": patch
+  "@radix-ui/react-use-escape-keydown": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


andy-hook (approved):

diegohaz:
@jaknas I haven't tested this, but you could try stopping Radix from escaping when an Ariakit popup is open within it:

```jsx
<Dialog.Content
  ref={ref}
  onEscapeKeyDown={(event) => {
    if (ref.current?.contains("[data-open]")) {
      event.preventDefault();
    }
  }}
>
```

SebKranz:
We're also running into this issue with the HeadlessUI-Combobox. 

I'm not sure if handling escape in the capturing-phase is a good idea. From what I understand, it breaks compatibility with every component that doesn't use Radix's `<DismissableLayer>`.

By being a capturing listener at the root, there's basically no way for other components to intercept the event unless we manually add the `onEscapeKeyDown` to every component using `<DismissableLayer>` everywhere.

From the discussion, it seems like this change was made to support the inverse of this use-case, under the assumption that it would be a harmless change (https://github.com/radix-ui/primitives/issues/2653#issuecomment-1980661195). Maybe it would be worth reconsidering that decision.

diegohaz:
> This seems to work when `event.target` (or focus) is on the combobox input, but it fails to work when `event.target` is on combobox option in the popover. In that case, the popover doesn't close at all – as if Ariakit internally checks the `preventedDefault` property?

Yes, Ariakit checks `event.defaultPrevented`, and other libraries likely do the same. It seems that Radix could benefit from something like `hideOnEscape`, where you can return `true` or `false` for this specific behavior instead of relying on `event.preventDefault()`, which is meant to prevent all side effects on cancelable events.

PR #2771, opened 2024-03-11
[ScrolAreaViewport] Add `data-radix` attribute to the content element

nklhtv:
I need this data attribute to query the inner div element. Trying to avoid usage of `>div`-like queries.


nklhtv:
@vladmoroz @lucasmotta ping

Issue #1966, opened 2023-02-20
[Slider] Incorrect range mapping


latin-1:
## Bug report

Dragging the thumb or clicking on the track.

### Current Behavior

![Current Behavior](https://user-images.githubusercontent.com/44946699/220063805-838032ef-1354-4bce-8a7f-f42a2c9d753f.png)

### Expected behavior

![Expected Behavior](https://user-images.githubusercontent.com/44946699/220063843-a748ec71-9795-4e0a-88fb-839900c437fc.png)

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/p/sandbox/35y3ku)

### Suggested solution

Modify the mapping function.

$$
\frac12W_{thumb}+V_{percent}(W_{track}-W_{thumb})
$$

### Additional context

Other implementations:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range
- https://react-spectrum.adobe.com/react-spectrum/Slider.html

Note: There is also another type of slider:

![Another Type](https://user-images.githubusercontent.com/44946699/220066292-858fcedf-4ab3-492a-a3d6-ff4183b17b6e.png)

- https://mui.com/material-ui/react-slider/
- https://baseweb.design/components/slider/

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | @radix-ui/react-slider | 1.1.0 |
| React            | n/a     |         |
| Browser          |         |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

benoitgrelard:
Hey @latin-1, the current implementation is by design. We interpolate the size of the thumb over the whole range so it doesn't get out of bounds (outside of its natural bounding box).

What is "correct" is up to interpretation, for example, our implementation matches the native input range on MacOS:

<img width="253" alt="CleanShot 2023-02-20 at 15 57 22@2x" src="https://user-images.githubusercontent.com/1539897/220152921-e4a2e68a-16f1-42df-aa83-bb2f941a3dbf.png">

That said, I agree that this is perhaps a little opinionated, and something we have discussed about revisiting at some point in the future, by potentially providing a way to opt into different positioning strategies.

latin-1:
@benoitgrelard 
You misunderstood what I meant.

[...]

PR #2806, opened 2024-03-27
[Slider] Add 'thumbAlignment' option to align thumb center with slider edges

mark-gl:
### Description

Added a `thumbAlignment` prop that provides a choice between containing thumb sliders within the track or allowing them to centre on the edges of the track. This replicates the behaviour of the [Ark](https://ark-ui.com/docs/components/slider)/[Zag.js](https://zagjs.com/components/react/slider) slider prop with the same name.

 By default, Radix's slider thumb position interpolates to ensure it stays within the track. The centred alignment option might be more convenient in some use cases, since the mouse will always be at the centre of the thumb when dragging. [MUI](https://mui.com/material-ui/react-slider/) and [Base Web](https://baseweb.design/components/slider/) use this approach as well.

This PR also adjusts the default thumb behaviour to prevent the 'jump' that occurs when dragging the thumb from the slider edges. This is achieved by ignoring cursor movements for half of the current thumb's width (or height for vertical sliders) on either edge of the track.

The two options can be visualised by the images from #1966:

#### thumbAlignment: "contain" (default)

image

#### thumbAlignment: "overflow"
image


ADTC:
Is it possible for you to add a third option to do this?

Expected Behavior

I think the `contain` option should be fixed to do this, but if the library maintainers are saying the wrong version is "by design" then let us at least have the correct version "by option" 🙂 

Maybe call it `compensate`. (Referring to the option compensating for the thumb width. See my comment in the issue.)

denzellgh (commented):
Nice work!🚀

evadecker:
Thanks for this PR! I'm currently running into the same issue — track width and border radius creates a gap at the start of the track with the current implementation. This proposal would fix the problem.

CleanShot 2024-07-08 at 10 25 23@2x

To summarize, there are two main issues with the default implementation today:

1. At small ranges the width of the `Range` creates a gap with the `Thumb`.
2. Due to the way the thumb position is interpolated across the range of the slider, it creates [cursor drift](https://github.com/radix-ui/primitives/issues/1966#issuecomment-2028195429).

Having 3 different options for a prop where two look nearly-identical and only differ in behind-the-scenes implementation feels like a confusing experience for users.

Instead of having three `thumbAlignment` options, could the default implementation (which is buggy, not actually a desirable behavior) be replaced with what @ADTC is proposing as "compensate"?

Thus we'd have two options for `thumbAlignment`:

1. `contain` (visually remains almost identical to the current implementation, but behind the scenes uses the `compensate` positioning logic proposed in #1966 to fix the gap and cursor drift)
2. ~`center`~ `overflow` (allows the thumb to break out of the slider's bounding box)*

_The name `center` feels ambiguous in that it's unclear which direction (vertical or horizontal) it refers to—`overflow` feels like a better name to me, and pairs better with `contain`._

mark-gl:
The problem I'm having with implementing a 'compensate' mode is that the slider range would either have to be always partially filled or never full in order for it to line up with the thumb consistently (here the thumb is transparent):

**Slider range at minimum**
slider-empty

**Slider range at maximum**
slider-full

To try and get around this I've put together a concept for a 'compensate' mode that keeps the original interpolation so that the slider range is still empty at the minimum and full at the maximum, but the thumb now behaves in the way you’ve described (at the slider edge, dragging the first half of the thumb has no effect). Try it out and let me know what you think:

https://mmxt9p-9009.csb.app/?path=/story/components-slider--thumb-alignment

This implementation won’t solve the issue @evadecker is having with the border radius, but I think you can work around that with some CSS pseudoelement tricks to fill the area next to the thumb similar to what I’ve done [here](https://4hhskk-6006.csb.app/?path=/story/example-mediaslider--primary) (see [MediaSlider.module.css](https://github.com/mark-gl/soprano-ui/blob/main/src/media-slider/MediaSlider.module.css)).

mark-gl:
Given that I can't think of a scenario where anyone would prefer the 'contain' behaviour over the proposed 'compensate' behaviour I linked above, I've gone ahead with @evadecker's suggestion of replacing the default behaviour and renaming the alternative mode to 'overflow'.

This does increase the scope of the PR, but hopefully it will be a suitable resolution for #1966.

ADTC:
I apologize for not responding sooner but I had kinda moved on to other things. But yes, I don't think anyone in the right mind wants a "contain" option where the mouse behaviour is weird. The "compensate" option (as I have originally described it, with two half-thumb-width fake tracks added to "center") is the right way to go and it should actually be the default behavior IMO, or at least an option.

jbobrow:
Would love to see this one resolved. It is really egregious in our use-case.

https://github.com/user-attachments/assets/76dfda51-667f-4cd9-86b8-e1b127eead57

PR #2827, opened 2024-04-10
[Tooltip] Prevent state change on unmounted component

hanselabreu:
### Description

This pull request addresses an issue where the Tooltip attempts to set state on an unmounted component, leading to potential memory leaks and unexpected behavior.

### Issue:
Currently the Tooltip attempts to update its state even after it has been unmounted. This results in warnings in the console and can lead to unpredictable behavior, particularly in asynchronous operations.

### Proposed Solution:
To fix this issue, I have implemented a check to ensure that state is only updated if the component is still mounted. This is achieved by utilizing a boolean flag (`isMountedRef`) to track the mounted state of the component. When the component is unmounted, this flag is set to `false`, preventing any state updates.


lucaspolizeli:
Good catch. I'm facing the warnings you mentioned, and I probably think this is the reason.

fslc92:
+1

dajs3010:
+1

PR #2849, opened 2024-04-18
[Dialog] Align disableOutsidePointerEvents to trapFocus prop

pedrosousa13:
### Description

We were trying to use `forceMount` on our Dialog implementation of the component. Our idea is to have the modals be in the DOM but not visible until the Trigger is clicked.

Doing this, we noticed that the `trapFocus` worked correctly but the `disableOutsidePointerEvents` prop was adding `pointer-events: none` to a lot of elements so you couldn't interact with the page. 

By aligning this behavior with `trapFocus`, everything works as intended and the pointer event styles aren't added if the modal is closed.

### DIFF
--- a/.yarn/versions/ff2592d9.yml
+++ b/.yarn/versions/ff2592d9.yml
@@ -1,0 +1,2 @@
+releases:
+  "@radix-ui/react-dialog": patch
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -273,7 +273,7 @@ const DialogContentModal = React.forwardRef<DialogContentTypeElement, DialogCont
         // we make sure focus isn't trapped once `DialogContent` has been closed
         // (closed !== unmounted when animating out)
         trapFocus={context.open}
-        disableOutsidePointerEvents
+        disableOutsidePointerEvents={context.open}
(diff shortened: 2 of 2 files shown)


pedrosousa13:
Can someone take a look at this please?

jbgomez:
Bump for consideration 🙏

andreasss55:
bump for this  🙏 @chaance

pedrosousa13:
I will fix the conflicts with this. Would be lovely for someone to take a look from workos.

pedrosousa13:
This is such an easy win, could someone please review it?

andreasss55:
is Radix-ui dead already? they never fix this issue for years

schardev:
Bumping this up. We're also running into the same issue. We have a library that initializes maps in the background, and using `forceMount` on Portal ends up disabling the entire page, which breaks that flow.

What we need is focus to be locked only while the dialog is actually open. That way, we can load async data in the background and render the dialog only when the user explicitly opens it, without blocking the rest of the page.

michaelchangdk:
Please bump - also need a portal for a dialog to stay mounted in the background to receive data streams until it is opened.

chaance:
Closed with https://github.com/radix-ui/primitives/commit/cec4b8820dea7b96c2aa2abac6f9079aabf6c50b

PR #2865, opened 2024-04-24
fix: :bug: fill height of parent container scroll area

flixlix:
### Description

Ran into an issue, where the scroll area wouldn't take up the whole height of the parent container, only then did I notice there was an intermediate `div` being placed with display table. I get this is to match the height of the children, but this way I can't fill the height of the parent.

Here is an example where I need this:

In case the ScrollArea ViewPort does not overflow, I want a button to be at the bottom of the page, but in case the viewport gets bigger (scrolling gets enabled) I want the button to stay at the bottom of this child.

#### Before:

ScrollArea Content inherits the size from its children, so the button gets appended immediately after the cards (this would be the expected behaviour if we would've needed scrolling)

Screenshot 2024-04-24 at 19 30 30

#### After:

ScrollArea Content fills the whole page's height and thus the button can be at the very bottom of the page.

Screenshot 2024-04-24 at 19 16 49

> [!NOTE]
> I've tried around with lots of other css styles but ultimately only this ended up working, if anyone has other suggestions, feel free :)

### DIFF
--- a/packages/react/scroll-area/src/ScrollArea.tsx
+++ b/packages/react/scroll-area/src/ScrollArea.tsx
@@ -183,7 +183,10 @@ const ScrollAreaViewport = React.forwardRef<ScrollAreaViewportElement, ScrollAre
            * widths that change. We'll wait to see what use-cases consumers come up with there
            * before trying to resolve it.
            */}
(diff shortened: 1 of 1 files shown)


nednella:
Also ran into this issue. Your fix is exactly what I'm looking for too. Any way to nudge code owners?

PR #2922, opened 2024-05-28
Build tooling changes

vladmoroz:
Changing build tooling to esbuild and (temporarily) tsup in order to support `'use client'` directives up next. (Currently, Parcel strips them out, see https://github.com/parcel-bundler/parcel/issues/9050). Also sets us free from a bunch of other workarounds related to the current setup.


package.json
```diff
@@ -18,16 +18,11 @@
     "cypress:dev": "cypress open",
     "dev": "yarn storybook",
     "build-storybook": "storybook build",
-    "// build": "For context on tsconfig replacements in build scripts, see https://github.com/radix-ui/primitives/pull/361#discussion_r555004944",
-    "build": "yarn build:config && yarn build:packages && yarn build:cleanup",
-    "build:config": "mv tsconfig.json tsconfig.tmp.json && mv tsconfig.production.json tsconfig.json",
```
vladmoroz, line 23: This dance with configs isn't needed anymore

packages/react/toolbar/src/Toolbar.test.tsx
```diff
@@ -1,8 +1,6 @@
-import React from 'react';
```
vladmoroz, line 1: Had to please the pre-push hooks (how has this stuff worked before?)

Other changes in this file are due to my formatting on save

package.json
```diff
@@ -18,16 +18,11 @@
...
-    "build:packages": "parcel build 'packages/*/*/' --no-cache && yarn build:fix-type-defs",
-    "build:fix-type-defs": "node ./scripts/fix-type-defs-imports",
```
vladmoroz, line 25: No need to fix the bugged imports anymore

build.mjs
```diff
@@ -0,0 +1,48 @@
...
+  await tsup.build({
+    entry: [file],
+    format: ['cjs', 'esm'],
+    dts: { only: true },
+    outDir: dist,
+    silent: true,
+    external: [/@radix-ui\/.+/],
+  });
+  console.log(`Built ${path}/dist/index.d.ts`);
+}
```
vladmoroz, line 46: tsup gets the job done, but this is temporary—we'll probably just use tsc to compile the type definitions when stuff is updated to not need bundling in the first place:
https://github.com/radix-ui/primitives/pull/2922#discussion_r1617016819

.storybook/main.ts
```diff
@@ -17,6 +17,16 @@ const config: StorybookConfig = {
...
+  swc: () => ({
+    jsc: {
+      transform: {
+        react: {
+          // Do not require importing React into scope to use JSX
+          runtime: 'automatic',
+        },
+      },
+    },
+  }),
```
vladmoroz, line 29: Somehow:
- pre-push hooks didn't like unused React imports in some stories (how did folks push to their branches before?)
- Stories without React in scope won't compile

hadihallak (approved):

vladmoroz:
@hadihallak yep, I confirmed that "splitting" and React in the external don't do anything right now, this was a remnant from a different setup I was exploring when these were useful

PR #2923, opened 2024-05-28
RSC compatibility

vladmoroz:
- Adds `'use client'` in key places
- Fixes a hydration issue with Portal SSR (https://github.com/radix-ui/primitives/issues/1386)


packages/react/primitive/src/Primitive.tsx
```diff
@@ -46,9 +46,9 @@ const Primitive = NODES.reduce((primitive, node) => {
     const { asChild, ...primitiveProps } = props;
     const Comp: any = asChild ? Slot : node;
 
-    React.useEffect(() => {
+    if (typeof window !== 'undefined') {
       (window as any)[Symbol.for('radix-ui')] = true;
-    }, []);
+    }
```
vladmoroz, line 51: Note this change, this is to allow simple primitives to be server components (like `AccessibleIcon`)

No context for how it's used in the original PR, but seems benign to do the assignment in render:
https://github.com/radix-ui/primitives/pull/853

packages/react/portal/src/Portal.tsx
```diff
@@ -16,11 +17,14 @@ interface PortalProps extends PrimitiveDivProps {
   /**
    * An optional container where the portaled content should be appended.
    */
-  container?: HTMLElement | null;
+  container?: Element | null;
```
vladmoroz, line 20: Corresponds to the type expected by `React.createPortal`

packages/react/portal/src/Portal.tsx
```diff
@@ -16,11 +17,14 @@ interface PortalProps extends PrimitiveDivProps {
...
 }
 
 const Portal = React.forwardRef<PortalElement, PortalProps>((props, forwardedRef) => {
-  const { container = globalThis?.document?.body, ...portalProps } = props;
+  const { container: containerProp, ...portalProps } = props;
+  const [mounted, setMounted] = React.useState(false);
+  useLayoutEffect(() => setMounted(true), []);
+  const container = containerProp || (mounted && globalThis?.document?.body);
```
vladmoroz, line 27: When a container isn't provided, we must render twice to avoid hydration errors

hadihallak (approved):

packages/react/portal/src/Portal.tsx
dubzzz, line 27: From an end-user point-of-view don't we introduce an extra frame as React will not render the content at first execution but after a second render cycle?

dubzzz: Actually users can still pass the container to `globalThis?.document?.body` to drop that extra frame. Forget about my comment

PR #2931, opened 2024-06-03
Fix module resolution in edge cases

vladmoroz:
Fixes an issue when certain toolchains would be unable to resolve CJS packages produced by the new build tooling


build.mjs
```diff
@@ -12,6 +12,7 @@ async function build(path) {
     packages: 'external',
     bundle: true,
     sourcemap: true,
+    format: 'cjs',
```
vladmoroz, line 15: I didn't realise `'cjs'` was not the default option, instead it's `'iife'`:
https://esbuild.github.io/api/#format-iife

vladmoroz: Example output diff

image

Issue #1658, opened 2022-09-12
[Select] Selecting an option triggers a touch event on elements positioned behind


chrisarnold:
### Current Behavior

Using the Select primitive, when selecting an option, the touch event is interacting with any elements positioned underneath the list. This isn't happening on desktop on mouse click but can be replicated in Chrome when using the device simulation

https://user-images.githubusercontent.com/1044655/189703615-f4d8d640-81f4-4263-bc98-a157afce9e0a.mov

### Expected behavior

When an option is selected, the menu should close and no events should be triggered on other inputs

### Reproducible example

https://codesandbox.io/s/late-snowflake-39o7x7?file=/App.js

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | @radix-ui/react-select | 1.0.0 |
| React            | n/a     | 17 |
| Browser          | Chrome | 104 |

benoitgrelard:
Hi @chrisarnold,

Whilst I can reproduce it in emulation mode on Chrome, I cannot replicate it on an actual phone (iPhone iOS 16).
Are you seeing this issue natively anywhere?

chrisarnold:
Hey @benoitgrelard, hmmm that's strange... it's happening natively for me as well. On both Safari and Chrome on an iPhone running iOS 15.6.1

https://user-images.githubusercontent.com/1044655/191274645-f914c5c1-1476-495e-a85e-305c66ccc8eb.MOV

andy-hook:
I was also able to repro on 15.4.1

benoitgrelard:
Interesting, different behaviour in iOS 16 then?


https://user-images.githubusercontent.com/1539897/191276693-da30a99d-5417-4ca6-89ec-4ac3cf076041.mov

andy-hook:
> Interesting, different behaviour in iOS 16 then?

Seems so 🤔 the behaviour I'm seeing matches those posted by @chrisarnold .

benoitgrelard:
That's going to make testing trickier haha, at least we can reliably reproduce in emulation mode, although not sure what to make of that or the change between iOS15 and 16 😅

cpt-westphalen:
I have the same problem on desktop/Firefox right now. The click on any option starts a text selection on the elements behind. I've managed to work around this by using "user-select: none" on the CSS, but this shouldn't be the solution. :sweat_smile: 

[...]

PR #2403, opened 2023-09-22
fix(select): select click event passthrough underneath

xmliszt:
close #1658 

### Description

The `select` component has this behaviour when using mobile browsers, or in Chrome mobile simulator. The tapping of an item to select will close the select content wrapper and then trigger `onClick` event of the element underneath the select content wrapper.

I have experimented it in the storybook as shown in following screenshot:

As seen from screenshot, I put a button covering the entire view underneath the select component, in the second GIF recording, we can see that the button underneath is clicked when we try to tap to select an item (reflected by the increasing of the counter value).

### Hypothesis

In the original `handleSelect` event listener when an item is selected, the `context.onOpenChange(false)` was called before the `onClick` event was triggered. This is because the `handleSelect` event listener is listening to the `onPointerUp` event, which will be fired first before `onClick` event gets fired. Therefore, this results in our select content wrapper being **closed** before `onClick` event is fired, causing the `onClick` event to **pass through** to the component underneath.

### Solution

My proposed solution in this PR is to **make sure that `onClick` event is captured by the select item before the `onOpenChange(false)` is called which closes the select content wrapper**. This can be done by set up two states, one will get updated when `onClick` on the select item is captured, the other will be updated when `onPointerUp` on the select item is captured.

Then we just listen to the two states with a `useEffect` that will handle the effects caused by the changing two states. Once we get both states to be `true`, basically means our select item has both `onPointerUp` and `onClick` event received, then we can safely call `onOpenChange(false)` to close the select content wrapper.

### Result

The result is the `onClick` event no longer pass through to the component below, as shown in below GIF image:

PR #2939, opened 2024-06-06
[Select] Fix touch devices bugs

vladmoroz:
- Fix an issue when clicking a Select Item would click an element behind it on Android touch devices (https://github.com/radix-ui/primitives/issues/1658)
  - Reproducible in Chrome’s touch device simulator
- Fix an issue when Select Trigger would open inadvertently on touch-based scroll (https://github.com/radix-ui/primitives/issues/1912) 
  - Reproducible on iOS and in Chrome’s touch device simulator
- Similarly to the above, but in an open menu, prevent touch-based scrolling over Select Items from yielding a weird half-working focus interactions, and change Select Items to active on "click" for touch devices instead

The dropdown menu and potentially a couple other primitives would need similar fixes up next (can be done separately)

Touch, before:

https://github.com/radix-ui/primitives/assets/8441036/238b0fc6-3b05-45fd-9e54-71011dcba752

Touch, after:

https://github.com/radix-ui/primitives/assets/8441036/678a8fd2-1e78-432f-b55c-f153f315f67b

Behaviour with mouse devices is preserved:

https://github.com/radix-ui/primitives/assets/8441036/8042f00d-5d7a-4609-9a8b-b784f08122a4

Tested on real devices as well

### DIFF
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -220,6 +220,7 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
     const isDisabled = context.disabled || disabled;
     const composedRefs = useComposedRefs(forwardedRef, context.onTriggerChange);
     const getItems = useCollection(__scopeSelect);
+    const pointerTypeRef = React.useRef<React.PointerEvent['pointerType']>('touch');
 
     const [searchRef, handleTypeaheadSearch, resetTypeahead] = useTypeaheadSearch((search) => {
       const enabledItems = getItems().filter((item) => !item.disabled);
@@ -230,11 +231,18 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
       }
     });
 
-    const handleOpen = () => {
+    const handleOpen = (pointerEvent?: React.MouseEvent | React.PointerEvent) => {
       if (!isDisabled) {
         context.onOpenChange(true);
         // reset typeahead when we open
         resetTypeahead();
+      }
+
+      if (pointerEvent) {
+        context.triggerPointerDownPosRef.current = {
+          x: Math.round(pointerEvent.pageX),
+          y: Math.round(pointerEvent.pageY),
+        };
       }
     };
 
@@ -262,8 +270,15 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
             // because we are preventing default in `onPointerDown` so effectively
             // this only runs for a label "click"
             event.currentTarget.focus();
+
+            // Open on click when using a touch or pen device
+            if (pointerTypeRef.current !== 'mouse') {
+              handleOpen(event);
+            }
           })}
           onPointerDown={composeEventHandlers(triggerProps.onPointerDown, (event) => {
+            pointerTypeRef.current = event.pointerType;
+
             // prevent implicit pointer capture
             // https://www.w3.org/TR/pointerevents3/#implicit-pointer-capture
             const target = event.target as HTMLElement;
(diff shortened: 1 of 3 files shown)


packages/react/select/src/Select.tsx
```diff
@@ -262,8 +270,15 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
             // because we are preventing default in `onPointerDown` so effectively
             // this only runs for a label "click"
             event.currentTarget.focus();
+
+            // Open on click when using a touch or pen device
+            if (pointerTypeRef.current !== 'mouse') {
+              handleOpen(event);
+            }
```
vladmoroz, line 277: For posterity, about pen devices in this PR:
- Using an Apple Pencil on iOS yields `pointerType: 'pen'`, but behaves exactly like touch (e.g. opening on pointerdown when intending to scroll the page)
- Using a Wacom tablets on macOS also yields `pointerType: 'pen'`, but behaves exactly like mouse (which don't need pointerdown to scroll, so the current behaviour isn't a problem in this case).

Since we can disambiguate the two, I'm being a bit more defensive here and assuming `pen` is equivalent to `touch`

packages/react/select/src/Select.tsx
```diff
@@ -272,13 +287,10 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
...
-            if (event.button === 0 && event.ctrlKey === false) {
-              handleOpen();
-              context.triggerPointerDownPosRef.current = {
-                x: Math.round(event.pageX),
-                y: Math.round(event.pageY),
-              };
+            // but not when the control key is pressed (avoiding MacOS right click); also not for touch
+            // devices because that would open the menu on scroll. (pen devices behave as touch on iOS).
+            if (event.button === 0 && event.ctrlKey === false && event.pointerType === 'mouse') {
+              handleOpen(event);
```
vladmoroz, line 293: Retain open on pointerdown for mouse devices

packages/react/select/src/Select.tsx
```diff
@@ -1242,11 +1255,24 @@ const SelectItem = React.forwardRef<SelectItemElement, SelectItemProps>(
...
             onPointerMove={composeEventHandlers(itemProps.onPointerMove, (event) => {
+              // Remember pointer type when sliding over to this item from another one
+              pointerTypeRef.current = event.pointerType;
               if (disabled) {
                 contentContext.onItemLeave?.();
-              } else {
+              } else if (pointerTypeRef.current === 'mouse') {
                 // even though safari doesn't support this option, it's acceptable
                 // as it only means it might scroll a few pixels when using the pointer.
                 event.currentTarget.focus({ preventScroll: true });
```
vladmoroz, line 1278: Check for a mouse pointer is added because on iOS gliding through the items doesn't quite work the same way—focus gets stuck on the first item that received focus on pointer move and won't budge

packages/react/select/src/Select.tsx
```diff
@@ -1242,11 +1255,24 @@ const SelectItem = React.forwardRef<SelectItemElement, SelectItemProps>(
...
-            onPointerUp={composeEventHandlers(itemProps.onPointerUp, handleSelect)}
+            onClick={composeEventHandlers(itemProps.onClick, () => {
+              // Open on click when using a touch or pen device
+              if (pointerTypeRef.current !== 'mouse') handleSelect();
+            })}
+            onPointerUp={composeEventHandlers(itemProps.onPointerUp, () => {
+              // Using a mouse you should be able to do pointer down, move through
+              // the list, and release the pointer over the item to select it.
+              if (pointerTypeRef.current === 'mouse') handleSelect();
+            })}
```
vladmoroz, line 1266: In case of https://github.com/radix-ui/primitives/issues/1658, what appears to happen when you click an item on a touch device is that touch and pointer events are fired on the Select Item correctly, but then also a click event is emulated, but by that time pointerup has fired and the select is closed, so the simulated click lands on the item behind

Moving the handler that closes the select to onclick avoids that issue, but we keep pointerup for mouse devices to retain the ability to glide through the items without releasing the pointer (which has never worked on iOS anyway).

chaance (approved):

jasongerbes:
@chaance, it seems PR #2403 can be closed as #1658 was fixed in this PR.

Issue #1666, opened 2022-09-15
[ScrollArea] `asChild` on `Viewport` is not working correctly


benoitgrelard:
## Bug report

When using `asChild` on `Viewport`, the functionality is lost.
This is probably due to the fact that there's an extra node around `children`.
I think the solution is to use `Slottable` probably.

### Reproducible example

https://codesandbox.io/s/scrollarea-viewport-aschild-bug-y46g5q?file=/App.js

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |         |         |
| React            | n/a     |         |
| Browser          |         |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

exah:
I think we can remove the additional `div` when the `asChild` prop is used and rely on implementation in userland. It will allow adding `scroll-snap-type` wrappers inside, for example.

deltasierra96:
Any updates on this issue?

gwu:
I just ran into this issue because I'm using https://www.framer.com/motion/ to animate some elements inside the viewport of a ScrollArea, and in order to allow Framer Motion's automatic "layout" animations within scrollable containers, the scrollable container (in this case, the ScrollArea's Viewport) needs to become a <motion.div> instead of a plain <div>.

finkrer:
@benoitgrelard I have just encountered this issue. This is due to the fact that when the viewport, which has `overflow: scroll`, is merged with the children wrapper, which has `display: table`, the resulting element has both `display: table` and `overflow: scroll`.

The problem here is that `overflow` does not work with `display: table` elements, so scrolling doesn't work.

benoitgrelard:
Hey @finkrer we may do this, it's related to #926

PR #2945, opened 2024-06-10
[ScrollArea] Viewport fixes

### LINKED ISSUE #926: [ScrollArea] Content with ellipsis CSS bursts out of container.
Issue: https://codesandbox.io/s/frosty-cerf-poy3y?file=/src/styles.css
Example of native scrollable without this issue: https://codepen.io/jjenzz/pen/jOLMjvJ

I'm wondering if we still need the `display: table` stuff now so will investigate getting rid of that as a potential solution at some stage.

vladmoroz:
- Fix `asChild` prop not working as expected on the Viewport part
- Rework the inner content element styles on the Viewport part
  - Replace `display: table` with flex styles on the parent that allow the viewport to fill container height (based on the [Radix Themes implementation](https://github.com/radix-ui/themes/blob/main/packages/radix-ui-themes/src/components/scroll-area.css#L40))
  - Replace most related inline styles with 0,0,0 specificity styles for easier overrides via `[data-radix-scroll-area-content]`
  - Fix `text-overflow: ellipsis` not working when used with a vertical scrollbar

Closes #926 #1666

### DIFF
--- a/packages/react/scroll-area/src/ScrollArea.tsx
+++ b/packages/react/scroll-area/src/ScrollArea.tsx
@@ -148,10 +148,29 @@ const ScrollAreaViewport = React.forwardRef<ScrollAreaViewportElement, ScrollAre
     const composedRefs = useComposedRefs(forwardedRef, ref, context.onViewportChange);
     return (
       <>
-        {/* Hide scrollbars cross-browser and enable momentum scroll for touch devices */}
         <style
           dangerouslySetInnerHTML={{
-            __html: `[data-radix-scroll-area-viewport]{scrollbar-width:none;-ms-overflow-style:none;-webkit-overflow-scrolling:touch;}[data-radix-scroll-area-viewport]::-webkit-scrollbar{display:none}`,
+            __html: `
+[data-radix-scroll-area-viewport] {
+  scrollbar-width: none;
+  -ms-overflow-style: none;
+  -webkit-overflow-scrolling: touch;
+}
+[data-radix-scroll-area-viewport]::-webkit-scrollbar {
+  display: none;
+}
+:where([data-radix-scroll-area-viewport]) {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  height: 100%;
+}
+:where([data-radix-scroll-area-content]) {
+  flex-grow: 1;
+  min-width: 100%;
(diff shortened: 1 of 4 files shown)


packages/react/scroll-area/src/ScrollArea.tsx
```diff
@@ -1010,6 +1035,26 @@ function useResizeObserver(element: HTMLElement | null, onResize: () => void) {
...
+) {
+  const { asChild, children } = options;
+  if (!asChild) return typeof content === 'function' ? content(children) : content;
+
+  const firstChild = React.Children.only(children) as React.ReactElement;
+  return React.cloneElement(firstChild, {
+    children: typeof content === 'function' ? content(firstChild.props.children) : content,
+  });
+}
+
```
vladmoroz, line 1057: We worked this out in https://github.com/radix-ui/themes/pull/311

mboudreau:
Any progress on this?  I just hit this issue today...

fkysly:
Any progress on this too...

Crisis2010:
Same here, have tmp fix, but need proper

chaance (approved):

lazakrisz:
thank you so much folks 🙏

kognise:
FYI this caused a breaking regression for us. `width: min-content` on the inner container caused horizontal layouts inside vertical scroll areas to not be full width. For now we are doing:

```css
.rt-ScrollAreaViewport>* {
  width: 100%;
}
```

but it doesn't seem very correct? Not entirely sure.

TimChild:
I have the same issue as @kognise

chaance:
Reverted in 1.2.1 to fix the regression. Will revisit this at a later date.

Georgegriff:
That's unfortunate, the removal of display table in this pr fixed other things, such as being able to properly use flexbox to dynamically size the scroll height, was relying on it. hopefully there's a way of bringing it back

chaance:
@Georgegriff I know. I'm planning to come back to this but I think the splash radius here for breaking apps is a little too big to justify in a patch release. We'll either re-release it in a major or adjust it to avoid breakage.

Georgegriff:
Any luck on coming back to this at all? this component being flex box rather than table was so nice

Issue #2978, opened 2024-06-25
TooltipProviderProps no longer exported after update to @radix-ui/react-tooltip v1.1.1


ivy-lli:
## TooltipProviderProps no longer exported after update to @radix-ui/react-tooltip v1.1.1

Using shadcn/ui Tooltip component: https://ui.shadcn.com/docs/components/tooltip

### Current Behavior

![image](https://github.com/radix-ui/primitives/assets/42733123/71e6eef3-b2c9-49e6-b787-0aade3b679ad)

### Expected behavior

![image](https://github.com/radix-ui/primitives/assets/42733123/e1542d0d-02ce-4a28-af10-132290ec87fe)

### Reproducible example

Try to reexport the TooltipProvider

### Suggested solution

Not sure but maybe this change causes this problem? https://github.com/radix-ui/primitives/pull/2934/files#diff-da7e63370e718104438e4866d2cd43b3390689aa2c030321312c0be78602f544

### Additional context

I think this is also related to other issues like: https://github.com/radix-ui/primitives/issues/2973

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | @radix-ui/react-tooltip | v1.1.1 |
| React            | n/a | v18.3.1 |
| TypeScript | n/a | v5.4.5 |
| Node             | n/a     | v20.10.0        |
| npm/yarn         | npm        | v10.2.3        |
| Operating System | macOs        | 14.5        |

vladmoroz:
Released 1.1.2

PR #2968, opened 2024-06-21
[Checkbox] export type CheckedState

bdsqqq:
### Description

Adds back the export for the type `CheckedState`, it was exported in 1.0.* and used by some to make typesafe APIs that interact with Checkbox. But this export was removed in 1.1, now consumers need a type like: `type CheckedState = Parameters<NonNullable<CheckboxProps['onCheckedChange']>>[0]` to keep their APIs in sync with the Radix internal types.

PR #2980, opened 2024-06-25
[Tooltip] Export type tooltip provider props

rczobor:
### Description

Exports back previously exported type. It's also used when generating [shadcn-ui/tooltip](https://ui.shadcn.com/docs/components/tooltip).
Fixes [#2978](https://github.com/radix-ui/primitives/issues/2978)
Similar to [#2968](https://github.com/radix-ui/primitives/pull/2968), but for Tooltip.

### DIFF
--- a/packages/react/tooltip/src/Tooltip.tsx
+++ b/packages/react/tooltip/src/Tooltip.tsx
@@ -754,6 +754,7 @@ export {
   Arrow,
 };
 export type {
+  TooltipProviderProps,
   TooltipProps,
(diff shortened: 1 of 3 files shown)


packages/react/tooltip/src/Tooltip.tsx
vladmoroz, line 1: Could you export it from packages/react/tooltip/src/index.ts as well?

vladmoroz (approved):

Issue #2230, opened 2023-06-26
[Avatar] Allow `Avatar.Image` to be supplied a `component` to render the image


mririgoyen:
### Overview

The `Avatar.Image` primitive assumes I want to use an `img` without the ability to override that. I'd like to use the [Next.js Image component](https://nextjs.org/docs/pages/api-reference/components/image) instead for its optimizations. If I try to use the Next Image component instead of the primitive, the Fallback fails to work properly.

### Examples in other libraries

Other libraries, such as Material UI, allow you to pass an optional `component` as a prop.

```
<Image
  alt={altText}
  className="aspect-square h-full w-full"
  component={NextImage} // Provide a React component or a string value of an HTML element, default would be "img"
  src={src}
/>
```

### Who does this impact? Who is this for?

This would affect anyone utilizing the Next.js framework who wants to use it's image optimizations.

### Other Notes

If there are any other primitives that make use of images, allowing overrides on them would be greatly appreciated.

joaom00:
Did you try this?

```jsx
<Avatar.Image asChild>
  <NextImage />
</Avatar.Image>
```

zaunermax:
> Did you try this?
> 
> ```js
> <Avatar.Image asChild>
>   <NextImage />
> </Avatar.Image>
> ```

For me this does not throw type errors, it just outright does not work. Instead it renders the `AvatarFallback` instead.

raffizulvian:
I faced a similar issue. My workaround was to also supply `src` to `Avatar.Image` and `NextImage`.
```jsx
<Avatar.Image asChild src='image.svg'>
  <NextImage src='image.svg' alt='Some alt text' {...otherProps} />
</Avatar.Image>
```
It's a little bit redundant, but it could be wrapped inside a wrapper component.

I think it would be nice to make `src` not required if `asChild` is present or use a solution like other libraries as stated above by the author.

trm217:
Another issue with having to supply the src on Avatar.Image, is that asset imports, which enables further build-time optimisation, is not possible as Avatar.Image would not accept it for the src value.

akhan619:
> I faced a similar issue. My workaround was to also supply `src` to `Avatar.Image` and `NextImage`.
> 
> ```js
> <Avatar.Image asChild src='image.svg'>
>   <NextImage src='image.svg' alt='Some alt text' {...otherProps} />
> </Avatar.Image>
> ```
> 
[...]

PR #2999, opened 2024-07-04
Adding support for framework image component in Radix Avatar

rubenskj:
### Description

Hello there, how you guys doing?

As discussed in issue #2230, we would love to enhance image optimization by integrating the Next.js Image component. I've implemented a solution that allows us to utilize the Next.js Image component while maintaining the existing functionality.

In this update, I've removed the image load logic from useLayoutEffect and delegated it directly to the respective components. This means that Primitive.img will handle image loading as usual, while the Next.js Image component will load images with its optimizations.

I would greatly appreciate your feedback on this solution. Let's collaborate to refine this feature and ensure it meets our needs.

### DIFF
--- a/packages/react/avatar/src/Avatar.stories.tsx
+++ b/packages/react/avatar/src/Avatar.stories.tsx
@@ -50,10 +50,11 @@ export const Styled = () => (
       <Avatar.Image
         className={imageClass()}
         alt="John Smith"
-        src={otherSrc}
         onLoadingStatusChange={console.log}
-        component={FakeFrameworkImage}
-      />
+        asChild
+      >
+        <FakeFrameworkImage src={otherSrc} />
+      </Avatar.Image>
       <Avatar.Fallback className={fallbackClass()}>
         <AvatarIcon />
       </Avatar.Fallback>
@@ -61,13 +62,9 @@ export const Styled = () => (
 
     <h1>With image framework component & with fallback (but broken src)</h1>
     <Avatar.Root className={rootClass()}>
-      <Avatar.Image
-        className={imageClass()}
-        alt="John Smith"
-        src={srcBroken}
-        onLoadingStatusChange={console.log}
-        component={FakeFrameworkImage}
-      />
+      <Avatar.Image className={imageClass()} alt="John Smith" onLoadingStatusChange={console.log}>
+        <FakeFrameworkImage src={srcBroken} />
+      </Avatar.Image>
       <Avatar.Fallback className={fallbackClass()}>
         <AvatarIcon />
       </Avatar.Fallback>
@@ -100,13 +97,9 @@ export const Chromatic = () => (
 
     <h1>With image framework component</h1>
     <Avatar.Root className={rootClass()}>
-      <Avatar.Image
-        className={imageClass()}
-        alt="John Smith"
-        src={otherSrc}
-        onLoadingStatusChange={console.log}
-        component={FakeFrameworkImage}
-      />
+      <Avatar.Image className={imageClass()} alt="John Smith" onLoadingStatusChange={console.log}>
+        <FakeFrameworkImage src={otherSrc} />
+      </Avatar.Image>
       <Avatar.Fallback className={fallbackClass()}>
         <AvatarIcon />
       </Avatar.Fallback>
@@ -114,13 +107,9 @@ export const Chromatic = () => (
 
     <h1>With image framework component & with fallback (but broken src)</h1>
     <Avatar.Root className={rootClass()}>
-      <Avatar.Image
-        className={imageClass()}
-        alt="John Smith"
-        src={srcBroken}
-        onLoadingStatusChange={console.log}
-        component={FakeFrameworkImage}
-      />
+      <Avatar.Image className={imageClass()} alt="John Smith" onLoadingStatusChange={console.log}>
+        <FakeFrameworkImage src={srcBroken} />
+      </Avatar.Image>
       <Avatar.Fallback className={fallbackClass()}>
         <AvatarIcon />
       </Avatar.Fallback>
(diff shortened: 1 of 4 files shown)


chaance:
Thanks for the suggestion! I'm not quite sure the implementation here is what we're looking for. A few things:

First: we shouldn't need the `component` prop since we already have `asChild`. This is our preferred API for polymorphic components, and you can do this today:

```tsx
<Avatar.Image asChild src="https://whatever.com/img.jpg" alt="cute kitten">
  <NextImage width={500} height={300} />
</Avatar.Image>
```

While you might get type errors since Next requires `src` and `alt`, you should be able to suppress them since we merge props internally when we clone the child element.

> while the Next.js Image component will load images with its optimizations

Not everyone uses Next JS, and we shouldn't assume as much. I think I'd rather have a boolean prop that disables our internal optimizations rather than removing them completely.

dragospaulpop:
New to Next, but I found that supplying `src` to Avatar.Image also downloads the original file, which defeats the purpose of the optimizations provided by NextImage. I got it working by supplying AvatarImage with a small 1x1 transparent image and NextImage with the actual image url:
```javascript
<AvatarImage src="/images/blank.png" asChild>
  <Image alt={user.username} width={48} height={48}  src={user.image}  />
</AvatarImage>
```

Udhayarajan:
I've improved this code and raised PR in @RubensKj fork.
Let's wait for a week, if he isn't active I will rebase the changes from this repository and raise PR here.

Changes: 
https://github.com/RubensKj/primitives/pull/1

RubensKj:
@Udhayarajan thanks for the PR! I've merged and also fixed the merge problems with main branch.

Thanks for contributing together 🚀 . Now let's wait for the review ❤️

packages/react/avatar/src/Avatar.tsx
```diff
@@ -54,25 +54,55 @@ const IMAGE_NAME = 'AvatarImage';
...
     });
 
     useLayoutEffect(() => {
-      if (imageLoadingStatus !== 'idle') {
-        handleLoadingStatusChange(imageLoadingStatus);
+      if (!src) {
+        handleLoadingStatusChange('error');
+        return;
       }
-    }, [imageLoadingStatus, handleLoadingStatusChange]);
```
Udhayarajan, line 75: These changes may not required. If this happens as an effect of rebasing please revert this.

packages/react/avatar/src/Avatar.stories.tsx
```diff
@@ -33,6 +44,31 @@ export const Styled = () => (
...
+        <FakeFrameworkImage src={otherSrc} />
+      </Avatar.Image>
+      <Avatar.Fallback className={fallbackClass()}>
+        <AvatarIcon />
+      </Avatar.Fallback>
+    </Avatar.Root>
+
+    <h1>With image framework component & with fallback (but broken src)</h1>
+    <Avatar.Root className={rootClass()}>
+      <Avatar.Image className={imageClass()} alt="John Smith" onLoadingStatusChange={console.log}>
```
Udhayarajan, line 65:
```suggestion
      <Avatar.Image className={imageClass()} alt="John Smith" onLoadingStatusChange={console.log} asChild>
```

if `asChild` is not passed it will still render default `Primitive.img`

packages/react/avatar/src/Avatar.stories.tsx
```diff
@@ -58,6 +94,26 @@ export const Chromatic = () => (
         <AvatarIcon />
       </Avatar.Fallback>
     </Avatar.Root>
+
+    <h1>With image framework component</h1>
+    <Avatar.Root className={rootClass()}>
+      <Avatar.Image className={imageClass()} alt="John Smith" onLoadingStatusChange={console.log}>
```
Udhayarajan, line 100:
```suggestion
      <Avatar.Image className={imageClass()} alt="John Smith" onLoadingStatusChange={console.log} asChild>
```

packages/react/avatar/src/Avatar.stories.tsx
```diff
@@ -58,6 +94,26 @@ export const Chromatic = () => (
...
+        <FakeFrameworkImage src={otherSrc} />
+      </Avatar.Image>
+      <Avatar.Fallback className={fallbackClass()}>
+        <AvatarIcon />
+      </Avatar.Fallback>
+    </Avatar.Root>
+
+    <h1>With image framework component & with fallback (but broken src)</h1>
+    <Avatar.Root className={rootClass()}>
+      <Avatar.Image className={imageClass()} alt="John Smith" onLoadingStatusChange={console.log}>
```
Udhayarajan, line 110:
```suggestion
      <Avatar.Image className={imageClass()} alt="John Smith" onLoadingStatusChange={console.log} asChild>
```

packages/react/avatar/src/Avatar.tsx
```diff
@@ -54,25 +54,55 @@ const IMAGE_NAME = 'AvatarImage';
...
+      const childProps = child.props;
+
+      // Clone the child to add onLoad and onError event listeners
+      return React.cloneElement(child, {
+        ...restProps,
+        ...child.props,
+        onError: (event: React.SyntheticEvent<HTMLImageElement, Event>) => {
+          console.log('error');
+          handleLoadingStatusChange('error');
+          if (childProps.onError) childProps.onError(event);
```
Udhayarajan, line 97: question: @chaance do we also need to trigger `onError` callback from `props`?

RubensKj: For me make sense, considering that the child component might have some implementation of it 🤔.

But Chance can give us a better overview if it's necessary or not

Udhayarajan: IMO, to mimic the natural CSS flow, we should check and send the event to both the child's and parent's callback, so that if any using the parent callback the code won't break thus avoiding chaos.

chaance:
Closing this as stale, as the implementation still makes assumptions about the child component that we don't want to make. We will add a prop to disable optimizations and find another path towards supporting non-standard image components.

Issue #3005, opened 2024-07-07
Define translateX on `NavigationBar.Indicator` as a custom CSS property


terrymun:
### Overview

It would be really helpful if the `translateX` value for `<NavigationBar.Indicator>` is set as a CSS custom property as with calculations/measurements made available on other primitive subcomponents.

An example name of the CSS custom property can be: `--radix-navigation-menu-indicator-translate-x`.

### Examples within the same `<NavigationMenu>` component

- The `<NavigationMenu.Viewport>` component has its width and height made available via `--radix-navigation-menu-viewport-width` and `--radix-navigation-menu-viewport-height`, which allows us to use it for advanced calculations if needed

### Who does this impact? Who is this for?

This is mostly for advanced users who wants to perform additional CSS transforms on the indicator (i.e. entry animation involving `translateY`, for example).

Right now the `transform` is inlined for the element and it will overwrite whatever CSS transform defined on the element. By making it a variable, it will be possible to do this for the following template:

```tsx
<NavigationMenu.Indicator className={styles.indicator} forceMount />
```

```scss
.indicator {
  background-color: blue;
  height: 2px;
  /** NOTE: This is needed to tween when user hovers from one item to another without triggering hide/show of the indicator */
  transition: transform 200ms ease-in-out;

  &[data=state='visible'] {
    animation: enterFromBottom 200ms ease-in-out;
  }

  &[data-state='hidden'] {
    animation: exitToBottom: 200ms ease-in-out;
  }
}

@keyframes enterFromBottom {
  from {
    opacity: 0;
    transform: translateX(var(--radix-navigation-menu-indicator-translate-x, 0)) translateY(-2rem);
  }

  to {
    opacity: 1;
    transform: translateX(var(--radix-navigation-menu-indicator-translate-x, 0)) translateY(0);
  }
}

@keyframes exitToBottom {
  from {
    opacity: 1;
    transform: translateX(var(--radix-navigation-menu-indicator-translate-x, 0)) translateY(0);
  }

  to {
    opacity: 0;
    transform: translateX(var(--radix-navigation-menu-indicator-translate-x, 0)) translateY(-2rem);
  }
}
```

chaance:
Closed in https://github.com/radix-ui/primitives/commit/84ebe2cd4ddc76d1d7434ec6c09bb7b43aba9d25

PR #3024, opened 2024-07-19
[NavigationMenu] Add NavigationMenu.Indicator translate values as css vars

kylemcd:
### Description
In issue #3005 it was requested that the translate values defined in `NavigationMenu.Indicator` be added as css variables so that this value can be used to make additional calculations.

This PR takes the `position.offset` defined in `NavigationMenu.Indicator` and assigns it to a css variable based on the if the menu is horizontal or vertical.


leecobaby:
I need this pr

PR #3051, opened 2024-08-06
refactor(useControllableState): add usePrevious instead useRef

sukvvon:
### Test

image
* test is successful! 👍

### Description

* we provide `usePrevious` custom hook in `radix-ui/primitives` so i add our `usePrevious` hook instead react's useRef in `useControllableState`. 
* i think using our product(`usePrevious`) is more pretty and cool! 🤗.


chaance:
Thanks for the PR! In this case I think we're fine using the ref approach, as it requires one less effect to run.

PR #3062, opened 2024-08-10
refactor(useControllableState): merge useUncontolledState in useControllableState

sukvvon:
### Test

image

* test is successful!

### Description

* handleChage always run both(useControllableState, useUncontrolledState), we can merge.
* both always use nextValue (which is executed function case or not), so we dont need to seperate it. in useUncontrolledState, run useEffect when useUncontrolledState's value changed. but if merge, it does not need to use useEffect.
* in both, compare prevValue and nextValue if they are same or not, this case also we can merge.

### DIFF
--- a/.yarn/versions/261b859b.yml
+++ b/.yarn/versions/261b859b.yml
@@ -1,0 +1,25 @@
+releases:
+  "@radix-ui/react-use-controllable-state": patch
+
+declined:
+  - primitives
+  - "@radix-ui/react-accordion"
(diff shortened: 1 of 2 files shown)


chaance:
Unless there's a bug with our current implementation, I'm going to close this for now. We need to prioritize bugs and reduce noise in the repo in the mean time. If there is a bug that this fixes, please link to an issue that reproduces the bug and I will reconsider.

PR #3072, opened 2024-08-15
fix: Stop ignoring RadioGroup.Item onKeyDown prop

david-arteaga:
### Description

`RadioGroup.Item` ignores the `onKeyDown` prop.
This change fixes that.


chaance (approved):

chaance:
@david-arteaga Do you mind running the version command so that this change is captured in the next release? Check the details here: https://github.com/radix-ui/primitives/blob/main/.github/CONTRIBUTING.md#development

PR #3080, opened 2024-08-22
refactor(context): add React.FC type in Provider

sukvvon:
### Description

* add React.FC type in Provider
  * just Function has not displayName type but React.FC has displayName type
  * make the format the same with CollectionProvider


chaance (approved):

sukvvon:
@chaance in UI Tests, is there any problem?

PR #3093, opened 2024-09-01
[Select] Missing className in the SelectValue component

andreperegrina:
### Description

The `SelectValue` component was missing the `className` property, which caused issues when attempting to customize the element's appearance.


chaance:
This is by design. See the comment a few lines above
```ts
// We ignore `className` and `style` as this part shouldn't be styled.
```

andreperegrina:
@chaance  Could you elaborate on why you wouldn't want to modify this element?
I understand it may have been discussed before, and it would be helpful for me to understand as I might be using it incorrectly.

Issue #2436, opened 2023-10-09
The element in dialog lost focus state.


taoliujun:
### Current Behavior

There is an input A element with a custom clear icon in the dialog component, and in its focusout event, the focus behavior of another input B element in the dialog component is triggered. It is found that the trigger failed and the cursor does not enter the B element.

### Expected behavior

The cursor enters the B element.

### Suggested solution

I looked at the context source code of the dialog component and read the following in the focus-scope file:

```typescript
// When the focused element gets removed from the DOM, browsers move focus
// back to the document.body. In this case, we move focus to the container
// to keep focus trapped correctly.
function handleMutations(mutations: MutationRecord[]) {
  const focusedElement = document.activeElement as HTMLElement | null;
  if (focusedElement !== document.body) return;
  for (const mutation of mutations) {
    if (mutation.removedNodes.length > 0) focus(container);
  }
}
```

With the MutationObserver feature, if an element is removed (when the input A loses focus, the clear icon is removed), the focus will be placed on the container (react-dialog root) node, causing the focus of the B element to be lost immediately.

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |         |         |
| React            | n/a     |         |
| Browser          |         |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

enkot:
RadixVue also has this issue, it uses the same logic with `MutationObserver` - https://github.com/radix-vue/radix-vue/issues/518

benoitgrelard:
Hi @taoliujun could you provide a reproduction in a sandbox?

BenLorantfy:
@benoitgrelard I fixed this issue and created a PR: https://github.com/radix-ui/primitives/pull/3115

The PR contains a reproduction as a [storybook story](https://github.com/radix-ui/primitives/pull/3115/files#diff-4ae1552c8355d94b3a69609fc03170678c835486f59edd0f5931882b9c9707f3R480-R498) / [cypress test](https://github.com/radix-ui/primitives/pull/3115/files#diff-cef61c0d2141f9d5f77a7386e60456769dfdc8d23dba85afccdb47a68ca4901eR125-R131)

Also please note the vue fork fixed this already here with a similar fix: https://github.com/radix-vue/radix-vue/pull/519

flex-haegul:
“How is this issue progressing?”

[...]

PR #3115, opened 2024-09-11
[FocusScope] Fix bug where focus is moved to container when it shouldn't be

BenLorantfy:
### Description

Fixes bug described in https://github.com/radix-ui/primitives/issues/2436 where focus was moved to the container unnecessarily.  The bug occurs when an element is removed via an `onBlur` callback.

## Before
https://github.com/user-attachments/assets/91f653b8-5b44-4f94-b24b-69ae68fea157

## After
https://github.com/user-attachments/assets/57105819-95f3-40f3-84ef-4ee5a4b45d10

## More technical details
Let's say we have some error text "This field is required" that is removed when the field is filled in and blurred.  The current code checks if _any_ node is removed in the mutation observer (not just if the focused element is removed), and if so focuses the container.  The text getting removed triggers this condition: https://github.com/radix-ui/primitives/blob/660060a765634e9cc7bf4513f41e8dabc9824d74/packages/react/focus-scope/src/FocusScope.tsx#L115

Additionally, the early return is not triggered here: https://github.com/radix-ui/primitives/blob/660060a765634e9cc7bf4513f41e8dabc9824d74/packages/react/focus-scope/src/FocusScope.tsx#L113

Because during any blur, the [activeElement is document.body](https://codepen.io/mrburger/pen/jOjJmBZ?editors=1010).  This means checking `document.activeElement === document.body` is not a robust way of checking if the focus was lost.

image

The fix was to prevent focusing the container if the last focused element is still in the container:

https://github.com/radix-ui/primitives/blob/a3c1e51aa9d5d6552e13f2aa5fdf79a78bebbcad/packages/react/focus-scope/src/FocusScope.tsx#L124-L131

I left the previous checks in place because I was a bit scared to remove them, but it may not be required.  The vue fork fixed this by replacing the existing checks: https://github.com/radix-vue/radix-vue/pull/519/files


MantasMikal:
Would be great to get this out! This bug causes a lot of issues on the project I'm working on. 
Thank you for fixing this :)

theskillwithin:
+1

BenLorantfy:
Hey @chaance , I've fixed the conflicts and pushed a changeset.  Please lmk if I should do anything else

joeriddles:
Would be great to see this merged. Running into issues with react-hook-form inside a radix dialog that I think this would fix.

cgatian:
@joeriddles Radix is dead.

PR #3131, opened 2024-09-20
feat(injected-css): optionally export css for consumer to import

prests:
### Description

This is a POC for removing all of the inline styles for CSP safe alternatives. This is tied to [this discussion](https://github.com/radix-ui/primitives/discussions/3130).

This converts inline styles into css class equivalents, and exports them as part of the npm package. The idea is that consuming apps would import these css files as part of their build. I left this as a progressive enhancement where by providing `skipStyleInjection` it's assumed you're importing these CSS files, but consumers who are not can still use the injected inline styles.

**Looking for input on how to have one source of truth between the css file and the inline styles. I don't think there's an easy way to do it but was hoping to not have to change two places at once.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,7 @@ cypress/screenshots
 
 # So devs can maintain their own todo lists in the project
 TODO.md
+
+# For those using yalc to test publishes locally
+.yalc
+yalc.lock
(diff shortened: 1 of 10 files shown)


chaance:
I appreciate the contribution and consideration of the tradeoffs here! I'm going to close this PR for now because a) there's a huge backlog I'm trying to clear out, and b) I'm not sure I fully understand the problem and solution at this point. Glad to continue the discussion [here](https://github.com/radix-ui/primitives/discussions/3130) to see how we can move forward.

Issue #2975, opened 2024-06-24
[Select] Allow setting `form` prop


cmd-johnson:
## Feature request

Native `<select>`s (and all other input elements) have a `form` prop that lets us override which form the input belongs to. As far as I can tell, the `@radix-ui/react-select` primitives don't support this property right now.

### Overview

When using native `<select>` elements, you can render a select element outside of a `<form>` and later set the element's form using the form's id:
```html
<select name="selectOutsideOfForm" form="exampleForm">
  <option value="a" selected>A</option>
  <option value="b">B</option>
</select>
<form id="exampleForm">
  <button>Submit</button>
</form>
```
Clicking the submit button includes the `selectOutsideOfForm` value in the resulting submission.

Similarly you can exclude a `<select>` from a form by setting it's `form` attribute to an invalid form id:
```html
<form id="exampleForm">
  <select name="selectInsideOfForm" form="">
    <option value="a" selected>A</option>
    <option value="b">B</option>
  </select>
  <button>Submit</button>
</form>
```
Clicking the submit button now doesn't include the `selectInsideOfForm` element in the form.

### Who does this impact? Who is this for?

Supporting the `form` attribute would enable use-cases such as
- rendering a native `<select>` element using SSR and then replace it with a radix `<Select>` element after hydration, allowing users to use the app before it is hydrated (or even with JS disabled)
- allowing more ways to structure forms by rendering `<Select>` elements outside of a `<form>` element

### Additional context

[MDN reference for the `<select>` `form` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#form)

From looking through radix's code, I'd expect the implementation to work somewhat along these lines:
- adding a new `form` prop to the `Select.Root` component
- adding a new `form` property to the `SelectContextValue` type
- passing the `form` prop value to the `<SelectProvider>`
- set the `form` attribute of the `Primitive.button` element inside the `Select.Trigger` to the `form` value from `useSelectContext`
- set the `form` attribute of the `BubbleSelect` component to the `form` value from `useSelectContext`
- change the logic of determining `isFormControl` inside of the `Select.Root` component to check for the trigger's `form` (which will the browser sets to the actual form element reference at runtime) instead of using `trigger.closest("form")`.
[...]

PR #3134, opened 2024-09-23
[Select] add form prop (#2975)

cmd-johnson:
### Description

This PR adds a new `form` prop to the Select.Root primitive which serves the same purpose as the `form` attribute on a normal HTML `<select>` element.
This allows us to use radix Select elements outside of a form's markup and still have its value included in that form's submissions.

Apart from adding the `form` prop to the `Select.Root` primitive, this PR also changes how `isFormControl` is determined. Instead of using `element.closest('form')`, it now uses `element.form`. This shouldn't have an impact to controls that are located inside a `<form>` already, because the `form` attribute is automatically set to the parent form element. For elements outside a form, this attribute is unset, so it should be functionally the same as the previous behavior, with the added benefit of supporting overriding the associated form using the `form` attribute passed to the element.

There's more details in my original feature request #2975.

### DIFF
--- a/.yarn/versions/79af2124.yml
+++ b/.yarn/versions/79af2124.yml
@@ -1,0 +1,6 @@
+releases:
+  "@radix-ui/react-select": minor
+
+declined:
+  - primitives
+  - ssr-testing
--- a/cypress/integration/Select.spec.ts
+++ b/cypress/integration/Select.spec.ts
@@ -30,4 +30,18 @@ describe('Select', () => {
       cy.findByText('…').should('exist');
     });
   });
+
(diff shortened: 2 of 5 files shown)


chaance:
See https://github.com/radix-ui/primitives/pull/2484#issuecomment-2379643220. We'll need to revert to the `closet` check.

chaance:
I'm actually going to close this for now. Per https://github.com/radix-ui/primitives/issues/2530#issuecomment-2380105452 I'd like to introduce a more wholistic solution to cover all affected primitives, and I still need to do some manual testing for edge cases.

Tracking in https://github.com/radix-ui/primitives/issues/2530

PR #3154, opened 2024-09-28
[Avatar] Adding support for Framework image

Udhayarajan:
### Description
Hi Team,

I'm proposing a new suggestion to address issue #2230 on top of existing PR #2999 

I've implemented this solution by removing the prefetch image which helps to determine the image loading error.

Since `Primitive.img` already handles `OnLoad` and `OnError` events, it makes sense to leverage these callbacks for handling image load failures in a single fetch.

## Changes Introduced
- On the first render, `Primitive.img` is displayed, and its `OnLoad` and `OnError` callbacks are subscribed through an intermediate handler:
    - **Loading:** The image’s `style.display` is set to `none`.
    - **Success:** The `display` property is updated to the user-specified value.
    - **Failure:** The component re-renders, returning `null` to display the fallback.
- Removed the `useImageLoadingStatus` custom hook, simplifying the image loading process: https://github.com/radix-ui/primitives/blob/6469d41dcc6e84fe41d70a2703924338e7562dd1/packages/react/avatar/src/Avatar.tsx#L119

NOTE: I don't find any other use-case of `useImageLoadingStatus` hook, other than tracking image loading status

## What This Solves
- Enables better integration with framework-specific optimizations (e.g., Next.js’s Image component).

## Visual Comparison
**Before:**

https://github.com/user-attachments/assets/0c0bada4-cbfe-4542-bd8d-a559c60e7ea3

**After:**

https://github.com/user-attachments/assets/f0f8022d-9a1d-4177-96b1-9a836489f4f2

**EDIT:**
*Upon toggling `Disable Cache` off It makes only one request*

### DIFF
--- a/.yarn/versions/5ffaf749.yml
+++ b/.yarn/versions/5ffaf749.yml
@@ -1,0 +1,3 @@
+releases:
+  "@radix-ui/react-avatar": patch
+  primitives: patch
--- a/packages/react/avatar/src/Avatar.stories.tsx
+++ b/packages/react/avatar/src/Avatar.stories.tsx
@@ -5,6 +5,7 @@ export default { title: 'Components/Avatar' };
 
 const src = 'https://picsum.photos/id/1005/400/400';
 const srcBroken = 'https://broken.link.com/broken-pic.jpg';
+const otherSrc = 'https://picsum.photos/id/1006/400/400';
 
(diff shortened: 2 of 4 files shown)


chaance:
Planning to handle this a little differently https://github.com/radix-ui/primitives/pull/3943

Issue #3156, opened 2024-09-30
Unable to reset value in controlled mode in `useControllableState` 


flex-hyuntae:
## Bug report
1. Use controlled state in useControllableState hook
```
const [value, setValue] = useControllableState({
  prop: valueFromProps,
  defaultProp: valueFromDefaultProps,
  onChange: onChangeValueFromProps
})
```
2. but `valueFromProps` is undefindable like `valueFromProps?: string`
3. First `valueFromProps` is undefined
4. Second i will change value use `setValue` and call `onChangeValueFromProps` 
5. After 4, `valueFromProps` changed some value like `abc`
6. After 5, i want clear value. so i changed `valueFromProps` to `abc` to `undefined
7. but `value` from `useControllableState` is `abc`

I debug this issue. The cause is step 3.
when step 3, `useControllableState` work like uncontrolled state.
because in `useControllableState` check controlled or not like `prop === undefined`.
but this is bug because i want use controlled `useControllableState`

### Current Behavior
- `useControllableState` prop changed undefined to value. And init to undefined not work

### Expected behavior
- `useControllableState` prop changed undefined to value. And init to undefined work well

### Reproducible example

[CodeSandbox](https://codesandbox.io/p/sandbox/priceless-danilo-g3vlq4?workspaceId=78021ced-56d6-42ef-833f-3082d0c64b8a)

### Suggested solution
- `useControllableState`  check controlled or not like `prop in params`

### Additional context
- Maybe this issue is related
  - https://github.com/radix-ui/primitives/issues/1588 
  - https://github.com/radix-ui/primitives/issues/1569

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |         |         |
| React            | n/a     |         |
| Browser          |         |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

chaance:
Works as intended. See the reply in the linked PR.

PR #3157, opened 2024-09-30
feat: Change decision logic in useControllableState

flex-hyuntae:
### Description
Fix. https://github.com/radix-ui/primitives/issues/3156

### DIFF
--- a/.yarn/versions/05b8625b.yml
+++ b/.yarn/versions/05b8625b.yml
@@ -23,3 +23,6 @@ releases:
   "@radix-ui/react-toolbar": minor
   "@radix-ui/react-tooltip": minor
   "@radix-ui/react-use-controllable-state": minor
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


chaance:
I'm not entirely sure we want this. Our intent is to mirror the behavior of controlled/uncontrolled components in React itself. Take this, for example:

```tsx
function App() {
  const [value, setValue] = React.useState();
  return (
    <input
      value={value}
      onChange={(event) => {
        setValue(event.target.value);
      }}
    />
  );
}
```

If you type in the input, you'll get a console warning for switching between controlled/uncontrolled components since the initial value, while technically provided, is undefined. Attempting to set it back to undefined with the clear button does not clear the input.

The same holds true in Radix: if a value is undefined, the field is uncontrolled. Do not switch between controlled and uncontrolled components. If the nullable value reverts to undefined, pass a fallback eg: `value={nullableValue ?? ""}`

flex-hyuntae:
Ok i understand

```
Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components%s
```

Issue #3158, opened 2024-09-30
Radio Input Attribute checked can be faulty if set to boolean


maxkarkowski:
Hi, 
i ran into issue when styling radio group items with the new :has css styling.
i have a css like this 
```tsx
// sets interactive color if element is checked
// aria checked attribute is necessary if component is used without a form
      &:has(*[checked="true"]):not(:has(*:disabled)),
      &:has(*[data-state="checked"]):not(:has(*:disabled)) {
        background-color: red;
      }
```

if i switch between two radio button and then switch routes and go back via back button, both radio input are getting the style even though only one input radio has the attribute checked in the web inspector.

But if i select the items via the browser console, the "unchecked" radio item still hast checked true on the element.
 
After a little bit of digging it seems that setting the checked attribute is not the valid way.

https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

> The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

> If the attribute is present, its value must either be the empty string or a value that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the attribute's canonical name, with no leading or trailing whitespace.

Here is the code in your primitive. 
https://github.com/radix-ui/primitives/blob/6469d41dcc6e84fe41d70a2703924338e7562dd1/packages/react/radio-group/src/Radio.tsx#L80
I guess it would be better to
1) not setting the checked attribute entirely 
or
2) change the way of setting the attribute on the input radio

```tsx
 {isFormControl && (
          <BubbleInput
            control={button}
            bubbles={!hasConsumerStoppedPropagationRef.current}
            name={name}
            value={value}
            // only set attribute 'checked' if true to checked="checked"
           // if false don't render attribute completely
            checked={checked ? "checked" : undefined}
            required={required}
            disabled={disabled}
            // We transform because the input is absolutely positioned but we have
            // rendered it **after** the button. This pulls it back to sit on top
            // of the button.
            style={{ transform: 'translateX(-100%)' }}
          />
        )}
```

This is maybe also the case for checkboxes
https://github.com/radix-ui/primitives/blob/6469d41dcc6e84fe41d70a2703924338e7562dd1/packages/react/checkbox/src/Checkbox.tsx#L108

chaance:
[...]

PR #3160, opened 2024-10-01
Attribute checked has faulty types

maxkarkowski:
### Description

Possible solution for issue https://github.com/radix-ui/primitives/issues/3158

I added the type of undefined for the property checked in RadioItems, Checkboxes and switches to either set the html-attribute to true or undefined(to completely get rid of the attribute if set to false)


chaance:
This is something that would ultimately need to be addressed in `react-dom`. React itself requires that an input component to either be controlled or uncontrolled for the lifetime of the component, and passing `undefined` would break that rule and result in buggy behavior. We're following the rules of React here.

PR #3163, opened 2024-10-02
feat: Tooltip - Performance - avoid calling clearTimeout on unmount if timeout id is 0

carloseustaquio:
#### Context:
We have a product list page with products with multiple color variants. These color variants have the Tooltip component to display the color name on hover.

#### Issue:
We have noticed a significant amount of `clearTimeout` calls when the component with the Tooltips _unmounts_. That's interesting because it happens even when we don't trigger any of the Tooltips in the page.

6f5c359a-9b20-4739-82e8-41b2c0e2a1d8

After some investigation, we noticed this is caused by a `window.clearTimeout` that's happening in the component unmount.

#### Fix:
The fix proposed is quite simple. I'm checking `openTimerRef.current` and only calling this if it's not `0`. I'm also resetting `openTimerRef.current` to `0` after each `clearTimeout()`. We have noticed a reduction in the number of `clearTimeout` calls in the unmount of this component. This means a significant performance improvement in our scenario given we have dozens of Tooltips in some of our pages.

4ef24b80-ef0b-4659-a354-bec155e0e8cc

### DIFF
--- a/.yarn/versions/ae1744f4.yml
+++ b/.yarn/versions/ae1744f4.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-tooltip": patch
+
+declined:
+  - primitives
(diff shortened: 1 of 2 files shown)


chaance (approved):

oliviertassinari:
Nice, we did the same fix in https://github.com/mui/material-ui/pull/37512.

PR #3172, opened 2024-10-09
[DropdownMenu][ContextMenu][Menu] Only register a single set of event listeners for the entire document

### LINKED ISSUE #1239: Each Menu adds its own document event listeners for detecting keyboard state

mleibman-db:
### Description

Make `Menu` components more efficient when used in large numbers by only registering a single set of event listeners on the `document` to track whether a user is using a keyboard or a mouse.

This properly fixes https://github.com/radix-ui/primitives/issues/1239 and https://github.com/radix-ui/primitives/issues/1139.

I also saw that a similar change was attempted in https://github.com/radix-ui/primitives/pull/1252, but I'm proposing a much simpler implementation here. 

P.S. I did read through the discussion in https://github.com/radix-ui/primitives/issues/1239, but we unfortunately do seem to be running into some performance issues due to having a huge number of React components including Menus on the page. While I agree that this is an issue in and of itself, I also think that it's great for UI libraries to be efficient and be able to scale well with a large number of component instances.

### DIFF
--- a/packages/react/menu/src/Menu.tsx
+++ b/packages/react/menu/src/Menu.tsx
@@ -19,6 +19,7 @@ import { Slot } from '@radix-ui/react-slot';
 import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
 import { hideOthers } from 'aria-hidden';
 import { RemoveScroll } from 'react-remove-scroll';
+import { useIsUsingKeyboard } from './useIsUsingKeyboard';
 
 import type { Scope } from '@radix-ui/react-context';
 
@@ -69,7 +70,6 @@ const [MenuProvider, useMenuContext] = createMenuContext<MenuContextValue>(MENU_
 
 type MenuRootContextValue = {
   onClose(): void;
-  isUsingKeyboardRef: React.RefObject<boolean>;
   dir: Direction;
   modal: boolean;
 };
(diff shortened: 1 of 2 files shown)


mleibman-db:
Friendly ping.

strazan:
please merge 🙏 😭

PR #3174, opened 2024-10-10
[Select] fix: disabled select when fieldset has `disabled` attr

### LINKED ISSUE #2417: [Select] Not inheriting `disabled` from parent fieldset
## Bug report
Hi, 
I just noticed a small, but important issue with Select Primitive. 
If you wrap it with a `<fieldset disabled></fieldset>` it'll not inherit the disabled attribute. This works like this with every other "form" primitive (radio group, switch, checkbox etc) so I'd also think it should be the same for Select.

### Current Behavior

This works:
```tsx
<Select.Root disabled>
  (...)
</Select.Root>
```

This doesn't:
```tsx
<fieldset disabled>
  <Select.Root>
    (...)
  </Select.Root>
</fieldset>
```

### Expected behavior

A select root to inherit disabled attribute from parent fieldset.

ojj1123:
### Description

close #2417 
- When there is disabled `fieldset`, `Select` component does not be disabled.
- This is because [`Select.Trigger` has `onPointerDown` handler](https://github.com/radix-ui/primitives/blob/74b182b401c8ca0fa5b66a5a9a47f507bb3d5adc/packages/react/select/src/Select.tsx#L281) and even if button is disabled, that handler can be triggered.
- So I added the code that check if there is `fieldset` disabled.

#### AS-IS

https://github.com/user-attachments/assets/eb77a71e-9848-4070-8e6f-5f6ba98d092b

#### TO-BE

https://github.com/user-attachments/assets/632ab57c-4b5c-44dd-9a9a-19eb693918c4

### DIFF
--- a/.yarn/versions/a4fe86bf.yml
+++ b/.yarn/versions/a4fe86bf.yml
@@ -1,0 +1,5 @@
+releases:
+  "@radix-ui/react-select": patch
+
+declined:
+  - primitives
--- a/packages/react/select/src/Select.tsx
+++ b/packages/react/select/src/Select.tsx
@@ -288,10 +288,18 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
               target.releasePointerCapture(event.pointerId);
             }
 
+            const fieldset = target.closest('fieldset');
+            const fieldsetDisabled = fieldset ? fieldset.disabled : false;
+
             // only call handler if it's the left button (mousedown gets triggered by all mouse buttons)
             // but not when the control key is pressed (avoiding MacOS right click); also not for touch
             // devices because that would open the menu on scroll. (pen devices behave as touch on iOS).
-            if (event.button === 0 && event.ctrlKey === false && event.pointerType === 'mouse') {
+            if (
+              event.button === 0 &&
+              event.ctrlKey === false &&
+              event.pointerType === 'mouse' &&
+              !fieldsetDisabled
+            ) {
               handleOpen(event);
               // prevent trigger from stealing focus from the active item after opening.
               event.preventDefault();


RndUsername:
Will this react to changes? For example if the fieldset is dynamically disabled?

ojj1123:
> Will this react to changes? For example if the fieldset is dynamically disabled?

@RndUsername 
Yeah, It works well. This is because `disabled` state of `fieldset` is checked whenever triggering `onPointerDown` handler.
FYI, HTML select tag works as `pointerdown` event is triggered using `mouse`.

```tsx
export const Styled = () => {
  const [disabled, setDisabled] = React.useState(false);

  return (
    <div>
      <button type="button" onClick={() => setDisabled((prev) => !prev)}>
        {disabled ? 'enabled' : 'disabled'}
      </button>
      <div>
        <p>Radix `Select` component</p>
        <fieldset disabled={disabled}>
          {POSITIONS.map((position) => (
            <Label key={position}>
              Choose a number:
              <Select.Root defaultValue="two">
                ...
              </Select.Root>
            </Label>
          ))}
        </fieldset>
      </div>
    </div>
  );
};
```

https://github.com/user-attachments/assets/87e54158-ec68-4cc8-98d7-a32881739b5a

Sparticuz:
Ran into this today, can we get a review and merge? 🙏

ryanagillie:
Hey all, it's been 4 months can someone take a look at this?

ojj1123:
I feel like some people want it to be merged. So if need any help, I will be contributing or revewing:)

(I want to fix conflicts, but there is no permission to write this branch since I was not assigned on this MR)

BShaq:
Hi @hadihallak @chaance is there any update on this PR? Would love to see it move forward when you have time 🙏

@ojj11123: In the related issue (https://github.com/radix-ui/primitives/issues/2417), janruo mentioned that this problem affects additional components as well. Do you think it would make sense to address those in this PR too, or should that be handled separately?

ojj1123:
@BShaq I'm not sure if this repository is maintained. Even if i open some PRs or push some changes into this, maintainers wouldn't check it. So if you'd like to address this issue, it would be okay to do it.

PR #3183, opened 2024-10-20
[Select] fix: onClick event propagation

lucaserly:
### Description

Prevent Select.Root's onValueChange event to propagate to other event handlers.

### DIFF
--- a/.yarn/versions/04e78d65.yml
+++ b/.yarn/versions/04e78d65.yml
@@ -1,0 +1,2 @@
+releases:
+  "@radix-ui/react-select": patch
--- a/.yarn/versions/04e78d65.yml
+++ b/.yarn/versions/04e78d65.yml
@@ -1,5 +1,0 @@
-releases:
-  "@radix-ui/react-select": patch
-
-undecided:
-  - primitives
--- a/.yarn/versions/04e78d65.yml
+++ b/.yarn/versions/04e78d65.yml
@@ -1,2 +1,5 @@
 releases:
   "@radix-ui/react-select": patch
+
+undecided:
+  - primitives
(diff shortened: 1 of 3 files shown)


akselinurmio:
Why would this be needed? Is there some reason you can't do this on your own codebase? I would be very careful when planning to call `stopPropagation` anywhere, as it makes it harder to predict the behaviour of other event handlers listening to the same event.

lucaserly:
This would be needed because Select's onValueChange click event bubbles up to outer components, triggering unwanted events on a parent element. The reason I cannot do it in my codebase is because onValueChange does not have event as a parameter. Do you have an alternative in mind?

akselinurmio:
I understand and I would like to try and help you. You can add event listeners to the `event.target` element itself, or any parent of the `event.target` and the event stops bubbling to any further parents. Radix Primitives components call the event handler props you pass to them in the event handlers they add ([see here](https://github.com/radix-ui/primitives/blob/74b182b401c8ca0fa5b66a5a9a47f507bb3d5adc/packages/react/select/src/Select.tsx#L1275))). So instead of calling `event.stopPropagation()` in Radix UI code, could you try out if adding a new event listener to the target element as `onClick` prop in your app code would solve the issue?

For example:

```tsx
<Select.Root>
  {...}
  <Select.Portal>
    <Select.Content>
      {...}
      <Select.Item onClick={event => event.stopPropagation()}>
        {...}
      </Select.Item>
    </Select.Content>
  </Select.Portal>
</Select.Root>
```

lucaserly:
@akselinurmio thanks for the suggestion. Tried but still doesn't work.

PR #3225, opened 2024-11-12
Revert "[ScrollArea] Viewport fixes (#2945)"

chaance:
Reverts changes in #2945 that caused a regression where scroll area elements without horizontal overflow do not span with width of their container.


sdelhomm:
Reverting back the `display: table` causes so much issues... please consider providing a way to control the behavior via a prop

PR #3228, opened 2024-11-15
chore(.github/workflows): update github actions 'uses' to latest version

sukvvon:
### Description

image

* github actions is now using node version 20 but `actions/checkout@v2`, `actions/setup-nodes@v2` are using node verion 16 which is deprecated in github actions, so i update it all v4, you can see [GitHub Actions: Transitioning from Node 16 to Node 20](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/)
* also update `actions/cache@v1` to latest version v4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,20 +10,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


sukvvon:
@chaance Could you review this PR?

Issue #3224, opened 2024-11-12
SlotClone should not be passed a `ref` if it is a `React.Fragment`


terrymun:
### Current Behavior

Consuming `<Dialog.Portal>` (and likely other components that uses `<SlotClone>` under the hood)) will cause the following error to be thrown when using Next.js 15:

> Invalid prop `ref` supplied to `React.Fragment`

![Screenshot 2024-11-12 at 22 08 17](https://github.com/user-attachments/assets/382e4304-6fd9-468f-a32f-778a93bd4e0f)

### Expected behavior

Using `<Dialog.Portal>` should not be causing this error.

### Reproducible example

_TBD_

### Suggested solution

We should probably avoid passing `ref` to `<SlotClone>` component if it is a React Fragment:

https://github.com/radix-ui/primitives/blob/74b182b401c8ca0fa5b66a5a9a47f507bb3d5adc/packages/react/slot/src/Slot.tsx#L35

https://github.com/radix-ui/primitives/blob/74b182b401c8ca0fa5b66a5a9a47f507bb3d5adc/packages/react/slot/src/Slot.tsx#L35

### Additional context

_n.a._

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | `@radix-ui/react-dialog` | 1.1.2 |
| React            | n/a     | 19 RC   |
| Browser          | Chrome  | 130.0.6723.117  |
| Assistive tech   | n/a     | n/a     |
| Node             | n/a     | 20.17.0 |
| pnpm             | n/a     | 9.12.3  |
| Operating System | macOS   | 15.1 (24B83) |

terrymun:
This seems to be related to https://github.com/radix-ui/primitives/issues/3211

PR #3229, opened 2024-11-16
fix: do not pass ref into SlotClone component if children is a React.Fragment

terrymun:
### Description

This PR fixes the issue with consuming any Radix UI Primitive in React 19, which will throw the following error in the console (as reported in #3224):

385466239-382e4304-6fd9-468f-a32f-778a93bd4e0f

The error seems to arise from not checking the type of the `children` prop, which could potentially be a React Fragment that does not accept `ref` prop.

Therefore, the `ref` prop is only conditionally added to the props of the `<SlotClone>` element if `children` is not of type `React.Fragment`.

### DIFF
--- a/.yarn/versions/63262d25.yml
+++ b/.yarn/versions/63262d25.yml
@@ -1,0 +1,45 @@
+releases:
+  "@radix-ui/react-accessible-icon": patch
+  "@radix-ui/react-accordion": patch
+  "@radix-ui/react-alert-dialog": patch
+  "@radix-ui/react-announce": patch
+  "@radix-ui/react-arrow": patch
+  "@radix-ui/react-aspect-ratio": patch
+  "@radix-ui/react-avatar": patch
+  "@radix-ui/react-checkbox": patch
+  "@radix-ui/react-collapsible": patch
+  "@radix-ui/react-collection": patch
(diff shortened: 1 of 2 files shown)


terrymun:
Hi @chaance, any idea if we could get this merged? That will fix a lot of the extraneous errors to those consuming React 19 with Radix :)

V-iktor:
Currently waiting on this to be able to upgrade my project 🚀

terrymun:
@V-iktor For the moment being I am simply using the modified `<Slot>` component (as per my PR) and configure it as an alias for my webpack config:

```ts
webpack: (config) => {
  config.resolve = {
    ...config.resolve,
    alias: {
      ...config.resolve.alias,
      // TODO: Remove this shim once Radix fixes the issue with incorrect ref prop drilling
      //       https://github.com/radix-ui/primitives/issues/3224
      '@radix-ui/react-slot': path.resolve(__dirname, '/path/to/custom/slot.tsx'),
    },
  };

  return config;
},
```

V-iktor:
Hello world 🌍

terrymun:
@chaance Thanks a lot for looking into this and getting the PR merged in. Much appreciated 🙌

PR #3234, opened 2024-11-20
Fix Presence component memory leak

### LINKED ISSUE #3202: Dialog memory leak (detached Dialog.Content)
### Current Behavior

Every time a Dialog gets opened and then closed, a Detached `<div>` element (the `Dialog.Content`) gets detached but is never garbage collected, as if a reference was kept, preventing it. This represents a memory leak because opening and closing the dialog 20 times means 20 Detached `Dialog.Content` elements **as long as their whole tree of children**, which makes the memory usage grow. 

### Expected behavior

No reference should be held with the `Dialog.Content` element upon closing to allow proper garbage collection.

### Reproducible example

- Go to the following CodeSandbox: [https://codesandbox.io/p/sandbox/cool-meadow-h6vftg](https://codesandbox.io/p/sandbox/cool-meadow-h6vftg?file=%2Fsrc%2FApp.js%3A14%2C15-14%2C22) and click the "Open in a new tab" button at the top right of the Preview tab (or directly open the preview link: [https://h6vftg.csb.app/](https://h6vftg.csb.app/))
- Open the Chrome Devtools, go to the "Memory" tab, and click the "Detached elements" checkbox
- Click the blue "Start" button at the bottom of the panel
- Notice that no element is detached yet (or at least none related to Radix's Dialog)
- Click the "Edit profile" button of the app and close the dialog 4 of 5 times in a row
- Run a new "Memory"/"Detached elements" analysis in the devtools
- Notice multiple `<div>` elements with the `role=dialog` attribute get leaked (this is actually the Dialog.Content element)

### Suggested solution

Find which reference might not get removed when the dialog gets closed. I tried to search in [the sources](https://github.com/radix-ui/primitives/blob/main/packages/react/dialog/src/Dialog.tsx) but am not able to notice anything.

### Your environment 

| Radix Package(s) | @radix-ui/react-dialog        | 1.1.2        |
| React            | n/a     | ^18.0.0        |
| Browser          | Google Chrome        | 130.0.6723.92        |
| Operating System | Apple M1 Pro        |  Sonoma 14.7       |

lauri865:
### Description

Fixes #3202 

Took a while to track this down, but here goes:
When the child of Presence component gets unmounted, `stylesRef` is not clean up, which in turn prevents the detached nodes from being garbage collected, as it's holding a reference to the node. Affects any radix component that depends on Presence.
* Tooltip, HoverCard, etc. – leaves a new detached node after every hover/unhover
* Dialog, Popover, DropdownMenu, etc.  – leaves detached nodes around after closing

Note when testing: please build React for production first, as the dev-mode itself can have memory leaks.

Alternative could be to use WeakRef, but I wasn't sure how the Radix team feels about its browser support.

### DIFF
--- a/packages/react/presence/src/Presence.tsx
+++ b/packages/react/presence/src/Presence.tsx
@@ -153,7 +153,12 @@ function usePresence(present: boolean) {
   return {
     isPresent: ['mounted', 'unmountSuspended'].includes(state),
     ref: React.useCallback((node: HTMLElement) => {
-      if (node) stylesRef.current = getComputedStyle(node);
+      if (node) {
+        stylesRef.current = getComputedStyle(node);
+      }
+      else {
+        stylesRef.current = {} as any;
+      }
       setNode(node);
     }, []),
   };


maecapozzi:
It would be so fantastic if we could merge this PR, since it's preventing our usage of this component and causing severe performance degradation for us. Is there any way that I can help to get this PR merged?

chaance (approved):

chaance:
Fixed in https://github.com/radix-ui/primitives/commit/f2e6c688035d880a832d44755d31d39ba87c1d35, co-authored the commit accordingly!

PR #3250, opened 2024-11-24
[Switch] fix : Hidden input for Switch does not render "over the button"

### LINKED ISSUE #3235: Hidden input for Switch does not render "over the button"
### Current Behavior

https://github.com/user-attachments/assets/9d4f7414-dacd-418b-b1cf-5bb48cc243ef

Also in the docs it seems not to be laid over the button 
Screenshot 2024-11-20 at 13 32 16

### Expected behavior
The hidden input is correctly rendered above the button without the consumer having to add relative position to the parent element

### Suggested solution
maybe wrap with a div that is relative 

### Your environment

Brokyeom:
Closes #3235 

### Description

The BubbleInput within the Switch component had an absolute positioning applied, but there was no parent node with relative positioning, causing the component to overflow its intended area.

- Added a div wrapper around `<Primitive.button />` and `<BubbleInput />` and applied the necessary styles.
- Removed the `transform: translate(-100%)` property previously applied as an inline style on BubbleInput.

#### AS-IS

image

#### TO-BE

image

#### Test Result
image

#### Storybook 
image

### DIFF
--- a/packages/react/switch/src/Switch.tsx
+++ b/packages/react/switch/src/Switch.tsx
@@ -57,6 +57,10 @@ const Switch = React.forwardRef<SwitchElement, SwitchProps>(
 
     return (
       <SwitchProvider scope={__scopeSwitch} checked={checked} disabled={disabled}>
+        {/*
+        Wrap with div to prevent input from escaping component bounds
+        link : https://github.com/radix-ui/primitives/issues/3235
+        */}
         <div style={{ position: 'relative', display: 'inline-flex' }}>
           <Primitive.button
             type="button"
--- a/packages/react/switch/src/Switch.tsx
+++ b/packages/react/switch/src/Switch.tsx
@@ -59,7 +59,6 @@ const Switch = React.forwardRef<SwitchElement, SwitchProps>(
       <SwitchProvider scope={__scopeSwitch} checked={checked} disabled={disabled}>
         {/*
         Wrap with div to prevent input from escaping component bounds
-        link : https://github.com/radix-ui/primitives/issues/3235
         */}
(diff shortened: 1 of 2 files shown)


packages/react/switch/src/Switch.tsx
```diff
@@ -57,44 +57,46 @@ const Switch = React.forwardRef<SwitchElement, SwitchProps>(
...
-            bubbles={!hasConsumerStoppedPropagationRef.current}
-            name={name}
-            value={value}
-            checked={checked}
-            required={required}
+        {/*
+        Wrap with div to prevent input from escaping component bounds
+        link : https://github.com/radix-ui/primitives/issues/3235
+        */}
+        <div style={{ position: 'relative', display: 'inline-flex' }}>
```
Brokyeom, line 64: Add div element.

packages/react/switch/src/Switch.tsx
```diff
@@ -57,44 +57,46 @@ const Switch = React.forwardRef<SwitchElement, SwitchProps>(
...
+            aria-checked={checked}
+            aria-required={required}
+            data-state={getState(checked)}
+            data-disabled={disabled ? '' : undefined}
             disabled={disabled}
-            form={form}
-            // We transform because the input is absolutely positioned but we have
-            // rendered it **after** the button. This pulls it back to sit on top
-            // of the button.
-            style={{ transform: 'translateX(-100%)' }}
```
Brokyeom, line 95: Remove transform inline style.

chaance:
You can add the wrapper or add position: relative to an existing element directly. Adding a parent DOM element would cause other issues for existing users.

Issue #3037, opened 2024-07-27
Collapsible: `--radix-collapsible-content-height` CSS variable missing when equal to `0` 


fuma-nama:
### Current Behavior

When the Collapsible Content height is zero (empty content), `--radix-collapsible-content-height` isn't added to CSS variables.
This is problematic because in CSS, variables are inherited from parents. 

In a nested structure of collapsible, this causes **animation that depends on `--radix-collapsible-content-height`** to use the value from parent, which is an incorrect value.

https://github.com/user-attachments/assets/87a11952-8afa-440d-a1ce-5f331250dd72

For example, `math` item is a collapsible with zero height, it inherits the variable `--radix-collapsible-content-height` from parent, because it is missing on the `math` item itself.

### Expected behavior

`--radix-collapsible-content-height` should be defined on the content node (`--radix-collapsible-content-height: 0` in element styles)

### Reproducible example

https://codesandbox.io/p/sandbox/radix-primitives-template-2r30e

The nested item with empty content doesn't have the variable defined.

### Suggested solution

N/A

### Additional context

N/A

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |    `@radix-ui/react-collapsible`     |     1.1.0    |
| React            | n/a    | 18.3.0        |
| Browser          | chrome        | latest        |
| Assistive tech   |         |         |
| Node             | n/a     |      18   |
| npm/yarn         |   pnpm      |    9.5     |
| Operating System |       MacOS  |  14.5       |

PR #3270, opened 2024-12-04
Fixed "Collapsible: --radix-collapsible-content-height CSS variable m…

neo-schobert:
Description
This pull request fixes a bug where the --radix-collapsible-content-height CSS variable is not set when the Collapsible content's height equals zero. This issue caused incorrect behavior in nested collapsible components, as they would inherit the variable from their parent, leading to unexpected animations.

Problem
When a Collapsible content node has no height (e.g., empty content), the --radix-collapsible-content-height variable is missing. In a nested collapsible structure, child nodes without content inherit the variable from their parent, resulting in incorrect animations or unexpected visual behavior.

Solution
This PR ensures that the --radix-collapsible-content-height CSS variable is always defined on the content node, even when the height is zero. Specifically:
When the height of a Collapsible content is zero, the variable is explicitly set to --radix-collapsible-content-height: 0.

Changes
Updated logic to always define the --radix-collapsible-content-height variable on collapsible content, regardless of its height.
Added tests to verify that the variable is correctly applied when content height is zero.
Included an update to the documentation for developers to clarify the behavior of --radix-collapsible-content-height in nested structures.

### DIFF
--- a/packages/react/collapsible/src/Collapsible.tsx
+++ b/packages/react/collapsible/src/Collapsible.tsx
@@ -1,12 +1,12 @@
-import * as React from 'react';
 import { composeEventHandlers } from '@radix-ui/primitive';
+import { useComposedRefs } from '@radix-ui/react-compose-refs';
 import { createContextScope } from '@radix-ui/react-context';
+import { useId } from '@radix-ui/react-id';
+import { Presence } from '@radix-ui/react-presence';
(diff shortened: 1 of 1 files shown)


neo-schobert:
Fixed https://github.com/radix-ui/primitives/issues/3037#issue-2433522727

Issue #3274, opened 2024-12-06
Export useDialogContext in radix dialog


olafur164:
### Overview

I would like to have access to the useDialogContext. I'm using shadcn for components and one of the components is the Sheet component that uses Radix Dialog. What I want to do is https://www.radix-ui.com/primitives/docs/components/dialog#close-after-asynchronous-form-submission
But we are using the Sheet in the layout using nextjs combination of parallel and intercepting routes ( This might be the wrong way to do this but I'm investigating stuff) but then a child of this layout is what triggers closing the sheet. So in order to do this it is best to use react context. Which I can create myself but I don't any reason why this wouldn't just be exported from radix, is this a possibility?

chaance:
> We do not currently expose internal state hooks for our components. If we decide to change this, it's something we'd want to do consistently across the library to ensure the patterns are predictable. If you want to access the state I'd recommend controlling it and replicating this pattern in a wrapper component.

PR #3275, opened 2024-12-06
[Dialog]: feature export useDialogContext

olafur164:
### Description

Exporting useDialogContext for the purpose of having access to triggering open/close dialog in a child component of Dialog without making it controlled by the parent component

Closes: https://github.com/radix-ui/primitives/issues/3274

### DIFF
--- a/packages/react/dialog/src/Dialog.tsx
+++ b/packages/react/dialog/src/Dialog.tsx
@@ -555,6 +555,7 @@ const Close = DialogClose;
 
 export {
   createDialogScope,
+  useDialogContext,
   //
   Dialog,
   DialogTrigger,


chaance:
We do not currently expose internal state hooks for our components. If we decide to change this, it's something we'd want to do consistently across the library to ensure the patterns are predictable. If you want to access the state I'd recommend controlling it and replicating this pattern in a wrapper component.

xxxijustwei:
@chaance Thank you for your response, but I have a different perspective on this matter. You mentioned that internal state hooks for components are not currently exposed, and if this were to change, it would be done consistently across the library to ensure predictable patterns. However, you did not specify why exposing useDialogContext would be problematic or what potential risks and challenges it might introduce.

Not exposing internal context can limit developers' flexibility, especially in scenarios where finer-grained control over dialog state is needed. For instance, in complex UI interactions, a child component might need to directly trigger the opening or closing of a dialog without relying on the parent component to pass controlled state. This is not an uncommon requirement in real-world development, and exposing useDialogContext would allow developers to achieve more flexible customization without breaking the existing API.

If the concern is that exposing internal context might compromise encapsulation or lead to unpredictable behavior, I believe this can be mitigated by clearly documenting usage scenarios and limitations. 

Could you please elaborate on the specific reasons or potential issues with exposing internal context? This would help us have a more comprehensive discussion about whether this feature is worth adding and how it can be implemented while maintaining consistency and maintainability across the library.

programming-with-ia:
@xxxijustwei 
For additional context and supporting arguments, you might find this discussion helpful: [#3536](https://github.com/radix-ui/primitives/discussions/3536). It highlights why this feature could improve flexibility in real-world scenarios.

PR #3283, opened 2024-12-12
Fix composeRefs in React 19

dmitri-gb:
### Description

Ref callbacks [can now return cleanup functions](https://react.dev/blog/2024/12/05/react-19#cleanup-functions-for-refs) in React 19. Currently it's not always possible to use the new cleanup-returning ref callbacks with Radix components. For instance, the following example with `TextField` from Radix Themes isn't working properly.

```tsx
<TextField.Root
  ref={(el) => {
    console.log('created', el);
    return () => console.log('removed');
  }}
/>
```

Instead of logging `removed` when unmounting, this logs `created null`.

This PR fixes the issue. I added a single test for the `Switch` component to test the behaviour.

I also updated the main workspace React version to 19 and fixed a handful of type-checking issues. However, all packages should still be compatible with earlier React versions.

Storybook also needed an update, as the previous version was not compatible with React 19. I performed the update by simply running their codemod.


chaance (approved):

chaance:
Important to note that React 17/18 will log an error in the console when returning a function from a callback ref, but afaict it doesn't actually break anything. We'll need to add some annoying hack check for React 19 before returning the function.

dmitri-gb:
> Important to note that React 17/18 will log an error in the console when returning a function from a callback ref

Hmm, you're right. I missed that.

I think an easy fix would be to only return our own clean-up function if any of the passed refs also returned one. Then the error would only (and rightfully) show up, if someone were to actually return a function from their ref callback in React <19.

chaance:
@dmitri-gb I like that solution much better than trying to trick compilers 👍 Mind making that adjustment to the PR and I can merge?

dmitri-gb:
Since this is already merged, here's a new PR: https://github.com/radix-ui/primitives/pull/3285

PR #3303, opened 2024-12-28
Rearrange style prop to override unused CSS Variable

seongminn:
### Description

packages: `DropdownMenu`, `ContextMenu`, `Menubar`

Radix-ui allows us to easily add styles by passing a special CSS Variable. This is one of the most useful parts during development, but it can be a disadvantage in terms of debugging and readability because it is expressed as a style prop on the DOM. So we rearranged the order so that `...props.style` covers existing unused CSS Variables.


seongminn:
@lucasmotta @hadihallak @chaance Can I get your feedback about this PR?

Issue #1760, opened 2022-11-06
[Slider] onValueCommit not being called properly


vctormb:
## Bug report
Sometimes the `onValueCommit` prop is not being called properly when moving the thumb.

### Current Behavior
When moving the thumb sometimes the `onValueCommit` prop is not called.

### Expected behavior
To always call the `onValueCommit` prop when the user stops sliding

### Reproducible example
I'm using the [same example](https://codesandbox.io/s/5s1o0m?module=App.js&file=/App.js:1434-1447) from the documentation.

https://user-images.githubusercontent.com/13953703/200179160-12c1437c-1a3f-4c91-bcc7-6ef191ca45d7.mov

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |     slider    |   1.1.0      |
| React            | n/a     |   18.2      |
| Browser          |    chrome     |    107.0.5304.87 (Official Build) (arm64)     |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |    macOS     |    Version 13.0     |

benoitgrelard:
Hey @vctormb,

I am unable to reproduce this on my side, what browser/OS are you using and how do your replicate it?

vctormb:
> Hey @vctormb,
> 
> I am unable to reproduce this on my side, what browser/OS are you using and how do your replicate it?

Hey @benoitgrelard! chrome v107 / macOS v13 ventura.

I just opened up the slider docs and moved the thumb really fast. Keep sliding until you notice that the onValueCommit wasn't called

benoitgrelard:
I'm not yet on Ventura, but I cannot reproduce the issue with the same chrome v107.

I notice in the video that when it happens if when you released outside the codesandbox iframe.
Although, this doesn't break things for me, could you try to do this but slowly and see if it still breaks for you?
Perhaps it could be a pointer capture bug in Ventura?

vctormb:
> I'm not yet on Ventura, but I cannot reproduce the issue with the same chrome v107.
> 
> I notice in the video that when it happens if when you released outside the codesandbox iframe. Although, this doesn't break things for me, could you try to do this but slowly and see if it still breaks for you? Perhaps it could be a pointer capture bug in Ventura?

Oh, right! Here it is another video doing it slowly

[...]

PR #3329, opened 2025-01-18
[Slider] onValueCommit not called when start/end are equal

emmavray:
### Description

For multi-value sliders where two values are allowed to overlap, `onValueCommit` will not be called for overlapping values.

#1760

### Steps to reproduce

```
<Slider.Root
	defaultValue={[10, 30]}
	minStepsBetweenThumbs={0}
>
	<Slider.Track>
		<Slider.Range />
	</Slider.Track>
	<Slider.Thumb />
	<Slider.Thumb />
</Slider.Root>
```

Dragging the END value equal to the START value will trigger `onValueChange` but will not trigger `onValueCommit`, despite the end value having changed.

https://github.com/user-attachments/assets/106c1611-d459-48fa-93a1-0ba5394dc292

### Cause

`updateValues()` is responsible for setting `valueIndexToChangeRef`, which is used later by `handleSlideEnd()` to determine if changes have been made. The problem is that `updateValues()` will attempt to find the relevant index by searching for the current value -- using **the first index found**. If the array contains multiple identical values, this strategy will fail.

Effectively when multiple Thumbs have identical values, the first index is always used. This means that `handleSlideEnd()` will only check for changes in the first Thumb's value.

Changing values using the keyboard works as expected, this bug only affects dragging behavior. Changing the first value always works as expected.

### Fix

`updateValues()` already receives an index parameter; it's used to validate that the minSteps has been satisfied, and in some cases used for change detection. Using this index instead of trying to infer it by searching the first occurrence of the value solves the issue.

### DIFF
--- a/.yarn/versions/a5937b9e.yml
+++ b/.yarn/versions/a5937b9e.yml
@@ -1,0 +1,5 @@
+releases:
+  primitives: decline
+
+declined:
+  - "@radix-ui/react-slider"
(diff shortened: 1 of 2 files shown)


drazik (commented):
This seems to fix the issue. I don't have approve permission. Can somebody with the permissions review it please?

PR #3347, opened 2025-01-23
Update README.md links

ahsalar:
Hey
It seems that the links are stale. Also the website is not forwarding addresses without "www" to the correct place at the moment. As a result links are not working.


TheGreyfeather:
This also fixes this issue: https://github.com/radix-ui/website/issues/846

chaance:
This was a temporary DNS issue that should be resolved.

PR #3354, opened 2025-01-29
 Configure adaptive positioning for small screens

ritikpal1122:
feat(popover): improve responsive positioning behavior

- Add collision detection and boundary constraints
- Configure adaptive positioning for small screens
- Implement fallback positions when space is limited
- Add padding and sticky behavior for better UX

Whart I've fixed in this PR 
## Changes
- Added collision detection using `avoidCollisions` prop
- Set document.body as collision boundary
- Configured adaptive positioning with fallback positions
- Added 10px collision padding for better spacing
- Implemented sticky behavior for smooth transitions

## Testing
- Tested on small screens
- Verified tooltip repositioning when space is limited
- Confirmed proper behavior on all viewport sizes

### DIFF
--- a/apps/ssr-testing/app/popover/page.tsx
+++ b/apps/ssr-testing/app/popover/page.tsx
@@ -6,7 +6,16 @@ export default function Page() {
     <Popover.Root>
       <Popover.Trigger>open</Popover.Trigger>
       <Popover.Portal>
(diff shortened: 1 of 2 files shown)


ritikpal1122:
hi @chaance can you please check this out

PR #1845, opened 2022-12-18
Use Owner document for selected components

ritz078:
### Description

Related to #1677 and #1721

Use `ownerDocument` instead of `document` in the following components:

- PopOver
- FocusScope
- Toolbar
- Menu
- Popper
- RovingFocusGroup
- Select
- Toast
- HoverCard
- FocusGuard

The stitch styles do not get applied in the popup window because of how stitch is used here. Due to this issue, it's very hard to ensure that components like `NavigationMenu` are working correctly in the popup window.

PR #3357, opened 2025-02-03
feat(DocumentContext): Add support for React Portals in new windows & document picture-in-picture

### LINKED ISSUE #1721: [All] Ensure all `document` handlers are attached to `ownerDocument`
### Overview

Currently, wherever we attach document handlers, we do so at the global `document` level.
Following up from #1677 where it was patched only for `DismissableLayer` we need to find a solution to do this throughout the entire library.

### Additional context

Here's another recent discussion which demonstrates similar issues with tooltips rendered in iframes: #1715

arturbien:
### Description

This PR adds support for using Radix Primitives in a new window popups and Document Picture in Picture.

https://github.com/user-attachments/assets/203fcba7-c61c-471c-b79f-d5340876e806

## Use cases

One of the common use cases are popup windows and [Document Picture-in-Picture ](https://developer.mozilla.org/en-US/docs/Web/API/Document_Picture-in-Picture_API). Document Picture-in-Picture API is used for example by apps like Google Meet where users can enter PIP mode with livestream and chat portalled to a floating window.

This is a use case we're currently facing at [Whop](https://whop.com/), we're building a live-streaming app with Document PIP support.

## Why doesn't it work currently in Radix

Since JS and React is running in the parent window, and bunch of components rely on calling `document`,  they break when being portalled to a new document. For example when you try to open `<Dialog />` or `<Tooltip />` in a new document, it will actually open in the parent window, because all event listeners and portals are attached to the parent window.

## Solution

There were attempts at fixing this before , but the solutions mostly relied on tweaking individual components instead of creating a more global and simpler solution:
https://github.com/radix-ui/primitives/issues/1721
https://github.com/radix-ui/primitives/issues/1676
https://github.com/radix-ui/primitives/pull/1677
https://github.com/radix-ui/primitives/pull/1845

This PR adds an optional `<DocumentContext.Provider value={documentElement} />` component where users can pass reference to the popup window document, and `useDocument()` hook which returns document passed to the provider with a fallback to `globalThis.document`. The hook is used internally within all primitives that rely on the `document`. All that user has to do to make Radix work with portals to another window is to wrap portalled content with the provider like so:

```tsx
createPortal(
   <DocumentContext.DocumentProvider document={portalElement.ownerDocument}>
        {content} // Radix components here
   </DocumentContext.DocumentProvider>,
   portalElement
 )
```

**This PR is not introducing any breaking changes or new props to any of the modified components. The provider is completely optional.** 

## TODO:
- [x] Swap `document` references with `providedDocument`
- [x] Swap `window` references with `providedDocument.defaultView` 
- [x] Storybook example
- [x] SSR page example
- [x] Tests
- [ ] Bump package versions (minor or or patch?) for all updated primitives
- [ ] Add `<DocumentContext />` page to the docs website

(cc @chaance @benoitgrelard @hugotiger @seleckis @2DTW @gregorybolkenstijn @william-will-angi @LordZardeck @ritz078 @joaodematte)

Closes #1721 
Closes #1715


chaance:
Thanks for this! It's been on my todo list for a while to fix document references for a while.

Curious though why this needs to be a separate context and not a utility function that checks the `ownerDocument` of a given ref node. I feel like there's a good bit of bloat with the React tree in Radix components as-is, so if we can accomplish this without a new provider I think that would be simpler and a bit more efficient. Thoughts?

arturbien:
> Curious though why this needs to be a separate context and not a utility function that checks the ownerDocument of a given ref node

- Not every component renders an actual HTML element (e.g., Portal, DescriptionWarning, FocusGuards), making it difficult to reliably access a DOM reference. A utility function wouldn’t work in cases where there’s no direct element to check.

- Some components rely on document even before rendering anything (e.g., attaching event listeners, reading window size, using MutationObserver/ResizeObserver, or reading attributes). The provider approach is more future-proof since it ensures access to the correct document in all scenarios.

- I think adding utility function would require additional prop drilling and managing refs across multiple components, adding unnecessary complexity.

- useDocument improves DX. In our case, the same component is rendered both on the main page and in a new window via a portal. Some of its child components also rely on document (for event listeners etc), so accessing it via a hook is far more convenient than manually handling refs. Otherwise, developers would have to implement their own DocumentProvider anyway.

- It's easier to explain how to build PIP and popup windows with Radix - "just wrap portalled content in provider, then use `useDocument` whenever you need access to the relavant `document`). Otherwise a developer will build let's say a Dialog that renders their own custom component that relies on the document, and we'll get bunch of questions about listeners being attached to the wrong window. Exposing and documenting a hook for getting relevant document solves this

> I feel like there's a good bit of bloat with the React tree in Radix components as-is

- The Provider is optional and will only be used in this rare case of building popups and Document Picture in Picture, so I don't think it's a big issue

chaance:
All valid points when it comes to DX _inside of Radix itself_ but we've made similar decisions with that mindset and I think it's been at the expense of both DX on the consumer's side (the Raect Dev tools components tab is a giant mess) and end user performance. I'd like to recalibrate how we think about and prioritize DX in general.

Might not have time to dig into this in depth right away. Competing priorities and whatnot. It's on my TODO list.

arturbien:
> All valid points when it comes to DX _inside of Radix itself_ but we've made similar decisions with that mindset and I think it's been at the expense of both DX on the consumer's side (the Raect Dev tools components tab is a giant mess) and end user performance. I'd like to recalibrate how we think about and prioritize DX in general.
> 
> Might not have time to dig into this in depth right away. Competing priorities and whatnot. It's on my TODO list.

Not sure I understand the impact of the DX on the consumer side though  🤔 - the provider will only ever be used to wrap components passed to `ReactDom.createPortal()` and only for the niche use case for cross-document popups and DPIP like:

```tsx
ReactDom.createPortal(<DocumentContext.Provider value={popupWindow.document}>{content} </DocumentContext.Provider>)
```

So 99% of consumers won't ever use the provider in their apps 🤔

chaance:
@arturbien to be clear, I haven't reviewed in depth yet so take my feedback at this stage with a grain of salt :) Initial reaction was assuming that the provider would be used internally but that does not appear to be the case, so I'm open to it.

packages/react/document-context/src/document-context.tsx
```diff
@@ -0,0 +1,23 @@
...
+const DocumentContext = React.createContext<Document | null>(null);
+
+interface DocumentProviderProps {
+  document?: Document;
+  children: React.ReactNode;
+}
+
+export function DocumentProvider({ document: doc, children }: DocumentProviderProps) {
+  const value = React.useMemo(
+    () => doc ?? (typeof document !== 'undefined' ? globalThis?.document : null),
```
chaance, line 13: I don't think we need to fallback to the global document in this case. This is already handled in `useDocument`.

In either event, memoization in this context isn't really needed since the reference is outside the React tree and stable.

packages/react/document-context/src/document-context.tsx
```diff
@@ -0,0 +1,23 @@
...
+    [doc]
+  );
+  return <DocumentContext.Provider value={value}>{children}</DocumentContext.Provider>;
+}
+
+export function useDocument() {
+  const doc = React.useContext(DocumentContext);
+  // Return default document if available and no context value
+  return doc ?? (typeof document !== 'undefined' ? globalThis?.document : null);
+}
```
chaance, line 23: We should check that JS is hydrated before accessing the global document. This prevents potential bugs if referenced during render.
```suggestion
const subscribe = () => () => {};
export function useDocument() {
  const doc = React.useContext(DocumentContext);
  const isHydrated = useSyncExternalStore(
    subscribe,
    () => true,
    () => false
  );
  return doc ?? (isHydrated ? document : null);
}
```

Also want to use the `use-sync-external-store` package since we support older versions of React.
```ts
import { useSyncExternalStore } from 'use-sync-external-store/shim';
```

arturbien:
hey @chaance ! 🫡 any chance we could get this in this week? I think the sooner the better since this PR touches lots of components

chaance:
@arturbien probably not this week, got a few other things going on that take priority. Probably won't be merging any other changes in the mean time so you shouldn't need to deal with conflicts before I get to it!

arturbien:
hey @chaance ! any updates on this?  🥹

leweyse:
Hi @chaance! Do you have any updates regarding this PR?

chaance:
Fixing up a few things over at #3516. Had a bunch of conflicts since I overhauled the repo setup and didn't want to ask you to fix things again!

PR #3380, opened 2025-02-13
Close tooltip on pointer down

ajworkos:
### Description

This PR closes the tooltip automatically `onPointerDown`.

Attached are the before and after behaviours:

## Before

https://github.com/user-attachments/assets/5752322f-4afe-41c3-9d50-91d72d50b909

## After

https://github.com/user-attachments/assets/16326be9-4a4b-4abb-b2b9-ad2dc470d6fb

### DIFF
--- a/packages/react/tooltip/src/tooltip.tsx
+++ b/packages/react/tooltip/src/tooltip.tsx
@@ -303,6 +303,7 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
             hasPointerMoveOpenedRef.current = false;
           })}
           onPointerDown={composeEventHandlers(props.onPointerDown, () => {
+            context.onClose();
             isPointerDownRef.current = true;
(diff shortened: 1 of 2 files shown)


packages/react/tooltip/src/tooltip.tsx
```diff
@@ -303,6 +303,7 @@ const TooltipTrigger = React.forwardRef<TooltipTriggerElement, TooltipTriggerPro
             hasPointerMoveOpenedRef.current = false;
           })}
           onPointerDown={composeEventHandlers(props.onPointerDown, () => {
+            context.onClose();
```
chaance, line 306: I think we should call this conditionally in the event that the user's handler triggers some sort of side-effects or unrelated state changes.
```suggestion
            if (context.open) {
              context.onClose();
            }
```

chaance (changes requested):

chaance (approved):

PR #3383, opened 2025-02-16
list primitive

abernier:
### Description

A [Listbox](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/) primitive for radix

https://github.com/user-attachments/assets/b1e60d64-5827-4a76-bd4a-38e627c126bf

### DIFF
--- a/.yarn/versions/b270b59b.yml
+++ b/.yarn/versions/b270b59b.yml
@@ -1,0 +1,3 @@
+declined:
+  - primitives
+  - "@radix-ui/react-list"
--- a/packages/react/list/package.json
+++ b/packages/react/list/package.json
@@ -1,0 +1,72 @@
+{
+  "name": "@radix-ui/react-list",
+  "version": "1.0.0",
+  "license": "MIT",
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.mts",
+        "default": "./dist/index.mjs"
+      },
+      "require": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      }
+    }
+  },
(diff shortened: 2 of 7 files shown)


packages/react/list/package.json
abernier, line 1: i've just duplicated it from another one and adapted

packages/react/list/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "@radix-ui/react-list",
+  "version": "1.0.0",
```
abernier, line 3: not sure about the version you want

packages/react/list/package.json
```diff
@@ -0,0 +1,72 @@
...
+  },
+  "homepage": "https://radix-ui.com/primitives",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/radix-ui/primitives.git"
+  },
+  "bugs": {
+    "url": "https://github.com/radix-ui/primitives/issues"
+  },
+  "stableVersion": "1.0.0"
```
abernier, line 71: not sure about the version you want (bis)

packages/react/list/src/list.stories.tsx
abernier, line 1: will add some more stories

.yarn/versions/b270b59b.yml
abernier, line 1: i've chosen `declined` to workaround the pre-push hook, but not familiar with changeset

chaance:
Appreciate the contribution, but we are not looking externally for new primitives at the moment. I'd generally advise opening a discussion to propose additions to the library before opening a PR.

PR #3392, opened 2025-02-24
Make sure we never try to assign `undefined` to pointerEvents

etodanik:
### Description

I'm using a non-standard browser (https://coherent-labs.com/) for game UI. It handles unset `pointerEvents` unusually by returning `undefined` (where the spec would require an empty string). This fix makes this code resilient to such behavior.

A bug was also submitted to their team, but I thought this is a nice harmless way to make sure `undefined` won't ever make its' way to `pointerEvents` (which would be a no-op)


etodanik:
@chaance would you mind taking a look?

chaance:
@etodanik Please add a changeset to make sure this is captured in the next release.

PR #3407, opened 2025-03-07
fix(VisuallyHidden): Prevent unexpected scroll overflow issue by adding explicit position

### LINKED ISSUE #3406: VisuallyHidden component causes unnecessary scroll in overflow containers
### Current Behavior

When using the `VisuallyHidden` component inside containers with `overflow: hidden` or `overflow-y: scroll`, the component extends beyond its parent container boundaries, causing unwanted scrollbars to appear or content to overflow. This creates unexpected scrolling behavior in parent components.

### Expected behavior

The `VisuallyHidden` component should remain within its parent container boundaries and not trigger any scrolling behavior in parent components, while still maintaining its accessibility functionality.

### Reproducible example

[CodeSandbox Example](https://codesandbox.io/p/sandbox/cjq654)

```jsx
import { VisuallyHidden } from "@radix-ui/react-visually-hidden";

export default function App() {
  return (
    <div
      style={{ height: "200px", overflow: "scroll", border: "1px solid red" }}
    >
      {Array.from({ length: 200 }, (_, index) => (
        <p key={index}>test</p>
      ))}
      <VisuallyHidden />
    </div>
  );
}

```

### Suggested solution

The issue can likely be fixed by adding explicit positioning (`top: 0` and `left: 0`) to the VisuallyHidden component's inline styles. This would properly constrain the element within its parent container.

```jsx
<span
  style={{
    position: 'absolute',
    border: 0,
    width: 1,
    height: 1,
    padding: 0,
    margin: -1,
    overflow: 'hidden',
    clip: 'rect(0, 0, 0, 0)',
    whiteSpace: 'nowrap',
    wordWrap: 'normal',
    top: 0, // Add this
    left: 0, // Add this
  }}
>
  {children}
</span>
```

### Additional context

This issue impacts layouts where precise control over scrolling behavior is important, such as fixed-height containers or modals with scrollable content.

### Your environment

| Software         | Name(s)                    | Version |
| ---------------- | -------------------------- | ------- |
| Radix Package(s) | @radix-ui/react-visually-hidden | 1.1.2   |
| React            | n/a                        | 18.2.0  |
| Browser          | Chrome                     | Latest  |
| Assistive tech   | N/A                        | N/A     |
| Node             | n/a                        | N/A     |
| npm/yarn         | npm                        | N/A     |
| Operating System | CodeSandbox Environment    | N/A     |
| TypeScript       | n/a                        | 4.4.4   |
| React Scripts    | n/a                        | 5.0.1   |

cham0287:
### Description

Fixes #3406

This PR addresses an issue where the VisuallyHidden component could trigger unwanted scrollbars when used within containers that have `overflow-y-scroll` applied. By adding explicit `top: 0` and `left: 0` positioning to the component, it properly constrains the element within its parent container boundaries.

### Changes made:
Added `top: 0` and `left: 0` to the inline style properties of the VisuallyHidden component

### DIFF
--- a/packages/react/visually-hidden/src/visually-hidden.tsx
+++ b/packages/react/visually-hidden/src/visually-hidden.tsx
@@ -20,6 +20,8 @@ const VisuallyHidden = React.forwardRef<VisuallyHiddenElement, VisuallyHiddenPro
         style={{
           // See: https://github.com/twbs/bootstrap/blob/main/scss/mixins/_visually-hidden.scss
           position: 'absolute',
+          top:0,
+          left:0,
           border: 0,
           width: 1,
           height: 1,


ahaubenstock:
+1 just ran into this problem today as well with select inside a scrolling component

PR #3408, opened 2025-03-07
dispatch slideEnd when pointerCapture is lost

lukeingalls:
This is meant to address https://github.com/radix-ui/primitives/issues/1760.

When testing I noticed that `pointerUp` fires unreliably. In order to reliably commit updates when a slide ends I also need to monitor `lostPointerCapture`. I don't know how to consistently force this event to emit but it seems more likely when I drag outside of the browser window.

This demo is shot from this branch on the storybook view. The log is associated with the event listener that is calling `onSlideEnd`.

https://github.com/user-attachments/assets/08004e45-5f59-4249-b28f-5728d98fabce


denzellgh (commented):
Looks good

guy-pimloc:
This looks good - is this ready to merge?

PR #3425, opened 2025-03-20
[DismissableLayer] Fix race condition between layers closing at the same time

trevorharwell:
### Description

I encountered a bug where multiple layers closing at the same time would both determine that `context.layersWithOutsidePointerEventsDisabled.size > 1` which prevented the pointer events from being reset back to it's original value.  By moving the `context.layersWithOutsidePointerEventsDisabled.delete(node)` to inside of the clean up method for the outer events hook, the race condition is resolved.

I see the comment about "We purposefully prevent combining this effect with the `disableOutsidePointerEvents` effect" which this PR might not be aligned with.  My thought is that the context.layers could represent the "creation order" while context.layersWithOutsidePointerEventsDisabled would not need to adhere to that.


trevorharwell:
@chaance could you take a look at this?  I think it's pretty straight forward and solves a really nasty bug.

zb-sj:
I confirm that this PR fixes what it describes.

Issue #3422, opened 2025-03-19
[DismissableLayer/useEscapeKeydown] Using "capture" in useEscapeKeydown breaks compatibility with non-radix components


noahnu:
### Current Behavior

_I was going back and forth on whether to report this as a bug or feature request since it was an intentional change, ultimately settled on a bug since it's unexpected behaviour / causes compatibility issues._

We're using radix's popover component which internally uses useEscapeKeydown (transitively via dismissible layer) to listen to the "Escape" key and close the popover. If we have a non-radix popover-like component rendered inside (e.g. react-select or react-datepicker) we would expect hitting the "Escape" key to close just react-select or react-datepicker, i.e. the inner dialog/menu. In fact both of these examples, react-select and react-datepicker listen to the "keydown" and call `event.preventDefault()`. Since radix's useEscapeKeydown uses "capture: true", it runs before any of the other components. It doesn't give the other components an opportunity to stop the event.

This behaviour regressed in https://github.com/radix-ui/primitives/pull/2761 and worked previously.

### Expected behavior

`useEscapeKeydown` should not use "capture", thus giving other html elements an opportunity to call "preventDefault" on the event.

### Suggested solution

Ideally radix doesn't use "capture: true". Alternatively if that's too much of a breaking change as folks have already migrated code and adapted (it's been out for a while), perhaps some flag to override this without having to resort to a patch? In the meantime, we've patched the file to unblock ourselves.

### Additional context

On both the original pull request https://github.com/radix-ui/primitives/pull/2761 and the associated GitHub issue https://github.com/radix-ui/primitives/issues/2653 a number of folks reported how the new behaviour broke compatibility with other tools.
- https://github.com/oxidecomputer/console/issues/2557
- https://github.com/radix-ui/primitives/pull/2761#issuecomment-1994574961 (ariakit)

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |         |         |
| React            | n/a     |    18     |
| Browser          |         |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |     yarn    |    4.7.0     |
| Operating System |         |         |

PR #3438, opened 2025-04-01
Make useEscapeKeydown's event capture optional

bhowell2-fleetio:
### Description
This was originally implemented to address #2653. 
This was introduced here: https://github.com/radix-ui/primitives/pull/2761
There are other's impacted based on this issue: https://github.com/radix-ui/primitives/issues/3422

This was quite the sweeping change and makes it quite difficult for anyone to intercept this. Many libraries/components rely on checking various states of the event to determine if they will perform some function and now internal components cannot handle this event first and Radix sets the defaultPrevented state of this. From a sanity POV, does it really make sense to have some higher level component handle this before the component has a chance? 

I'm more of the opinion that this behavior should be rolled back completely - or even keep this PR, but flip the defaults (i.e., capture is disabled by default rather than enabled - which I only did to keep this inline with current functionality).

In this PR we can expose this behavior at least. Without the "ignore/disabling" of the capture in this PR, as far as I can tell, it would necessitate creating a Dialog/Dismissable Context so lower level components can hook into this behavior - and even still it's not going to be great b/c useEscapeKeydown calls `preventDefault` which other components are going to check.

E.g., react-select's menu cannot be closed with the escape key, becasue the Dialog/DismissableLayer eats the event and prevents default on it and closes the dialog. Alternatively, it leaves the dialog open and the menu open as the Dialog/DismissableLayer still prevents default, preventing react-select from closing.

Change: allow control of escape keydown capture.

### DIFF
--- a/packages/react/dismissable-layer/src/dismissable-layer.tsx
+++ b/packages/react/dismissable-layer/src/dismissable-layer.tsx
@@ -37,6 +37,15 @@ interface DismissableLayerProps extends PrimitiveDivProps {
    */
   onEscapeKeyDown?: (event: KeyboardEvent) => void;
   /**
+   * If escape key event handler should avoid capturing the event before it is dispatched
+   * to the EventTarget.
+   *
+   * See: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#capture
+   *
+   * @default false (i.e., capture the event).
+   */
+  ignoreEscapeKeyCapture?: boolean;
+  /**
    * Event handler called when the a `pointerdown` event happens outside of the `DismissableLayer`.
    * Can be prevented.
    */
@@ -62,6 +71,7 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
   (props, forwardedRef) => {
     const {
       disableOutsidePointerEvents = false,
+      ignoreEscapeKeyCapture,
       onEscapeKeyDown,
       onPointerDownOutside,
       onFocusOutside,
@@ -99,15 +109,19 @@ const DismissableLayer = React.forwardRef<DismissableLayerElement, DismissableLa
       if (!event.defaultPrevented) onDismiss?.();
     }, ownerDocument);
 
-    useEscapeKeydown((event) => {
-      const isHighestLayer = index === context.layers.size - 1;
-      if (!isHighestLayer) return;
-      onEscapeKeyDown?.(event);
-      if (!event.defaultPrevented && onDismiss) {
-        event.preventDefault();
-        onDismiss();
-      }
-    }, ownerDocument);
+    useEscapeKeydown(
+      (event) => {
+        const isHighestLayer = index === context.layers.size - 1;
+        if (!isHighestLayer) return;
+        onEscapeKeyDown?.(event);
+        if (!event.defaultPrevented && onDismiss) {
+          event.preventDefault();
+          onDismiss();
+        }
+      },
+      ownerDocument,
+      ignoreEscapeKeyCapture
+    );
 
     React.useEffect(() => {
       if (!node) return;
--- a/packages/react/use-escape-keydown/src/use-escape-keydown.tsx
+++ b/packages/react/use-escape-keydown/src/use-escape-keydown.tsx
@@ -6,7 +6,8 @@ import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
  */
 function useEscapeKeydown(
   onEscapeKeyDownProp?: (event: KeyboardEvent) => void,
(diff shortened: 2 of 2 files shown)


chaance:
While I am not sold on the approach here (seems like a leaky abstraction) I have felt similar pain points when working with third-party libraries. I feel like there are a few alternatives we should consider to make it easier to work with and intercept events handled by Radix without interfering with browser defaults.

A few initial ideas (open to others):

### Event handler return values

Support return values for event handlers that we can use to determine whether or not to call internal handlers without calling `event.preventDefault`. That would look something like this:

```ts
<DIsmissableLayer
  onEscapeKeyDown={(event) => {
    if (isThirdPartyPopoverOpen) {
      // returning `false` tells us to skip calling internal handlers
      return false;
    }
  }}
/>
```

This approach would probably be the simplest to implement, but has a few drawbacks. We'd probably have to consider this a breaking change since users often pass arrow functions with implicit returns that could interfere with our logic. Also, while it's simple enough _if you know the API_, I think it's somewhat unintuitive as it deviates from standard event handler logic.

### Add new `RadixEvent`

I've been mulling this idea for a while and the more I think about it, the more I like it. Basically all event handlers passed down to Radix components would receive an additional argument of a new type `RadixEvent` that would look something lke this:

```ts
interface RadixEvent {
  defaultPrevented: boolean;
  preventDefault: () => void;
}

interface RadixEventHandler<E extends { defaultPrevented: boolean }> {
  // preserve the original event handler type for backwards compatibility
  (event: E): void;
  (event: E, radixEvent: RadixEvent | undefined): void;
}
```

This would result in user code that looks something like this:

```ts
<DIsmissableLayer
  onEscapeKeyDown={(event, radixEvent) => {
    if (isThirdPartyPopoverOpen) {
      radixEvent?.preventDefault();
    }
  }}
/>
```

bhowell2-fleetio:
@chaance thanks for your response. Yeah I tried to work around this and I didn't love how much hacking I was having to do to get it to work - and even then the easier hacking wasn't fullproff b/c of Radix's forced handling (still calling `preventDefault`) of the event, so would have to take it further into controlling the 3rd party lib states rather than the lib being able to control the event first. 

Probably any change to this is going to be breaking at this point. 

Either sounds fine to me, and agree both are better than my PR here in terms of functionality and overall library health.

I will say I'm still wondering if this `capture: true` is a good idea in the Dialog at all? Even if you don't want to expose the capture ability outside of Radix via prop, should the modal be capturing this before internal elements/components have the chance? I understand the need to capture this on the DropdownMenu that originated this change, but on the Dialog i'm struggling to see a case that makes sense for the Dialog to handle it before its internals - this works with Radix's components that use the `DismissableLayer` which uses a stack of sorts to determine if it should handle the event - so the Dialog is not closing if, for instance, a `DropdownMenu` is open within the Dialog, but this still fails for external libs. 

Regarding the first change, checking for explicit false/true rather than `falsy` will help some, but yeah agreed i'm sure many are using `() => functionCall()` that returns booleans as well. Would be breaking for sure, but least it's an easy fix. 

In your second example are you then requiring the user to do this prevent default rather than the dismissable layer (`useEscapeKeydown`) automatically calling it? W/r to the capture change here, then that would not be much different than exposing the prop here (i'm only talking about the capture and I think the 2nd is defintely more well rounded overall that this PR). 

In the first and second example, if capture is still occurring first, we're going to have to create a blacklist or whitelist of targets that should/should not close the dialog - a bit annoying and prone to things slipping through. Or use a context/etc to hook into this onEscapeKeyDown, because many times the components are nested deep within others and it's not the component that is mounting the Dialog that is going to know if a Select menu is open. Without a simple whitelist/blacklist (looking for target elements) and having to use a hook-type behavior it's probably going to complicate the internals of other components - e.g., if we had to do this via hook then a component might now have to set an onFocus/onBlur to set that the Dialog should not close b/c the SelectInput is focused and the menu is open.

chaance:
@bhowell2-fleetio I hear you, and to be perfectly honest I haven't had a chance to look deeply into alternative approaches that don't rely on handling this event during the capture phase.

I think your best bet for now is to use `patch-package` or some other mechanism to patch this temporarily to meet your needs. We'll need to discuss the right approach here internally and get some consensus on the right patterns before moving forward, and unfortunately we have some competing priorities at the moment that will delay any immediate action. I've got this issue flagged and will follow up when we have a chance to consider it more deeply.

bhowell2-fleetio:
@chaance thanks again and understood. Yeah we've just avoided the upgrade to the Dialog version that has this capture. Just got around to "fixing" it here, but totally understand this likely isn't the correct approach for the library overall. Might patch this locally and use that for now or may keep chugging along with the previous version for now. I'll close this out.

PR #3441, opened 2025-04-02
Slottable: Improve displayName for easier debugging

chaance:
Something that has bugged me for a while is how Radix components add a lot of noise to the component tree view in React dev tools. This largely has to do with our use of the `Slot` component, which results in a view that is hard to grok and makes it more annoying to debug with so many components just named `Slot`.

CleanShot 2025-04-02 at 14 32 42

This PR introduces functions to create `Slot` and `Slottable` components with unique display names to make things a little less chaotic.

CleanShot 2025-04-02 at 14 32 25

In the future I'd like to propose eliminating internal usage of `Slot` for a functional approach to actually clean up the component tree as much as possible. This is an interim solution that avoids major refactoring.

### DIFF
--- a/packages/react/slot/src/slot.tsx
+++ b/packages/react/slot/src/slot.tsx
@@ -117,7 +117,10 @@ function isSlottable(
   child: React.ReactNode
 ): child is React.ReactElement<SlottableProps, typeof Slottable> {
   return (
-    React.isValidElement(child) && '__radixId' in child && child.__radixId === SLOTTABLE_IDENTIFIER
+    React.isValidElement(child) &&
+    typeof child.type === 'function' &&
+    '__radixId' in child.type &&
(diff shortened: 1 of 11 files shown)


packages/react/slot/src/slot.tsx
```diff
@@ -57,40 +63,65 @@ interface SlotCloneProps {
...
-    return React.cloneElement(children, props);
-  }
 
-  return React.Children.count(children) > 1 ? React.Children.only(null) : null;
-});
+    return React.Children.count(children) > 1 ? React.Children.only(null) : null;
+  });
 
-SlotClone.displayName = 'SlotClone';
+  SlotClone.displayName = `${ownerName}.SlotClone`;
```
ajworkos, line 83: 👍

ajworkos (approved):

PR #3454, opened 2025-04-10
ci(.github/workflows): simplify 'pnpm install' command

sukvvon:
### Description

* In CI `pnpm install` means `pnpm install --frozen-lockfile`, so i simplify it.
  * https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data


sukvvon:
@chaance 

Could you review this PR?

PR #3455, opened 2025-04-10
Minor `useControllableState` improvements

chaance:


packages/react/use-controllable-state/src/use-controllable-state.tsx
```diff
@@ -1,57 +1,96 @@
...
-function useControllableState<T>({
+export function useControllableState<T>({
   prop,
   defaultProp,
   onChange = () => {},
-}: UseControllableStateParams<T>) {
-  const [uncontrolledProp, setUncontrolledProp] = useUncontrolledState({ defaultProp, onChange });
+  caller,
+}: UseControllableStateParams<T>): [T, SetStateFn<T>] {
+  const [uncontrolledProp, setUncontrolledProp, onChangeRef] = useUncontrolledState({
```
chaance, line 24: Returning the saved callback ref removes the need to double-save it internally, saving a needless effect cycle.

packages/react/use-controllable-state/src/use-controllable-state.tsx
```diff
@@ -1,57 +1,96 @@
...
+  Value: T,
+  setValue: React.Dispatch<React.SetStateAction<T>>,
+  OnChangeRef: React.RefObject<ChangeHandler<T> | undefined>,
+] {
+  const [value, setValue] = React.useState(defaultProp);
   const prevValueRef = React.useRef(value);
-  const handleChange = useCallbackRef(onChange);
+
+  const onChangeRef = React.useRef(onChange);
+  useInsertionEffect(() => {
```
chaance, line 80: Using `useInsertionEffect` (or pre-19: `useLayoutEffect`) ensures we save the callback as early as possible. There may be some edge cases where the reference is updated outside of the React lifecycle for some reason, but that would be the case with our current implementation as well.

Issue #3365, opened 2025-02-06
Ability to opt-out of `BubbleInput` / `BubbleSelect`


willo-icon:
### Overview

There are a lot of feature requests related to the inconsistent and inaccessible rendering of `<input />` and `<select />` elements inside of components like `<Checkbox />` and `<Select />` (#874, #3356, #1333, #2987, #1270 to list a few) and while I look forward to the upcoming refactors that have been mentioned in many of those issues, I'd like to suggest / request an alternative approach. A great way to work around this would be to simply render your own `<input />` or `<select />` components so that you can choose how to manage them. This should be pretty straight forward by just putting your own children in there, but the problem comes from the selectively rendered `<BubbleInput />` and `<BubbleSelect />` components. Because there's no deterministic way to stop these from rendering, you could end up with two conflicting input elements on your components.

As a stop-gap for all of these problems, I would propose a new `disableNativeInput?: boolean` prop that lets prevent rendering the `Bubble*` components. As far as I can tell this should not negatively impact the components, since they already handle not having the native elements when placed outside of a form.

### Examples in other libraries

This is somewhat of a Radix-specific issue so I don't know if there are relevant links, but the majority of component libraries that have `Checkbox` or `Select` components (e.g. MUI, React Aria) either consistently render a real `<input />` or `<select />` under the hood, or provide an option to use native elements.

### Who does this impact? Who is this for?

As a new optional prop, this shouldn't impact anyone. This is for people who want to leverage the excellent Radix components without sacrificing the ability to use native input elements.

### Additional context

Radix is dope.

chaance:
I've got several PRs to complete this for all of our form control components. Closing since these will all get done by the next release.

- https://github.com/radix-ui/primitives/pull/3901
- https://github.com/radix-ui/primitives/pull/3904
- https://github.com/radix-ui/primitives/pull/3902

PR #3457, opened 2025-04-10
Use primitives for bubble input components

chaance:
A long-standing issue is the lack of control users have over the internal "bubble input" components we render in our form controls. Because these components are rendered as implementation details rather than exposed directly, users struggle to access or control them when they cause unexpected behaviors.

Ultimately we want users to take control of these components, and even opt-out these implementations altogether when conflicts arise. Related: https://github.com/radix-ui/primitives/issues/3365.

This PR in a small first step in that direction. Each primitive's `BubbleInput` will now use a `Primitive` node rather than the built-in JSX DOM component, and accepts forwarded refs to match the behavior of other component parts.

### DIFF
--- a/packages/react/select/src/select.tsx
+++ b/packages/react/select/src/select.tsx
@@ -19,7 +19,7 @@ import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
 import { useControllableState } from '@radix-ui/react-use-controllable-state';
 import { useLayoutEffect } from '@radix-ui/react-use-layout-effect';
 import { usePrevious } from '@radix-ui/react-use-previous';
-import { VisuallyHidden } from '@radix-ui/react-visually-hidden';
+import { VISUALLY_HIDDEN_STYLES } from '@radix-ui/react-visually-hidden';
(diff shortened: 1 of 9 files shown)


packages/react/select/src/select.tsx
```diff
@@ -1659,26 +1663,35 @@ const BubbleSelect = React.forwardRef<HTMLSelectElement, React.ComponentPropsWit
...
+     * We use visually hidden styles rather than `display: "none"` because
+     * Safari autofill won't work otherwise.
      */
     return (
-      <VisuallyHidden asChild>
-        <select {...selectProps} ref={composedRefs} defaultValue={value} />
-      </VisuallyHidden>
+      <Primitive.select
+        {...props}
+        style={{ ...VISUALLY_HIDDEN_STYLES, ...props.style }}
```
chaance, line 1680: Very minor change, but if we can just expose the style object it removes a tiny bit of overhead and excess noise in the component tree (which is nicer for dev tools IMO)

lucasmotta (approved):

Issue #3356, opened 2025-02-02
onChange event on checkbox that exposes the input synthetic event


spassvogel:
I'm using the Checkbox in a form together with server actions in NextJs. The form lives on a server component, the Checkbox is in a client child component.

What I would like to do is submit the form as soon as one of the values change, using the input type="checkbox"'s synthetic event to access the containing form,  like so.

`<Checkbox name="tos" onChange={(e) => e.currentTarget.form?.requestSubmit()} />`

However, the only event the Checkbox seems to have is a `onCheckedChange` with a boolean, which is quite limited

Issue #2987, opened 2024-07-02
keyPressed event handler for select component


shahinghasemi:
Hello. is it possible to expose an `keyPressed`  event handler for `Select.Content` ? 
This is the scenario why it can be usefull:

I've added a prop to the `Select.Content` as `Header` which is a react component and I've passed an `input` to call my api and search through the provided input(Something like combobox). However when The content is open, when any key  in input is pressed  it first searches through the available items. What I want is to prevent the `Content` to search through the available items instead send the provided input to my API and get the new items. 

![Screenshot from 2024-07-02 13-17-42](https://github.com/radix-ui/primitives/assets/72931905/fc627241-8acb-405a-9277-77179389455b)

Issue #1333, opened 2022-04-15
Add native checkbox element


Haberkamp:
### Overview

I'm creating this issue because it currently is not possible to change the state of the checkbox component when javascript is disabled. For example this allows these user to opt-out of something like a registration. Because this does not work I suggest that the button element should be swapped out with a native html checkbox element.

The html checkbox element also provides an `indeterminate` attribute to cover this case as well.
### Examples in other libraries

The react library [Mantine](https://mantine.dev/core/checkbox/) does this where they use an input element and add the `type="checkbox"` attribute to it.

### Who does this impact? Who is this for?

This would be a break for users that either directly access the button element with Javascript via the DOM-API as well as for users that select the button element via css like:

```css
.checkbox button {
  /* some styles */
}
```

benoitgrelard:
Hey @Haberkamp,

The fact it isn't a checkbox under the hood is an implementation detail that shouldn't matter.
We do add a hidden native checkbox in cases where the checkbox is used in a form for example, but that's only to handle change events bubbling in forms.

I'm not clear what your issues are.

Haberkamp:
> @benoitgrelard it sounds like another progressive enhancement request.

Yes that exactly what I meant.

henrik-stahl:
> The fact it isn't a checkbox under the hood is an implementation detail that shouldn't matter.

Why did you decide to implement a checkbox that isn't a checkbox under the hood, though? Just curious to hear the reasoning behind this decision. :)

pixeliner:
> Because of this reason when using react-hook-form and uncontrolled input - register returns On/Off instead of boolean

Just because of this, I created my own CheckBox component. I can't think of a reasoning why On/Off are being used while a simple boolean could have suficed. I mean even indeterminate should be a false state and be controlled internally like any other component library solved

PR #3459, opened 2025-04-10
checkbox: Add unstable `Provider`, `Trigger` and `BubbleInput`

chaance:
This PR introduces new parts for the Checkbox component. This change is intended to test out a new pattern that we can replicate for other form controls.

Currently our form control components will try to detect whether or not they are associated with a form, and if so, render a hidden input so that native events bubble to the form itself. While this can make behavior feel more like native form elements, the downside is that it adds an additional element that is hard to control and not always what users expect.

The goal is the hand control back to the user. So instead of `Checkbox.Root` rendering both a provider and an input, users can opt in and render the parts that they want.

```tsx
// before
<Checkbox.Root>
  <Checkbox.Indicator />
</Checkbox.Root>

// after
<Checkbox.Provider>
  <Checkbox.Trigger>
    <Checkbox.Indicator />
  </Checkbox.Trigger>
</Checkbox.Provider>
```

Currently these parts are prefixed with `unstable_` since we may want to consider the API more deeply before committing to these patterns. Should we decide to move forward, in a future version we'd probably rename `Checkbox.Provider` to `Checkbox.Root` and get rid of the pre-composed version entirely.

Related:

- #3365
- #874
- #3356
- #1333
- #2987

### DIFF
--- a/packages/react/checkbox/src/checkbox.test.tsx
+++ b/packages/react/checkbox/src/checkbox.test.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react';
 import { axe } from 'vitest-axe';
 import type { RenderResult } from '@testing-library/react';
-import { cleanup, render, fireEvent } from '@testing-library/react';
-import * as Checkbox from './checkbox';
+import { cleanup, render, fireEvent, screen, waitFor, act } from '@testing-library/react';
+import * as Checkbox from '.';
 import { afterEach, describe, it, beforeEach, vi, expect } from 'vitest';
 
 const CHECKBOX_ROLE = 'checkbox';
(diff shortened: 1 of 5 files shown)


packages/react/checkbox/src/checkbox.test.tsx
```diff
@@ -20,184 +20,464 @@ global.ResizeObserver = class ResizeObserver {
...
+      const toggleButton = screen.getByText('Toggle checkbox');
+      act(() => fireEvent.click(toggleButton));
+      expect(onChange).toHaveBeenCalledWith(false);
+    });
   });
 });
 
-describe('given a controlled Checkbox in a form', () => {
-  afterEach(cleanup);
+describe('Legacy Checkbox', () => {
```
chaance, line 301: Left these tests unchanged but moved them into a nested describe block, which is why the diff looks so big.

PR #3464, opened 2025-04-13
Add new `PasswordToggleField` primitive

chaance:
See Changeset for details

### DIFF
--- a/packages/react/password-toggle-field/src/password-toggle-field.tsx
+++ b/packages/react/password-toggle-field/src/password-toggle-field.tsx
@@ -276,8 +276,8 @@ const PasswordToggleFieldToggle = React.forwardRef<
 });
 
 interface PasswordToggleFieldSlotDeclarativeProps {
-  visible: React.ReactElement;
-  hidden: React.ReactElement;
+  visible: React.ReactNode;
+  hidden: React.ReactNode;
 }
 
 interface PasswordToggleFieldSlotRenderProps {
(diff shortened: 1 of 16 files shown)


packages/react/password-toggle-field/src/password-toggle-field.tsx
```diff
@@ -0,0 +1,479 @@
...
+    }, [idProp, syncInputId]);
+
+    // We want to reset the visibility to `false` to revert the input to
+    // `type="password"` when:
+    // - The form is reset (for consistency with other form controls)
+    // - The form is submitted (to prevent the browser from remembering the
+    //   input's value.
+    //
+    // See "Keeping things secure":
+    //   https://technology.blog.gov.uk/2021/04/19/simple-things-are-complicated-making-a-show-password-option/)
```
dcwither, line 143: 🫢

packages/react/password-toggle-field/src/password-toggle-field.tsx
```diff
@@ -0,0 +1,479 @@
...
+        onClick={(event) => {
+          onClick?.(event);
+          if (event.defaultPrevented) {
+            focusState.current = INITIAL_FOCUS_STATE;
+            return;
+          }
+
+          flushSync(() => {
+            setVisible((s) => !s);
+          });
```
dcwither, line 326: why does this one need `flushSync`? I could see why it might be necessary for the form onSubmit, but having a hard time understanding the use here.

dcwither (commented):
Wow, lots of considerations! Nice work, just a few questions.

packages/react/password-toggle-field/src/password-toggle-field.tsx
chaance: It just ensures that the state is updated before moving focus. It's possible for the toggle button's DOM node to get blown away and replaced when this state changes depending on how the consumer renders it, so not flushing the update could focus an element right before it leaves the DOM.

PR #3472, opened 2025-04-15
fix: click callback inside `SelectItem` shouldn't be omitted

### LINKED ISSUE #3460: `<a/>` element in SelectItem is not accessable
### Current Behavior

We have tooltip in SelectItem, which includes a link to document, but it's not clickable inside SelectItem for some reasons.

### Expected behavior

Should clickable

### Reproducible example

https://stackblitz.com/edit/vitejs-vite-v5kpt7vv?file=src%2FDemo.tsx

himself65:
Fixes: https://github.com/radix-ui/primitives/issues/3460

Wrap with `startTransition` will allow consecutive logic can be handled correctly

### DIFF
--- a/packages/react/select/src/select.tsx
+++ b/packages/react/select/src/select.tsx
@@ -1269,8 +1269,13 @@ const SelectItem = React.forwardRef<SelectItemElement, SelectItemProps>(
 
     const handleSelect = () => {
       if (!disabled) {
-        context.onValueChange(value);
-        context.onOpenChange(false);
+        const startTransition = React.startTransition
+          ? React.startTransition
+          : window.requestAnimationFrame;
+        startTransition(() => {
+          context.onValueChange(value);
+          context.onOpenChange(false);
+        });
       }
     };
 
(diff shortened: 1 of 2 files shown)


packages/react/select/src/select.tsx
```diff
@@ -1269,8 +1269,13 @@ const SelectItem = React.forwardRef<SelectItemElement, SelectItemProps>(
...
       if (!disabled) {
-        context.onValueChange(value);
-        context.onOpenChange(false);
+        const startTransition = React.startTransition
+          ? React.startTransition
+          : window.requestAnimationFrame;
+        startTransition(() => {
+          context.onValueChange(value);
+          context.onOpenChange(false);
+        });
```
SukkaW, line 1278: This seems wrong. The value change handler should have the highest priority (higher than render/paint) to quickly respond to the user interaction. Thus, you can't use `requestAnimationFrame` here.

The `startTransition` also feels wrong, as this schedules the value change to the next re-render.

PR #3477, opened 2025-04-16
[Slot] use stable composed refs in SlotClone

radnan:
### Description

right now we are passing an anonymous function (returned from `composeRefs()` as `ref` to cloned slot. but this causes the ref callback to re-trigger on every render since react is seeing a new function being passed on every new render. so we replace the anonymous function with the stable `useComposedRefs` hook instead which uses `useCallback` under the hood and only re-creates the callback when the arguments change.

added two tests - one using native `<input />` and another using `Slot` to showcase how both should behave the same.

### DIFF
--- a/packages/react/slot/src/slot.test.tsx
+++ b/packages/react/slot/src/slot.test.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
 import { cleanup, render, screen, fireEvent } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { Slot, Slottable } from './slot';
 import { afterEach, describe, it, beforeEach, vi, expect } from 'vitest';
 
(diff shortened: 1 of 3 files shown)


packages/react/slot/src/slot.test.tsx
```diff
@@ -160,3 +191,24 @@ const Button = React.forwardRef<
...
 });
+
+const Input = React.forwardRef<
+  React.ElementRef<'input'>,
+  React.ComponentProps<'input'> & {
+    asChild?: boolean;
+  }
+>(({ asChild, children, ...props }, forwardedRef) => {
+  const Comp = asChild ? Slot : 'input';
+  const [value, setValue] = useState('');
```
radnan, line 202: using controlled value here to force component to re-render as value is typed in the input

chaance (approved):

packages/react/slot/src/slot.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { composeRefs } from '@radix-ui/react-compose-refs';
+import { useComposedRefs } from '@radix-ui/react-compose-refs';
```
radnan, line 2: Fixed lint issue @chaance

Issue #3424, opened 2025-03-20
[DropdownMenu] Cannot open the menu with a ref


mleralec:
## Bug report

On the [DropdownMenu](https://www.radix-ui.com/primitives/docs/components/dropdown-menu), we cannot trigger the menu by clicking the trigger with a ref (like `triggerRef.current.click()`).
It's because the DropdownMenuTrigger component use a `onPointerDown` instead of a `onClick` callback, as you can see [there](https://github.com/radix-ui/primitives/blob/main/packages/react/dropdown-menu/src/dropdown-menu.tsx#L116).
I tested on the other primitives (Dialog, Popover) and it works well with a ref so that's why it's a bug for me.

### Current Behavior

```tsx
<DropdownMenu.Root>
  <DropdownMenu.Trigger ref={triggerRef}>Trigger</DropdownMenu.Trigger>
  <DropdownMenu.Content>
    <DropdownMenu.Item>Item A</DropdownMenu.Item>
    <DropdownMenu.Item>Item B</DropdownMenu.Item>
    <DropdownMenu.Item>Item C</DropdownMenu.Item>
  </DropdownMenu.Content>
</DropdownMenu.Root>
```
Based on this code 🔝, the snippet `triggerRef.current.click()` won't open the menu

### Expected behavior

Clicking the ref opens the dropdown menu

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/p/sandbox/great-swanson-w8xmsj)

### Additional context

The workaround I did is to rebind the `onClick` on the DropdownMenu.Trigger:
```tsx
const [open, setOpen] = useState(false)

return (
  <DropdownMenu.Root open={open} onOpenChange={setOpen}>
    <DropdownMenu.Trigger 
      onPointerDown={event => {
        event.preventDefault()
      }}
      onClick={() => {
        setOpen(!open)
      }}
    >
      Trigger
    </DropdownMenu.Trigger>
    <DropdownMenu.Content>
      <DropdownMenu.Item>Item A</DropdownMenu.Item>
      <DropdownMenu.Item>Item B</DropdownMenu.Item>
      <DropdownMenu.Item>Item C</DropdownMenu.Item>
    </DropdownMenu.Content>
  </DropdownMenu.Root>
)
```

### Your environment

It's reproducible everywhere at the moment

---

I'll be happy to help if needed 😃

jordan-sussman:
[...]

PR #3481, opened 2025-04-17
[DropdownMenu] Replace onPointerDown handler by onClick

mleralec:
Fixes https://github.com/radix-ui/primitives/issues/3424

### Description

This pull request will fix the issue https://github.com/radix-ui/primitives/issues/3424
I'm not sure about the usage of the `onPointerDown` on the DropdownMenu since the `onClick` handler is used on all radix-ui components with trigger (Dialog, Popover)

I keep the "prevent trigger focusing when opening" behaviour but removed the "only call handler if it's the left button (mousedown gets triggered by all mouse buttons)" behaviour by using the `onClick` instead of the `onPointerDown`


chaance:
This is by design. We probably want a ref to track whether or not the menu is triggered already by the pointer event and toggling in the click handler if not.

Issue #3479, opened 2025-04-17
useSyncExternalStore breaks react 17


RodBrowning:
export function useIsHydrated() {
   return useSyncExternalStore(
     subscribe,
     () => true,
     () => false
   );
 }

chaance:
Please try updating to the latest version and let me know if you still have any issues.

PR #3482, opened 2025-04-17
fix: use internal useSyncExternalStore instead of the React one

mleralec:
Fixes https://github.com/radix-ui/primitives/issues/3479
Use internal hook `packages/react/use-is-hydrated/src/use-is-hydrated.tsx` instead of the `useSyncExternalStore` from React (it breaks on React version < 18)


chaance:
Already patched in the latest release

Issue #3489, opened 2025-04-20
use-sync-external-store/shim/index.js breaking in latest related


davecarlson:
### Current Behavior
Cannot buid application

### Expected behavior
Build application

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/2r30e)

### Suggested solution
Had to downgrade @radix-ui/react-avatar to 1.1.3 from 1.1.6

### Additional context

I believe this causes it: 
https://github.com/radix-ui/primitives/commit/74cb95a787a302d68e970f16acb3fdec8b2154ad

### Your environment
```
  "@radix-ui/react-avatar": "^1.1.6",
    "@radix-ui/react-checkbox": "^1.2.2",
    "@radix-ui/react-dialog": "^1.1.10",
    "@radix-ui/react-dropdown-menu": "^2.1.11",
    "@radix-ui/react-icons": "^1.3.2",
    "@radix-ui/react-label": "^2.1.4",
    "@radix-ui/react-navigation-menu": "^1.2.9",
    "@radix-ui/react-popover": "^1.1.10",
    "@radix-ui/react-progress": "^1.1.4",
    "@radix-ui/react-select": "^2.2.2",
    "@radix-ui/react-separator": "^1.1.4",
    "@radix-ui/react-slot": "^1.2.0",
    "@radix-ui/react-switch": "^1.2.2",
    "@radix-ui/react-tabs": "^1.1.8",
    "@radix-ui/react-tooltip": "^1.2.3",

```

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |         |         |
| React            | n/a     |         |
| Browser          |         |         |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn/pnpm    |         |         |
| Operating System |         |         |

tmcw:
Getting the same issue trying to run tests with vitest. The line, afaict:

https://github.com/radix-ui/primitives/blob/f2e6c688035d880a832d44755d31d39ba87c1d35/packages/react/use-is-hydrated/src/use-is-hydrated.tsx#L1

It looks like use-sync-external-store _does_ have an exports field in package.json:

https://github.com/facebook/react/blob/bc6184dd993e6ea0efdee7553293676db774c3ca/packages/use-sync-external-store/package.json#L5-L20

But that was added in 1.5.0, and use-sync-external-store has a semver spec of 1.4.0 and up:

[...]

PR #3491, opened 2025-04-21
deps: Bump dependency on use-sync-external-store

tmcw:
This PR bumps the semver range for use-sync-external-store to ensure that the installed version has an `exports` field in package.json so that it can be successfully imported

### Description

Details from the related issue #3489:

```
 Error: Directory import '/home/runner/work/val.town/val.town/node_modules/use-sync-external-store/shim' is not supported resolving ES modules imported from /home/runner/work/val.town/val.town/node_modules/@radix-ui/react-use-is-hydrated/dist/index.mjs
Did you mean to import "use-sync-external-store/shim/index.js"?
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { code: 'ERR_UNSUPPORTED_DIR_IMPORT', url: 'file:///home/runner/work/val.town/val.town/node_modules/use-sync-external-store/shim' }
```

Getting the same issue trying to run tests with vitest. The line, afaict:

https://github.com/radix-ui/primitives/blob/f2e6c688035d880a832d44755d31d39ba87c1d35/packages/react/use-is-hydrated/src/use-is-hydrated.tsx#L1

It looks like use-sync-external-store _does_ have an exports field in package.json:

https://github.com/facebook/react/blob/bc6184dd993e6ea0efdee7553293676db774c3ca/packages/use-sync-external-store/package.json#L5-L20

But that was added in 1.5.0, and use-sync-external-store has a semver spec of 1.4.0 and up:

https://github.com/radix-ui/primitives/blob/f2e6c688035d880a832d44755d31d39ba87c1d35/packages/react/use-is-hydrated/package.json#L38

So this probably needs to be bumped to 1.5.0.

### DIFF
--- a/.changeset/ripe-rockets-hide.md
+++ b/.changeset/ripe-rockets-hide.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-use-is-hydrated': minor
+---
+
+Update the dependency specification for use-sync-external-store
--- a/packages/react/use-is-hydrated/package.json
+++ b/packages/react/use-is-hydrated/package.json
@@ -35,7 +35,7 @@
     "build": "radix-build"
   },
   "dependencies": {
-    "use-sync-external-store": "^1.4.0"
+    "use-sync-external-store": "^1.5.0"
   },
   "peerDependencies": {
     "@types/react": "*",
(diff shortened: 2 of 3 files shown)


chaance (approved):

tmcw:
Whoops, forgot to bump the pnpm lockfile, just pushed that.

PR #3500, opened 2025-04-25
Popper: Fix infinite re-renders

raymondkneipp:
Fixes the following error:

Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render. Stack: React 4 PopperAnchor popper.tsx:84

### Description

Add context.onAnchorChange to useEffect deps array to prevent infinite re-renders


nayounsang:
`close #3546`

c-ehrlich:
Can confirm these changes fix the issue for us

chaance:
Should be addressed here: https://github.com/radix-ui/primitives/pull/3614

PR #3426, opened 2025-03-22
refactor(react): replace deprecated 'ElementRef' with 'ComponentRef'

sukvvon:
### Description

image

* `ElementRef` is now deprecated, so i replaced `ElementRef` with `ComponentRef`

PR #3506, opened 2025-04-28
refactor: replace ElementRef to ComponentRef in dialog

nayounsang:
### Description

- `ElementRef` is deprecated but used in `dialog`.
- Replace `ElementRef` to `ComponentRef`.
- To limit to DOM elements, have to use `HTMLxxxElement` directly, but I think that's the most appropriate way for now, logically.
- There is no logical difference, as I simply converted the old type to a modern one.


nayounsang:
dupe #3426

PR #3516, opened 2025-04-29
feat(DocumentContext): document provider and hook

chaance:
Wrapping up https://github.com/radix-ui/primitives/pull/3357


packages/react/avatar/src/avatar.tsx
```diff
@@ -100,16 +100,12 @@ const AvatarFallback = React.forwardRef<AvatarFallbackElement, AvatarFallbackPro
...
-    const providedDocument = useDocument();
-    const documentWindow = providedDocument?.defaultView;
-
     React.useEffect(() => {
-      if (!documentWindow) return;
       if (delayMs !== undefined) {
-        const timerId = documentWindow.setTimeout(() => setCanRender(true), delayMs);
-        return () => documentWindow.clearTimeout(timerId);
+        const timerId = globalThis.window.setTimeout(() => setCanRender(true), delayMs);
+        return () => globalThis.window.clearTimeout(timerId);
```
arturbien, line 106: @chaance afaik browsers throttle setTimeout and setInterval  by 1000ms when browser tab is inactive or hidden. So if the timeout is set on the parent window, and user opens Document PIP from it with an avatar in it, then navigates to a different tab or app, these timeouts might get delayed or paused. that's why I've attached all timeouts to the provided document window

fannychan:
Hi @chaance! Is still something that will be worked on or has a new solution come up? 🙏

chaance:
Decided to close this for now and reintroduce in v2, since this feels worthy of a major change.

PR #3518, opened 2025-04-30
refactor: resolved incorrect accessibility from Select

### LINKED ISSUE #3636: Select: Implement `aria-activedescendant` for improved accessibility
### Current Behavior

The SelectContent with `role="listbox"` **lacks the `aria-activedescendant` attribute**. Instead, the `aria-selected` attribute on each SelectItem element is incorrectly being set based on both its selection and **focus** state.

### Expected behavior

[The `role="listbox"` element **should possess an `aria-activedescendant` attribute**](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role#states_and_properties). This attribute's value must dynamically update to the `id` of the `option` element that is currently receiving virtual focus as the user navigates the listbox via keyboard.

Crucially, the `aria-selected` attribute on `option` elements should **only** reflect whether an item is actually selected, independent of its focus state.

### Suggested solution

To address this, the `listbox` component's logic needs adjustment:

1. **Add `aria-activedescendant`**: Include the `aria-activedescendant` attribute on the element with `role="listbox"`.
    
2. **Dynamic Updates**: Implement logic to dynamically update the `aria-activedescendant` attribute's value to the `id` of the currently "virtually focused" `option` whenever a user navigates using keyboard input (e.g., arrow keys).
    
3. **Correct `aria-selected` Usage**: Ensure that the `aria-selected` attribute on `option` elements solely indicates a selection state and is not influenced by focus.
    

### Your environment

|Software|Name(s)|Version|
|---|---|---|
|Radix Package(s)|||
|React|n/a||
|Browser|||
|Assistive tech|||
|Node|n/a||
|npm/yarn/pnpm|||
|Operating System|||

seongminn:
closed #3636

### Description

- I removed isFocused from the condition that determines the aria-selected attribute in the SelectItem component.
- Added `aria-activedescendant` from SelectContent

### DIFF
--- a/.changeset/early-books-tap.md
+++ b/.changeset/early-books-tap.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-select': patch
+---
+
+resolved incorrect accessibility from Select
--- a/packages/react/select/src/select.tsx
+++ b/packages/react/select/src/select.tsx
@@ -1302,7 +1302,7 @@ const SelectItem = React.forwardRef<SelectItemElement, SelectItemProps>(
             aria-labelledby={textId}
             data-highlighted={isFocused ? '' : undefined}
             // `isFocused` caveat fixes stuttering in VoiceOver
-            aria-selected={isSelected && isFocused}
+            aria-selected={isSelected}
             data-state={isSelected ? 'checked' : 'unchecked'}
             aria-disabled={disabled || undefined}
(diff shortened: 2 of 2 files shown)


nathanael-ruf:
@chaance Is there a person we can tag to have this reviewed?

PR #3522, opened 2025-05-01
Implement multi select

### LINKED ISSUE #1270: [Select] Support multiple selections

Zach-Jaensch:
### Description
closes: #1270 

Implements the `multiple` option for Select.
image

Changes the form implementation (SelectBubbleInput) to update the option selected prop, instead of the select value prop. this is how the multiple select maintains the value, instead of having an array implementation in the select value. The single select also supports this, and as shown in the added tests, will still update the form as needed.

Externally, the prop type are more limiting then the HTMLSelectElement as it allows value to be an array on single, and string on multiple, however, react generates an error for these cases. 

Internally, not have complex typing and type checks throughout the component, I've elected instead to do some basic checks and assumptions.

I added some basic tests to the Select component. Ideally I would use `userEvent` instead of `fireEvent`, however, due to the many event handlers and functions missing from jsdom, I've left the tests as very basic tests. Perhaps using vitest browser mode could help with testing the complex components like Select?

Thank you for taking a look at this PR. Any and all feedback is welcome!

### DIFF
--- a/.changeset/quiet-schools-leave.md
+++ b/.changeset/quiet-schools-leave.md
@@ -1,0 +1,6 @@
+---
+"@repo/storybook": patch
+"@radix-ui/react-select": minor
+---
+
+Support `multiple` prop and array `value` and `defaultValue` on Select (#3522)
--- a/apps/storybook/stories/select.stories.tsx
+++ b/apps/storybook/stories/select.stories.tsx
@@ -59,6 +59,53 @@ export const Styled = () => (
   </div>
 );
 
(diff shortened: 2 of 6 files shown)


Zach-Jaensch:
Hey @chaance and team! My work has been somewhat delayed by not having this in the library, and we have been pushing that task down the road. I was wondering if there is a timeline for having this reviewed? I see it has been self-assigned some time ago. It would be great to know if we should wait or move on to other workarounds.

martpie:
Gentle poke :)

PR #3551, opened 2025-05-19
fix: [One-Time Password Field] truncate values ​​that exceed the length when pasting

### LINKED ISSUE #3550: [OneTimePasswordField] Pasting Input Exceeds Maximum Length
### Current Behavior

Regardless of how many `<OneTimePasswordField.Input>` elements are rendered, if you copy and paste text into the field, all of the pasted characters are inserted, even beyond the intended length.

### Expected behavior

Any characters beyond the configured number of inputs should be truncated—excess text should not be added to the field.

### Reproducible example

1. Navigate to the Radix UI One Time Password Field demo:
   [https://www.radix-ui.com/primitives/docs/components/one-time-password-field](https://www.radix-ui.com/primitives/docs/components/one-time-password-field)
2. Copy a string of more than 6 digits.
3. Click into the OTP field and paste.
4. Observe that all digits appear, overflowing the intended 6-character limit.

### Suggested solution

On value change (e.g. in the `onValueChange` handler), verify the new value’s length and slice it to the maximum allowed length before updating state.

### Your environment

| Radix Package(s) |@radix-ui/react-one-time-password-field|^0.1.6|
| React            | n/a     |^19.1.0|
| npm/yarn/pnpm    |pnpm|10.11.0|

nayounsang:
### Description
close #3550
- truncate values exceed the length of collections when pasting
- There is a concern that SRP may be violated when inserting the corresponding logic into `sanitizeValue`, and therefore, side effects are a concern, so verification is performed in the paste logic.
- Test code is skipped because the pasted values ​​of `@testing-library/user-event` are not assigned to input.

**Test with Storybook**
Case: paste `124868392`
```html
<input autocomplete="off" autocapitalize="off" autocorrect="off" autosave="off" spellcheck="false" readonly="" type="hidden" 
value="124868" 
name="code">
```
image

### DIFF
--- a/.changeset/famous-icons-arrive.md
+++ b/.changeset/famous-icons-arrive.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-one-time-password-field': patch
+---
+
+truncate values ​​that exceed the length when pasting
--- a/packages/react/one-time-password-field/src/one-time-password-field.test.tsx
+++ b/packages/react/one-time-password-field/src/one-time-password-field.test.tsx
@@ -64,6 +64,17 @@ describe('given a default OneTimePasswordField', () => {
     await act(async () => await user.paste('1,2,3,4,5,6'));
     expect(getInputValues(inputs)).toBe('1,2,3,4,5,6');
   });
+
+  // TODO: userEvent paste not behaving as expected. Debug and unskip.
+  it.todo('should truncate pasted characters to the number of inputs', async () => {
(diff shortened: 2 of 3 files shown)


chaance:
Thanks! Mind adding a unit test to cover this case?

nayounsang:
@chaance I add **todo** test because `user.paste` not linked to input data. This is same to the paste test above.

PR #3581, opened 2025-05-28
refactor(react/use-callback-ref): replace 'useMemo' with 'useCallback'

sukvvon:


sukvvon:
@chaance Could you review this PR?

Issue #3560, opened 2025-05-21
aria-controls id on Dialog.Trigger is invalid while the dialog is not open


nathanael-ruf:
### Current Behavior

The trigger is assigned a `aria-controls` id that doesn't exist in the DOM until the user clicks on the trigger. IBM Equal Access Accessibility Checker reports
```
The 'id' "radix-:r18:" specified for the ARIA property 'aria-controls' value is not valid
The ARIA property must reference a non-empty unique id of an existing element that is visible
```

### Expected behavior

No invalid ids

### Suggested solution

Not sure, perhaps just omit `aria-controls` here?

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) | @radix-ui/react-dialog        | 1.1.11  |
| React            | n/a     | 19.0.0        |
| Browser          | Chrome        | 136.0.7103.114         |
| Assistive tech   | IBM Equal Access Accessibility Checker | 4.0.4        |
| Node             | n/a     | 22.15.0        |
| npm/yarn/pnpm    | yarn        | 4.9.1        |
| Operating System | MacOS        | 15.5        |

gnarlex:
@chaance could you link to the PR / change that addresses this? Since you closed this as completed

PR #3591, opened 2025-06-10
refactor: adjust aria-controls to triggers according to conditions

seongminn:
### Description

related #3560

Applied the aria-controls attribute conditionally to trigger components. 
The aria-controls attribute points to the id of elements whose content or presence is controlled by the trigger. [reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls)

The key consideration is that the referenced id must always exist in the DOM.
The dropdown-menu component already implements this approach, but the other components did not follow this pattern, so I have updated them accordingly.

Modified components:
- dialog
- menu
- navigation-menu
- popover
- select

p.s) The collapsible and tabs components were not modified because the elements referenced by their aria-controls attributes are always present in the DOM.

### DIFF
--- a/.changeset/brown-falcons-battle.md
+++ b/.changeset/brown-falcons-battle.md
@@ -1,0 +1,9 @@
+---
+'@radix-ui/react-navigation-menu': patch
+'@radix-ui/react-popover': patch
+'@radix-ui/react-dialog': patch
+'@radix-ui/react-select': patch
+'@radix-ui/react-menu': patch
+---
+
+adjust aria-controls to triggers according to conditions.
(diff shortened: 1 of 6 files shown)


Jacksonmills:
looks good hope we can get this fixed asap!

nathanael-ruf:
Is there anything we can do to get this approved, merged and released?
Currently this requires a bunch of hacks/ignores to keep our a11y CI happy.

seongminn:
@nathanael-ruf Sorry, I have no idea 🥲

@chaance Could you please review this request?

adamgreg:
I just encountered #3560 . I hope this PR gets picked up soon. Good work @seongminn !

PR #3595, opened 2025-06-13
feat(tooltip): add isLabelledBy prop for screen reader label support

alizehkhan:
### Description

Accessibility improvement: Adds an optional `isLabelledBy` prop to the `TooltipTrigger` component, allowing the tooltip content to be announced by screen readers as a label (via `aria-labelledby`) instead of the default description (`aria-describedby`). This is useful when the tooltip should serve as the accessible name of the trigger element.

### AriaLabel story was adjusted to use and explain the new prop.

https://github.com/user-attachments/assets/184986da-4673-43b5-b951-7d61b9f8907a

### DIFF
--- a/apps/storybook/stories/tooltip.stories.tsx
+++ b/apps/storybook/stories/tooltip.stories.tsx
@@ -510,19 +510,25 @@ export const Positions = () => (
 
 export const AriaLabel = () => (
   <Tooltip.Provider>
-    <p>The first button will display AND enunciate the label.</p>
-    <p>The second button will display the label, but enunciate the aria label.</p>
-    <div style={{ display: 'flex' }}>
+    <p>
+      In the first example, the <code>isLabelledBy</code> prop is set to <code>true</code> on the
+      Tooltip trigger. This means the Tooltip content is treated as a <strong>label</strong> for the
+      trigger element. In this scenario, screen readers will read only the Tooltip content as the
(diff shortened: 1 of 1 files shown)


chaance:
> Unfortunately, that results in buggy behaviour where the screen reader announces the label three times
In this example you are not overriding the `aria-describedby` prop. I guess what I'm suggesting is that users can do both on their own side to achieve the same result.

I do see the larger point, which is that perhaps associating the tooltip is more appropriate in some scenarios as the desired behavior. This is certainly common, but I'm not sure if it should be encouraged. Tooltips are already clunky from an accessibility/usability perspective, and they are least problematic when used to provide context rather than serve as a label.

**There are cases IMO where API friction is a good thing.** If your product needs demand it, we don't stop you from implementing the behavior you want. We just don't want to make doing the _usually bad_ thing too easy. It's a tightrope to be sure. You can pass `null` to effectively unset `aria-describedby`, and if you want a prop that simplifies this in your app you can always wrap `Tooltip.Trigger` in your own component library.

```tsx
<Tooltip.Trigger aria-labelledby="foo" aria-describedby={null!}>
  <button id="foo" />
</Tooltip.Trigger>
```

I'll leave this open for now. I could be convinced that it's worthwhile, but I'm ambivalent. I'll seek some outside opinions and come back to it if I decide one way or another.

PR #3602, opened 2025-06-24
[Presence] Export usePresence hook

Sh031224:
### Description

Hello,
First of all, thank you for your amazing work on this project. I frequently use the utilities provided by radix-ui/primitives in my projects, and I truly appreciate the quality and flexibility of your components.

This pull request adds an export for the usePresence hook in `presence.tsx`.

### Motivation

The reason for this change is that I wanted to create new, more flexible components by utilizing the internal isPresent state directly. Exporting the usePresence hook allows for greater control and composability when building custom components that need to respond to presence logic.

### Summary of changes

Exported the `usePresence` hook from `presence.tsx.`

Thank you for considering this contribution! Please let me know if there are any changes or additional steps required.

### DIFF
--- a/.changeset/small-wings-punch.md
+++ b/.changeset/small-wings-punch.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-presence': patch
+---
+
+Add export for the usePresence hook in presence.tsx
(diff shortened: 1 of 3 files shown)


chaance:
We intentionally do not expose these hooks as they are considered internal implementation details. Can you make a minimal repro of what you're trying to accomplish so we can see whether or not there is alternative approach with existing APIs?

PR #3608, opened 2025-07-01
feat: add the possibility to set the item key of a select item

svenliebig:
to enable more control for the developer over the key property of the select item. the initial behaviour is setting the item value directly as key property, when working with values that need to be listed separately but have the same values (for example countries that have the same country code, but need to be displayed separately) the implemenation before created a react error in the console because of duplicated keys in the select item.

this change is backwards compatible to the initial implementation.

### DIFF
--- a/.changeset/early-steaks-mate.md
+++ b/.changeset/early-steaks-mate.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-select': minor
+---
+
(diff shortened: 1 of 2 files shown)


chaance:
We are actually working on splitting out the bubble select entirely so developers will have full control over how the hidden form field is rendered. This prop would be immediately deprecated once we do that, so I'm inclined to pass on this for now. I'd suggest patching the package as an interim solution until we make that change.

PR #3654, opened 2025-08-13
[OneTimePasswordField] Fix iOS Chrome autocomplete

### LINKED ISSUE #3641: [OneTimePasswordField] autoComplete not working in Chrome on iOS
### Current Behavior

When using autoComplete in Chrome on iOS, only the first character of the password is entered.

### Expected behavior

When using autoComplete in Chrome on iOS, the entire password is input.

### Reproducible example

1. Navigate to the Radix UI One Time Password Field demo in Chrome on iOS device:
    https://www.radix-ui.com/primitives/docs/components/one-time-password-field
2. Send an SMS to your device with the following contents:
```
Test OTP: 123456

@www.radix-ui.com #123456
```
3. Observe only the first character '1' is entered and the second input is selected.

### Suggested solution

Remove the max length property and if more than one character is entered, spread them into the other boxes.

### Your environment

| Radix Package(s) | @radix-ui/react-one-time-password-field | 0.1.7 |
| Browser          | chrome | 139.7.7258.60 |
| Operating System | iOS | 18.5 |

controversial:
Autocomplete in `OneTimePasswordField`s is broken on iOS Chrome, see #3641.

This is because:
- `OneTimePasswordFieldInput` **expects** a password manager / autocomplete provider to enter the full code in the first input. When it encounters this case in the `input` event handler, it dispatches the `PASTE` action.
- **However**, iOS chrome autocomplete appears to limit the length of its `input` event’s value for the `maxLength` property set on the target input

The solution is to increase the `maxLength` on the input that expects to receive autocomplete input (the input that has `supportsAutoComplete`) so that it can receive up to `collection.size` characters of input, rather than just one.

Closes #3641.

Adds additional handling for a bug caused by iOS Chrome firing `onChange` on the first input _after_ it fires the `input` event—this had caused OneTimePasswordField to focus the _second_ input where it should be focusing the _final_ input (of the “pasted” value).

(I’m not sure how to add tests for this PR as it’s specific to the iOS chrome environment, but I can try)


controversial:
Compared to #3643, this PR:
- Fixes autocomplete automatically, without requiring that the user knows how/where to override `maxLength` to get correct behavior.
- Handles a secondary bug where a rogue onChange event resulted in the _second_ input being focused after autocomplete (instead of the _final_ input)

chaance (approved):

chaance:
@controversial Had an accidental lint error slip through on main. Mind updating your branch so tests can re-run in CI?

controversial:
@chaance just rebased off the latest main; lints and tests are now passing

controversial:
@chaance It looks like this was released under `1.4.3-rc.1755205432459` even though the stable release is already 1.4.3; do the experimental releases need to be bumped to 1.4.4?

PR #3657, opened 2025-08-14
Updated Twitter to X as Twitter now no longer exists

coultonluke:
### Updated Twitter to X as Twitter now no longer exists

Updated link in readme from Twitter to X as Twitter now no longer exists


coultonluke:
Apologies but this was closed but it says it was unmerged and the .MD still says "Twitter". Was the commit no good? I thought I did it properly. Cheers

PR #3658, opened 2025-08-21
Allow aria-description to describe Dialog and AlertDialog

jannikbertram:
### Description

Currently, the `Dialog` and `AlertDialog` components log a warning if neither a `DialogDescription` / `AlertDialogDescription` component nor a custom component with `aria-describedby` attribute is used.

It shouldn't be necessary to provide a visible description to a dialog, as some of them can be 
self-explanatory. 

With this PR, the aria-description is also accepted to provide a description to screen readers.

### DIFF
--- a/.changeset/new-otters-bet.md
+++ b/.changeset/new-otters-bet.md
@@ -1,0 +1,6 @@
+---
+'@radix-ui/react-alert-dialog': patch
+'@radix-ui/react-dialog': patch
+---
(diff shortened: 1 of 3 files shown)


chaance:
We'll consider this in the future if/when `aria-description` is widely supported. For now it's much better to stick with `aria-describedby` which can point to a visually hidden element.

PR #3687, opened 2025-09-27
fix(select): select item firing click events when disabled

### LINKED ISSUE #3683: [Select] - Disabled `SelectItem` fires click events
### STR
1. Use the Stackblitz link below to proceed to the example where the issue can be reproduced
2. Click the Radix `Select` to pop the content open
3. Attempt to select the disabled item with the name `Item 2`
4. Observe a result

### Current Behavior

Alert is triggered. Looks like the disabled `SelectItem` is clickable (`onClick` fired)

### Expected behavior

Disabled `SelectItem` should block users from interacting with it.
See a native `select` example below the Radix one.

### Reproducible example

[Stackblitz](https://stackblitz.com/edit/vitejs-vite-kfowjcsq?file=src%2FApp.tsx)

### Suggested fix

`SelectItem` to have `pointer-events: none;` when disabled

OmkarJ13:
### Description
Fixes #3683 

- Click events should not be fired when disabled prop is passed on the select item. 
- To achieve this, disabling all pointer events on the select item when it is disabled, if not passed pointer events will be set to 'auto', which is the default.

### DIFF
--- a/.changeset/legal-lights-visit.md
+++ b/.changeset/legal-lights-visit.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-select': patch
+---
+
+Disable pointer events on select item when disabled. Fixes #3683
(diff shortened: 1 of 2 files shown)


chaance:
I don't think this is the right fix, and it could be an unexpected breaking change for some users. I'm going to punt on this until a major release since it's easy to manage in userland as-is.

PR #3690, opened 2025-10-02
fix(select): prevent defaulting to first option when no value selected

### LINKED ISSUE #3198: Unselected Controlled Select Element Submits First Value by Default
## Bug report
Using a controlled Select element with a placeholder in a form causes the form to submit the first item as the selected value, even if the user hasn’t made a selection.

### Current Behavior
When submitting a form with a controlled Select component, the first item in the Select is submitted as the value, even if no selection has been made. See screenshot below:
Screenshot 2024-10-28 at 8 06 59 AM

Attempting to use a placeholder item for Select results in an error:
```
A <Select.Item /> must have a value prop that is not an empty string. This is because the Select value can be set to an empty string to clear the selection and show the placeholder.
```

### Expected behavior

* If Select is not required: The form should submit an empty value ("") when no selection is made.
* If Select is required: The form should throw an [invalid event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/invalid_event) when attempting to submit without a selection.

### Reproducible example
[Select example](https://codesandbox.io/p/sandbox/radix-ui-select-demo-forked-fkh9sy?workspaceId=ab60572b-7df1-42e4-8c39-b111e78064ad)

### Suggested solution
Ideally, if the user makes no selection, the Select value should be "". However, to avoid a breaking change for current users, an optional flag to control this behavior could be provided.

### Your environment

| Radix Package(s) |     @radix-ui/react-select |    2.1.2     |
| React            | n/a     |   18      |
| Browser          |    Chrome     |    latest     |
| Assistive tech   |     None    |         |
| npm/yarn         |     npm    |         |
| Operating System |     OSX    |         |

ritujachopade:
### Description

Fixes #3198

This PR updates the hidden input inside the `Select` component to handle cases 
where no option is selected.

### Changes
- Changed `value={value}` to `value={value ?? ""}` in the hidden input
- Ensures that when no value is selected, the form submits an empty string (`""`) 
  instead of `undefined` or defaulting to the first option

### Before
When a `<Select>` was rendered with no value selected, the hidden input could 
default to the first option’s value, causing forms to submit an unintended result.

### After
If no value is selected, the hidden input now submits an empty string (`""`), 
allowing form submissions to correctly reflect that no selection was made.

### DIFF
--- a/.changeset/eager-cycles-slide.md
+++ b/.changeset/eager-cycles-slide.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-select': major
+---
+
+Fix Select hidden input so it submits empty string when no value is selected
--- a/packages/react/select/src/select.tsx
+++ b/packages/react/select/src/select.tsx
@@ -223,8 +223,7 @@ const Select: React.FC<SelectProps> = (props: ScopedProps<SelectProps>) => {
             tabIndex={-1}
             name={name}
             autoComplete={autoComplete}
-            value={value}
-            // enable form autofill
+            value={value ?? ""}
             onChange={(event) => setValue(event.target.value)}
             disabled={disabled}
             form={form}


ritujachopade:
@chaance @benoitgrelard Can one of you please review?

PR #3692, opened 2025-10-02
fix(select): scroll button bumpiness

antlio:
### Description

this pr intends to fix bumpiness caused by the scroll button when reaching top/bottom of the content getting hidden. the pr updates the style of the button and display them accordingly to the first/last element height.

| state | preview |
| -------|------|
| before | <video src="https://github.com/user-attachments/assets/13156545-ad90-4d52-911a-50bc18a1042b" /> |
| after | <video src="https://github.com/user-attachments/assets/ecc3e4a9-da33-44d0-b913-307fb93194ce" /> |

### DIFF
--- a/.changeset/bumpy-jokes-judge.md
+++ b/.changeset/bumpy-jokes-judge.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-select': patch
+---
+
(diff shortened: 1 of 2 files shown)


antlio:
@chaance any thoughts on that one or should i close to avoid stale?

PR #2197, opened 2023-06-09
[Slot] Add support for nested children (#1825)

ayloncarrijo:
### Description

This PR modifies the way `Slottable` works, adding support for nested children without the need to parse them.

Fixes #1825

### Example

```tsx
import React from 'react';
import { Slot, Slottable } from '@radix-ui/react-slot';

function Button({ asChild, children, ...props }) {
  const Component = asChild ? Slot : "button";

  return (
    <Component {...props}>
      <Slottable child={children}>
        {(child) => (
          <div>
            <div>icon</div>
            <div>{child}</div>
            <div>icon</div>
          </div>
        )}
      </Slottable>
    </Component>
  );
}
```

PR #3695, opened 2025-10-06
[Slot] support slotting onto nested children

jjenzz:
### Description

This is inspired by #2197 and should resolve #3626, but I've also improved the implementation of `Slot` slightly:

- Avoids rendering redundant extra `Slot.SlotClone` comps
- Replaces `React.Children.only()` error with a more intuitive warning
- Re-introduces a working version of the `useComposedRefs` change that was [reverted](https://github.com/radix-ui/primitives/pull/3554)
- Adds a whole bunch of tests including thorough use-case examples in `ssr-testing`


packages/react/slot/src/slot.tsx
```diff
@@ -10,43 +10,26 @@ interface SlotProps extends React.HTMLAttributes<HTMLElement> {
...
-        <SlotClone {...slotProps} ref={forwardedRef}>
-          {React.isValidElement(newElement)
-            ? React.cloneElement(newElement, undefined, newChildren)
-            : null}
-        </SlotClone>
-      );
+    const slottableElement = findSlottableElement(children);
+    const slottableElementRef = slottableElement ? getElementRef(slottableElement) : undefined;
+    // fine in RSC as it's a React.useCallback under the hood
+    const composedRefs = useComposedRefs(forwardedRef, slottableElementRef);
```
jjenzz, line 18: This broke in the previous attempts because Slot was changed to a `'use client'` boundary. We can compose them without that, and all is well :)

PR #3729, opened 2025-10-31
[Slot] add nested slottable support

jjenzz:
### Description

This is inspired by #2197 and should resolve #3626, but I've made it backwards compatible and also improved the implementation of `Slot` slightly:

- Avoids rendering redundant extra `Slot.SlotClone` comps
- Replaces `React.Children.only()` error with a more intuitive warning
- Adds tests including thorough use-case examples in `ssr-testing`

### DIFF
--- a/.changeset/sixty-pets-clap.md
+++ b/.changeset/sixty-pets-clap.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-slot': patch
+---
+
+Support slotting onto nested children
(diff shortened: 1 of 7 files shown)


jjenzz:
tbh, something like this might be more apt. and would allow an API without render props https://github.com/jjenzz/slot-jsx-pragma

```tsx
export const IconButton = ({ children, asChild = false, ...props }) => {
  const Comp = asChild ? Slot : 'button';
  return (
    <Comp {...props}>
      <svg /> <span><Slottable>{children}</Slottable></span>
    </Comp>
  );
};
```

chaance:
Made some tweaks and merged with https://github.com/radix-ui/primitives/commit/cac030cd7e880567f600e2f029fab2c6fa0d2cc9

Issue #3709, opened 2025-10-27
Radix One-Time Password Field - Values Not Persisting and Keyboard Navigation Broken


Guilherme-gso:
I recently integrated the Radix UI One-Time Password Field library into my project and am experiencing critical functionality issues:

1. **Values not persisting**: When typing digits, the focus correctly moves to the next input field, but the previously entered values disappear or don't display. The hidden input field doesn't seem to be capturing the complete OTP value.

2. **Broken keyboard navigation**: 
   - **Tab key**: Doesn't properly navigate between input fields
   - **Backspace/Delete**: Fails to clear the current field or move focus back to the previous input as expected

The component appears to handle focus management but fails at value retention and standard keyboard interactions.

**Environment:**
- Package: `@radix-ui/react-one-time-password-field`
- Version: 0.1.8
- Framework: React
- OS: MacOS
- Browser: Google Chrome

**Code Example:**
```tsx
<OneTimePasswordField.Root name="otp" >
  <OneTimePasswordField.Input />
  <OneTimePasswordField.Input />
  <OneTimePasswordField.Input />
  <OneTimePasswordField.Input />
  <OneTimePasswordField.HiddenInput />
</OneTimePasswordField.Root>
```

https://github.com/user-attachments/assets/c9a3c0e1-02dd-4e49-ae72-044639725c96

Has anyone else experienced similar problems? Any guidance would be appreciated.

WilcoBakker:
@chaance, managed to investigate this further?

WilcoBakker:
As far as I understand the collection items do not update properly for the dispatch function of the `OneTimePasswordField` as can be seen in the following screenshots:

React 19.1:

<img width="1531" height="451" alt="Image" src="https://github.com/user-attachments/assets/bfd4b6a3-f23d-4197-a3ca-118cf784d67a" />

React 19.2:

<img width="1525" height="291" alt="Image" src="https://github.com/user-attachments/assets/f99da422-f13e-4e5d-980f-60e42750b9f3" />

Perhaps the fact dispatch is defined by using the `useEffectEvent` is problematic as it uses the state value collection for example, the documentation of React states `useEffectEvent` shouldn't be used in this way as it could lead to unexpected problems:

<img width="912" height="316" alt="Image" src="https://github.com/user-attachments/assets/ca5b67c5-a570-40e5-9aec-2a8fd7fef859" />

Will try to see if `useCallback` fixes the issue,

WilcoBakker:
[...]

PR #3770, opened 2025-12-09
Fixed OneTimePasswordField's dispatch function.

WilcoBakker:
As described by issue #3709, starting with React 19.2, the `OneTimePasswordField` component no longer functions properly.

The issue being React's `useEffectEvent` hook being used to define the `dispatch` function instead of using React's `useCallback` hook which allows dependencies to be supplied. Particularly the `collection` state value wouldn't update properly causing the value inside the `dispatch` function to be out of sync with the latest state.

React 19.1:

524183999-bfd4b6a3-f23d-4197-a3ca-118cf784d67a

React 19.2:

524184971-f99da422-f13e-4e5d-980f-60e42750b9f3

Caveats of the `useEffectEvent` hook according the [documentation of React](https://react.dev/reference/react/useEffectEvent#caveats):

524189928-ca5b67c5-a570-40e5-9aec-2a8fd7fef859

So I have updated the `dispatch` to make use of React's `useCallback` hook instead restoring the functionality of `OneTimePasswordField`.

### DIFF
--- a/.changeset/neat-ends-read.md
+++ b/.changeset/neat-ends-read.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-one-time-password-field': patch
+---
+
+Fixed the dispatch handler's funcionality.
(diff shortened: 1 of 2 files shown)


fuerst:
Thanks for your work! After applying the patch: entering the OT code now works. Submitting it does not. The OTP field is using the props autoSubmit and autoFocus. There is no error at the console.

chaance:
Done in https://github.com/radix-ui/primitives/pull/3831

PR #3773, opened 2025-12-11
Popover: only set aria-controls if open

kosta:
### Description

At the moment, we always set `aria-controls` in the `Popover.Trigger`, even if `context.open` is false, i.e. when there should be no element with the given id. This is illegal: We must only provide this id if the element exists. 

Fix: Only render `aria-controls` if `context.open`.

Similar logic already exists e.g. in [`TooltipTrigger aria-describedby`](https://github.com/radix-ui/primitives/blob/2bab24a811e45c7a83198659272097f6cfa5a165/packages/react/tooltip/src/tooltip.tsx#L288)

I could not find tests for this component (I can create a new test file if you want).

The functionality is already covered in the existing Storybook.

### DIFF
--- a/.changeset/stale-clubs-teach.md
+++ b/.changeset/stale-clubs-teach.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-popover': patch
+---
+
+Popover: only set aria-controls if open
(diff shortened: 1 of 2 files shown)


kosta:
Hi! It's been a good month since I created this bugfix PR. Maybe it was overlooked?

kosta:
Hi! It's April now. Any feedback for this small bugfix PR?

kosta:
Who's still not responding to this PR in May?

PR #3777, opened 2025-12-13
fix(slot): Handle RSC lazy references with non-PromiseLike payloads

### LINKED ISSUE #3776: asChild with lazy references fails in React 19 + RSC production builds
# Issue: `asChild` with lazy references fails in React 19 + RSC production builds

## Description

When using `DialogTrigger` (or other components) with `asChild`, the trigger element doesn't render in production builds when React Server Components serialization creates lazy references (`Symbol(react.lazy)`) for the trigger prop.

The fix in `@radix-ui/react-slot@1.2.4` that unwraps lazy components using `use()` doesn't work because it's applied too late - inside `Slot`, after `DialogTrigger` has already received the lazy reference.

## Environment

- `@radix-ui/react-dialog`: 1.1.14
- `@radix-ui/react-slot`: 1.2.4
- `next`: 16.x (App Router with RSC)
- `react`: 19.x
- **Only occurs in production builds** (not in development)

## Steps to Reproduce

1. Create a wrapper component that accepts a `trigger` prop and passes it to `DialogTrigger`:

```tsx
'use client';

export const NestedDialog = ({ trigger, children }) => {
  return (
    <Dialog>
      <DialogTrigger asChild>{trigger}</DialogTrigger>
      <DialogContent>{children}</DialogContent>
    </Dialog>
  );
};
```

2. Use it from a Server Component where React Flight's deduplication optimization triggers:

```tsx
// Server Component
export default async function Page() {
  const sharedData = await getData(); // Returns []

  return (
    <div>
      <SiblingComponent data={sharedData} />
      <NestedDialog trigger={<Button>Open</Button>}>
        <FormComponent data={sharedData} /> {/* Same reference triggers deduplication */}
      </NestedDialog>
    </div>
  );
}
```

3. Build for production: `npm run build && npm run start`

4. The button doesn't render.

## Root Cause Analysis

We traced the issue through the RSC payload:

**Working case (dev or simple pages):**
```json
{"trigger": ["$", "button", null, {"className": "...", "children": "Open"}]}
```

**Broken case (production with deduplication):**
```json
{"trigger": "$L32"}  // Lazy reference
```

When `trigger` arrives at the client component, it has:
```javascript
{
  $$typeof: Symbol(react.lazy),
  _payload: { status: "resolved_model", value: ..., reason: ... },
  _init: [Function]
}
```

**The fix in `Slot` (v1.2.4) correctly detects this:**
```javascript
if (isLazyComponent(children) && typeof use === "function") {
  children = use(children._payload);
}
```

**But it doesn't work because:**
- `DialogTrigger` receives the lazy reference as `children`
- `DialogTrigger` may process/wrap children before passing to `Slot`
- By the time `Slot` receives it, the structure has changed

## Workaround

Unwrap lazy references **before** passing to `DialogTrigger`:

```tsx
import { use } from 'react';

function unwrapLazy(element: ReactNode): ReactNode {
  const el = element as any;
  if (
    el?.$$typeof === Symbol.for('react.lazy') &&
    el?._payload !== null &&
    typeof el?._payload === 'object' &&
    'then' in el._payload
  ) {
    return use(el._payload);
  }
  return element;
}

export const NestedDialog = ({ trigger, children }) => {
  const unwrappedTrigger = unwrapLazy(trigger);

  return (
    <Dialog>
      <DialogTrigger asChild>{unwrappedTrigger}</DialogTrigger>
      <DialogContent>{children}</DialogContent>
    </Dialog>
  );
};
```

## Suggested Fix

The lazy unwrapping logic should be applied in `DialogTrigger` (and similar components like `TooltipTrigger`, `PopoverTrigger`, etc.) **before** any processing of children, not just in `Slot`.

Alternatively, components using `asChild` should document that wrapper components need to handle lazy references themselves.

## Related Issues

- [React #32392](https://github.com/facebook/react/issues/32392) - `cloneElement` with async server components
- [Next.js #82527](https://github.com/vercel/next.js/issues/82527) - cloneElement not working in production builds
- [Radix #2537](https://github.com/radix-ui/primitives/issues/2537) - Plans for asChild with cloneElement being legacy

## Additional Context

React maintainers have indicated that `cloneElement` is "soft deprecated" due to conflicts with RSC optimization strategies. However, until an alternative pattern for `asChild` exists, this lazy reference handling is necessary for production RSC builds.

sleitor:
## Summary

Use `Children.toArray` in `SlotClone` to properly unwrap React lazy components that have non-PromiseLike payloads (common in RSC context).

**Problem:** The existing `isLazyComponent` check requires `isPromiseLike(element._payload)`, but in RSC context the `_payload` may have a different structure, causing lazy references to slip through and break `cloneElement`. This results in components not rendering in production builds.

**Solution:** Use `Children.toArray` which internally resolves lazy children regardless of payload structure, making it a robust solution for RSC compatibility.

Fixes #3776

## Changes

- Replace `isLazyComponent` + `isValidElement(children)` approach with `Children.toArray` in `createSlotClone`
- Early return for multiple children error (maintains original behavior)
- Clean separation of concerns with clear comments

## Test plan

- [x] Existing unit tests pass (10 passed, 2 skipped)
- [x] Tested in real Next.js 16 application with Turbopack production build
- [x] Verified nested Dialog trigger renders correctly in production

## Related issues

- https://github.com/facebook/react/issues/32392
- https://github.com/vercel/next.js/issues/82527

### DIFF
--- a/.changeset/upset-candies-sell.md
+++ b/.changeset/upset-candies-sell.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-slot': patch
+---
+
+fix: Handle RSC lazy references with non-PromiseLike payloads
--- a/packages/react/slot/src/slot.tsx
+++ b/packages/react/slot/src/slot.tsx
@@ -99,22 +99,28 @@ interface SlotCloneProps {
 
 /* @__NO_SIDE_EFFECTS__ */ function createSlotClone(ownerName: string) {
   const SlotClone = React.forwardRef<any, SlotCloneProps>((props, forwardedRef) => {
-    let { children, ...slotProps } = props;
-    if (isLazyComponent(children) && typeof use === 'function') {
-      children = use(children._payload);
+    const { children, ...slotProps } = props;
+
+    // Multiple children is an error
+    if (React.Children.count(children) > 1) {
+      return React.Children.only(null);
     }
 
-    if (React.isValidElement(children)) {
-      const childrenRef = getElementRef(children);
-      const props = mergeProps(slotProps, children.props as AnyProps);
+    // Use Children.toArray to properly unwrap RSC lazy references
+    // (which may have non-PromiseLike payloads that bypass isLazyComponent)
+    const childrenArray = React.Children.toArray(children);
+    if (childrenArray.length === 1 && React.isValidElement(childrenArray[0])) {
+      const child = childrenArray[0];
+      const childRef = getElementRef(child);
+      const mergedProps = mergeProps(slotProps, child.props as AnyProps);
       // do not pass ref to React.Fragment for React 19 compatibility
-      if (children.type !== React.Fragment) {
-        props.ref = forwardedRef ? composeRefs(forwardedRef, childrenRef) : childrenRef;
+      if (child.type !== React.Fragment) {
+        mergedProps.ref = forwardedRef ? composeRefs(forwardedRef, childRef) : childRef;
       }
-      return React.cloneElement(children, props);
+      return React.cloneElement(child, mergedProps);
     }
 
-    return React.Children.count(children) > 1 ? React.Children.only(null) : null;
+    return null;
   });
 
   SlotClone.displayName = `${ownerName}.SlotClone`;


sleitor:
Friendly ping 👋 This PR has been open since December. There's also a related regression report in #3780 (silent failure in RSC dev mode after Next.js 16.0.10 security patch) that likely shares the same root cause.

The `Children.toArray` approach used here was independently validated by Mantine ([mantinedev/mantine#8522](https://github.com/mantinedev/mantine/pull/8522)) for the same class of issues. Would love to get a maintainer review when you have a chance!

GeeWizWow:
I know these types of comments are generally not welcome, but I'd really appreciate if we could see some movement on this PR, I'm seeing more and more widespread occurrences of this issue.

@chaance and @nicknisi you guys seem to have been most recently active in this project, eternal gratitude if you could take a look.

Beyond this, there also seems to be issues with certain packages not resolving the latest workspace versions correctly, for instance: `@radix-ui/react-tooltip` version 1.2.8 (latest) references slot 1.2.3, despite the changelog explicitely citing that it has been updated to 1.2.4

```md
# @radix-ui/react-tooltip

## 1.2.8

- Updated dependencies: `@radix-ui/react-presence@1.1.5`, `@radix-ui/react-slot@1.2.4`, [...truncated]
```

```cli
npm view @radix-ui/react-tooltip --registry https://registry.npmjs.org

@radix-ui/react-tooltip@1.2.8 | MIT | deps: 12 | versions: 359
View docs [here](https://radix-ui.com/primitives/docs/components/tooltip).
https://radix-ui.com/primitives

dist
[...truncated]

dependencies:
[...truncated]                 @radix-ui/react-slot: 1.2.3
```

sleitor:
👋 Gentle ping — just checking if this is still on the radar for review. Happy to address any feedback or rebase if needed!

Issue #3786, opened 2026-01-08
`NavigationMenuContent` when `forceMount` has no `data-state` attribute when used with `NavigationMenuViewport`


tholander:
### Current Behavior

Currently when you use `NavigationMenuContent` with `forceMount` and a `NavigationMenuViewport`, the `data-state` attribute is not set on `NavigationMenuContent`. 

### Expected behavior

I think the `NavigationMenuContent` should always have the `data-state` attribute set.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/p/sandbox/jqs8qn)

### Suggested solution

The `data-state` attribute should be pass to `ViewportContentMounter` in [NavigationMenuContent](https://github.com/radix-ui/primitives/blob/907513701a75b11a115563f9554ac6e8147bf2db/packages/react/navigation-menu/src/navigation-menu.tsx#L787)

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |   @radix-ui/react-navigation-menu      | 1.2.14 |
| React            | n/a     | 19.2.0 |
| Browser          | Chrome | 143.0.7499.170 |
| Assistive tech   | n/a |         |
| Node             | n/a     |         |
| npm/yarn/pnpm    | npm | 10.9.3 |
| Operating System | MacOS | 26.1 |

PR #3787, opened 2026-01-08
fix(navigation-menu): set data-state attribute to `NavigationMenuContent` when using `NavigationMenuViewport`

tholander:
Issue : https://github.com/radix-ui/primitives/issues/3786

### Description

Passing `data-state` to `ViewportContentMounter` to enable style/animation control when used with `forceMount` and `NavigationMenuViewport`


dimitriBouteille:
Nice PR :rocket: 

Must be merge quickly !!!!

PR #3801, opened 2026-02-02
Add explicit role to child div of ScrollAreaViewport

andrewhayward:
### Description

The `ScrollAreaViewport` renders a `div` with a `display: table` style. This can cause issues in browsers (such as Chrome) that implicitly generate an element role based on specific heuristics; in this case, `role: LayoutTable`.

By adding an explicit `role: presentation` to this node, we prevent any guessing by the browser, ensuring that no additional and unexpected semantics are included in the accessibility tree.


HardikDewra (commented):
Small but important accessibility fix. The inner div with `display: table` is a pure layout hack for width calculation and has zero semantic meaning, so `role="presentation"` is exactly right here. This prevents screen readers from announcing an extra container that confuses users navigating by landmarks.

Would be nice to add a brief comment in the code explaining WHY `role="presentation"` is needed (because `display: table` creates an implicit table role in the a11y tree), so future contributors don't remove it thinking it's unnecessary.

PR #3804, opened 2026-02-05
fix(slot): use memoized useComposedRefs in SlotClone to prevent React 19 infinite loops

### LINKED ISSUE #3799: installHook.js:1 Error: Maximum update depth exceeded - React 19 + Radix
## Bug report

### Current Behavior

When using Radix components with React 19, an infinite loop occurs causing the error:                                                                                          
                                                                                                                                                                                 
  Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of  
  nested updates to prevent infinite loops.                                                                                                                                      
      at setRef (composeRefs.tsx:11:5)                                                                                                                                           
      at composeRefs.tsx:22:45                                                                                                                                                   
      at Array.forEach (<anonymous>)                                                                                                                                             
      at composeRefs.tsx:22:28                                                                                                                                                   
      at setRef (composeRefs.tsx:11:5)                                                                                                                                           
      at composeRefs.tsx:22:45                                                                                                                                                   
      ...                                                                                                                                                                        
                                                                                                                                                                                 
  The setRef function in @radix-ui/react-compose-refs recursively triggers itself, causing React to hit the maximum update depth limit.                                          
                                                                                                                                                                                 
  In React 19, ref callback behavior changed - ref callbacks can now return cleanup functions, and React may call ref callbacks differently than in React 18. When setRef calls  
  ref(value) at line 11, it triggers a re-render which invokes setRef again via the composed ref callback at line 22, creating an infinite loop.    


### Expected behavior

  Composed refs should work without causing infinite update loops in React 19.                                                                                                    

### Reproducible example

I have platejs plugins with radix popovers and radix tooltips on them. If I write fast inside plate, the page breaks. 
I'm on a Macbook M1 Pro with 32 GB. One way to trigger this error is to inspect chrome and start a performance recording and start typing real fast. On Safari, for example, it doesn't break. I tried on restarted browser, guest mode. I'm not the only one

[CodeSandbox Template](n/a)

### Suggested solution

  The composeRefs and useComposedRefs functions need to handle React 19's new ref callback behavior. Possible approaches:                                                        
                                                                                                                                                                                 
  1. Memoize the composed ref callback to prevent unnecessary re-invocations                                                                                                     
  2. Add a guard to prevent recursive setRef calls                                                                                                                               
  3. Handle the cleanup function return value from React 19 ref callbacks       

### Additional context
Right now, it works fine with react 18 (even on next.js 16+) and we had to downgrade.

I also saw this reddit post https://www.reddit.com/r/react/comments/1nnvwsp/react_19_causes_maximum_update_depth_exceeded/ where a user raised a similar case

React 19 introduced changes to ref callback behavior, including support for cleanup functions returned from ref callbacks. This may be causing the ref to be re-invoked on each render cycle.                                                                                                                                                                 
                                                                                                                                                                                 
  Related stack trace also shows Slate editor involvement:                                                                                                                       
  at Slate.useCallback[onContextChange] (slate.tsx:62:7)                                                                                                                         
  at e.onChange (with-dom.ts:371:7)                                                                                                                                              
  at with-react.ts:54:7                                                                                                                                                          
                                                                                                                                                                                 
  This suggests the issue manifests when Radix components are used alongside Slate/Plate editors, where both libraries compose refs. 

### Your environment

| Software         | Name(s)                               | Version       |
| ---------------- | ------------------------------------- | ------------- |
| Radix Package(s) | @radix-ui/react-compose-refs          | 1.1.0         |
| Radix Package(s) | @radix-ui/react-accordion             | 1.2.12        |
| Radix Package(s) | @radix-ui/react-avatar                | 1.1.11        |
| Radix Package(s) | @radix-ui/react-checkbox              | 1.3.3         |
| Radix Package(s) | @radix-ui/react-collapsible           | 1.1.12        |
| Radix Package(s) | @radix-ui/react-dialog                | 1.1.15        |
| Radix Package(s) | @radix-ui/react-dropdown-menu         | 2.1.16        |
| Radix Package(s) | @radix-ui/react-hover-card            | 1.1.15        |
| Radix Package(s) | @radix-ui/react-icons                 | 1.3.2         |
| Radix Package(s) | @radix-ui/react-popover               | 1.1.15        |
| Radix Package(s) | @radix-ui/react-progress              | 1.1.8         |
| Radix Package(s) | @radix-ui/react-select                | 2.2.6         |
| Radix Package(s) | @radix-ui/react-separator             | 1.1.8         |
| Radix Package(s) | @radix-ui/react-toolbar               | 1.1.11        |
| Radix Package(s) | @radix-ui/react-tooltip               | 1.2.8         |
| React            | react, @types/react                   | 19.2.3        |
| Browser          | Chrome                                | Latest        |
| Assistive tech   | n/a                                   |               |
| Node             | node                                  | 24.13.0       |
| npm/yarn/pnpm    | pnpm                                  | 10.26.0       |
| Operating System | macOS                                 | Darwin 24.6.0 |


  - Next.js: 16.1.1                                                                                                                                                              
  - TypeScript: 5.9.3                                                                                                                                                            
  - platejs: 52.0.15

yannisgu:
> [!NOTE]
> This PR was generated by an AI agent (Claude Code). Human verification of the fix and its implications is still outstanding.

## Summary

- `SlotClone` now uses `useComposedRefs()` (memoized via `React.useCallback`) instead of raw `composeRefs()` to compose forwarded and children refs
- This keeps the composed ref callback identity stable across renders, preventing React 19's ref unmount/remount cycle from triggering infinite loops

## Reproduction

**Minimal repro repo**: https://github.com/yannisgu/radix-tanstack-form-react19-repro

```bash
git clone https://github.com/yannisgu/radix-tanstack-form-react19-repro
cd radix-tanstack-form-react19-repro/radix-select-tanstack-form
pnpm install && pnpm dev
# Open http://localhost:5173 — tab freezes
```

## Problem

`SlotClone` calls `composeRefs(forwardedRef, childrenRef)` directly on every render, creating a **new function** each time. In React 19, `setRef()` returns `ref(value)`, which React treats as a cleanup function. When React sees a new callback ref with cleanup on each render, it unmounts the old ref (calling cleanup) and mounts the new one — triggering any state setters passed as refs (e.g. `SelectTrigger`'s `onTriggerChange`, or `Checkbox`'s `useSize`).

This state update during commit causes a re-render, which creates another new `composeRefs` function, which triggers another unmount/remount — an infinite synchronous loop that **freezes the browser tab**.

The bug is particularly severe when combined with libraries that trigger re-renders during the commit phase (e.g. `@tanstack/react-form`'s `useField` layout effect with no dependency array), but can occur with any component that has effects or state updates during commit.

## Root cause chain

1. `SlotClone` calls `composeRefs()` → new function identity every render
2. `setRef(ref, value)` returns `ref(value)` → React 19 treats it as cleanup ref
3. New ref identity + cleanup → React unmounts old ref, mounts new ref
4. Ref mount triggers state setter (e.g. `onTriggerChange`) → re-render
5. Re-render → step 1 → infinite loop

## Fix

Replace raw `composeRefs()` with `useComposedRefs()`, which wraps the call in `React.useCallback`. This was already the intended pattern — `useComposedRefs` exists for exactly this purpose but wasn't used in `SlotClone`.

## Affected components

Any Radix component using `Slot`/`asChild` that passes a state setter as a ref:
- `@radix-ui/react-select` (SelectTrigger + onTriggerChange)
- `@radix-ui/react-checkbox` (useSize state setter)
- `@radix-ui/react-popover`, `@radix-ui/react-tooltip`, `@radix-ui/react-dialog`, etc.

Fixes #3799

### DIFF
--- a/.changeset/fix-slot-clone-composed-refs.md
+++ b/.changeset/fix-slot-clone-composed-refs.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-slot': patch
+---
+
+Fixed infinite re-render loop in React 19 caused by `SlotClone` creating a new `composeRefs` callback on every render. Now uses memoized `useComposedRefs` to keep ref identity stable.
--- a/packages/react/slot/src/slot.tsx
+++ b/packages/react/slot/src/slot.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { composeRefs } from '@radix-ui/react-compose-refs';
+import { composeRefs, useComposedRefs } from '@radix-ui/react-compose-refs';
 
 declare module 'react' {
   interface ReactElement {
@@ -104,12 +104,23 @@ interface SlotCloneProps {
       children = use(children._payload);
     }
 
+    const childrenRef = React.isValidElement(children)
+      ? getElementRef(children)
+      : undefined;
+
+    // useComposedRefs memoizes the composed callback so that ref identity is
+    // stable across renders.  The previous raw `composeRefs()` call created a
+    // new function every render, which in React 19 causes an unmount/remount
+    // cycle (because React treats a new callback-ref-with-cleanup as a
+    // different ref), leading to infinite re-render loops when a state setter
+    // is passed as one of the refs (e.g. SelectTrigger's onTriggerChange).
+    const composedRef = useComposedRefs(forwardedRef, childrenRef);
+
     if (React.isValidElement(children)) {
-      const childrenRef = getElementRef(children);
       const props = mergeProps(slotProps, children.props as AnyProps);
       // do not pass ref to React.Fragment for React 19 compatibility
       if (children.type !== React.Fragment) {
-        props.ref = forwardedRef ? composeRefs(forwardedRef, childrenRef) : childrenRef;
+        props.ref = composedRef;
       }
       return React.cloneElement(children, props);
     }


abductedMind:
need this, thank you!

chaance:
The slot implementation recently changed which introduces some conflicts, but I wanted to prioritize a fix for the next release so I went ahead and re-implemented the solution in https://github.com/radix-ui/primitives/pull/3899 (with co-author credit).

PR #3806, opened 2026-02-06
fix(tooltip): handle Window as scroll event target

### LINKED ISSUE #3805: TypeError: target?.contains() is not a function
### Current Behavior

I've never seen this error before and Sentry didn't have a screen recording for this one.
Based on what I can decipher from the stack trace on Sentry, I believe that the following error is thrown when automatically closing a tooltip by scrolling.

<details>
<summary><b>Stack trace (screenshot)</b></summary>

Image

</details>

Errors thrown in event handlers will not have any effect and it was the last statement of the callback so it's safe to assume the user wasn't impacted at all except for the automatic closing not working. I suppose that they could still use the close button though.

### Expected behavior

The edge case (explained further) should be handled properly to avoid this error.

### Reproducible example

Sorry, I don't have a reproducible example.
This is the first time ever I see this error, I have no clue how the my project's code could cause the error, I didn't trigger it myself either, but this error does make sense on paper (see further).

### Suggested solution

I don't know the exact solution, but I do know why the error happens.

There is a cast `const target = event.target as HTMLElement` on [line 531 of tooltip.tsx](https://github.com/radix-ui/primitives/blob/main/packages/react/tooltip/src/tooltip.tsx#L531).
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget), the event target can be any DOM element, its children, `Document` or `Window`.
The `scroll` event handler is added to the `window` object on line 534 so it is definitely a valid event target.
However, it's the only one in the list that does not have the method `.contains()`.

In other words: the cast is almost always fine, but not when it comes to this edge case.

<details>
<summary><b>The event data on Sentry confirms it (screenshot)</b></summary>

Image

</details>

### Your environment

| Radix Package(s) | @radix-ui/react-tooltip  | 1.2.8 |
| React            | n/a     | 19.2.0-canary-0bdb9206-20250818 |
| Browser          | Chrome | 144.0.0 |
| Assistive tech   | NextJS (with Turbopack) | 15.5.11 |
| Node             | n/a     | 22.22.0 |
| npm/yarn/pnpm    | npm | unknown - compatible with NodeJS v22.22 |
| Operating System | Amazon Linux 2 (hosted by Vercel) | 5.10.247-277.989.amzn2.x86_64 |

kaigritun:
## Summary

Fixes #3805

When a scroll event is captured with `{ capture: true }`, the `event.target` can be `Window` instead of an `HTMLElement`. The `Window` object doesn't have a `.contains()` method, which causes a `TypeError: target?.contains() is not a function` error.

## Changes

Added an `instanceof Node` check before calling `.contains()` to ensure the target is a valid DOM node that supports the method.

### Before
```ts
if (target?.contains(context.trigger)) onClose();
```

### After
```ts
const target = event.target;
// event.target can be Window when scrolling, which doesn't have .contains()
if (target instanceof Node && target.contains(context.trigger)) onClose();
```

## Testing

- All existing tooltip tests pass
- The fix handles the edge case where `event.target` is `Window` or `Document` (neither implements `Node`)

### DIFF
--- a/packages/react/tooltip/src/tooltip.tsx
+++ b/packages/react/tooltip/src/tooltip.tsx
@@ -528,8 +528,9 @@ const TooltipContentImpl = React.forwardRef<TooltipContentImplElement, TooltipCo
     React.useEffect(() => {
       if (context.trigger) {
         const handleScroll = (event: Event) => {
-          const target = event.target as HTMLElement;
-          if (target?.contains(context.trigger)) onClose();
+          const target = event.target;
+          // event.target can be Window when scrolling, which doesn't have .contains()
+          if (target instanceof Node && target.contains(context.trigger)) onClose();
         };
         window.addEventListener('scroll', handleScroll, { capture: true });
         return () => window.removeEventListener('scroll', handleScroll, { capture: true });


MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

PR #3820, opened 2026-02-25
fix(select): guard against null ref in typeahead focus

### LINKED ISSUE #3597: Error Cannot read properties of null (reading 'focus') - radix select
### Current Behavior

[](https://github.com/user-attachments/assets/a839a177-eb98-4f58-a121-28e75ec3681e)

There's a lag/nondetected input when clicking on the select component.

### Expected behavior

As seen in the video, the customer should be able to smoothly click through the component without an error.

### Reproducible example

N/A

### Suggested solution

Add an if statement to see if nextItem.ref.current is null. Seems like this line might be:

This error was emitted by Sentry. 

`../../node_modules/@radix-ui/react-select/dist/index.mjs in <anonymous> at line 384:47

    const [searchRef, handleTypeaheadSearch] = useTypeaheadSearch((search) => {
      const enabledItems = getItems().filter((item) => !item.disabled);
      const currentItem = enabledItems.find((item) => item.ref.current === document.activeElement);
      const nextItem = findNextItem(enabledItems, search, currentItem);

     if (nextItem) {
        setTimeout(() => nextItem.ref.current.focus()); <----
      }
    });
    const itemRefCallback = React.useCallback(
      (node, value, disabled) => {
        const isFirstValidItem = !firstValidItemFoundRef.current && !disabled;`

### Your environment

| Radix Package(s) |      select   |      2.2.5   |
| React            | n/a     |    18.2     |
| Browser          |    Chrome     |    136     |
| npm/yarn/pnpm    |    npm     |     10.8.2    |
| Operating System |    windows     |     >= 10    |

Tourniercy:
### Description
	
Fixes #3597

Adds a null check on nextItem.ref.current before calling .focus() in the typeahead search handler within SelectContentImpl.

When using the Select component on mobile browsers (observed on Android Chrome), the DOM element referenced by nextItem.ref.current can be null if the element hasn't mounted yet or has been unmounted due to timing issues.

This causes an unhandled TypeError: Cannot read properties of null (reading 'focus').

### DIFF
--- a/packages/react/select/src/select.tsx
+++ b/packages/react/select/src/select.tsx
@@ -669,7 +669,7 @@ const SelectContentImpl = React.forwardRef<SelectContentImplElement, SelectConte
          * Imperative focus during keydown is risky so we prevent React's batching updates
          * to avoid potential bugs. See: https://github.com/facebook/react/issues/20332
          */
-        setTimeout(() => (nextItem.ref.current as HTMLElement).focus());
+        setTimeout(() => (nextItem.ref.current as HTMLElement)?.focus());
       }
     });
 


chaance:
Already resolved on main 👍

PR #3834, opened 2026-03-19
fix(direction): respect document-level dir attribute in useDirection when no DirectionProvider is present

### LINKED ISSUE #3830: feat(direction): auto-detect RTL from HTML document when no DirectionProvider is present
## Problem

`useDirection()` currently falls back to `'ltr'` when no `DirectionProvider` exists in the React tree:

```typescript
// packages/react/direction/src/direction.tsx
function useDirection(localDir?: Direction) {
  const globalDir = React.useContext(DirectionContext);
  return localDir || globalDir || 'ltr'; // ← hardcoded fallback
}
```

This means every Radix component renders `dir="ltr"` on its DOM elements — **ignoring** `<html dir="rtl">` entirely.

### Real-world impact

We discovered this while contributing Arabic/RTL support to [AFFiNE](https://github.com/toeverything/AFFiNE) (65k+ GitHub stars). Even after correctly setting `<html lang="ar" dir="rtl">` at the document level, every Radix dropdown, dialog, popover, and scroll area continued to render `dir="ltr"` — effectively overriding the document direction.

The workaround: manually wrap the entire app in `DirectionProvider` with a `MutationObserver`. This works, but it requires **every application** to implement this boilerplate. With 130M+ monthly downloads and Radix powering shadcn/ui (used by Vercel, Linear, Supabase), the RTL community repeats this fix across every project.

## Proposed fix

```typescript
function useDirection(localDir?: Direction) {
  const globalDir = React.useContext(DirectionContext);
  const documentDir =
    typeof document !== 'undefined'
      ? document.documentElement.dir === 'rtl' ? 'rtl' : 'ltr'
      : 'ltr';
  return localDir || globalDir || documentDir;
}
```

### Why this is safe

- **No breaking changes**: `DirectionProvider` still takes precedence — existing apps are unaffected
- **LTR apps unchanged**: Most apps don't set `<html dir="rtl">`, so `documentDir` defaults to `'ltr'`
- **SSR safe**: `typeof document !== 'undefined'` guard handles server-side rendering
- **Dynamic**: Works with runtime language switching (when `dir` changes on `<html>`)
- **Zero configuration**: RTL apps work correctly without any wrapper

## Who this helps

Arabic, Hebrew, Persian, and Urdu speakers — ~500 million people. Every app using Radix that sets `<html dir="rtl">` currently has broken RTL layout for dropdowns, dialogs, and popovers.

## Context

- Validated during AFFiNE Arabic localization: PR [#14646](https://github.com/toeverything/AFFiNE/pull/14646)
- shadcn added RTL support in January 2026 — this change would make it automatic for all Radix consumers
- The fix is a **single line change** in one file

ahmadaurwahaj:
Fixes #3830

### Description

`useDirection()` currently falls back to `'ltr'` when no `DirectionProvider` exists in the React tree - completely ignoring `<html dir="rtl">` set at the document level. This causes every Radix component (DropdownMenu, Dialog, Popover, ScrollArea, etc.) to render `dir="ltr"` on its DOM elements, overriding the document's RTL setting.

Teams building Arabic, Hebrew, Persian, or Urdu apps must manually wrap their entire application in `<DirectionProvider dir="rtl">` just to get correct layout - boilerplate that should not be required.

**Fix:** Read `document.documentElement.dir` as a fallback before defaulting to `'ltr'`. Added a `getDocumentDirection()` helper with an SSR guard.

Priority order (unchanged, additive only): `localDir` > `DirectionProvider` context > `<html dir>` > `'ltr'`

**Why this is safe:**
- No breaking changes - `DirectionProvider` and `localDir` still take full precedence. Existing apps are unaffected.
- LTR apps unchanged - if `<html dir>` is not `"rtl"`, result is `'ltr'` as before.
- SSR safe - `typeof document === 'undefined'` guard falls back to `'ltr'` on the server.

**Tests:** Added the first unit tests for the `direction` package (6 tests) covering all priority levels - no provider + document rtl, provider overrides document, local dir overrides provider.

**Storybook not required:** This change is to an internal hook (`useDirection`) with no visual component of its own. There is no UI to demonstrate - the fix is a logic change in a single fallback value.

### DIFF
--- a/packages/react/direction/src/direction.tsx
+++ b/packages/react/direction/src/direction.tsx
@@ -20,8 +20,15 @@ const DirectionProvider: React.FC<DirectionProviderProps> = (props) => {
 
 function useDirection(localDir?: Direction) {
   const globalDir = React.useContext(DirectionContext);
-  return localDir || globalDir || 'ltr';
+  const documentDir = getDocumentDirection();
+  return localDir || globalDir || documentDir;
 }
+
+function getDocumentDirection(): Direction {
+  if (typeof document === 'undefined') return 'ltr';
+  return document.documentElement.dir === 'rtl' ? 'rtl' : 'ltr';
+}
+
 
 const Provider = DirectionProvider;
 
--- a/.changeset/free-crabs-watch.md
+++ b/.changeset/free-crabs-watch.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-direction': patch
+---
+
+Respect document-level dir attribute in useDirection when no DirectionProvider is present
--- a/packages/react/direction/src/direction.test.tsx
+++ b/packages/react/direction/src/direction.test.tsx
@@ -1,0 +1,107 @@
+import React from 'react';
+import type { RenderResult } from '@testing-library/react';
+import { cleanup, render } from '@testing-library/react';
+import { DirectionProvider, useDirection } from './direction';
+import { afterEach, describe, it, beforeEach, expect } from 'vitest';
+
+// Helper component that renders the resolved direction as text
+function DirectionDisplay({ localDir }: { localDir?: 'ltr' | 'rtl' }) {
+  const direction = useDirection(localDir);
+  return <span data-testid="direction">{direction}</span>;
+}
+
+describe('useDirection', () => {
+  let rendered: RenderResult;
+
+  afterEach(() => {
+    cleanup();
+    // Reset document dir after each test
+    document.documentElement.dir = '';
+  });
+
+  describe('given no DirectionProvider and no local dir', () => {
+    describe('when document dir is not set', () => {
+      beforeEach(() => {
+        document.documentElement.dir = '';
+        rendered = render(<DirectionDisplay />);
+      });
+
+      it('should default to ltr', () => {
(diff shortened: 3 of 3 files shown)


packages/react/direction/src/direction.tsx
```diff
@@ -20,9 +20,16 @@ const DirectionProvider: React.FC<DirectionProviderProps> = (props) => {
...
 function useDirection(localDir?: Direction) {
   const globalDir = React.useContext(DirectionContext);
-  return localDir || globalDir || 'ltr';
+  const documentDir = getDocumentDirection();
+  return localDir || globalDir || documentDir;
 }
 
+function getDocumentDirection(): Direction {
+  if (typeof document === 'undefined') return 'ltr';
+  return document.documentElement.dir === 'rtl' ? 'rtl' : 'ltr';
```
mazipan-wego, line 29: Can be simplified like this @ahmadaurwahaj.

```suggestion
  return document.documentElement.dir || 'ltr';
```

chaance:
No plans to address at this time. The implementation here would result in a hydration mismatch. This is better handled by consumer code since folks may have reasonably different approaches as to how they handle the fallback on the server if the value provided to the root node is not accessible.

PR #3839, opened 2026-04-03
fix(dialog): use getRootNode() for Shadow DOM accessibility check

### LINKED ISSUE #3814: [Dialog] validation errors in console when used inside Shadow-DOM
### Current Behavior

When rendering the Dialog inside a Shadow-DOM the accessibility validations for the Dialog-Title always fail as document.getElementById cannot find elements in the Shadow-DOM.

### Expected behavior

The error logs in the console are not printed.

### Reproducible example

can be provided if required

### Suggested solution

As you can only render the Dialog in a Shadow-DOM by using the DialogPortal, it would make sense, to use the provided container element to find the root element in such cases by using [getRootNode](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode).

### Additional context

I need to use a Shadow-DOM to not be effected by the styles of the Host-Application.

### Your environment

| Radix Package(s) |     @radix-ui/react-dialog    |     1.1.15    |
| React            | n/a     |    19.2.3     |
| Browser          |     Chrome    |    144.0.7559.133     |
| Assistive tech   |     x    |     x    |
| Node             | n/a     |   v24.13.1      |
| npm/yarn/pnpm    |     npm    |      11.8.0   |
| Operating System |     macos    |   26.3      |

sujiu222:
## Summary

Fixes #3814

## Problem

`Dialog` component uses `document.getElementById` to find the `DialogTitle` and `DialogDescription` elements for accessibility validation. When Dialog is rendered inside a Shadow DOM, elements are not accessible via `document.getElementById`, causing false-positive accessibility warnings in the console:

```
DialogContent requires a DialogTitle for the component to be accessible for screen reader users.
```

## Root Cause

Both `TitleWarning` and `DescriptionWarning` components called `document.getElementById(id)` to check for the presence of the title/description elements. In Shadow DOM, elements reside in a `ShadowRoot`, not in the top-level `document`, so `document.getElementById` always returns `null`.

## Solution

Replace `document.getElementById` with `contentRef.current?.getRootNode()` to obtain the correct root node, then call `getElementById` on that root. This correctly handles both:

- **Regular DOM**: `getRootNode()` returns the `Document`, same behavior as before
- **Shadow DOM**: `getRootNode()` returns the `ShadowRoot`, which has its own `getElementById` (via the `DocumentOrShadowRoot` interface)

The `TitleWarning` component was also updated to receive `contentRef` (already available in `DialogContentImpl`) alongside the existing `contentRef` already passed to `DescriptionWarning`.

## Changes

- `packages/react/dialog/src/dialog.tsx`
  - Added `contentRef` prop to `TitleWarningProps` and `TitleWarning` component
  - Pass `contentRef` to `<TitleWarning>` in `DialogContentImpl`
  - Replace `document.getElementById` with `(contentRef.current?.getRootNode() ?? document).getElementById` in both warning components

## Testing

- Existing behavior is preserved for regular DOM usage (`getRootNode()` returns `document` in non-shadow context)
- Shadow DOM context now correctly resolves accessibility checks

### DIFF
--- a/.changeset/fix-dialog-shadow-dom-accessibility.md
+++ b/.changeset/fix-dialog-shadow-dom-accessibility.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-dialog': patch
+---
+
+Fix accessibility check for `DialogTitle` and `DialogDescription` when rendered inside Shadow DOM. Previously used `document.getElementById` which fails in Shadow DOM contexts; now uses `element.getRootNode()` to search within the correct document scope.
--- a/packages/react/dialog/src/dialog.test.tsx
+++ b/packages/react/dialog/src/dialog.test.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { axe } from 'vitest-axe';
 import type { RenderResult } from '@testing-library/react';
-import { render, fireEvent, cleanup } from '@testing-library/react';
+import { render, fireEvent, cleanup, act } from '@testing-library/react';
 import * as Dialog from './dialog';
 import type { Mock, MockInstance } from 'vitest';
 import { describe, it, afterEach, beforeEach, vi, expect } from 'vitest';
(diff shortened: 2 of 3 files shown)


sujiu222:
Thanks for the review! Addressed all three points in the latest commit:

1. **Runtime guard for `getRootNode()`**: Added `instanceof Document || instanceof ShadowRoot` check before calling `getElementById`. Falls back to `contentRef.current?.ownerDocument ?? document` for detached subtrees or `DocumentFragment` nodes. Applied to both `TitleWarning` and `DescriptionWarning`.

2. **Tests**: Will add a Shadow DOM regression test — noted as a follow-up in this PR or a separate test-only PR depending on the test setup complexity.

PR #3840, opened 2026-04-03
fix(dialog): ensure scroll lock is cleared on unmount during navigation

### LINKED ISSUE #3797: [Dialog] Scroll lock (overflow: hidden) persists after navigating away via Link inside Dialog
### Current Behavior

When navigating away from a page via a Link inside a Dialog, the scroll lock (overflow: hidden on body) persists. When returning to the original page, scrolling is broken.

Steps to reproduce:

On Page A, open a Dialog
Click a Link inside the Dialog to navigate to Page B
On Page B, click a Link to navigate back to Page A
Result: Page A has overflow: hidden on body and cannot be scrolled
After these steps, document.body.style.overflow returns "hidden" and body has data-scroll-locked="1" attribute.

### Expected behavior

The scroll lock should be properly removed when the Dialog unmounts during navigation.

### Reproducible example

```ts
// Page A - Dialog with Link
<Dialog>
  <Dialog.Trigger asChild>
    <Button>Open Dialog</Button>
  </Dialog.Trigger>
  <Dialog.Content>
    <Dialog.Close asChild>
      <Button asChild>
        <Link to="/page-b">Go to Page B</Link>
      </Button>
    </Dialog.Close>
  </Dialog.Content>
</Dialog>

// Page B - Link back to Page A
<Link to="/page-a">Back to Page A</Link>
// After clicking this, Page A cannot scroll

```

### Suggested solution / Workaround

```ts
// router.tsx
import { createRouter, RouterProvider } from "@tanstack/react-router";
import { useEffect } from "react";

const cleanupDialogScrollLock = () => {
  document.body.style.overflow = "";
  document.body.style.pointerEvents = "";
  document.body.removeAttribute("data-scroll-locked");
};

const router = createRouter({ routeTree });

const Router = () => {
  useEffect(() => {
    const unsubBefore = router.subscribe("onBeforeNavigate", cleanupDialogScrollLock);
    const unsubLoad = router.subscribe("onLoad", cleanupDialogScrollLock);

    return () => {
      unsubBefore();
      unsubLoad();
    };
  }, []);

  return <RouterProvider router={router} />;
};
```

### Your environment

Software | Version
-- | --
radix-ui | 1.1.3
react | 18.2.0
@tanstack/react-router | 1.132.6
Browser | Chrome

sujiu222:
## Summary

Fixes #3797

## Problem

When a `Link` inside a `Dialog` triggers route navigation, the Dialog unmounts but the scroll lock (`overflow: hidden` on `body`) is not properly cleaned up. This leaves the page in a non-scrollable state — `data-scroll-locked="1"` persists on `document.body` — even after navigating to a different page and back.

## Root Cause

`react-remove-scroll` manages the `data-scroll-locked` attribute via `useEffect` (asynchronous). In certain SPA router scenarios (TanStack Router, Next.js App Router with transitions), the old route's async effect cleanups may be **deferred or skipped** before the new route is painted. This causes `data-scroll-locked` to remain on `document.body`, making the page non-scrollable.

## Solution

Add a `useLayoutEffect` cleanup in `DialogOverlayImpl` that **synchronously** decrements (or removes) the `data-scroll-locked` attribute when the overlay unmounts while the dialog is still open (i.e. a forced unmount caused by router navigation).

### Double-decrement protection

A `isOpenRef` guard prevents interfering with the normal close flow:
- **Normal close**: `context.open` transitions to `false` *before* the overlay unmounts (via `Presence`), so `isOpenRef.current === false` → our cleanup skips, leaving `react-remove-scroll`'s own cleanup to handle the decrement normally ✅
- **Forced unmount (navigation)**: `context.open` is still `true` at unmount time → our `useLayoutEffect` cleanup fires synchronously and decrements the counter ✅

### Nested dialogs

When multiple dialogs are stacked (counter > 1), our cleanup correctly decrements to the next value rather than blindly removing the attribute. `react-remove-scroll`'s subsequent async decrement reads 0 and calls `removeAttribute`, which is a safe no-op on an already-absent attribute.

## Testing

Added a regression test (`scroll lock cleanup on forced unmount`) that:
1. Renders a controlled Dialog with `open={true}`
2. Waits for `react-remove-scroll`'s `useEffect` to set the attribute
3. Calls `root.unmount()` directly (simulating a router tearing down the page)
4. Asserts that `data-scroll-locked` is no longer present on `document.body`

The test passes with this fix and fails without it.

> **Note**: The existing test suite has pre-existing failures due to a `React.act is not a function` incompatibility between `@testing-library/react@16.3.0` and React 19. Those failures are unrelated to this change. The new regression test avoids `@testing-library/react`'s render path (which triggers the issue) and uses `ReactDOM.createRoot` directly.

### DIFF
--- a/.changeset/fix-dialog-scroll-lock-cleanup-on-unmount.md
+++ b/.changeset/fix-dialog-scroll-lock-cleanup-on-unmount.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-dialog': patch
+---
+
+Fix scroll lock not being released when Dialog is forcefully unmounted during SPA navigation. Added a `useLayoutEffect` cleanup in `DialogOverlayImpl` that synchronously removes `data-scroll-locked` from `document.body` when the overlay unmounts while the dialog is still open (e.g. route change triggered by a Link inside the Dialog).
--- a/packages/react/dialog/package.json
+++ b/packages/react/dialog/package.json
@@ -47,6 +47,7 @@
     "@radix-ui/react-primitive": "workspace:*",
     "@radix-ui/react-slot": "workspace:*",
     "@radix-ui/react-use-controllable-state": "workspace:*",
+    "@radix-ui/react-use-layout-effect": "workspace:*",
     "aria-hidden": "^1.2.4",
     "react-remove-scroll": "^2.6.3"
   },
--- a/packages/react/dialog/src/dialog.test.tsx
+++ b/packages/react/dialog/src/dialog.test.tsx
@@ -139,3 +139,62 @@ describe('given a default Dialog', () => {
     });
   });
 });
+
+describe('scroll lock cleanup on forced unmount', () => {
+  /**
+   * Regression test for: https://github.com/radix-ui/primitives/issues/3797
+   *
+   * When a Dialog is open and its parent component is forcefully unmounted
+   * (e.g. during SPA route navigation), the `data-scroll-locked` attribute
+   * should be removed from `document.body` synchronously, preventing the
+   * page from remaining in a non-scrollable state.
+   *
+   * We use ReactDOM.createRoot directly to avoid @testing-library/react's
+   * dependency on the deprecated ReactDOM.act from react-dom/test-utils.
+   */
+  let container: HTMLDivElement;
+  let root: import('react-dom/client').Root;
+
+  beforeEach(() => {
+    container = document.createElement('div');
+    document.body.appendChild(container);
+  });
+
+  afterEach(() => {
+    // Ensure no leftover attributes between tests
+    document.body.removeAttribute('data-scroll-locked');
+    if (root) {
+      root.unmount();
+    }
+    container.remove();
+  });
+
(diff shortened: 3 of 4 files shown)


sujiu222:
Thanks for the review! Addressed all three points:

1. **SSR-safe `useLayoutEffect`**: Switched from `React.useLayoutEffect` to `@radix-ui/react-use-layout-effect` for consistency with the rest of the codebase (e.g. `portal.tsx`). Also added `@radix-ui/react-use-layout-effect` to the `dialog` package dependencies.

2. & 3. **`getRootNode()` guard for `DocumentFragment`**: Added `instanceof Document || instanceof ShadowRoot` check before calling `getElementById`. Falls back to `contentRef.current?.ownerDocument ?? document` when the root node is a `DocumentFragment` or any other node type that doesn't implement `getElementById`. Applied to both `TitleWarning` and `DescriptionWarning`.

chaance:
@sujiu222 can you rebase this on main and isolate your changes to only the single scroll-lock cleanup you're targeting?

PR #3862, opened 2026-04-20
fix: remove dialog description warning when aria-describedby is optional

### LINKED ISSUE #3855: Dialog component shouldn't warn when there's no description
### Current Behavior

When a dialog is created without a description, there's a warning (in non-production environments) that there's a missing description ([code reference](https://github.com/radix-ui/primitives/blob/main/packages/react/dialog/src/dialog.tsx#L536)):
```
Warning: Missing `Description` or `aria-describedby={undefined}` for {DialogContent}
```

### Expected behavior

Creating a dialog without a description shouldn't have a warning. 

In the [APG dialog pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/), `aria-describedby` is optional, and there are legitimate reasons to not need it:
> Optionally, the [aria-describedby](https://w3c.github.io/aria/#aria-describedby) property is set on the element with the dialog role to indicate which element or elements in the dialog contain content that describes the primary purpose or message of the dialog. Specifying descriptive elements enables screen readers to announce the description along with the dialog title and initially focused element when the dialog opens, which is typically helpful only when the descriptive content is simple and can easily be understood without structural information. It is advisable to omit specifying aria-describedby if the dialog content includes semantic structures, such as lists, tables, or multiple paragraphs, that need to be perceived in order to easily understand the content, i.e., if the content would be difficult to understand when announced as a single unbroken string.

Since it's not something that's required by the pattern, and there are reasons that it wouldn't be used, having a warning logged adds extra noise that can confuse developers into adding a description when it's not necessary.

### Reproducible example

[Basic modal example sandbox](https://codesandbox.io/p/devbox/nervous-sun-38p6dl?workspaceId=ws_G4MKkVzFwB1BaysuMEtUk7)

### Suggested solution

Only apply the `aria-describedby` attribute if a description is provided, and don't warn when a description isn't provided.

### Your environment

| Radix Package(s) | `react-dialog`        |  1.1.7       |
| Browser          |  n/a       |         |
| Assistive tech   |  n/a       |         |
| npm/yarn/pnpm    | n/a        |         |
| Operating System | n/a        |         |

mixelburg:
Fixes #3855

## Problem

When a Dialog is created without a `DialogDescription` component, a console warning fires:
```
Warning: Missing `Description` or `aria-describedby={undefined}` for {DialogContent}.
```

However, per the [APG dialog pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/), `aria-describedby` is optional. There are legitimate reasons to omit it, especially when dialog content includes semantic structures (lists, tables, multiple paragraphs) that would be difficult to understand as a single unbroken string by screen readers.

## Fix

- `aria-describedby` is now only set on `DialogContent` when a `DialogDescription` component is actually rendered
- `DialogDescription` registers itself with the Dialog context via `useEffect` on mount
- The console warning for missing description is removed entirely
- The `aria-labelledby` attribute (for `DialogTitle`) is unchanged — title remains required

## Changes

- `DialogContextValue` — added `hasDescription: boolean` and `setHasDescription`
- `Dialog` root — creates `hasDescription` state, passes through context
- `DialogContentImpl` — conditionally sets `aria-describedby`
- `DialogDescription` — registers itself via useEffect
- `dialog.test.tsx` — updated test expectation (no warning when no description)
- All 8 existing tests pass

### DIFF
--- a/.changeset/dialog-description-warning.md
+++ b/.changeset/dialog-description-warning.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-dialog': patch
+---
+
+Remove console warning when dialog has no description; only set aria-describedby when DialogDescription is rendered
--- a/packages/react/dialog/src/dialog.test.tsx
+++ b/packages/react/dialog/src/dialog.test.tsx
@@ -87,8 +87,8 @@ describe('given a default Dialog', () => {
     });
 
     describe('when no description has been provided', () => {
-      it('should warn to the console', () => {
-        expect(consoleWarnMockFunction).toHaveBeenCalledTimes(1);
+      it('should not warn to the console', () => {
+        expect(consoleWarnMockFunction).not.toHaveBeenCalled();
       });
     });
 
--- a/packages/react/dialog/src/dialog.tsx
+++ b/packages/react/dialog/src/dialog.tsx
@@ -31,6 +31,8 @@ type DialogContextValue = {
   contentId: string;
   titleId: string;
   descriptionId: string;
+  hasDescription: boolean;
+  setHasDescription(value: boolean): void;
   open: boolean;
   onOpenChange(open: boolean): void;
   onOpenToggle(): void;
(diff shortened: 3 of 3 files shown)


mixelburg:
Closing this PR as it's been superseded by upstream changes. The  and  components have been removed entirely from the dialog implementation, making this fix no longer necessary. Happy to reopen with a fresh approach if the issue persists on the current main branch.

PR #3864, opened 2026-04-22
feat(tooltip): add overflow-only trigger support

mayank1513:
### Description

Add `onlyShowOnOverflow` and `overflowTolerance` to `Tooltip.Trigger` to enhance UX for overflow-only tooltip use cases. `onlyShowOnOverflow` defaults to `false`, preserving existing behavior and accessibility unless explicitly enabled. Even when explicitly enabled it preserves accessibility - only gates hover.

### DIFF
--- a/.changeset/smart-laws-drum.md
+++ b/.changeset/smart-laws-drum.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-tooltip': minor
+---
+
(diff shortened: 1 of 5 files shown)


mayank1513:
I would prefer `overflorOnly` instead of `onlyShowOnOverflow`. However, I saw some longer descriptive names in the code. So preferred clarity.

What do you prefer?

mayank1513:
https://github.com/radix-ui/primitives/issues/3865

chaance:
No plans to implement at this time.

PR #3866, opened 2026-04-28
feat(direction): auto-detect RTL from HTML document when no DirectionProvider is present (#3830)

### LINKED ISSUE #3830: feat(direction): auto-detect RTL from HTML document when no DirectionProvider is present

mixelburg:
## What

Fixes #3830

When no `DirectionProvider` exists in the React tree, `useDirection()` currently falls back to a hardcoded `'ltr'` — ignoring `<html dir="rtl">` entirely. This commit adds a fallback to `document.documentElement.dir` so RTL apps work correctly out of the box.

## How

Changed `useDirection` to read `document.documentElement.dir` as a last-resort fallback before defaulting to `'ltr'`:

```ts
function useDirection(localDir?: Direction) {
  const globalDir = React.useContext(DirectionContext);
  const documentDir =
    typeof document !== 'undefined'
      ? document.documentElement.dir === 'rtl' ? 'rtl' : 'ltr'
      : 'ltr';
  return localDir || globalDir || documentDir;
}
```

## Why this is safe

- **No breaking changes**: `DirectionProvider` still takes precedence — existing apps are unaffected
- **LTR apps unchanged**: Most apps don't set `<html dir="rtl">`, so `documentDir` defaults to `'ltr'`
- **SSR safe**: `typeof document !== 'undefined'` guard handles server-side rendering
- **Dynamic**: Works with runtime language switching (when `dir` changes on `<html>`)
- **Zero configuration**: RTL apps work correctly without any wrapper

## Who this helps

Arabic, Hebrew, Persian, and Urdu speakers (~500 million people). Every app using Radix that sets `<html dir="rtl">` currently has broken RTL layout for dropdowns, dialogs, and popovers.

## Checks

- [x] Lint passes (`pnpm -r --filter @radix-ui/react-direction run lint`)
- [x] Typecheck passes (`pnpm -r --filter @radix-ui/react-direction run typecheck`)

### DIFF
--- a/packages/react/direction/src/direction.tsx
+++ b/packages/react/direction/src/direction.tsx
@@ -20,7 +20,11 @@ const DirectionProvider: React.FC<DirectionProviderProps> = (props) => {
 
 function useDirection(localDir?: Direction) {
   const globalDir = React.useContext(DirectionContext);
-  return localDir || globalDir || 'ltr';
+  const documentDir =
+    typeof document !== 'undefined'
+      ? document.documentElement.dir === 'rtl' ? 'rtl' : 'ltr'
+      : 'ltr';
+  return localDir || globalDir || documentDir;
 }
 
 const Provider = DirectionProvider;


chaance:
No plans to address at this time. The implementation here would result in a hydration mismatch. This is better handled by consumer code since folks may have reasonably different approaches as to how they handle the fallback on the server if the value provided to the root node is not accessible.

PR #3874, opened 2026-05-09
fix(tooltip): skipDelayDuration of 0 actually prevents delay skipping

### LINKED ISSUE #3873: [Tooltip] `skipDelayDuration` of 0 does not actually prevent the delay from being skipped
### Current Behavior

When using a `skipDelayDuration` of 0 it is still possible for tooltips to open instantly when moving between triggers because of the usage of setTimeout.

### Expected behavior

A `skipDelayDuration` of 0 should never allow the delay to be skipped.

### Context

Not a duplicate of #1110 because I am interested in completely deactivating the skip behaviour and dont care about acurate durations.

### Suggested solution

In the [TooltipProvider](https://github.com/radix-ui/primitives/blob/22473d16404bfd446305db5b6c9308aece99fdec/packages/react/tooltip/src/tooltip.tsx#L88) the onOpen and onClose logic is always run even if the skip duration is 0. Everytime a tooltip opens `isOpenDelayedRef.current` is set to `true` and onClose the ref is set to `false` again. But because a setTimeout of 0 will cause the callback to be executed in the next event cycle or if onClose fires a little late - it is possible for a different tooltip to open instantly, even if the `skipDelayDuration` is 0.

By simply not performing any action onClose or onOpen if the `skipDelayDuration` is 0, the `isOpenDelayedRef` would always remain in its default state of `true`. 

Simply adding `if(skipDelayDuration == true) return` to both onOpen and onClose would do the trick.

mixelburg:
Fixes #3873

## Problem

When `skipDelayDuration` is set to 0, the onOpen and onClose handlers still toggled `isOpenDelayedRef` (via a 0ms setTimeout in onClose). This meant that moving between two tooltip triggers could cause the second tooltip to open instantly, even though the user explicitly set skipDelayDuration=0 to deactivate skip behavior entirely.

## Root cause

The `onOpen` callback always set `isOpenDelayedRef.current = false`, and `onClose` always scheduled a setTimeout(0) to set it back to `true`. With skipDelayDuration=0, the setTimeout fires in the next event-loop tick — so any tooltip opened before that tick sees the skip state and opens without delay.

## Fix

Added an early return in both `onOpen` and `onClose` when `skipDelayDuration <= 0`. When skip is disabled, the ref stays at its default `true` and no timer is scheduled, so every open always goes through the normal delay path.

Added `skipDelayDuration` to the dependency arrays of both callbacks (previously missing from onOpen).

### DIFF
--- a/packages/react/tooltip/src/tooltip.tsx
+++ b/packages/react/tooltip/src/tooltip.tsx
@@ -86,10 +86,12 @@ const TooltipProvider: React.FC<TooltipProviderProps> = (
       isOpenDelayedRef={isOpenDelayedRef}
       delayDuration={delayDuration}
       onOpen={React.useCallback(() => {
+        if (skipDelayDuration <= 0) return;
         window.clearTimeout(skipDelayTimerRef.current);
         isOpenDelayedRef.current = false;
-      }, [])}
+      }, [skipDelayDuration])}
       onClose={React.useCallback(() => {
+        if (skipDelayDuration <= 0) return;
         window.clearTimeout(skipDelayTimerRef.current);
         skipDelayTimerRef.current = window.setTimeout(
           () => (isOpenDelayedRef.current = true),


chaance:
@mixelburg Happy to merge this. Can you add a changeset?

PR #3876, opened 2026-05-12
fix(Select): prevent scroll from hidden bubble input in constrained layouts

### LINKED ISSUE #3875: [Select] Document scrolls due to hidden `<select>` element in forms with constrained layouts
### Current Behavior
When Radix's Select is used in a `<form>` with a constrained layout (i.e. nested flex containers with `height: 100dvh`), Radix positions the native hidden `<select>` element low enough on the page that it causes the page to extend past its normal scrollable area.

Video demo:

https://github.com/user-attachments/assets/71192612-a6af-42e5-a580-50698b2270aa

### Expected behavior

The Select wouldn't cause this scroll to occur, regardless of what the configuration is of its containing elements is.

### Reproducible example

[CodeSandbox example](https://codesandbox.io/p/sandbox/boring-dijkstra-g83zkh?file=%2Fsrc%2FApp.tsx)

Note: to reproduce, use a smaller viewport and scroll the preview down.

### Suggested solution

Potentially ensuring the wrapper clips the content within it properly; this is something we've had to do as a workaround to avoid the issue.

### Additional context

Seems to be partially a combination of it being within a `<form>` and partially due to `min-height: 0` properties

### Your environment

| Radix Package(s) |  radix-select   | @1, @2 (in sandbox)         |
| React            | 18     |         |
| Browser          |   Chrome, Safari      |         |
| npm/yarn/pnpm    | yarn        |         |
| Operating System |   MacOS   |         |

mixelburg:
## Description

The hidden native select element (SelectBubbleInput) uses position: absolute (from VISUALLY_HIDDEN_STYLES) but does not pin its location to the top-left of its containing block. In constrained flex layouts (e.g. nested flex containers with height: 100dvh inside a form), the element is positioned low on the page, causing it to extend the scrollable area.

## Fix

Add top: 0 and left: 0 to the SelectBubbleInput style to pin it to the top of its containing block, preventing the layout extension.

Fixes #3875

### DIFF
--- a/packages/react/select/src/select.tsx
+++ b/packages/react/select/src/select.tsx
@@ -1808,7 +1808,7 @@ const SelectBubbleInput = React.forwardRef<SelectBubbleInputElement, SelectBubbl
         form={form}
         onChange={(event) => onValueChange(event.target.value)}
         {...props}
-        style={{ ...VISUALLY_HIDDEN_STYLES, ...props.style }}
+        style={{ ...VISUALLY_HIDDEN_STYLES, top: 0, left: 0, ...props.style }}
         ref={composedRefs}
         defaultValue={selectValue}
       >


mixelburg:
Closing as this hasn't received any feedback. Happy to reopen if there's interest.

PR #3879, opened 2026-05-22
feat(select): support multiple selection via `multiple` prop

### LINKED ISSUE #1270: [Select] Support multiple selections

MaxQian888:
### Description

Implements `multiple` support on `Select.Root`. Closes #1270.

This builds on the design explored in #3522 (thanks @Zach-Jaensch) and addresses gaps in that PR around accessibility, positioning, and trigger behavior so multi-select stays consistent with the rest of Radix.

#### Changes

- **API** — `Select.Root` accepts a `multiple` prop. With `multiple: true`, `value` / `defaultValue` are `string[]` and `onValueChange` receives the new array. The single-select branch is unchanged — no migration required.
- **Selection** — Selecting an item in multi-mode toggles it without closing the popover. Single-select close-on-select behavior is preserved.
- **Native form** — `SelectBubbleInput` renders the hidden `<select>` with the `multiple` attribute and sets `option.selected` imperatively, so the same code path serves single and multi. `FormData` carries one entry per selected option.
- **A11y** — `aria-multiselectable="true"` is set on the listbox, and `aria-selected` on items reflects real selection state in multi (the existing VoiceOver `isFocused` workaround is preserved only for single-select).
- **Positioning** — `Select.Content` defaults to `position="popper"` in multi-select because the item-aligned positioner has no single anchor.
- **Trigger label** — The `Select.ItemText` portal into `Select.Value` is skipped in multi-select to avoid collisions on the same DOM node; multi-select callers pass \`children\` to `Select.Value` (see the new stories for the pattern).
- **Typeahead** — Closed-state typeahead on the trigger is a no-op in multi-select since there is no implicit \"current\" value to cycle.

#### Storybook

Three new stories were added next to the existing ones:

- \`MultiSelectStyled\` — uncontrolled-style demo with default value
- \`MultiSelectControlled\` — controlled, showing the recommended \`Select.Value\` children pattern
- \`MultiSelectWithinForm\` — verifies \`FormData\` produces one entry per selected option

#### Tests

The Select package currently has no unit tests in-repo; verification was done via the new Storybook stories plus \`pnpm typecheck\` and \`pnpm lint\` (both clean across the workspace, including \`@repo/storybook\`). Happy to add tests if you'd like — the test file from #3522 could be a starting point.

#### Compared to #3522

This PR keeps the API shape from #3522 (the \`multiple\` discriminated union and the \`option.selected\` form-integration trick) and adds the changes needed to make multi-select land cleanly: \`aria-multiselectable\`, the \`aria-selected\` fix, the popper fallback, the no-op trigger typeahead, and the \`ItemText\` portal skip. Happy to coordinate with #3522 if you'd prefer to land that one first and layer these fixes on top.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/multi-select-clouds-sing.md
+++ b/.changeset/multi-select-clouds-sing.md
@@ -1,0 +1,6 @@
+---
+'@radix-ui/react-select': minor
+'@repo/storybook': patch
+---
+
+Add multi-select support via a new `multiple` prop on `Select.Root`. When `multiple` is `true`, `value` / `defaultValue` accept `string[]`, `onValueChange` receives the new array, and selecting an item toggles it without closing the popover. The hidden native `<select>` is rendered with the `multiple` attribute so `FormData` carries every selected option. The listbox advertises `aria-multiselectable="true"`, `aria-selected` reflects the real selection regardless of focus, and `Select.Content` defaults to `position="popper"` because the item-aligned positioner has no single anchor in multi-select.
--- a/apps/storybook/stories/select.stories.tsx
+++ b/apps/storybook/stories/select.stories.tsx
@@ -59,6 +59,119 @@ export const Styled = () => (
   </div>
 );
 
+export const MultiSelectStyled = () => {
+  const [value, setValue] = React.useState<string[]>(['two']);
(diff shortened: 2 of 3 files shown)


chaance:
No plans on adding at this time.

PR #3927, opened 2026-06-04
fix(select): use overflowX/overflowY instead of overflow shorthand in Viewport

### LINKED ISSUE #2059: ScrollArea inside a Select generates overflow warnings
### Current Behavior

If you compose a ScrollArea within a Select Primitive, as shown in the docs a lot of warnings are being generated in the console
```
Warning: Updating a style property during rerender (overflowY) when a conflicting property is set (overflow) can lead to styling bugs. To avoid this, don't mix shorthand and non-shorthand properties for the same value; instead, replace the shorthand with separate values.
```
image

### Expected behavior

I would expect no warnings for this case.

### Reproducible example

[Sandbox with a no-styles select generating the warning](https://codesandbox.io/p/sandbox/sad-river-bjzupf?file=%2FApp.jsx&selection=%5B%7B%22endColumn%22%3A37%2C%22endLineNumber%22%3A6%2C%22startColumn%22%3A37%2C%22startLineNumber%22%3A6%7D%5D)

### Suggested solution

I am not 100% sure why this warning is being showed, but maybe if the `data-radix-scroll-area-viewport` component used a non-shorthand syntax it would be more safe?

### Your environment

| Software         | Name(s)                             | Version
| ---------------- | ----------------------------------- | ----------
| Radix Package(s) | react-select/react-scroll-area      | latest
| React            | n/a                                 |            
| Browser          | Chrome                              | Version 111.0.5563.110 (Official Build) (arm64)|
| Operating System | macOS Ventura                       | 13.1

LongTangGithub:
Closes #2059 

### Description

`Select.Viewport` was setting `overflow: 'hidden auto'` as a CSS shorthand. 

When a `ScrollArea` is composed inside `Select.Content` (as shown in the Radix docs), `ScrollArea.Viewport` sets `overflowX` and `overflowY` as separate longhand properties. 

React detects this as a conflict between a shorthand and longhands on the same element during re-render and logs a warning:

> Warning: Updating a style property during rerender (overflowY) when a conflicting property is set (overflow). To avoid this, don't mix shorthand and non-shorthand properties for the same value; instead, replace the shorthand with separate values.

### Fix

- Replace the `overflow` shorthand in `Select.Viewport` with explicit `overflowX: 'hidden'` and `overflowY: 'auto'` longhand properties. 

- This matches how `ScrollArea.Viewport` already sets its own overflow styles and eliminates the conflict.

### Manual Testing

A new Storybook story `WithScrollArea` has been added to reproduce this scenario. To verify:

1. Run `pnpm dev` and open **Components/Select → WithScrollArea**
2. Switch your DevTools console context to `storybook-preview-iframe`
3. **Before fix**: open the Select dropdown — the console shows a React style conflict warning
4. **After fix**: open the Select dropdown — no warning appears in the console

### Screenshot of Fix
no-console-warning

### DIFF
--- a/.changeset/five-zoos-smash.md
+++ b/.changeset/five-zoos-smash.md
@@ -1,0 +1,5 @@
+---
+"@radix-ui/react-select": patch
+---
+
+Fix React warning when composing ScrollArea inside Select by replacing overflow shorthand with overflowX/overflowY longhand properties in Select.Viewport
--- a/apps/storybook/stories/select.stories.tsx
+++ b/apps/storybook/stories/select.stories.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable react/jsx-pascal-case */
 import * as React from 'react';
 import { createPortal } from 'react-dom';
-import { Dialog, Select, Label as LabelPrimitive } from 'radix-ui';
+import { Dialog, Select, ScrollArea, Label as LabelPrimitive } from 'radix-ui';
 import { foodGroups } from '@repo/test-data/foods';
 import styles from './select.stories.module.css';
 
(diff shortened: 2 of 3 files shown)


apps/storybook/stories/select.stories.tsx
```diff
@@ -864,6 +864,48 @@ export const WithVeryLongSelectItems = () => (
   </div>
 );
 
+export const WithScrollArea = () => {
+  <div  style={{ padding: 50 }}>
```
chaance, line 868: You aren't actually returning the JSX here so this story renders nothing.

.changeset/five-zoos-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@radix-ui/react-select": patch
+---
+
+Fix React warning when composing ScrollArea inside Select by replacing overflow shorthand with overflowX/overflowY longhand properties in Select.Viewport
```
chaance, line 5:
```suggestion
Fixed React warning when composing ScrollArea inside Select due to conflicting inline `overflow` styles
```

chaance (changes requested):

PR #3943, opened 2026-06-07
avatar: Add `mode` prop

### LINKED ISSUE #2230: [Avatar] Allow `Avatar.Image` to be supplied a `component` to render the image

### LINKED ISSUE #2512: [Avatar] Avatar.Image doesn't implement lazy loading
### Current Behavior
The `<Avatar.Image />` accepts `loading="lazy"` but it actually has no effect, the image loads on page loading time.

Screenshot 2023-11-08 at 20 15 37
Screenshot 2023-11-08 at 20 09 59

### Expected behavior

The `<Avatar.Image />` with `loading="lazy"` provided loads image only when user navigates to it.

### Your environment

| Radix Package(s) |    @radix-ui/react-avatar    |    ^1.0.4     |

chaance:
This PR introduces a new `mode` prop for `Avatar.Image` for more control over loading state and rendering.

- `mode="default"` (default): Matches the existing behavior by rendering an `img` element conditionally based on the loading state of an `Image` object constructed after hydration. This ensures backwards compatibility.
- `mode="native"`: Renders an `img` element unconditionally using its event handlers to update loading state. In `native` mode, you can use CSS to target the image element and style it based on its loading state. Closes https://github.com/radix-ui/primitives/issues/2512 since the `lazy` attribute would dictate when the loading handler is called.
  ```tsx
  <Avatar.Root className="image-root">
    <Avatar.Image
      mode="native"
      className="image"
      alt="John Smith"
      src="./avatar.png"
    />
    <Avatar.Fallback className="fallback">
      <AvatarIcon />
    </Avatar.Fallback>
  </Avatar.Root>
  ```
  ```css
  .image-root {
    position: relative;
  }
  .image[data-radix-avatar-loading-status]:not(
      [data-radix-avatar-loading-status="loaded"]
    ) {
    /* hide the element visually until it's loaded to reveal the fallback */
    position: absolute;
    inset: 0;
    opacity: 0;
  }
  ```
- `mode="custom"`: Allows for more control over the image rendering. The `render` prop must be provided to render the image. This mode is useful when you want to use a custom image component, such as framework-specific image components or a design-system implementation that uses non-standard props. Closes https://github.com/radix-ui/primitives/issues/2230

  ```tsx
  <Avatar.Image
    mode="custom"
    className="image"
    alt="John Smith"
    src="./avatar.png"
    render={({ props, ref, context }) => (
      <CustomImage
        {...props}
        ref={ref}
        data-loading-status={context.loadingStatus}
        onLoadingComplete={({ naturalWidth, naturalHeight }) => {
          context.onLoadingStatusChange("loaded");
        }}
        onError={() => {
          context.onLoadingStatusChange("error");
        }}
      />
    )}
  />
  ```

### DIFF
--- a/.changeset/dry-cups-reply.md
+++ b/.changeset/dry-cups-reply.md
@@ -1,0 +1,58 @@
+---
+"@radix-ui/react-avatar": minor
+"radix-ui": minor
+---
+
+Added a new `mode` prop to `Avatar.Image` to allow for more control over the image rendering.
+
+- `mode="default"` (default): Matches the existing behavior by rendering an `img` element conditionally based on the loading state of an `Image` object constructed after hydration.
+- `mode="native"`: Renders an `img` element unconditionally using its event handlers to update loading state. In `native` mode, you can use CSS to target the image element and style it based on its loading state.
+  ```tsx
+  <Avatar.Root className="image-root">
+    <Avatar.Image
+      mode="native"
+      className="image"
+      alt="John Smith"
+      src="./avatar.png"
+    />
+    <Avatar.Fallback className="fallback">
+      <AvatarIcon />
+    </Avatar.Fallback>
+  </Avatar.Root>
+  ```
+  ```css
+  .image-root {
+    position: relative;
+  }
+  .image[data-radix-avatar-loading-status]:not(
+      [data-radix-avatar-loading-status="loaded"]
+    ) {
+    /* hide the element visually until it's loaded to reveal the fallback */
+    position: absolute;
+    inset: 0;
+    opacity: 0;
+  }
+  ```
+- `mode="custom"`: Allows for more control over the image rendering. The `render` prop must be provided to render the image. This mode is useful when you want to use a custom image component, such as framework-specific image components or a design-system implementation that uses non-standard props.
+
+  ```tsx
+  <Avatar.Image
+    mode="custom"
+    className="image"
+    alt="John Smith"
+    src="./avatar.png"
+    render={({ props, ref, context }) => (
+      <CustomImage
+        {...props}
+        ref={ref}
+        data-loading-status={context.loadingStatus}
+        onLoadingComplete={({ naturalWidth, naturalHeight }) => {
+          context.onLoadingStatusChange("loaded");
+        }}
+        onError={() => {
+          context.onLoadingStatusChange("error");
+        }}
+      />
+    )}
+  />
+  ```
--- a/packages/react/avatar/src/avatar.tsx
+++ b/packages/react/avatar/src/avatar.tsx
@@ -1,10 +1,10 @@
 import * as React from 'react';
-import { createContextScope } from '@radix-ui/react-context';
+import { createContextScope, type Scope } from '@radix-ui/react-context';
 import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
 import { useLayoutEffect } from '@radix-ui/react-use-layout-effect';
 import { Primitive } from '@radix-ui/react-primitive';
-
-import type { Scope } from '@radix-ui/react-context';
+import { composeEventHandlers } from '@radix-ui/primitive';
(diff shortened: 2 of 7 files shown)


thompsongl (commented):
LGTM. The `mode` enum makes sense to me and the implementation appears to provide the necessary & requested styling hooks

packages/react/avatar/src/avatar.tsx
```diff
@@ -65,45 +65,191 @@ const IMAGE_NAME = 'AvatarImage';
...
+
+  return (
+    <Primitive.img
+      {...imageProps}
+      data-radix-avatar-loading-status={imageLoadingStatus}
+      ref={useComposedRefs(
+        forwardedRef,
+        React.useCallback(
+          (node: HTMLImageElement | null) => {
+            if (!src) {
```
dcwither, line 181: default mode sets status error + fires onLoadingStatusChange when src is missing; native stays idle and never fires. Should align them or document the difference.

packages/react/avatar/src/avatar.tsx
```diff
@@ -65,45 +65,191 @@ const IMAGE_NAME = 'AvatarImage';
...
+          },
+          [setImageLoadingStatus, src],
+        ),
+      )}
+      src={src}
+      onError={composeEventHandlers(imageProps.onError, () =>
+        context.setImageLoadingStatus('error'),
+      )}
+      onLoad={composeEventHandlers(imageProps.onLoad, (event) => {
+        if (!src) {
```
dcwither, line 206: This `if (!src) return;` guard and comment are copied from the ref callback. onLoad can only fire once the image has a src and loads, so the branch looks unreachable here, and the comment (about not updating from error) describes the ref-callback case rather than this one.

packages/react/avatar/src/avatar.tsx
```diff
@@ -65,45 +65,191 @@ const IMAGE_NAME = 'AvatarImage';
...
+  useLayoutEffect(() => {
+    const previousLoadingStatus = loadingStatusRef.current;
+    loadingStatusRef.current = imageLoadingStatus;
+    if (imageLoadingStatus !== previousLoadingStatus) {
+      handleLoadingStatusChange(imageLoadingStatus);
+    }
+  }, [imageLoadingStatus, handleLoadingStatusChange]);
+
+  return (
+    <Primitive.img
```
dcwither, line 174: Always renders, which can lead to duplicate information between this and fallback on non-loaded states - should this be aria-hidden on non-loaded states?

.changeset/dry-cups-reply.md
```diff
@@ -0,0 +1,58 @@
...
+    <Avatar.Fallback className="fallback">
+      <AvatarIcon />
+    </Avatar.Fallback>
+  </Avatar.Root>
+  ```
+  ```css
+  .image-root {
+    position: relative;
+  }
+  .image[data-radix-avatar-loading-status]:not(
```
dcwither, line 27: `data-radix-avatar-loading-status` diverges from the `data-state` convention for consumer-facing state elsewhere (Progress exposes its loading lifecycle that way too) — I'm guessing this is an intentional direction change?

Also these are necessary to avoid duplicate UI, so it should maybe be called out more loudly.

PR #3964, opened 2026-06-13
fix(tooltip): don't reopen on window/tab refocus (#1800)

### LINKED ISSUE #1800: [Tooltip] Tooltip is opened on active buttons when the browser tab is switched
### Current Behavior

Steps:
- Go to https://www.radix-ui.com/docs/primitives/components/tooltip
- Click on the button
- Change browser tab
- Go back to the tooltip tab
- The tooltip state is open

This happens also when the browser window is minimized/maximized.

### Expected behavior

The tooltip shouldn't be open when the tab becomes active again

### Reproducible example

https://www.radix-ui.com/docs/primitives/components/tooltip

### Suggested solution

The issue is triggered by the [onFocus handler](https://github.com/radix-ui/primitives/blob/main/packages/react/tooltip/src/Tooltip.tsx#L301) because the browser triggers a focus event when the page becomes visible.

I've applied this workaround on my tooltips:

```js
  const prevActiveElement = useRef<EventTarget | null>(null);

  useEffect(() => {
    const handleFocus = (evt: Event) => {
      const { target } = evt;

      prevActiveElement.current = target;
    };

    window.addEventListener('focusin', handleFocus);

    return () => {
      window.removeEventListener('focusin', handleFocus);
    };
  }, []);

  return (
    <Tooltip.Root>
      <Tooltip.Trigger
        onFocus={(evt) => {
          // Blocks the tooltip open during the page visibility change
          // based on the assumption that we want to show the tooltip only when the focus
          // is triggered on a new element
          if (prevActiveElement.current === evt.target) {
            evt.preventDefault();
          }
        }}
```

To make the example small I've placed everything in the component, but it's better to track the previous active element inside a provider.

I'm migrating from [tippy.js](https://atomiks.github.io/tippyjs/) and while they handle focus they don't have the same issue.
I would also check there to find a possible solution.

### Your environment

| Radix Package(s) |     @radix-ui/react-tooltip    |    1.0.2    |
| React            | n/a     |    18.2     |
| Browser          |    Chrome     |  105       |
| Operating System |     Linux & MacOS    |         |

effinrich:
## Summary

Fixes #1800. A `Tooltip` whose trigger is focused reopens itself when you switch away from and back to the browser tab (or minimize/restore the window).

## Root cause

Returning to the tab refocuses the previously-active element, firing a synthetic `focus` event on the trigger. `TooltipTrigger`'s `onFocus` opens on any non-pointer focus, so it can't distinguish that refocus from a genuine keyboard focus — and opens the tooltip.

## Fix

Track window blur via a ref (mirrors the recently-merged window-blur handling for menus, #3938). When the trigger is focused immediately after the window was blurred, treat it as the tab/window regaining focus and skip opening — consuming the flag so the next genuine focus opens normally.

## Test

Adds a regression test using `window.dispatchEvent(new FocusEvent('blur'))` then `fireEvent.focus(trigger)`, asserting `data-state` stays `closed`, and that a subsequent genuine focus still opens. Changeset included (`patch`: `@radix-ui/react-tooltip`, `radix-ui`).

## Tradeoff (open to feedback)

If the window is blurred and the user later keyboard-focuses a *different, never-focused* trigger, that trigger's first focus is suppressed once. Minor — happy to switch to a `requestAnimationFrame`-on-window-`focus` reset to eliminate it if preferred.

### DIFF
--- a/.changeset/tooltip-window-refocus.md
+++ b/.changeset/tooltip-window-refocus.md
@@ -1,0 +1,6 @@
+---
+'@radix-ui/react-tooltip': patch
+'radix-ui': patch
+---
+
+Fixed a bug where the tooltip opened when switching back to the browser tab while its trigger was focused.
--- a/packages/react/tooltip/src/tooltip.test.tsx
+++ b/packages/react/tooltip/src/tooltip.test.tsx
@@ -83,6 +83,41 @@ describe('Tooltip', () => {
     });
   });
 
+  // Regression test for https://github.com/radix-ui/primitives/issues/1800
+  // Switching away from and back to the browser tab refocuses the previously-focused
+  // trigger, firing a synthetic `focus` event that must not reopen the tooltip.
+  it('does not open on focus caused by the window regaining focus', () => {
+    render(
+      <Tooltip.Provider>
+        <Tooltip.Root delayDuration={0}>
+          <Tooltip.Trigger>Tooltip Trigger</Tooltip.Trigger>
+          <Tooltip.Portal>
+            <Tooltip.Content>Tooltip Content</Tooltip.Content>
+          </Tooltip.Portal>
+        </Tooltip.Root>
+      </Tooltip.Provider>,
+    );
+
+    const trigger = screen.getByText('Tooltip Trigger');
+
+    // A genuine focus opens the tooltip.
+    act(() => void fireEvent.focus(trigger));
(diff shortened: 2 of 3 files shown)


effinrich:
@chaance When you get a chance to review this, can you give me some feedback regarding suggested improvements in my fix?

PR #3968, opened 2026-06-15
Stabilize all composed refs

chaance:
Filling in the gaps in https://github.com/radix-ui/primitives/pull/3967 and adding test coverage.

### DIFF
--- a/.changeset/quick-readers-cheer.md
+++ b/.changeset/quick-readers-cheer.md
@@ -1,0 +1,14 @@
+---
+'@radix-ui/react-dismissable-layer': patch
+'@radix-ui/react-dropdown-menu': patch
+'@radix-ui/react-navigation-menu': patch
+'@radix-ui/react-focus-scope': patch
+'@radix-ui/react-scroll-area': patch
+'@radix-ui/react-select': patch
+'@radix-ui/react-popper': patch
+'@radix-ui/react-slider': patch
+'@radix-ui/react-toast': patch
+'@radix-ui/react-menu': patch
+---
+
+Fixed infinite re-render loop in React 19 caused by unstable composed ref callbacks being recreated on every render.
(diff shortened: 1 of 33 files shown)


petrarca:
Thanks for this work — #3967 and this PR fix the known call sites, but `useComposedRefs` itself still has the unstable identity:

```ts
function useComposedRefs<T>(...refs: PossibleRef<T>[]): React.RefCallback<T> {
  // eslint-disable-next-line react-hooks/exhaustive-deps
  return React.useCallback(composeRefs(...refs), refs);
}
```

The `refs` spread creates a new array on every render, so the `useCallback` dep check always fails and the composed ref gets a new identity every commit. Any caller that passes an inline ref or a state setter not already stabilized via `useCallbackRef` will reproduce the infinite loop.

We've been running this fix in production across Tooltip, Select, Popover and Dialog with React 19.2:

```ts
function useComposedRefs<T>(...refs: PossibleRef<T>[]): React.RefCallback<T> {
  const refsRef = React.useRef(refs);
  refsRef.current = refs;
  return React.useCallback((node: T) => composeRefs(...refsRef.current)(node), []);
}
```

The `refsRef` holds the latest refs without being a dep, so the callback identity is stable across renders. Would be happy to add this to the PR if it fits the direction you're going.

chaance:
@petrarca I don't believe your comment is quite right. `useCallback(fn, deps)` does not compare the deps array _by reference_, and it doesn't care about the function identity at all (that's why we need the deps array in the first place). Every function passed into `useCallback` gets a new identity every commit, which is the entire reason we need useCallback in the first place.

> Any caller that passes an inline ref or a state setter not already stabilized via `useCallbackRef` will reproduce the infinite loop.

This is correct, and it's the known footgun I'm addressing with a script to check call-site usage in https://github.com/radix-ui/primitives/pull/3968/commits/fd0d78e72d0db30e6c568a5e03eea6ea5d83f721.

PR #3974, opened 2026-06-23
fix(select): prevent text selection on items

### LINKED ISSUE #1488: [Dialog/Select] Selection mode is active after choosing an item in Select in Firefox
### Current Behavior

Following this discussion https://github.com/radix-ui/primitives/discussions/1484#discussioncomment-2993174.

Using Dialog `0.1.8-rc.47` and Select `0.1.2-rc.49` in `Firefox 101.0.1`, after choosing an item in Select, selection mode is activated and text content in Dialog is highlighted on hover.

Screenshot 2022-06-21 at 20 36 20

### Expected behavior

After selecting an item in Select component, text in Dialog is not highlighted on hover.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/dawn-wind-xohv43?file=/App.js)

Choose `option 1` from the select and hover on the text `Very basic modal`, it will be on selection mode and highlight the text.

fayonation:
## Summary

- Sets `userSelect: 'none'` on Select items to prevent accidental text selection when interacting with the dropdown

Fixes #1488

## Problem

In Firefox (especially inside dialogs), dragging over Select options can highlight item text instead of selecting the option.

## Test plan

- [x] `pnpm vitest run packages/react/select/src/select.test.tsx`
- [x] Storybook **Within Dialog** — options have `user-select: none`
- [ ] CI passes on PR

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/fix-select-item-user-select-1488.md
+++ b/.changeset/fix-select-item-user-select-1488.md
@@ -1,0 +1,5 @@
+---
+'@radix-ui/react-select': patch
+---
+
+Prevent text selection on Select items to avoid Firefox leaving selection mode active after closing inside a Dialog.
--- a/packages/react/select/src/select.test.tsx
+++ b/packages/react/select/src/select.test.tsx
@@ -140,3 +140,14 @@ describe('clearing an optional value (#2706)', () => {
     expect(emptyOptions).toHaveLength(1);
   });
 });
+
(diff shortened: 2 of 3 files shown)


chaance:
This is something you can easily handle in userland.

PR #3976, opened 2026-06-23
fix(menu): defer outside pointer interactions on menu content

### LINKED ISSUE #3971: Clicking outside a DropdownMenu inside a Dialog also dismisses the dialog
### Current Behavior

If a `Dialog` contains an open `DropdownMenu`, and you click outside the `DropdownMenu`, both the `DropdownMenu` and the `Dialog` are closed.

https://github.com/user-attachments/assets/66e6128b-be85-4d31-9f3b-32b8d142a259

### Expected behavior

Only the `DropdownMenu` should be closed. (The `Dialog` should stay open.)

### Reproducible example

https://stackblitz.com/edit/vitejs-vite-x8tgg65b?file=src%2FApp.tsx

### Suggested solution

I suspect this might be happening because the new `deferPointerDownOutside` logic in https://github.com/radix-ui/primitives/pull/3928 gives React time to re-render in between [this event handler](https://github.com/radix-ui/primitives/blob/71a7122a8f52cd2146a4fd0e5ac4016e394f6e9c/packages/react/dismissable-layer/src/dismissable-layer.tsx#L99-L113) running on the `DropdownMenu`'s `DismissableLayer` and the same event handler running on the `Dialog`'s `DismissableLayer`. So by the time the `Dialog`'s layer checks `isPointerEventsEnabled`, it sees the state where the `DropdownMenu` has already been closed.

I don't know if adding `deferPointerDownOutside` to `Menu` would be enough to fix this, or if you'd need to move where the `isPointerEventsEnabled` check happens to somewhere before you wait for the `click` event. (Then again, I'm not really familiar with this code and I don't totally understand what it's trying to achieve, so maybe the problem is something else.)

### Your environment

| Radix Package(s) |  `@radix-ui/react-dialog`, `@radix-ui/react-dropdown-menu`     |  1.1.17, 2.1.18       |
| React            | n/a     |   19.2.7      |
| Browser          | Google Chrome        |  149.0.7827.115       |
| Assistive tech   |  n/a       |         |
| Node             | n/a     |    22.22.3     |
| npm/yarn/pnpm    |   npm      | 10.8.2        |
| Operating System |    macOS     | 26.5.1        |

fayonation:
## Summary

- Adds `deferPointerDownOutside` to menu `DismissableLayer`, matching Popover and Dialog behavior
- Adds a regression test and Storybook **Dropdown Within Dialog** reproduction story

Fixes #3971

## Problem

Clicking outside a DropdownMenu that is open inside a Dialog dismisses both the menu and the dialog, because the outside pointer event propagates to the dialog layer.

## Test plan

- [x] `pnpm vitest run packages/react/dismissable-layer/src/dismissable-layer.test.tsx`
- [x] Storybook **Dropdown Within Dialog** — clicking inside the dialog but outside the menu closes only the menu
- [ ] CI passes on PR

Made with [Cursor](https://cursor.com)


jwld:
We've just encountered this, a deployed fix would be great ❤️

fayonation:
looks like a different fix is merged from another user, closing this

PR #3990, opened 2026-07-01
feat(select, switch): Add optimistic async react actions to select and switch

### LINKED ISSUE #3989: Support react 19 action props
### Overview

Radix Select and Switch should expose action-style props for user-driven value changes:

```tsx
<Select.Root
  value={plan}
  valueChangedAction={async (nextPlan) => {
    await updateSubscriptionPlan(nextPlan);
  }}
>
  {/* ... */}
</Select.Root>
```

```tsx
<Switch.Root
  checked={notificationsEnabled}
  checkedChangedAction={async (nextChecked) => {
    await updateNotificationPreference(nextChecked);
  }}
>
  <Switch.Thumb />
</Switch.Root>
```

The existing synchronous callbacks, such as `onValueChange` and `onCheckedChange`, should continue to work for local state updates, analytics, and other synchronous observation. The action props are for changes that start async work, such as persistence, server mutations, async validation, or settings updates.

When an action prop is used, the primitive should optimistically display the requested value immediately using React's optimistic update semantics, invoke the action in a transition, and guard against duplicate/conflicting interactions while that action is pending. Styling remains user-owned through data attributes such as `data-pending` and `data-optimistic`; Radix should not prescribe loading visuals.

Because this behavior depends on modern React action and optimistic update APIs, action props should fail fast when used outside React 19.2 or newer.

- Astryx, Meta's design system, exposes action-style component APIs: https://astryx.atmeta.com/
- React Actions: https://react.dev/blog/2024/12/05/react-19#actions
- React `useOptimistic`: https://react.dev/reference/react/useOptimistic
- React `useTransition`: https://react.dev/reference/react/useTransition

This is for product teams building settings screens, dashboards, forms, and account-management flows where changing a Select or Switch often starts async work.

Without action props, users repeatedly bridge async work into primitives manually:

```tsx
const [isSaving, setIsSaving] = React.useState(false);

async function handleCheckedChange(nextChecked: boolean) {
  setIsSaving(true);
  await savePreference(nextChecked);
  setChecked(nextChecked);
  setIsSaving(false);
}

<Switch.Root
  checked={checked}
  disabled={isSaving}
  onCheckedChange={handleCheckedChange}
/>
```

That works, but every app has to reimplement pending state, duplicate-interaction guards, optimistic feedback, and reconciliation. Moving the interaction boundary into the primitive gives users a consistent async path while preserving existing synchronous callbacks.

### Additional context

This keeps Radix headless:

- Radix owns action invocation, transition wrapping, optimistic value display, duplicate-interaction guarding, accessibility semantics, and state data attributes.
- Users own loading visuals, pending text, spinners, toasts, disabled styling, and broader application UI.
- Errors from rejected actions should surface to React or framework-level error handling rather than being swallowed by the primitive.

The initial scope can stay focused on value-changing primitives where delayed visual feedback feels broken, starting with Select and Switch.

suneettipirneni:
Adds React 19.2+ async action props for value-changing primitives:

- `Select.Root` now supports `valueChangedAction`
- `Switch.Root` now supports `checkedChangedAction`
- Both controls use optimistic display state while the action is pending
- Both guard duplicate/conflicting interactions while their invoked action is pending

The new action props are intended for async mutations such as persistence, validation, or server updates. Userland owns action state with React’s `useActionState`, while the primitive owns the immediate optimistic control feedback and interaction dedupe.

The action props fail fast when used outside React 19.2+ or when required React action APIs are unavailable.

Closes #3989


chaance:
This is a really cool idea, but it adds a decent level of complexity and maintenance overhead I don't think we're currently able to manage. While not ideal, you can implement this behavior in your own code, then wrap Radix components with your own abstraction.

We may reconsider this in the future.

PR #4007, opened 2026-07-03
fix(checkbox): restore value when externally associated form is reset

koreahghg:
### Description

`Checkbox` only listened for the form `reset` event via `control?.form` (i.e. an ancestor `<form>` element). Unlike `RadioGroup`, `Select`, `Slider`, and `Switch` (fixed in #3958), it never resolved the `form` prop by id via `ownerDocument.getElementById(form)`, so a `Checkbox` associated with an external form (`<Checkbox.Root form="my-form" />` rendered outside `<form id="my-form">`) did not restore its value when that form was reset.

This brings `Checkbox` in line with the fix already applied to the other form controls in #3958.

- `packages/react/checkbox/src/checkbox.tsx`: resolve the associated form the same way `Switch`/`RadioGroup` do.
- `packages/react/checkbox/src/checkbox.test.tsx`: added controlled/uncontrolled tests for both ancestor-form and external (`form` prop) reset, mirroring `switch.test.tsx`. Ran `pnpm test` — 30/30 passing (26 existing + 4 new). Confirmed the 2 new external-form tests fail without the fix and pass with it.
- `apps/storybook/stories/checkbox.stories.tsx`: added a `WithinFormReset` story (mirrors Switch's) covering both the ancestor-form and external-form cases.
- Added a changeset.

### DIFF
--- a/.changeset/checkbox-form-reset.md
+++ b/.changeset/checkbox-form-reset.md
@@ -1,0 +1,6 @@
+---
+"@radix-ui/react-checkbox": patch
+"radix-ui": patch
+---
+
+Fixed a bug where `Checkbox` did not restore its value when an externally associated form (via the `form` prop) was reset. This brings `Checkbox` in line with the fix already applied to `RadioGroup`, `Select`, `Slider`, and `Switch`.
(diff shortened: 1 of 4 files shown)


chaance (approved):

koreahghg:
Hi @chaance ,  just following up on this — thanks for the review! Since it's approved and CI is green, could this be merged when you get a chance?

PR #4040, opened 2026-07-18
Only open Tooltip on genuine user-initiated focus

### LINKED ISSUE #2248: Tooltip within popover opens automatically due to trigger receiving focus
### Current Behavior

Hi radix-ui folks <3

Currently, if you create a popover with a tooltip inside, opening the popover automatically opens the tooltip. This is because the popover auto-focuses it's contents, and if the tooltip trigger is the first focusable element it receives that focus, which causes it to open.

This may not really be all that much of an issue - you can work around it by `.preventDefault()`ing the popover `onOpenAutoFocus`, but it tripped me up for a little while so I figured I'd let you know - feel free to close!

Kapture 2023-07-03 at 12 41 12

### Expected behavior

When opening a popover containing a tooltip, the tooltip should remain closed by default until its trigger is hovered

### Reproducible example

https://stackblitz.com/edit/stackblitz-starters-qgptek?file=src%2FApp.tsx

### Suggested solution

Possibly nothing! Although I'd guess the `.preventDefault()` approach has it's tradeoffs for accessibility.

### Your environment

| Radix Package(s) |  react-popover, react-tooltip       |  1.0.6       |
| React            | n/a     |     18.2    |
| Browser          |   Chrome  |    114.0.5735.133     |

kirthi-b:
Tooltip.Trigger opens the tooltip on any focus event, with no way to tell a real user interaction apart from focus it receives as a side effect of something else happening. The most common case: a Popover or Dialog auto-focuses its content on open, and if the tooltip trigger happens to be the first focusable element inside, the tooltip pops open immediately, before the user has hovered or tabbed to it themselves.

This fixes that by gating the open-on-focus behavior on `:focus-visible`. That pseudo-class already reflects the browser's own determination of whether a focus event came from a real keyboard/user interaction, as opposed to a plain scripted `.focus()` call, which is exactly what Popover's auto-focus (packages/react/focus-scope) does under the hood. Hover-triggered opening and the existing pointerdown guard are untouched.

I confirmed the bug with a Popover containing a Tooltip: before the fix, opening the popover left the tooltip in data-state="instant-open" the moment its content auto-focused. After the fix it stays closed until the user actually interacts with the trigger. I added three tests to Tooltip's suite covering: focus from a plain scripted .focus() call stays closed, real keyboard tab navigation still opens it, and tab-away still closes it. All pass, and the existing hover and click tests are untouched and still pass.

This only touches Tooltip's own trigger, not any shared focus-management code, so the blast radius should be limited to Tooltip itself. The one behavior change worth calling out: if anyone was relying on a scripted .focus() call to intentionally open a tooltip (outside of keyboard navigation), that will now no longer open it. That seems like exactly the behavior people have been asking for in this issue, but flagging it since Tooltip's focus handling is used everywhere.

Fixes #2248

### DIFF
--- a/.changeset/tooltip-focus-visible-open.md
+++ b/.changeset/tooltip-focus-visible-open.md
@@ -1,0 +1,6 @@
+---
+"radix-ui": patch
+"@radix-ui/react-tooltip": patch
+---
+
+Fixed `Tooltip` opening automatically when its trigger receives focus programmatically, e.g. when it is the first focusable element inside a `Popover` or `Dialog` that auto-focuses its content on open. The tooltip now only opens on focus that comes from a genuine user interaction (such as tabbing to the trigger), determined using `:focus-visible`. Hover-triggered opening is unaffected.
--- a/packages/react/tooltip/src/tooltip.test.tsx
+++ b/packages/react/tooltip/src/tooltip.test.tsx
@@ -258,6 +258,98 @@ describe('Tooltip', () => {
     expect(commitCounts.c! - initial.c!).toBe(0);
   });
 
+  // Regression tests for https://github.com/radix-ui/primitives/issues/2248
+  // A trigger can receive focus without the user directly interacting with
+  // it, e.g. when a `Popover` or `Dialog` auto-focuses its content on open
+  // and the trigger happens to be the first focusable descendant. The
+  // tooltip should only open for focus that genuinely originates from the
+  // user (such as keyboard navigation), not from that kind of programmatic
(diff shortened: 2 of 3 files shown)


chaance:
I think this solution is probably an over-correction and will result in accessibility issues. There's possibly a case to be made for making this opt-in via a prop, but even then I think we'd need to test extensively before assuming this is the better behavior. My recommendation is to redirect focus elsewhere in your dialog based on its content.

PR #4053, opened 2026-07-20
fix(Select): register content as ancestor FocusScope branch to prevent focus freeze inside Dialog

### LINKED ISSUE #3701: React 19: Select inside Dialog causes aria-hidden focus freeze
### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/2r30e)

### Your environment

gonzoblasco:
When a Select is used inside a Dialog (React 19), the Dialog's FocusScope traps focus with `trapped={true}`. The Select opens its content in a portal (outside the Dialog's DOM subtree), so the Dialog's FocusScope reclaims focus whenever it moves to the Select's portal — freezing the user in the Dialog with no way to interact with the Select.

**Root cause:** the Select's `FocusScope` doesn't register its content as a branch of the nearest ancestor `FocusScopeBranchProvider` (set up by Dialog). The Dialog's FocusScope sees focus leave its container and pulls it back.

**The fix:** added `useFocusScopeBranch(content)` in `SelectContentImpl`. This registers the Select's portal content as a branch of the ancestor FocusScope, so the Dialog's FocusScope knows the Select's portal is part of its scope and allows focus to move freely between the two.

This is the same pattern already used by other portalled components (Popover, Tooltip) that work correctly inside Dialogs.

**Testing:** all existing tests pass — Select (17), Dialog (15), FocusScope (11).

Fixes #3701

### DIFF
--- a/.changeset/slick-emus-show.md
+++ b/.changeset/slick-emus-show.md
@@ -1,0 +1,5 @@
+---
+"@radix-ui/react-select": patch
+---
+
+fix(Select): register content as ancestor FocusScope branch to prevent focus freeze inside Dialog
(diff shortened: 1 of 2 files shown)


gonzoblasco:

PR #4065, opened 2026-07-22
fix(slider): commit value on blur after keyboard navigation (#3619)

### LINKED ISSUE #3619: Slider: onValueCommit is not invoked onBlur with Keyboard Navigation
## Bug report
`Slider.Root` has a property `onValueCommit` that does not invoke any sort of `onBlur` expectedly when the user is incrementing, decrementing, focusing, and blurring using the keyboard.

### Current Behavior
When user blurs after keyboard-based updates, `onValueCommit` **_is not_** invoked.

I have a simple state:
- Turns red when `onValueCommit` is invoked
- Turns green when `onValueChange` is invoked

https://github.com/user-attachments/assets/a6936a73-7104-4958-80a8-bf856016d361

### Expected behavior
When user blurs after keyboard-based updates, `onValueCommit` **_is_** invoked.

### Suggested solution
I think there may need to be an `onBlur` explicitly specified on the input.

### Additional context
Otherwise, the component API design is really thoughtful and well-done - kudos.

### Your environment
- Chrome on MacOS
- "@radix-ui/react-slider": "1.2.0"

gonzoblasco:
## Description

Fixes #3619

When the slider value is changed via keyboard (Arrow keys, Home, End) and the user then tabs away (blur), `onValueCommit` was not being called.

### Root cause

The keyboard handlers (`onStepKeyDown`, `onHomeKeyDown`, `onEndKeyDown`) call `onValueCommit` immediately on each key press via `updateValues` with `{ commit: true }`. However, if the user blurs without pressing another key, the last committed value was never recorded — there was no blur handler to catch this case.

### Fix

Added a `focusout` event listener (capture phase) on the slider root element that compares the current value against the last committed value (`committedValuesRef`) and calls `onValueCommit` if they differ.

The `committedValuesRef` is updated whenever `onValueCommit` is called (both from keyboard handlers and pointer-based `handleSlideEnd`), so the blur handler only fires when there is an actual uncommitted change.

### Testing

- All 16 existing tests pass
- TypeScript: clean

### DIFF
--- a/.changeset/calm-paws-itch.md
+++ b/.changeset/calm-paws-itch.md
@@ -1,0 +1,15 @@
+---
+'@radix-ui/react-slider': patch
+---
+
+fix(slider): commit value on blur after keyboard navigation (#3619)
+
+When the slider value is changed via keyboard (Arrow keys, Home, End) and
+the user then tabs away (blur), `onValueCommit` was not being called.
+This happened because the keyboard handlers call `onValueCommit` immediately
+on each key press, but if the user blurs without pressing another key, the
+last committed value was never recorded.
+
+Added a `focusout` event listener (capture phase) on the slider root that
+compares the current value against the last committed value and calls
+`onValueCommit` if they differ.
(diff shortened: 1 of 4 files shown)


gonzoblasco:

PR #4067, opened 2026-07-22
fix(select): add aria-activedescendant and fix aria-selected (#3636)

### LINKED ISSUE #3636: Select: Implement `aria-activedescendant` for improved accessibility

gonzoblasco:
## Description

Fixes #3636

Two accessibility fixes for the Select component:

### 1. Added `aria-activedescendant` to the listbox

The `role="listbox"` element now has an `aria-activedescendant` attribute that dynamically points to the `id` of the currently focused option, as required by the [ARIA listbox pattern](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role#states_and_properties).

### 2. Fixed `aria-selected` on options

Previously `aria-selected` was set to `isSelected && isFocused`, meaning a selected item would not be announced as selected by screen readers unless it was also focused. Changed to just `isSelected` so it correctly reflects selection state independent of focus.

### Implementation

- Each `SelectItem` now has a unique `id` (via `useId`)
- A shared `activeItemIdRef` in `SelectContentContext` tracks the currently focused item
- The listbox reads `activeItemIdRef.current` for `aria-activedescendant`
- Focus/blur handlers on each item update the ref

### Testing

- All 19 existing tests pass
- TypeScript: clean

### DIFF
--- a/.changeset/calm-paws-itch.md
+++ b/.changeset/calm-paws-itch.md
@@ -1,0 +1,15 @@
+---
+'@radix-ui/react-slider': patch
+---
+
+fix(slider): commit value on blur after keyboard navigation (#3619)
+
+When the slider value is changed via keyboard (Arrow keys, Home, End) and
+the user then tabs away (blur), `onValueCommit` was not being called.
+This happened because the keyboard handlers call `onValueCommit` immediately
+on each key press, but if the user blurs without pressing another key, the
+last committed value was never recorded.
+
+Added a `focusout` event listener (capture phase) on the slider root that
+compares the current value against the last committed value and calls
+`onValueCommit` if they differ.
--- a/.changeset/forty-paws-rule.md
+++ b/.changeset/forty-paws-rule.md
@@ -1,0 +1,12 @@
+---
+'@radix-ui/react-dropdown-menu': patch
+---
+
+fix(dropdown-menu): open menu on ArrowUp and focus last item (#3640)
+
+According to the W3C APG pattern for menu buttons, pressing ArrowUp on a
+focused trigger should open the menu and move focus to the last item.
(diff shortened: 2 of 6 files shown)


gonzoblasco:

PR #4109, opened 2026-08-12
fix(select): announce option position and total for screen readers

### LINKED ISSUE #3962: [Select] VoiceOver announces wrong items count in Chrome
### Current Behavior

When using VoiceOver on macOS with **Chrome**, navigating a `Select` list announces
incorrect item counts. Tested on both a custom implementation and the Radix
documentation examples (fruit select).

**Scenario A — a value is already selected:**
Opening the list and focusing an item announces the label with a wrong total, e.g.
*"Banana, 1 of 3"* even when the list contains more items.

**Scenario B — no value is selected:**
Opening the list announces only the item label with no position or count at all,
e.g. *"Banana"* instead of *"Banana, 2 of 5"*.

VoiceOver + Safari is unaffected. NVDA on Windows is unaffected.

### Expected behavior

VoiceOver should announce the item label, its position, and the correct total for
every focused item regardless of whether a value is pre-selected,
e.g. *"Banana, 2 of 5"*.

### Reproducible example

Reproduced on the Radix documentation page using the fruit Select example.
https://www.radix-ui.com/primitives/docs/components/select?utm_source=chatgpt.com#accessibility

### Suggested solution

Needs investigation.

### Additional context

VoiceOver + Safari = works correctly
NVDA + Chrome = works correctly
VoiceOver + Chrome = problem

### Your environment

| Radix Package(s) |    @radix-ui/react-select     |     2.2.6    |
| React            | n/a     |     ^18    |
| Browser          |    Chrome     |    Version 149.0.7827.103     |
| Assistive tech   |    VoiceOver     |         |
| Operating System |     macOS    |    Tahoe 26.5 (25F71)     |

### Screen recording:

https://github.com/user-attachments/assets/38d195a3-1a8d-4abb-b620-799c0ee948cc

gonzoblasco:
## Summary

Fixes #3962 - `Select` options don't declare their position/total, so VoiceOver + Chrome announces an incorrect item count (based on the options visible in the scrollable viewport) or omits the count entirely when no value is preselected.

Example: a list of 5 fruits with `"Banana"` preselected announces `"Banana, 1 of 3"` instead of `"Banana, 2 of 5"`; with no preselection it announces just `"Banana"` with no position or count.

## Root cause

`Select.Item` renders `role="option"` but never sets `aria-posinset` / `aria-setsize`. Radix uses roving focus (real focus moves to each option) inside a scrollable `role="listbox"`, so the browser is left to derive the position from the DOM. Safari+VoiceOver and NVDA+Chrome happen to work, but VoiceOver+Chrome counts only the options currently visible in the viewport, giving a wrong total.

## Change

Each `Select.Item` now computes and declares its own position:

- `aria-setsize` - the number of options in its list container
- `aria-posinset` - its 1-based index within that container

The list container is the option's immediate parent: the enclosing `role="group"` when the item sits in a `Select.Group`, otherwise the viewport inside the `role="listbox"`. This keeps the count correct for grouped selects too (each group counts independently, per the [APG grouped listbox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/)).

The values are computed once on mount via `useLayoutEffect` from the DOM, so they're available before the first paint of the opened list.

## Verification

- New unit tests: ungrouped options get `aria-posinset`/`aria-setsize` for the full list; grouped options count within each group (count restarts per group).
- All existing Select tests pass.
- `lint`, `typecheck`, format check, and build (ESM/CJS/types) all pass.

Manual validation with VoiceOver + Chrome is still needed to confirm the announcement, but the attributes are now correct in the DOM.

### DIFF
--- a/.changeset/soft-tigers-glow.md
+++ b/.changeset/soft-tigers-glow.md
@@ -1,0 +1,6 @@
+---
+"@radix-ui/react-select": patch
+"radix-ui": patch
+---
+
+Fix `Select.Item` so screen readers announce the option's position and total (`aria-posinset` / `aria-setsize`). VoiceOver + Chrome previously announced an incorrect count (based on the options visible in the scrollable viewport) or omitted the count entirely when no value was preselected; the option now declares its own position within its list, including grouped selects.
--- a/packages/react/select/src/select.test.tsx
+++ b/packages/react/select/src/select.test.tsx
@@ -1602,6 +1602,124 @@ describe('Select.ItemText', () => {
   it.todo("applies the consumer's `className` and `style`");
 });
 
+describe('Select.Item aria-posinset/aria-setsize', () => {
+  afterEach(cleanupModal);
+
+  it('declares each ungrouped option position within the full list', () => {
+    render(
+      <Select.Root defaultOpen>
+        <Select.Trigger>
+          <Select.Value placeholder="Pick one" />
+        </Select.Trigger>
+        <Select.Portal>
+          <Select.Content>
+            <Select.Viewport>
+              <Select.Item value="apple">
+                <Select.ItemText>Apple</Select.ItemText>
+              </Select.Item>
+              <Select.Item value="banana">
+                <Select.ItemText>Banana</Select.ItemText>
+              </Select.Item>
+              <Select.Item value="cherry">
+                <Select.ItemText>Cherry</Select.ItemText>
+              </Select.Item>
+              <Select.Item value="date">
+                <Select.ItemText>Date</Select.ItemText>
+              </Select.Item>
+              <Select.Item value="elderberry">
+                <Select.ItemText>Elderberry</Select.ItemText>
+              </Select.Item>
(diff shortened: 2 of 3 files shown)


gonzoblasco:
**Manual validation with VoiceOver + Chrome** on the clearable story (`None/One/Two/Three`, default `Two`):

- ✅ Options `One`, `Two`, `Three` now announce correctly: "2 of 4", "3 of 4", "4 of 4" — previously the count was wrong ("1 of 3", based on visible options in the scrollable viewport) or missing entirely. This fixes #3962.
- ✅ Live captions confirm the announcements.

Found a related but **separate** issue while testing: the `None` item (the first option, with no preselection) announces nothing at all — no position, no total. The `None` item is also rendered as a real dropdown option while the trigger shows "Pick an option", so it behaves more like a placeholder than a selectable value. That's a different bug, worth its own investigation — I opened #4110 to track it.

PR #4112, opened 2026-08-13
fix(Select): set aria-posinset/setsize synchronously so first option announces correctly

### LINKED ISSUE #4110: select: first option without preselection announces nothing to screen readers
**Bug:** In `Select`, when no value is preselected, the first option in the list announces nothing (no position, no total) to VoiceOver + Chrome.

**Repro:** Use the clearable Select story (`None/One/Two/Three`, default `Two`). Open the dropdown — focus falls on `None` (the first option) and VoiceOver announces nothing. Moving to `One` correctly announces "2 of 4".

The `None` item is also rendered as a real dropdown option while the trigger displays "Pick an option", so it behaves like a placeholder rather than a selectable value.

**Related:** #4109 (fixes the count for the rest of the options) and #3962.

waterWang:
Fixes #4110

Sets `aria-posinset`/`aria-setsize` directly on the DOM element in a `useLayoutEffect` instead of via React state + re-render. This ensures the attributes are present when VoiceOver reads the first focused option — including when no value is preselected, which VoiceOver + Chrome otherwise announces before a state-driven re-render would have applied the attributes.

**Root cause:** The existing `aria-posinset`/`aria-setsize` approach (PR #4109) computes positions in `useLayoutEffect` + `useState`, which triggers a re-render. When the list opens with no selection, VoiceOver announces the first focused option *before* the re-render applies the attributes → silence. This fix sets attributes directly on the DOM node in the layout effect, bypassing the state round-trip entirely.

**Implementation:**
1. Added `itemRef` to track the DOM node
2. `useLayoutEffect` computes the option's position from its parent's `[role="option"]` children and sets `aria-posinset`/`aria-setsize` via `setAttribute`


gonzoblasco (commented):
Nice catch on the timing - the state round-trip in the `useLayoutEffect` + `useState` approach is exactly why the first option gets read before the attributes land. Setting them directly on the DOM node in the layout effect (which runs before paint) is a clean way to close that gap, and it's consistent with how the rest of the options behave once mounted.

A few things worth addressing before this is mergeable:

**1. Missing tests.** This PR changes `select.tsx` but adds no coverage. The behavior it fixes (first option announcing position/total when nothing is preselected) is exactly the kind of regression that needs a test - there's already a test suite for `aria-posinset`/`aria-setsize` in `select.test.tsx` (from #4109) covering the flat list, grouped list, and deferred-mount cases. Adding a case for the no-preselection scenario would lock this in. Without it, a future refactor could silently reintroduce the race.

**2. Imperative `setAttribute` vs declarative props.** Writing the attributes directly on the DOM works (React won't clobber attributes it never declared), but it's a pattern Radix generally avoids - the declarative `aria-posinset={...}`/`aria-setsize={...}` props are more idiomatic and keep the DOM in sync with React's mental model. The tradeoff here is real (the state round-trip is the bug), so I'm not saying revert - but it's worth a comment in the code explaining *why* the imperative path is necessary, since a future reader might "clean it up" back into state and reintroduce the race.

**3. Deferred-mount edge case.** The `[]` deps + compute-once-on-mount assumption holds when all options mount in the same commit, but if an option mounts later (conditional rendering, virtualization), its `aria-setsize` will be stale because the siblings already computed theirs. Same caveat applies to #4109, but worth documenting either way.

**4. `:scope` selector.** `:scope > [role="option"]` is well-supported in modern browsers, but it's worth confirming it's fine for the browsers Radix targets (it's not in the oldest supported set). A fallback or a comment would de-risk it.

Overall the approach is sound and the root-cause analysis is right. The main blocker is test coverage - happy to help if useful.

PR #3189, opened 2024-10-22
[ToggleGroup] fix: correct role for single value toggle group

akselinurmio:
### Description

Use correct role `"radiogroup"` instead of `"group"` for single value toggle groups.

Closes #3188

PR #4123, opened 2026-08-18
fix(toggle-group): check the item arrow keys move to when type="single"

### LINKED ISSUE #4119: `ToggleGroup` with `type="single"` announces `role="radiogroup"` but uses toolbar keyboard interaction
### Summary

`<ToggleGroup.Root type="single">` renders `role="radiogroup"` on the root and
`role="radio"` on the items, but arrow keys only move focus — they do not change
the selection. Per the [WAI-ARIA APG radio group
pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/), arrow keys in a
radiogroup move focus **and check** the newly focused radio.

`RadioGroup` from the same library implements this correctly, which makes the
inconsistency easy to see side by side.

### Why this is a mismatch and not a preference

The roles were added deliberately, in two steps, and neither discussion covered
keyboard interaction:

- #951 — "[ToggleGroup] Use `role=radio` when `type=single`" → PR #1118
- #3188 — "Toggle Group should have radiogroup role" → PR #3189

Both are closed and shipped. The result is that `type="single"` now carries the
roles of a radio group while keeping the roving-focus keyboard model of a
toolbar. Assistive technology announces "radio group, 1 of 3", and the
interaction that announcement implies is not there.

`type="multiple"` renders `role="toolbar"` and is unaffected — arrow-moves-only
is correct there. The mismatch exists only in the `single` case.

### Reproduction

```tsx
<ToggleGroup.Root type="single" defaultValue="a" aria-label="View">
  <ToggleGroup.Item value="a">A</ToggleGroup.Item>
  <ToggleGroup.Item value="b">B</ToggleGroup.Item>
</ToggleGroup.Root>
```

Focus item A, press ArrowRight.

- **Expected** (APG radio group): focus moves to B and B becomes checked.
- **Actual**: focus moves to B, A stays checked. A second Space/Enter is needed.

Swapping the same markup to `RadioGroup.Root` / `RadioGroup.Item` gives the
expected behaviour.

### A note on measuring this

Testing with a single `press("ArrowRight")` makes **both** primitives look
broken. `press` fires `keydown` and `keyup` back to back, and `RadioGroup`
clears on `keyup` the "arrow is down" flag that its `onFocus` handler reads.
Separating the events — `keyboard.down`, wait, `keyboard.up` — shows
`RadioGroup` checking and `ToggleGroup` not.

### Versions

`@radix-ui/react-toggle-group` 1.1.19 · `@radix-ui/react-radio-group` 1.4.7 ·
Chromium via Playwright.

solomonaustin:
Fixes #4119. The diagnosis in that issue is @lucasmatosdeoliveira's, including the two PRs that introduced the roles.

### Problem

`<ToggleGroup.Root type="single">` renders `role="radiogroup"` with `role="radio"` items and `aria-checked`, but keyboard handling comes from `RovingFocusGroup`, which only moves focus. The [WAI-ARIA radio group pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/) moves focus **and** checks the item focus lands on.

So assistive technology announces "radio group, 1 of 3" and the interaction that announcement implies is not there — arrow keys walk the group without ever changing the value. Keyboard and screen reader users get a control whose stated semantics do not match its behaviour.

The roles were added deliberately in #1118 and #3189, and neither discussion covered keyboard interaction.

### Fix

`RadioGroup` in this repo already solves exactly this, so I mirrored it rather than inventing anything: track arrow keys on `document`, and when focus lands on an item during arrow navigation, click it so the value changes and the change event fires.

Two deliberate differences from `RadioGroup`, both because `Toggle` toggles where a radio does not:

- An item that is already pressed is skipped — clicking it would turn it **off**, and a radio group never unchecks on navigation.
- The listeners are only attached for `type="single"`. `type="multiple"` keeps `role="toolbar"`, where move-only is correct, so it costs nothing there.

### Behaviour change

This is a real behaviour change for `type="single"`: arrow keys now change the value where before they only moved focus. That is the point of the fix — the alternative would be dropping the radio roles, which seems worse given they were added on purpose. Flagging it explicitly so it is a deliberate call rather than a surprise.

`Enter` handling is untouched. `RadioGroup` prevents default on `Enter` per the same pattern; that felt like separate scope from the reported bug, happy to fold it in if you would rather.

### Testing

Four tests added to the existing spec, using the same `fireEvent.keyDown` + manual `focus()` pattern the `RadioGroup` tests use:

- arrow key checks the item it moves to, and unchecks the previous one
- focus arriving without an arrow key leaves the value alone
- an arrow key landing back on the checked item keeps it checked
- `type="multiple"` still only moves focus

Verified the first fails without the source change; the other three pass either way and are there to pin the existing behaviour.

Full unit suite: 46 files / 984 tests pass. `typecheck` and `lint` clean on the package. Patch changeset included.


solomonaustin:
I went back over this myself before it gets reviewed, and there are two gaps worth putting on the record rather than leaving for someone to find.

**`Home` / `End` / `PageUp` / `PageDown` move focus without checking.** `MAP_KEY_TO_FOCUS_INTENT` in `roving-focus-group.tsx` maps all four to `first`/`last`, so they move the roving focus, but this PR only tracks the arrow keys. That leaves focus and selection out of sync — the same class of mismatch the PR is fixing, just on a different key.

I left it that way on purpose: `RadioGroup` tracks exactly the same four arrow keys and has the identical gap, so matching it seemed better than diverging in a bug fix. The APG radio pattern also doesn't specify `Home`/`End`. But if you'd rather both primitives handled them, it's a one-line change here and I'm happy to do it in both.

**The listeners are on the global `document`.** So in an iframe or a secondary window the keydown never reaches them and arrow keys won't select. `RadioGroup` does the same, which is why I followed it, but `ownerDocument` would be more correct and I do have a ref to the item in scope. Same offer: happy to change it here, or in both, or leave it for consistency.

Neither affects the reported bug — arrow-key selection works either way — so I haven't changed anything. Just flagging so the choice is yours rather than accidental.

solomonaustin:
Friendly ping — CI hasn't run on this yet. All six workflow runs are sitting at `action_required` (the first-time-contributor gate), so the `UNSTABLE` merge state is just "no checks reported", not a failure. It needs a maintainer to approve the runs.

In the meantime I ran both workflows locally against the PR head (`79567e1`), on `pnpm@11.5.0`:

**Build** (`.github/workflows/build.yml`)
- `pnpm lint` — clean across every package
- `node scripts/check-composed-refs.mjs` — no unstable composed-ref patterns (169 files)
- `pnpm build:ci` — all packages built (ESM/CJS/TSC)
- `vitest run` — **984 passed**, 2 skipped, 5 todo across 46 files
- `playwright test` — **104 passed**, 1 skipped, including `e2e/toggle-group.spec.ts`

**SSR build** (`.github/workflows/ssr.yml`)
- `@repo/ssr-testing` production build — all pages prerendered, `/toggle-group` included
- `test:merge` RSC boundary assertion — 17/17 pass

The 21 tests in `packages/react/toggle-group/src/toggle-group.test.tsx` all pass, 8 of which are new here.

Happy to rebase or adjust the approach if you'd rather solve it differently — no rush, just flagging that it's blocked on the workflow approval rather than on anything in the diff.

PR #4126, opened 2026-08-21
fix: apply deferPointerDownOutside consistently across overlay primitives

### LINKED ISSUE #4125: Select, HoverCard, Tooltip and NavigationMenu don't get the deferPointerDownOutside fix that Dialog/Popover got in #3928
PR #3928 added a `deferPointerDownOutside` prop to DismissableLayer and wired it into Dialog.Content and Popover.Content, so a pointerdown that lands outside the layer waits for the corresponding click before dismissing. This gives third-party UI injected by browser extensions (password managers, Grammarly, etc.) a chance to intercept the event, fixing #2055.

I checked every other primitive built on DismissableLayer and none of them received the same fix:
- select.tsx:858
- hover-card.tsx:332
- tooltip.tsx:555
- navigation-menu.tsx:1006

None of these pass `deferPointerDownOutside` to their `DismissableLayer`, so a pointerdown from extension UI still dismisses them immediately, same symptom as #2055 but for these components instead of Dialog.

I believe this is related to #2502 (HoverCard closing on Chrome autofill, closed as "Needs More Information") and #3175 (Select closing a nested HoverCard, still open) - both look like the same missing-defer root cause rather than being fixed already.

Note: Menu (DropdownMenu/ContextMenu/Menubar) has the same gap, but PR #3976 already adds `deferPointerDownOutside` there, so I'm leaving that one out of this report.

Expected: Select, HoverCard, Tooltip and NavigationMenu should get the same `deferPointerDownOutside` treatment as Dialog/Popover/Menu(#3976).

AbhiPra24:
Fixes #4125

### Summary
Extends the `deferPointerDownOutside` fix from Dialog/Popover (#3928) across Select, HoverCard, Tooltip, and NavigationMenu.

### Verification
- Component tests passing.


MILLERMARRU (commented):
Confirmed the diff is minimal and correct: it adds `deferPointerDownOutside` to `HoverCardContentImpl`, `NavigationMenuContentImpl`, `SelectContentImpl`, and `TooltipContentImpl`, mirroring exactly what #3928 did for Dialog/Popover. Nothing extraneous, changeset covers all four packages touched.

One thing worth flagging while this is open: I checked `packages/react/menu/src/menu.tsx` on current `main` and it still doesn't pass `deferPointerDownOutside` to its internal `DismissableLayer` anywhere. The sibling PR #3976 was meant to bring the same fix to the Menu family (DropdownMenu/ContextMenu/Menubar/Menu all share this file), but it's closed without merging. So after this PR lands, Select/HoverCard/Tooltip/NavigationMenu will be consistent but the Menu family will be the odd one out with the exact same browser-extension-interaction gap #3928/#4125 originally described. Might be worth reopening #3976's approach (or extending this PR) so the fix is complete across all overlay primitives rather than four of five.

AbhiPra24:
Thanks @MILLERMARRU for the thorough review and validation!

Good point on the Menu family (`packages/react/menu/src/Menu.tsx`). Keeping this PR scoped strictly to the 4 primitives (`Select`, `HoverCard`, `Tooltip`, `NavigationMenu`) ensures backward compatibility and matches the changeset cleanly without touching nested submenu interactions (which were discussed on #3976). Once this PR lands, we can open a dedicated follow-up specifically for the Menu family with tests for submenus and context menus.

PR #4132, opened 2026-08-29
fix(scroll-area): keep the thumb tracking in minified builds

kirkouimet:
### Description

The scrollbar thumb renders and fades correctly but never moves while scrolling, in production builds only. Dev is fine.

`addUnlinkedScrollListener` starts a `requestAnimationFrame` loop that polls the viewport and calls the handler that positions the thumb. The loop is started by an IIFE:

```js
(function loop() {
  ...
  rAF = window.requestAnimationFrame(loop);
})();
```

Built with esbuild's `keepNames`, that IIFE is emitted as a `/* @__PURE__ */` annotated call:

```js
(/* @__PURE__ */ __name((function loop() { ... }), "loop"))();
```

The annotation is true of `__name` itself, but the surrounding call is invoked for its side effect and its return value is unused, so a minifier may drop the entire expression. In `@radix-ui/react-scroll-area@1.2.18` built through Next 16, the shipped output is:

```js
var Z = m((e, t = () => {}) => (e.scrollLeft, e.scrollTop, () => window.cancelAnimationFrame(0)), "addUnlinkedScrollListener");
```

Two bare property reads and a cleanup that cancels frame `0`. The loop is gone.

The symptom is easy to misread. `ScrollAreaThumb`'s `onThumbPositionChange` still runs on mount and on the debounced scroll-end path, so the thumb appears in the correct position every time the scrollbar fades back in, and only fails to move *during* a scroll. It looks like a stale-position bug rather than a missing loop.

Measured on a deployed build, scrolling for six seconds:

| | scroll events | distinct thumb transforms |
| --- | --- | --- |
| dev | 215 | 185 |
| production | 253 | 4 |

### Fix

Declare `loop` and call it as a statement. Behaviour is identical, and there is no longer an annotated expression for the minifier to remove.

Verified against a real Next 16 production build. Before, the minified output contained no `requestAnimationFrame`; after, the loop survives and the thumb tracks normally.

This is the only IIFE in the file, so no other call sites are affected.


stefan-dospinescu-spektr:
Hope this will be merged soon ❤️

kachkaev:
Looks like a duplicate of https://github.com/radix-ui/primitives/pull/4106 which fixes the same SWC glitch in two more cases.

Issue #3836, opened 2026-03-24
DialogTrigger asChild still breaks with Server Components in published @radix-ui/react-dialog because it resolves @radix-ui/react-slot@1.2.3


maltesa:
### Current Behavior

The fix from [#3165](https://github.com/radix-ui/primitives/issues/3165) appears to exist in `@radix-ui/react-slot@1.2.4`, but the latest published `@radix-ui/react-dialog@1.1.15` still resolves `@radix-ui/react-slot@1.2.3`.

In our app, `DialogTrigger asChild` still breaks in a React Server Components setup when the trigger element is created on the server and passed into a client component.

With the published packages, the trigger works inconsistently in dev and disappears in our production build after hydration.

### Expected behavior

`DialogTrigger asChild` should work in production when the child trigger is created on the server and passed into a client component.

If the fix from #3165 is the intended fix for this class of problem, the published `@radix-ui/react-dialog` package should pick it up in practice.

### Reproducible example

We have not reduced this to a public repro yet, but the failing pattern is:

Server component:

```tsx
return <EditDialog trigger={<Button>Edit</Button>} />
```

Client component:

```tsx
'use client'

import * as Dialog from '@radix-ui/react-dialog'

export function EditDialog({ trigger }: { trigger: React.ReactNode }) {
  return (
    <Dialog.Root>
      <Dialog.Trigger asChild>{trigger}</Dialog.Trigger>
      <Dialog.Content>...</Dialog.Content>
    </Dialog.Root>
  )
}
```

### Suggested solution

Publish a `@radix-ui/react-dialog` release that picks up the Slot fix from #3165, or confirm the intended supported version/workaround.

Forcing `@radix-ui/react-slot@1.2.4` fixes the issue for us without any component-level workaround:

```json
{
  "pnpm": {
    "overrides": {
      "@radix-ui/react-slot": "1.2.4"
    }
  }
}
```

After that override, the same production build works correctly and the dialog opens as expected.

### Additional context

This looks like a published-package / release gap rather than a new root-cause bug:

- `@radix-ui/react-slot@1.2.4` appears to contain the fix from #3165
- latest published `@radix-ui/react-dialog@1.1.15` still resolves `@radix-ui/react-slot@1.2.3`
- overriding Slot to `1.2.4` fixes the production failure for us

Related: [#3165](https://github.com/radix-ui/primitives/issues/3165), [#3416](https://github.com/radix-ui/primitives/issues/3416)

### Your environment

[...]

Issue #3780, opened 2025-12-16
[Regression] Radix Slot/asChild prop merging silently fails in RSC dev mode after Next.js 16.0.10 security patch


paolo-dipasquale-intercom:
### Describe the bug
After upgrading to Next.js 16.0.10 (React2Shell CVE patch), the Radix UI `Slot` component 
silently fails to merge props onto child elements when used across a Server Component → 
Client Component boundary **in local development mode**.

When using `asChild={true}` on a Client Component (e.g., Button with `"use client"`) 
inside a Server Component, the Slot's className and other props are not serialized/merged 
onto the child element. **The component silently doesn't render at all** - no errors in 
the console, no hydration warnings, nothing. This makes it extremely difficult to debug.

### To Reproduce
1. Create a Server Component (no `"use client"`)
2. Inside it, render a Client Component that uses Radix `Slot` with `asChild={true}`
3. Pass className to the component expecting Slot to merge it onto the child
4. Run `next dev` (local development mode)
5. The **entire component fails to render** with no error messages

```
// Server Component (card.tsx - no "use client")
<Button className="text-white" theme="transparent" asChild>
  <span>Click me</span>  {/* ENTIRE BUTTON SILENTLY NOT RENDERED */}
</Button>

// Client Component (button/index.tsx - has "use client")
const Comp = asChild ? Slot : "button";
return <Comp className={cn(base(), className)} {...props}>{children}</Comp>### Expected behavior
The `Slot` component should merge the className onto the child `<span>` element, 
or at minimum throw an error/warning if serialization fails.
```

### Actual behavior
- The component silently doesn't render
- No console errors or warnings
- No React hydration mismatch warnings
- `document.querySelector()` returns `null` for the element
- Only discoverable by manually inspecting the DOM

### Workaround
Moving the className directly onto the child element bypasses the issue:
<Button theme="transparent" asChild>
  <span className="text-white">Click me</span>  {/* Works */}
</Button>### Environment
- Next.js: 16.0.10
- React: 19.1.2
- @radix-ui/react-slot: 1.1.2
- Node: [your version]
- OS: macOS

### Additional context
- This regression appeared after updating to Next.js 16.0.10 which includes patches 
  for CVE-2025-55182 (React2Shell)
- The security patch likely changed RSC serialization behavior in a way that breaks 
  Slot's prop merging across the RSC boundary
- **The silent failure makes this extremely difficult to debug** - we spent significant 
[...]

PR #4137, opened 2026-09-02
fix(slot): normalize single-element children arrays for React 19 compatibility

### LINKED ISSUE #4124: Popover: PopoverContent throws "Primitive.div failed to slot onto its children" on mount (React 19)
### Current Behavior

`PopoverContent` throws immediately on mount:

```
Error: Primitive.div failed to slot onto its children. Expected a single React element child or `Slottable`.
    at Primitive.div.Slot
    at updateForwardRef
    at beginWork
    ...
```

This happens with a minimal, standard usage — a single valid-element child, no fragments, no arrays, nothing unusual:

```tsx
import { Popover, PopoverContent, PopoverTrigger } from "radix-ui" // (also reproduces via the standalone @radix-ui/react-popover package)

function Demo() {
  return (
    <Popover.Root>
      <Popover.Trigger asChild>
        <button type="button">Open popover</button>
      </Popover.Trigger>
      <Popover.Portal>
        <Popover.Content>
          <div>Hello</div>
        </Popover.Content>
      </Popover.Portal>
    </Popover.Root>
  )
}
```

Clicking the trigger throws the error above the moment `Popover.Content` mounts. The trigger itself renders fine before that — the crash is specifically on `Content` mount, not on `Trigger`/`Anchor` composition.

### Expected behavior

The popover opens and renders its content, same as `Tooltip` and `Dialog` do with an equivalent structure.

### Reproduction / isolation notes

I spent a while isolating this before filing, since I initially assumed it was a mistake in my own code:

- **Reproduces with a single child** — not a multiple-children-into-`Slot` issue (ruled out the `Slottable` scenario from #3747, which is a different pattern).
- **Reproduces with `modal` both `true` and `false`** — not specific to the non-modal code path.
- **`Tooltip` and `Dialog` both work fine** in the exact same app/dependency tree with structurally similar composition (`Trigger asChild` + `Portal` + focus-managed `Content`).
- **Pinning `@radix-ui/react-slot` to `1.2.5` or `1.2.3` (full clean reinstalls each time) does not fix it** — so this isn't the Slot regression described in #3747 (which was fixed/worked around at the Slot level for a different pattern).
- Both `radix-ui` (`1.6.7`) and `@radix-ui/react-popover` (`1.1.23`) are already the latest stable releases at the time of filing — no newer version available to try (only `1.2.0-rc.*` prereleases for `@radix-ui/react-popover`, and `1.7.0-rc.*` for the `radix-ui` meta-package).

### Root cause (best guess from reading source)

`PopoverContentImpl` composes:

```
FocusScope (asChild) > DismissableLayer (asChild) > PopperPrimitive.Content
```

`Dialog` composes `FocusScope (asChild) > DismissableLayer (asChild) > Primitive.div` (no Popper layer) — **works fine**.

`Tooltip` composes `PopperPrimitive.Content` directly, with no `FocusScope`/`DismissableLayer` wrapping (tooltips aren't focusable, so this makes sense) — **works fine**.

`Popover` is the one primitive combining both `FocusScope`/`DismissableLayer` *and* `PopperPrimitive.Content` underneath — and it's the one that's broken. That intersection is my best guess at the actual fault line, though I haven't dug further into exactly which prop/ref is losing its single-child shape across that specific chain.

### Environment

- `react`: 19.2.8
- `react-dom`: 19.2.8
- `radix-ui`: 1.6.7 (unified package)
- `@radix-ui/react-popover` (resolved): 1.1.23
- `@radix-ui/react-slot` (resolved): 1.3.3 (also tried 1.2.5 and 1.2.3 via `overrides` — same failure both times)
- Bundler: Vite 8 (also reproduces under Storybook 10.5's Vite builder, separately from the app's own build)
- OS: Windows

### Workaround in use

Not using `Popover` for now — using `Dialog` with `modal={false}` and manually-positioned `Content` (no `Popper` anchoring) instead, which sidesteps the broken combination entirely.

VedAnt-1004:
Fixes #4124

### Problem Statement

In React 19, nested component compositions using `asChild` (such as `PopoverContentImpl` composing `FocusScope asChild` > `DismissableLayer asChild` > `PopperPrimitive.Content`) can pass `children` as a single-element array rather than a direct element reference, or alongside falsy values resulting from conditional logic.

`Slot` previously resolved candidate elements using:
```tsx
!hasSlottable &&
React.Children.count(children) === 1 &&
React.isValidElement(children)
```

Because an array itself is not a valid React element (`React.isValidElement([<Child />])` returns `false`), `slottableElement` remained `null`. As a result, `Slot` assumed invalid children were passed and threw:

```text
Error: Primitive.div failed to slot onto its children. Expected a single React element child or `Slottable`.
```

---

### Fix

1. **Normalized Candidate Children**: In `packages/react/slot/src/slot.tsx`, normalized and filtered `children` using `React.Children.toArray(children)` to strip out non-rendering primitives (`null`, `undefined`, boolean values).
2. **Safe Element Extraction**: If exactly one candidate remains in the normalized array and it satisfies `React.isValidElement(candidate)`, `Slot` assigns it to `slottableElement` for proper prop and ref merging.

---

### Benefits after the fix

* **Full React 19 Compatibility**: Resolves runtime slotting errors across composite primitives like `PopoverContent`, `DialogContent`, and `DropdownMenuContent`.
* **Resilient Child Handling**: Gracefully handles single-child arrays and falsy conditional siblings (e.g., `{condition && <span />} <Target />`) without breaking composition.
* **Preserves Slot Invariants**: Retains strict developer warnings and errors when multiple element children or invalid primitive types (like raw strings) are passed without a `Slottable`.

---

### Testing performed

* **Slot Unit Tests (`packages/react/slot/src/slot.test.tsx`)**:
  * Added regression test verifying `Slot` renders and merges props correctly when the child is wrapped in a single-element array (`[<button key="single">array child</button>]`).
  * Added regression test verifying `Slot` ignores falsy siblings (`false && <span />`, `null`) and correctly slots onto the single valid element.
  * Verified all existing slot test suites pass (34 passed, 2 skipped).

* **Popover Integration Tests (`packages/react/popover/src/popover.test.tsx`)**:
  * Added regression tests verifying `PopoverContent` successfully mounts on trigger click without throwing slot errors under both `modal={true}` and `modal={false}` modes (reproducing #4124).
  * Confirmed all 26 popover unit tests pass.

* **Test Execution Commands**:
  ```bash
  pnpm vitest run packages/react/slot
  pnpm vitest run packages/react/popover
  ```


mbelhadi:
This fixes a real case for us, and the numbers might help it get reviewed.

We have a pnpm monorepo with two production Next.js sites and an operator dashboard. On **Next 16.3.4**, `next dev` answered **500 on every route** of all three — under Turbopack *and* under `--webpack`:

```
Error: Slot failed to slot onto its children. Expected a single React element child or `Slottable`.
```

Probing `Slot` at runtime, with `@radix-ui/react-slot@1.3.3` untouched:

```json
{ "isLazy": true, "useType": "function", "dollar": "Symbol(react.lazy)",
  "lazySymbolMatches": true, "payloadThenable": true,
  "afterUse": "Symbol(react.lazy) valid=false" }
```

So the existing lazy handling *does* fire — `isLazyComponent` returns true and `use(children._payload)` runs — but what comes back is **another** `react.lazy`, so `isValidElement(children)` is still false and the guard below throws. Unwrapping one level is not enough; `Children.toArray` resolves it, which is exactly what this PR does.

Measured on our tree, Next 16.3.4, nothing else changed:

| | before | with this PR |
| --- | --- | --- |
| `/`, `/fr`, `/es`, `/fr/expertise`, `/fr/contact`, `/fr/insights`, `/fr/open-source` | 500 | **200** |
| `Slot failed` errors in the dev log | 9 | **0** |
| Playwright suite | could not start | **125 / 125** |

Two notes that may save someone else the search:

- **#3780, #3776 and #3836 are closed**, against `cac030c` ("[Slot] add nested slottable support"). That commit shipped in 1.3.3 and does not address this — 1.3.3 still throws.
- **`1.4.0-rc.1785512840124` does not fix it either.** It carries the single-level unwrap, which the probe above shows is insufficient.

We are running this patch through `pnpm.patchedDependencies` in the meantime, with a note to delete it rather than bump it once this lands. Happy to test any revision against the same repro.

VedAnt-1004:
Thanks for testing this against your production monorepo and sharing the benchmarks! The confirmation that this resolves the nested lazy wrapper issues on Next.js 16 will be super helpful for the maintainers reviewing this.

